PR #173, opened 2024-04-04
Fix Error Code: 1410 in mysql

DarkMaster007:
Error Code: 1410. You are not allowed to create a user with GRANT is fixed by first creating a user 'acore' with password 'acore' (password from config)


liyunfan1223:
I wonder in which situation this will happen, as I haven't encountered the error myself. I think the creation of the db user should be in the core, the create_mysql.sql file at https://github.com/liyunfan1223/azerothcore-wotlk/blob/Playerbot/data/sql/create/create_mysql.sql.

PR #231, opened 2024-05-20
Quests required for dungeons

yuan227301:
Added some tasks required to enter dungeons. Such as Pit of Saron, Halls of Reflection and so on.


liyunfan1223:
Thanks for your nice work!

staceyhanson62:
I don't get how this is supposed to work. I created a group, and we completed The Forge of Souls. All party members completed the quest. I entered the Pit of Saron - the bots did not. I summoned them, but the wont come in. When I check the dungeon finder, it says I'm eligible, but the bots are not. I tried everything I can think of, GM Commands etc.. But I can't figure it out.

Thanks

PR #236, opened 2024-05-25
 Add performance tuning

brian8544:
We spoke about it here (https://github.com/liyunfan1223/mod-playerbots/discussions/191), a long time ago. Today I tried implementing this myself.

The original implementation of which I used the references can be found here: https://github.com/celguar/mangosbot-bots/commit/ce4bd79d2dbc7e8453303f9f94c27761615c64a9

I have implemented the changes as described in this pull request. So far, everything seems to be running fine on my end. However, I would greatly appreciate it if @liyunfan1223 & others could test it out to ensure that everything works fine across different environments and use cases. The same goes for coding standards. Perhaps it can be used as a base / reference for a future proper implementation.

Thanks!

-Brian8544

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -449,6 +449,10 @@ AiPlayerbot.BotCheats = "taxi"
 # Enables/Disables password to bot account
 AiPlayerbot.RandomBotRandomPassword = 0
 
+# Diff with/without player in server. The server will tune bot activity to reach the desired server tick speed (in ms).
+# AiPlayerbot.DiffWithPlayer = 100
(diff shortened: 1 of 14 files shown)


liyunfan1223:
LGTM. So far it seems to only affect battlefield joining, with a slight impact on performance. But as you said, it can be a good base for future implementation.

Additional request, could you put a switch in the config to control this feature (perhaps preset to off) to avoid some of the issues introduced by this feature until we fully implement and test it?

brian8544:
A configuration option for this prototyping is indeed a very good idea! I'll add this & do some more testing. Thank you for your input! It seems I did make some minor mistakes, such as accounting for the battleground queue.

PR #243, opened 2024-05-27
Add performance tuning

brian8544:
((Some changes were not applied in the previous PR, I created a new PR, just to make sure everything is alright and clean this time.))

We spoke about it here (https://github.com/liyunfan1223/mod-playerbots/discussions/191), a long time ago. Today I tried implementing this myself.

The original implementation of which I used the references can be found here: https://github.com/celguar/mangosbot-bots/commit/ce4bd79d2dbc7e8453303f9f94c27761615c64a9

I have implemented the changes + on/off config switch as you asked in this new pull request. So far, everything seems to be running fine on my end, with all modified systems (such as battlegrounds) still working as it should.

However, I would greatly appreciate it if @liyunfan1223 & others could test it out to ensure that everything works fine across different environments and use cases. The same goes for coding standards. Perhaps it can be used as a base / reference for a future proper implementation.

Thanks!

-Brian8544

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -412,7 +412,7 @@ AiPlayerbot.MinRandomBotReviveTime = 60
 AiPlayerbot.MaxRandomBotReviveTime = 300
 AiPlayerbot.MinRandomBotTeleportInterval = 3600
 AiPlayerbot.MaxRandomBotTeleportInterval = 18000
-AiPlayerbot.RandomBotInWorldWithRotaionDisabled = 31104000
+AiPlayerbot.RandomBotInWorldWithRotationDisabled = 31104000
 
 # How far random bots are teleported after death
 AiPlayerbot.RandomBotTeleportDistance = 100
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -449,6 +449,11 @@ AiPlayerbot.BotCheats = "taxi"
 # Enables/Disables password to bot account
 AiPlayerbot.RandomBotRandomPassword = 0
 
+# Diff with/without player in server. The server will tune bot activity to reach the desired server tick speed (in ms).
+# AiPlayerbot.EnablePrototypePerformanceDiff = 0
+# AiPlayerbot.DiffWithPlayer = 100
+# AiPlayerbot.DiffEmpty = 200
+
 ##################################################################################
 #                                                                                #
 # Database Stuff 																 #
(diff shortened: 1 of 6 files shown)


brian8544:
Last time you've told me it only works for battlegrounds. You're right, it missed a crucial call. Check if this has done it the right way:

https://github.com/brian8544/mod-playerbots/commit/43cd01642a008622cfe7f1c0d660c140231e25cc

fuzzdeveloper:
@brian8544 hey I was just looking through these changes (whenever I update my server I like to have a quick look over what's changed) and I noticed that:
- that the code you've added to BGJoinAction::shouldJoinBg() is unreachable (both code blocks are immediately after a return statement)
- the noLag variable added to FreeBGJoinAction::shouldJoinBg() is unused

Those changes wont introduce problems but won't function as you were likely intending, I do however notice you've changed one of the if statements in BGJoinAction::shouldJoinBg() from:
if (!(hasPlayers || hasBots))
to:
if (!hasPlayers && !(isArena))

That will have an impact (one I'm not sure was intentional) and will have one regardless of the enablePrototypePerformanceDiff setting (it also renders the hasBots variable unused which is partially why I'm wondering if you intended to do this).

All changes mentioned can be found in BattleGroundJoinAction.cpp here:
https://github.com/liyunfan1223/mod-playerbots/pull/243/commits/3879b5116e344469b33d60e05bb7ad05a27cfe7f#diff-4009d1913f93054e322a12b0dadcfb719e16186e4a73d61bf5e2bcd969320653

brian8544:
> @brian8544 hey I was just looking through these changes (whenever I update my server I like to have a quick look over what's changed) and I noticed that:
> 
> - that the code you've added to BGJoinAction::shouldJoinBg() is unreachable (both code blocks are immediately after a return statement)
> 
> - the noLag variable added to FreeBGJoinAction::shouldJoinBg() is unused
> 
> 
> 
> Those changes wont introduce problems but won't function as you were likely intending, I do however notice you've changed one of the if statements in BGJoinAction::shouldJoinBg() from:
> 
> if (!(hasPlayers || hasBots))
> 
> to:
> 
> if (!hasPlayers && !(isArena))
> 
> 
> 
> That will have an impact (one I'm not sure was intentional) and will have one regardless of the enablePrototypePerformanceDiff setting (it also renders the hasBots variable unused which is partially why I'm wondering if you intended to do this).
> 
> 
> 
> All changes mentioned can be found in BattleGroundJoinAction.cpp here:
> 
> https://github.com/liyunfan1223/mod-playerbots/pull/243/commits/3879b5116e344469b33d60e05bb7ad05a27cfe7f#diff-4009d1913f93054e322a12b0dadcfb719e16186e4a73d61bf5e2bcd969320653

Oops, huge oversight 🙈.. Still learning. Can you create a PR to 'fix' this the proper way?

PR #245, opened 2024-05-28
Attack target

liyunfan1223:


src/strategy/values/TargetValue.cpp
```diff
@@ -100,6 +100,24 @@ void FindTargetStrategy::GetPlayerCount(Unit* creature, uint32* tankCount, uint3
     dpsCountCache[creature] = *dpsCount;
 }
 
+bool FindTargetStrategy::IsHighPriority(Unit* attacker)
+{
+    if (Group* group = botAI->GetBot()->GetGroup())
+    {
+        ObjectGuid guid = group->GetTargetIcon(4);
+        if (guid && attacker->GetGUID() == guid) {
```
yuan227301, line 108: @liyunfan1223 
I have some doubts about this part of the modification. The original logic is that rti4 represents the target not to be attacked, and rti7 represents the target to be attacked first (defined in RtiTargetValue). However, your modification here also changes rti4 to the target to be attacked first. Is there any reason for this?

src/strategy/values/AttackersValue.cpp
```diff
@@ -23,24 +23,25 @@ GuidVector AttackersValue::Calculate()
...
     GuidVector prioritizedTargets = AI_VALUE(GuidVector, "prioritized targets");
     for (ObjectGuid target : prioritizedTargets) {
         Unit* unit = botAI->GetUnit(target);
-        if (unit) {
+        if (unit && IsValidTarget(unit, bot)) {
             targets.insert(unit);
         }
     }
     if (Group* group = bot->GetGroup()) {
         ObjectGuid skullGuid = group->GetTargetIcon(4);
```
yuan227301, line 38: @liyunfan1223 
Also, I think “skullTarget” should be rti7, not rti4 (moonTarget). Is it a mistake?

src/strategy/values/InvalidTargetValue.cpp
```diff
@@ -27,6 +27,7 @@ bool InvalidTargetValue::Calculate()
                target->isFeared() ||
                target->HasUnitState(UNIT_STATE_ISOLATED) ||
                target->IsFriendlyTo(bot) ||
+               !AttackersValue::IsValidTarget(target, bot) ||
```
yuan227301, line 30: @liyunfan1223 
This modification will cause the bots no longer attack some units with rti7 (Such as the Bone Spike of Lord Marrowgar and the Kor'kron Battle-Mage on the battle Gunship), Although we can attack them through attach commands. Is there any way to optimize it?

PR #256, opened 2024-06-02
Add performance tuning in the overworld

brian8544:


liyunfan1223:
Hello @brian8544. Is this modification completed?

brian8544:
> Hello @brian8544. Is this modification completed?

I do _not_ know if this implementation was done correctly, so far it's all done (afaik), but **_please_** test & verify my code before pushing to main :)

liyunfan1223:
Thank you for your reply. I'm thinking of implementing something similar, maybe in a different way.

brian8544:
> Thank you for your reply. I'm thinking of implementing something similar, maybe in a different way.

You really should, it would improve performance around the player a lot :D

Issue #271, opened 2024-06-12
Issues with gems added to randomised bot gear


fuzzdeveloper:
Note: fix for this is in [PR272](https://github.com/liyunfan1223/mod-playerbots/pull/272)

**Describe the bug**
I noticed some bots in my raid had lower HP than they should (for level 70 prot warrior anyway). Upon inspecting their gear, I noticed none had blue sockets populated, I've also observed them having inappropriate (usually red) gems in the meta socket, and that sometimes sockets will be missing altogether (as far as inspection view ingame shows).

This happens with **AiPlayerbot.LimitEnchantExpansion = 0** and after re-rolling bot gear with **.playerbot bot init=99999** (so gear limitation should not be a factor).

**Commit hash**
7ffb188a984c5b95deaa8716bb239c0036b3fb22

**To Reproduce**
1. Create new character under admin account
2. .levelup 69 (to make character level 70)
3. .playerbots bot addclass warrior (can be any class)
4. /w [botname] summon
5. .playerbot bot init=99999
6. Inspect bot
7. May need to repeat 5+6 repeatedly until bot has gear with sockets to reproduce the issue.

**Expected behavior**
Bots should have all gem sockets appropriately populated, and gear should not be missing sockets.

**Screenshots**
![socket-bug-1](https://github.com/liyunfan1223/mod-playerbots/assets/50162646/8d977207-b9a9-4f71-b636-c72ad56ea203)
![socket-bug-2](https://github.com/liyunfan1223/mod-playerbots/assets/50162646/cb09b81d-5fc5-4938-872d-3c1b7c47158b)

**Desktop (please complete the following information):**
 - OS: Windows 10 ltsc 2019

**Additional context**
The cause of this bug is found on line 3470 of PlayerbotFactory.cpp - socketColor variable has a value of 1, 2, 4 or 8, yet it is used directly for accessing item in bestGemEnchantId[] array (where value of 0-3 is expected). This means meta socket gets red gem, red socket gets yellow gem, yellow/blue socket get array index out of bounds (or would in another language... It's quite fortunate this bug isn't causing larger issues than it is).

liyunfan1223:
Fix merged, thanks!

PR #272, opened 2024-06-12
fix for #271 - Issues with gems added to randomised bot gear

fuzzdeveloper:
#271 fix

socketColor variable has value of 1, 2, 4, 8, but was being used directly for accessing index of bestGemEnchantId[] array (causing wrong gems in meta and red sockets, no gems in blue sockets, and missing yellow socket).

Tested fix and it works as intended, has significant impact on quality of bot gear too (I notice level 70 prot warriors have gone from about 17K to 21K HP buffed - probably wont make them work in the harder TBC raids but its a start).


Dreathean:
Nice find!

liyunfan1223:
My mistakes, thanks!

PR #306, opened 2024-07-06
More Admin Power + Bots in LFG

BenWagner-UX:
* CHANGE: As an admin you are now able to initialize bots without being the master.
* FIX: Bot can now join LFG - The missing activation was introduced

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -352,10 +352,17 @@ void PlayerbotAI::UpdateAIInternal([[maybe_unused]] uint32 elapsed, bool minimal
         Player* owner = holder.GetOwner();
         if (!helper.ParseChatCommand(command, owner) && holder.GetType() == CHAT_MSG_WHISPER)
         {
-            std::ostringstream out;
-            out << "Unknown command " << command;
-            TellMaster(out);
-            helper.ParseChatCommand("help");
+            // To prevent spam caused by WIM
+            if (!(command.rfind("WIM", 0) == 0) &&
+                !(command.rfind("QHpr", 0) == 0)
+                )
+            {
+                std::ostringstream out;
+                out << "Unknown command " << command;
+                TellMaster(out);
+                helper.ParseChatCommand("help");
+            }
+          
         }
 
         chatCommands.pop();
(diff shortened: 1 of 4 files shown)


liyunfan1223:
Thank you for your efforts! The fix for join LFG is very good. But I want to discuss this change. I think allowing gm players to use the init command on any bot might not be an appropriate change. This is dangerous in my opinion, as it can inadvertently erase characters built by human players, such as in solo/friend servers where some characters are built by the players themselves despite having GM permissions.

To prevent such errors, I introduced the `maintenance` and `autogear` commands to distinguish them from the `init` command. In fact, it has a similar effect but is less likely to go wrong. What are your thoughts on this?

BenWagner-UX:
> Thank you for your efforts! The fix for join LFG is very good. But I want to discuss this change. I think allowing gm players to use the init command on any bot might not be an appropriate change. This is dangerous in my opinion, as it can inadvertently erase characters built by human players, such as in solo/friend servers where some characters are built by the players themselves despite having GM permissions.
> 
> To prevent such errors, I introduced the `maintenance` and `autogear` commands to distinguish them from the `init` command. In fact, it has a similar effect but is less likely to go wrong. What are your thoughts on this?

Heya and thank you for YOUR effort :) I wasn't aware of those commands to be honest and for some reason they aren't working for me ingame. No matter what I do, maintenance is always telling me that the character can't be found. Sometimes its replicating the commands I've added after "maintenance".
image

Dreathean:
for maintenance and autogear you just whisper (or say in party/raid chat for multiple bots) "maintenance" and "autogear", nothing else included.

liyunfan1223:
Thanks for the finding. Can you provide more information about the placement of the database table causing the crash? https://github.com/liyunfan1223/mod-playerbots/pull/306/commits/0f511d723cc6f8571811ef1cbf10d16aa6034804

BenWagner-UX:
> Thanks for the finding. Can you provide more information about the placement of the database table causing the crash? [0f511d7](https://github.com/liyunfan1223/mod-playerbots/commit/0f511d723cc6f8571811ef1cbf10d16aa6034804)

remove the playerbot_names table from the acore_characters database and create a bot ingame by using "addclass".

btw. I need to push into a different branch, this changes are not belonging to this PR - my bad :(

liyunfan1223:
Hello @BenWagner-UX, would you like to initiate a new PR containing only the LFG? We can merge it first.

PR #315, opened 2024-07-09
Conditions for Summon

noisiver:
Closes https://github.com/liyunfan1223/mod-playerbots/issues/312

Credit to @conFIGured65 for several suggestions, extensive testing, and more.

If AiPlayerbot.BotReviveWhenSummon is set to 0 or 1 the following requirements must be met:
- The master isn't dead
- The master isn't in combat
- The bot isn't dead or the bot is a ghost

We added error messages for all conditions to let the player know exactly why the bot(s) can't be summoned.

### DIFF
--- a/src/strategy/actions/UseMeetingStoneAction.cpp
+++ b/src/strategy/actions/UseMeetingStoneAction.cpp
@@ -176,8 +176,28 @@ bool SummonAction::Teleport(Player* summoner, Player* player)
 
             if (summoner->IsWithinLOS(x, y, z))
             {
-                bool allowed = sPlayerbotAIConfig->botReviveWhenSummon == 2 || (sPlayerbotAIConfig->botReviveWhenSummon == 1 && !master->IsInCombat() && master->IsAlive());
-                if (allowed && bot->isDead())
+                if (sPlayerbotAIConfig->botReviveWhenSummon < 2)
+                {
+                    if (master->IsInCombat())
+                    {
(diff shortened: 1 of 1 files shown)


liyunfan1223:
Should we add an additional option to control whether the summon command can be used in combat, such as "AllowSummonInCombat", and have "BotReviveWhenSummon" only control whether the bot is revived? This can make it clearer. After this PR, player cannot summon bots that are still alive in combat when BotReviveWhenSummon is set to 1, which is a bit ambiguous. @noisiver @Dreathean

conFIGured65:
# Enable/Disable bot revive when summon (0 = never, 1 = enable when non-combat and alive, 2 = enable always)

Isnt Option 1 dedicated to non combat though? If they are "in combat" whether bots alive but not there, still in combat as option suggests?
Option 2 would be they can summon whenever. in combat, out of combat, bots dead or alive.

Option 3 could be whenever but only alive bots? Would that allow them to still exploit like 2, to where if bot is dead, in combat, remove bot from group, add another bot and summon live bot? So in turn could be the same as 2?

Dreathean:
@liyunfan1223 You're right, I see the confusion. Because the config is specifically labeled "Bot revive when summon", it makes it seem like this config only impacts the revive function while summoning, and not the ability to summon in or out of combat period, which it does. I agree it would make it much clearer to have those be separated out, good idea!

noisiver:
@liyunfan1223 I tend to agree, actually. It should be amended. I can get on that this weekend.

PR #318, opened 2024-07-10
BG Pathing and AV fixes

fuzzdeveloper:
Big improvements for AV (and some for AB as well):
- New path selection code for non-WSG BG's, now prefers paths close to bot over paths close to destination (this is main reason why bots used to fly so much in AV, they don't seem to do that much anymore even with MM::MovePoint() in use)
- Added additional waypoints to vPath_AV_Horde_Cave_to_Tower_Point_Crossroad (main path used to exit cave, this is the other reason why horde used to fly, or more recently, get stuck near cave).
- Fixed wrong object targeted for most objective based targets (fixes all kinds of issues where bot waits next to wrong objective eg: already captured flag)
- Fixed wrong creature being targeted for captain and bosses (fix bot running back and forth near captain, fix bot running all the way back home instead of killing boss)
- Fixed wrong flag state being checked for some objectives (fixes bot staying in burning tower forever)
- Fix bot not moving close enough to flag in some towers, waiting forever instead
- Option to disable MoveSplinePath in BG or everywhere and use MovePoint instead (MoveSplinePath uses EscortMovementGenerator which does not allow bots to be stunned/snared/rooted/etc)
- Some minor arena fixes

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -248,6 +248,13 @@ AiPlayerbot.GlobalCooldown = 500
 # Max wait time when moving
 AiPlayerbot.MaxWaitForMove = 5000
 
+# Disables use of MoveSplinePath for bot movement, will result in more erratic bot movement but means stun/snare/root/etc
+# will work on bots (they wont reliably work when MoveSplinePath is enabled, though slowing effects still work ok)
+# Default: 0 - MoveSplinePath enabled
+#          1 - MoveSplinePath disabled in BG/Arena only
+#          2 - MoveSplinePath disabled everywhere
+AiPlayerbot.DisableMoveSplinePath = 0
+
 # Max search time for movement (higher for better movement on slopes)
 # default: 3
 AiPlayerbot.MaxMovementSearchTime = 3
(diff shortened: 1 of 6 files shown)


fuzzdeveloper:
Alternative patch to core to fix stun/snare/root/etc issue:
EscortMovementGenerator-fix

Perhaps even better would be to make custom MovementGenerator for the bots (would require MotionMaster::Mutate() to be public though, maybe other changes needed too).

Dreathean:
Awesome fixes! Very excited to try it out!

liyunfan1223:
Thank you @fuzzdeveloper! Your modifications to the battleground address many long-term issues and significantly improve the quality of the battleground.

Regarding movement issue, there still some problems to be resolved. Using `mm.MovePoint` may cause the bot go through walls or go down slopes, which is why i introduced a combination of `SearchForBestPath` and `mm.MoveSplinePath`. But my solution also causes the stun issue. Neither implementation is perfect so far, so i agree that we need a custom MovementGenerator, but requires more considerations and trials. Make it configurable on battleground is a good temporary solution.

fuzzdeveloper:
@liyunfan1223 I agree, and I know the problems you were addressing with  SearchForBestPath and mm.MoveSplinePath, I should mention that those problems (with mm.MovePoint) are much less significant when the paths and waypoints are good, and when the paths chosen properly (as they are in AB/AV now).

The flying/going through walls with mm.MovePoint, and the not going to waypoint at all with SearchForBestPath are symptoms of the same problem - bad waypoint (thats in air, or in the ground), or waypoint too far away - MovePoint just makes the problem a lot more obvious (but that did make it easier for me to fix things).

liyunfan1223:
Yes, you are absolutely right :). Bad waypoint is the real problem to be solved.

PR #320, opened 2024-07-10
SummonAction::Teleport Repairwhensummon changed to a .conf option

Chromie-WoW:
Removed repairwhensummon from everytime bot summon and added it as a .conf option giving server owners the choice to allow players to repair or not repair when summoning their bots.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -132,7 +132,7 @@ AiPlayerbot.GroupInvitationPermission = 1
 AiPlayerbot.BotReviveWhenSummon = 1
 
 # Enable/Disable bot repair gear when summon (0 = no, 1 = yes)
-# default: 1 (enable for non-combat)
+# default: 1
 AiPlayerbot.BotRepairWhenSummon = 1
 
 # Non-GM player can only use init=auto to initialize bots based on their own level and gear score
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -127,9 +127,13 @@ AiPlayerbot.EquipmentPersistenceLevel = 80
 # default: 1 (accept based on level)
 AiPlayerbot.GroupInvitationPermission = 1
 
-# Enable/Disable bot revive and repair gear when summon (0 = never, 1 = enable when non-combat and alive, 2 = enable always)
+# Enable/Disable bot revive when summon (0 = never, 1 = enable when non-combat and alive, 2 = enable always)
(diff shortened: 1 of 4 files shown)


src/PlayerbotAIConfig.cpp
```diff
@@ -273,6 +273,7 @@ bool PlayerbotAIConfig::Initialize()
     equipmentPersistenceLevel = sConfigMgr->GetOption<int32>("AiPlayerbot.EquipmentPersistenceLevel", 80);
     groupInvitationPermission = sConfigMgr->GetOption<int32>("AiPlayerbot.GroupInvitationPermission", 1);
     botReviveWhenSummon = sConfigMgr->GetOption<int>("AiPlayerbot.BotReviveWhenSummon", 1);
+    botRepairWhenSummon = sConfigMgr->GetOption<bool>("AiPlayerbot.BotRepairWhenSummon", false);
```
liyunfan1223: Default to be true

conf/playerbots.conf.dist
```diff
@@ -127,10 +127,14 @@ AiPlayerbot.EquipmentPersistenceLevel = 80
 # default: 1 (accept based on level)
 AiPlayerbot.GroupInvitationPermission = 1
 
-# Enable/Disable bot revive and repair gear when summon (0 = never, 1 = enable when non-combat and alive, 2 = enable always)
+# Enable/Disable bot revive when summon (0 = never, 1 = enable when non-combat and alive, 2 = enable always)
 # default: 1 (enable for non-combat)
 AiPlayerbot.BotReviveWhenSummon = 1
 
+# Enable/Disable bot repair gear when summon (0 = no, 1 = yes)
+# default: 1 (enable for non-combat)
```
liyunfan1223: remove `for non-combat`

liyunfan1223 (changes requested):
Thank you! Just a few minor fixes.

src/PlayerbotAIConfig.cpp
Chromie-WoW: If I set default to true, will that make .conf option#1 (on) set to false?

conf/playerbots.conf.dist
Chromie-WoW: removed (enable for non-combat)

src/PlayerbotAIConfig.cpp
Chromie-WoW: changed default to true

liyunfan1223 (approved):

PR #326, opened 2024-07-12
Summon cleanup

noisiver:
Three pull requests to do something so simple, it just goes to show that you shouldn't rush things - especially not when you're tired. Hopefully this is the last one for these options.

Changed the existing `BotReviveWhenSummon` to a bool. The other options take care of everything else so there's no need to add additional checks to this one. It's enabled by default.
I accidentally forgot to add the `isDead()` check before reviving the bots. I don't know if could cause crashes but I needed to add this back as soon as possible.


noisiver:
I'll give this some testing to make sure there are no issues with it and then merge it I suppose. I don't want to merge it until I know for sure it's ok though.

PR #332, opened 2024-07-12
changed: more rights for admins

BenWagner-UX:
Includes:
* Can player online as bots
* Can init these bots

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -538,8 +538,11 @@ std::string const PlayerbotHolder::ProcessBotCommand(std::string const cmd, Obje
         if (ObjectAccessor::FindPlayer(guid))
             return "player already logged in";
 
-        if (!sPlayerbotAIConfig->allowPlayerBots && !isRandomAccount && !isMasterAccount)
-            return "You cannot login another player's character as bot.";
+        if (!admin)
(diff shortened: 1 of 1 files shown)


liyunfan1223 (changes requested):
As we discussed in https://github.com/liyunfan1223/mod-playerbots/pull/306. I insist that allowing an init command on a player-created character is not a good idea. 

Imagine that you group with 2 bots in your account and 2 bots summoned by the addclass command for a dungeon fight, and then use ".playerbot bot init *" to initialize the 2 summoned bots (keeping bots in your account unchanged). This used to be work, but now it also irreversibly initiallizes 2 bots in your account and that's not what you want.

BenWagner-UX:
> As we discussed in #306. I insist that allowing an init command on a player-created character is not a good idea.
> 
> Imagine that you group with 2 bots in your account and 2 bots summoned by the addclass command for a dungeon fight, and then use ".playerbot bot init *" to initialize the 2 summoned bots (keeping bots in your account unchanged). This used to be work, but now it also irreversibly initiallizes 2 bots in your account and that's not what you want.

Do we find a constent by making it configurable? Otherwise, decline the PR :)

liyunfan1223:
It's ok if you make it configurable.

PR #333, opened 2024-07-12
fixed: Warcraft Instant Messenger is causing massive spam by sending …

BenWagner-UX:
…bots WIM-specific commands such as handshakes which can not be processed by the bot.


liyunfan1223:
Thank you! Making it configurable might be a good future work.

PR #348, opened 2024-07-16
Fix bot PvP bug

Jgoodwin64:
Added logic to check PvP flag. Made considerations for duels and PvE.  Fuz1111 tested it on PvP server with no issues and I tested it on PVE with no issues. I will continue to test on PvE too.


Jgoodwin64:
There may be a better way. Using ChooseTargetActions.cpp

Jgoodwin64:
I think we should move the check into ChooseTargetActions.cpp, So that the bot does not repeatedly keep checking if it should attack and say yes call the attack strategy then fail.

Jgoodwin64:
I'm developing this further. Better testing.

PR #352, opened 2024-07-17
Fixed pvp flag issue better than in #348

Jgoodwin64:
I've updated ChooseTargetAction.cpp to address an issue where the bots were not checking the PvP flag before engaging in combat. Added a function for PvP target validation.

Implemented bool IsValidPvPTarget(Player* bot, Unit* target) to ensure the target is flagged for PvP. 

The PR Fix bot PvP bug #348 worked too, but this approach also allows for adding more criteria to refine what an acceptable PvP target is. 
So, I canceled the other one. I've tested on a PVE server and everything works well.

### DIFF
--- a/src/strategy/actions/ChooseTargetActions.cpp
+++ b/src/strategy/actions/ChooseTargetActions.cpp
@@ -24,7 +24,7 @@ bool IsValidPvpTarget(Player* bot, Unit* target)
     }
     // If the target is not a player, it's always valid
     return true;
-
+}
 bool AttackEnemyPlayerAction::isUseful()
 {
     // if carry flag, do not start fight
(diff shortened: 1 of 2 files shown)


Jgoodwin64:
Side note bots wait to duel now for the 3 2 1.

Jgoodwin64:
@fuzzdeveloper on discord.
"if you want to reduce the duplicate duel checks i think this will work
`bool IsValidPvpTarget(Player* bot, Unit* target)
{
    if (Player* player = target->ToPlayer())
    {
        if (!bot->duel)
            return player->IsPvP();
        else
            return bot->duel->Opponent == player && time(nullptr) >= bot->duel->StartTime;
    }
    return true;
}`"

Jgoodwin64:
I'm looking in  AttackEnemyPlayersStrategy.h and AttackEnemyPlayersStrategy.cpp to see if the fix should be placed in there. :man_facepalming:

PR #354, opened 2024-07-17
Fix crash caused by summon to instance and avoid aoe

liyunfan1223:
https://github.com/liyunfan1223/mod-playerbots/issues/321

close https://github.com/liyunfan1223/mod-playerbots/issues/239

Slightly increase the performance overhead for avoid aoe.


liyunfan1223:
A summon invalid issue has been introduced here. Working in progress on it.

PR #360, opened 2024-07-21
Frost mage implementation

Bobblybook:
Hi,
Wondering if you can offer some advice on this.
I've worked on this yesterday, frost mage is now 90% working.

Procs are handled, rotation is handled, pet works.

Cold snap logic is not yet in, but the only issue I'm needing to fix is the deep freeze trigger is not working as intended.

I'm wanting to implement a 2-spell combo such as:

"If trigger, then cast frostbolt followed by deep freeze straight after".

The trigger is successfully being called, but I don't see this behaviour and deep freeze never gets cast.
I've verified the char has the skill learned.

The way I've implemented it is using a trigger with the continuer skill, set to deep freeze. Unsure if this is how I should be doing this, it may not work the way I think it does. Do you have any recommendations on how to set this up?

### DIFF
--- a/src/strategy/mage/FrostMageStrategy.cpp
+++ b/src/strategy/mage/FrostMageStrategy.cpp
@@ -10,6 +10,9 @@ class FrostMageStrategyActionNodeFactory : public NamedObjectFactory<ActionNode>
 public:
     FrostMageStrategyActionNodeFactory()
     {
+        creators["cold snap"] = &cold_snap;
+        creators["ice barrier"] = &ice_barrier;
+        creators["summon water elemental"] = &summon_water_elemental;
         creators["deep freeze"] = &deep_freeze;
         creators["frostbolt and deep freeze"] = &frostbolt_and_deep_freeze;
     }
--- a/src/strategy/mage/FrostMageStrategy.cpp
+++ b/src/strategy/mage/FrostMageStrategy.cpp
@@ -72,7 +72,7 @@ void FrostMageStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 
     triggers.push_back(new TriggerNode("no pet", NextAction::array(0, new NextAction("summon water elemental", ACTION_HIGH), nullptr)));
     triggers.push_back(new TriggerNode("has pet", NextAction::array(0, new NextAction("toggle pet spell", ACTION_HIGH + 1), nullptr)));
-    triggers.push_back(new TriggerNode("ice barrier", NextAction::array(0, new NextAction("ice barrier", ACTION_HIGH), nullptr)));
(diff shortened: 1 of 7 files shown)


liyunfan1223:
As far as I know, the implementation of 2-spell combo you wrote works. Maybe it's just that frostbolt consumes fingers of frost aura, thus making deep freeze unable to cast? You can check the logs to find out why, e.g. `grep "deep freeze" Playerbots.log`, there are probably logs about the reason for the failure. BTW, why did you make this spell combo haha, I'm not familiar with frost mage, does that make a higher dps?

Bobblybook:
> As far as I know, the implementation of 2-spell combo you wrote works. Maybe it's just that frostbolt consumes fingers of frost aura, thus making deep freeze unable to cast? You can check the logs to find out why, e.g. `grep "{spellID of deep freeze}" Playerbots.log`, there are probably logs about the reason for the failure. BTW, why did you make this spell combo haha, I'm not familiar with frost mage, does that make a higher dps?

Ok thanks - I was checking the logs and I couldn't actually see it being cast. There is reference to the first frostbolt with weight 20, but nothing comes after.
I'll take another look today and maybe swap it out for fire blast temporarily to see where it's going wrong.

Yes, it's a core part of playing frost mage - you get two stacks of FoF, but the stack doesn't get consumed until your spells hit, so for the final stack, you weave in a fast hitting spell before the second frostbolt lands (ice lance, or deep freeze if it's off cd). Both spells benefit from the buff so you can effectively use the last stack twice, you are basically getting 50% more fingers of frost stacks doing this.
It's a pretty core part of playing frost mage and it doesn't seem very difficult logic to implement so I figured I'd give it a go.

aronkleinhans:
There's this in the config

# Delay between two short-time spells cast
AiPlayerbot.GlobalCooldown = 500

Could be messing with the combo. I didn't check but I guess the delay triggers after the cast is done, there'd be no point to the delay anyway. 500ms is enough for the FB to land, and so it consumes the fof stack.

Bobblybook:
> There are these in the config
> 
> Delay between two short-time spells cast AiPlayerbot.GlobalCooldown = 500
> 
> Delay between two bot actions AiPlayerbot.ReactDelay = 100
> 
> Could be messing with the combo. I didn't check but I guess the delay triggers after the cast is done, there'd be no point to the delay anyway. 100-500ms could be enough for the FB to land, and so it consumes the fof stack.

That's fine, this shouldn't matter I don't think.

After much digging, I can't find any way to make the continuer action thing do anything at all. Not sure if it's actually implemented to be honest..
I realised that adding multiple actions to the array effectively queues them and seems to work perfectly.

See PR, I believe this is now done, works very well, this frost mage is now a bit of a beast!
Can be merged and closed if no other changes needed

PR #363, opened 2024-07-21
Fix: Bots attacking players that are not flagged for PvP

noisiver:
Credit to @Jgoodwin64 and @fuzzdeveloper  because most of it is based on their work along with some changes I've done while comparing how these functions are used in the core.

This should make it so bots will never attack players that are not flagged for PvP and not in a duel with the bot.

### DIFF
--- a/src/strategy/actions/AttackAction.cpp
+++ b/src/strategy/actions/AttackAction.cpp
@@ -95,6 +95,16 @@ bool AttackAction::Attack(Unit* target, bool with_pet /*true*/)
         return false;
     }
 
+    if (target->IsPlayer() && !target->IsPvP() && !target->IsFFAPvP() && (!bot->duel || bot->duel->Opponent != target || bot->duel->StartTime))
+    {
+        if (verbose)
+        {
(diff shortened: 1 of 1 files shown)


Jgoodwin64:
I tested it. I tried duels cross faction and same faction duel, I've tried telling horde bot to attack a nonflagged playerbot. I also tested running around as a horde in .go taxi 1 the human start area. Then I had a bot flag pvp and I told my bots to attack it and they did. I then tested if a flagged PvP bot could attack a non flagged bot. It can't. This fix is similar to the PR Fix bot PvP bug #348 that I closed due to trying to find the best spot for the fix, it is in the notes, but this code makes it where you can't order them to attack the nonflagged player. My first fix didn't I think. 

So, I think we should push this PR through as it is greatly need and has been tested a lot.

Jgoodwin64:
> Credit to @Jgoodwin64 because most of it is based on their work along with some changes I've done while comparing how these functions are used in the core.
> 
> This should make it so bots will never attack players that are not flagged for PvP and not in a duel with the bot.
"Their meaning mine and @fuzzdeveloper 's work.

PR #367, opened 2024-07-22
Fix .playerbot bot self

liyunfan1223:
This PR introduces some crucial changes, so I will keep it open for several days to monitor for any bugs that may arise. You are welcome to test it and provide feedback.

- Fixed `.playerbot bot self` command. Typing this command again to break. 
  - Closes https://github.com/liyunfan1223/mod-playerbots/issues/249.
- Fixed selling failure issue.
- Fixed a possible crash caused by invites.
  - https://github.com/liyunfan1223/mod-playerbots/issues/321#issuecomment-2235336394
- Fixed an addclass issue caused by decreasing RandomBotAccountCount.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -125,11 +125,11 @@ AiPlayerbot.BotAutologin = 0
 # Default: 0 (disabled)
 AiPlayerbot.AllowPlayerBots = 0
 
-# Bots will be summoned to player when accept group invitation
-AiPlayerbot.SummonWhenGroup = 1
-
 # Allow/deny bots from your guild
 AiPlayerbot.AllowGuildBots = 1
+
+# Bots will be summoned to player when accept group invitation
+AiPlayerbot.SummonWhenGroup = 1
 
 # Added following config
 # Selfbot permission level (0 = disabled, 1 = gm only (default), 2 = all players, 3 = activate on login)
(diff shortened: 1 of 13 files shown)


Jgoodwin64:
Tested and it works well, thanks for the effort. Still testing it. The command toggles well and I have not had the group with myself error.

Jgoodwin64:
I tested to see if I could still do " playerbot bot add [Name] " while being bot self-ed. If I did it correctly, the bug where you could not add bots or sometimes could not add bot, is gone. 
There is still an issue occasionally with .summon. Only sometimes, when you try .summon a player by targeting the party member, It will give an error. I also got an error once or twice saying that player is not online but it was clearly online.  This was all while bot self was on. It always turned off for me when I tried, and without any weirdness.

Jgoodwin64:
After further testing it seems the self bot doesn't want to walk by it's self anymore. Before, I would tell it stay and then I could take over walking now, I can't get it to walk to grind or do quests.

voarsh2:
> After further testing it seems the self bot doesn't want to walk by it's self anymore. Before, I would tell it stay and then I could take over walking now, I can't get it to walk to grind or do quests. Upon further, further testing. I see that when the targets are close enough the selfbot works fine. It is only getting to the area that is the issue.

Yh. It doesn't walk around when using selfbot

PR #369, opened 2024-07-22
Optimize the check for PvP flag

noisiver:
Moved the check for if players are flagged for PvP before attacking them. It's best if they never reach the attack action code if they won't be able to attack the target anyway. This looks to be the best solution.

Thank you @fuzzdeveloper for the help figuring this out.

I've done extensive tests in these areas:
- Walking around enemy bots both while being flagged for PvP and while not. Some of the bots were also flagged and some weren't to make sure it works in both cases.
- Being grouped up with bots and test the same as above as well as telling them to attack specific targets.
- Having the bots attack creatures to make sure that is unaffected.

I can't seem to test duels because even without any of this they won't actually attack me. They might take a swing at me but then they run off.

I'd appreciate if someone could give duels a whirl, someone who knows that duels worked for them before this.

Maybe @Jgoodwin64 could test that part?

If others feel like testing it all, that'd be great too. I figured I'd create the pull request so it could be seen and hopefully tested.

### DIFF
--- a/src/strategy/actions/AttackAction.cpp
+++ b/src/strategy/actions/AttackAction.cpp
@@ -95,16 +95,6 @@ bool AttackAction::Attack(Unit* target, bool with_pet /*true*/)
         return false;
     }
 
-    if (target->IsPlayer() && !target->IsPvP() && !target->IsFFAPvP() && (!bot->duel || bot->duel->Opponent != target || bot->duel->StartTime))
-    {
-        if (verbose)
-        {
-            botAI->TellError(Acore::StringFormat("%s is not flagged for pvp", target->GetName()));
-        }
-
-        return false;
-    }
-
     if (bot->IsMounted() && bot->IsWithinLOSInMap(target))
     {
         WorldPacket emptyPacket;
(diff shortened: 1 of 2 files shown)


fuzzdeveloper:
@noisiver regarding duels, both the code you've removed and the code you've added cause problems with duels on PVP server (I can get a duel going but I have to turn PVP flag on manually).

The cause of this is the StartTime part of the duel check:
`(!bot->duel || bot->duel->Opponent != target || bot->duel->StartTime)`
So basically you're preventing attack if not dueling, if duel opponent isn't the target, or if duel->StartTime is set at all (which it always will be when in a duel). Removing the StartTime condition fixes dueling (on PVP server anyway).

If the reason you had StartTime in there was because you were trying to avoid the bots starting duels early (I remember Orrytrout on discord was trying to achieve that), then you need to compare StartTime against GameTime::GetGameTime().count() and implement in a way that's totally separate to the PVP flag check (so that having PVP flag on doesn't bypass the time check). I also strongly suspect that the StartTime check would need to be implemented in AttackAction (targeting code may not get called often enough).

Aside from that, are we sure that implementing the PVP flag check in a single place is enough? I note that PlayerbotAIConfig::IsPvpProhibited() (which blocks PVP in certain areas, and therefore is doing something very similar) is checked in 3 places:
- AttackersValue::IsPossibleTarget() - which is where you've put the new check
- AttackEnemyPlayerAction::isUseful()
- NearestEnemyPlayersValue::AcceptUnit()

Orrytrout did mention issues with bots being able to attack non-flagged players by targetting flagged players then switching targets, if that's not happening then maybe doing the check in a single spot is enough.

noisiver:
Thank you for the comment about duels. Regarding switching from a flagged player to one that isn't flagged, I couldn't get the bot I used to test it to do that. I targeted a flagged bot and sent my bot in, and it did attack it, and then one that wasn't flagged and it didn't attack it.

PR #388, opened 2024-07-26
Apply Code Standard

noisiver:
Ran all the code through Visual Studio Code to apply the code standard from there but I replaced all tabs with spaces.

We have all sorts of styles here, and it should really follow the same layout to make it look cleaner and easier to follow.

If this is something you don't want then we can just leave it. I figured I'd give it a go at least.

I realize that if this is accepted it probably means I'll have to do this once in a while and I'm fine with that.


liyunfan1223:
Before we do that, it's a good idea to check what specific code style to use, and should it be aligned with azerothcore's code style. For developers using different platforms, code formatting tools that can be used across multiple platforms are also important.

liyunfan1223:
BTW, it is definitely good to set code style.

noisiver:
What do you reckon would be the best way to determine what it should look like? Should I compare it to parts of the AzerothCore code to structure the code in the same way? I know they have checks for codestyle but it's very minimal and doesn't cover much.

liyunfan1223:
I'm not actually sure what the best way is. Personally I would use clang-format as a formatting tool with google style (with the following .clang-format file).

```
BasedOnStyle: google
IndentWidth: 4
TabWidth: 4
ColumnLimit: 100
UseTab: Never
AllowShortFunctionsOnASingleLine: true
```
This kind of is a generic way to do it, or at least it can't go wrong.

noisiver:
Yeah, I'm trying `{ BasedOnStyle: Google, AllowShortFunctionsOnASingleLine: All, ColumnLimit: '100', TabWidth: '4', UseTab: Never }` but can't figure out how to make it use a new line before opening braces etc. Braces on new lines is what AzerothCore does so I figured we want that too.

noisiver:
I ended up using this: `{ BasedOnStyle: Google, AllowShortFunctionsOnASingleLine: All, AllowShortIfStatementsOnASingleLine: false, ColumnLimit: '100', IndentWidth: '4', UseTab: Never, BreakBeforeBraces: Allman }`
When it's done formatting all the files I'll commit it and call it a day. More tweaking can be done another day.

Jgoodwin64:
Thanks. I really think the two of you are really elevating this project to new levels.

noisiver:
I've been messing around, trying to make it look somewhat like AzerothCore but it's tough. This is what I used last and it looks a bit better if you ask me. Let me know if I should change something, I just compared AC to the results from clang-format to try and make it similar.

```
BasedOnStyle: Google
AllowShortFunctionsOnASingleLine: All
AllowShortIfStatementsOnASingleLine: false
AllowShortLoopsOnASingleLine: false
ColumnLimit: '250'
IndentWidth: '4'
UseTab: Never
BreakBeforeBraces: Allman
PointerAlignment: Left
```

PR #396, opened 2024-07-28
Add bots to transports

noisiver:
Add bots to transports they're on so they actually move with it. I've tested it with boats as well as airship and had no issues as a result of this change.

The bots sometimes have issues stepping onto airships because their pathing isn't perfect. As soon as they set foot on it and become a passenger that issue goes away.

Can be seen in action here: https://www.youtube.com/watch?v=WwfyZfSb6E4

### DIFF
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -849,6 +849,21 @@ void MovementAction::UpdateMovementState()
 
     if (bot->IsFlying())
         bot->UpdateSpeed(MOVE_FLIGHT, true);
+
+    Transport* newTransport = bot->GetMap()->GetTransportForPos(bot->GetPhaseMask(), bot->GetPositionX(), bot->GetPositionY(), bot->GetPositionZ(), bot);
+    if (newTransport != bot->GetTransport())
+    {
+        LOG_DEBUG("playerbots", "Bot {} is on a transport", IsMovingAllowed());
+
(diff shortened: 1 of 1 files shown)


noisiver:
I copy-pasted the debug message and silly me forgot to change the variable. Fixed now though.

liyunfan1223:
I am considering about whether there will be significant performance overhead, as frequent GetTransportForPos() may lead to this issue.

noisiver:
I have another idea and will get back to you with more about if it worked out or not.

noisiver:
Well, that was a waste of time. I can't reuse what the player does because it uses opcodes to determine if the player is on a transport. The piece of code I used actually came from `Creature::Update` so would it really be that heavy?

liyunfan1223:
I think it is indeed heavy, as `Creature::Update` uses a timer to ensure the function is triggered at most once per second. https://github.com/liyunfan1223/azerothcore-wotlk/blob/f7ad415b3ff9efdcf7418306ebeb907d352a9be6/src/server/game/Entities/Creature/Creature.cpp#L903C1-L928C2
We can use a similar method to increase the check interval and do some performance testing.

liyunfan1223:
The test result shows that there is no significant performance impact. It seems that I am overthinking it: D

noisiver:
I didn't even consider how much of a difference it would be using a timer. It's better safe than sorry so thanks a ton of testing it and making sure it wasn't horrible!

PR #406, opened 2024-07-28
Improve management and bot autojoining of battlegrounds/rated arena

CrypticAvocado:
Thanks to @fuzzdeveloper for helping out with getting bots playing rated arena.

I originally set out to just have bots in rated arena but found the use of factions for skirmish/rated not very intuitive. I also wanted the bots to auto play rated arena without players online. Sorry if this code sucks, I am still learning so appreciate feedback!

**Whats included** 
- Slightly better positioning based on bots classes in arena
- Fixed random bots not auto joining battlegrounds when the setting was enabled
- Ability to control how many bots auto join different rated arena types / warsong gulch
- Fixed arena teams being created but members not joining
- More control of arena team creation (pick how many 2s, 3s, 5s you want) + choose random rating range
- Fixed logging to be more accurate for battlegrounds/arena
- Battlegrounds/Arena handled as individual instances allowing for concise counts and real players across separate games with bots (also bots auto joining isn't impacted by real players playing)

Still see lots of room for improvement but my brains falling out so just need to open up the PR now for feedback.
I have tested a lot but will really appreciate it if someone else can try it out

**Testing**
If you set
AiPlayerbot.RandomBotAutoJoinBGWarsongCount to 1 in conf you should see bots fill up one warsong gulch instance.
If you play with the settings you should see the logs displayed like the below:
image

Rated arena team creation still has room for improvement, to get new teams if you already have bots - first make sure any existing teams are deleted then the bots need to be reinitialized.

### DIFF
--- a/src/RandomPlayerbotMgr.h
+++ b/src/RandomPlayerbotMgr.h
@@ -8,33 +8,33 @@
 #include "PlayerbotMgr.h"
 
 struct BattlegroundInfo {
-    std::vector<int> bgInstances;
-    std::vector<int> ratedArenaInstances;
-    std::vector<int> skirmishArenaInstances;
-    int bgInstanceCount = 0;
-    int ratedArenaInstanceCount = 0;
-    int skirmishArenaInstanceCount = 0;
-    int minLevel = 0;
-    int maxLevel = 0;
-    int activeRatedArenaQueue = 0; // 0 = Inactive, 1 = Active
-    int activeSkirmishArenaQueue = 0; // 0 = Inactive, 1 = Active
-    int activeBgQueue = 0; // 0 = Inactive, 1 = Active
+    std::vector<uint32> bgInstances;
+    std::vector<uint32> ratedArenaInstances;
+    std::vector<uint32> skirmishArenaInstances;
+    uint32 bgInstanceCount = 0;
+    uint32 ratedArenaInstanceCount = 0;
+    uint32 skirmishArenaInstanceCount = 0;
+    uint32 minLevel = 0;
+    uint32 maxLevel = 0;
+    uint32 activeRatedArenaQueue = 0; // 0 = Inactive, 1 = Active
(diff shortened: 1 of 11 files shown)


fuzzdeveloper:
Good to see this finally get a PR (I know you've been working on it a while), going to pull and give it a proper test later today

src/RandomPlayerbotMgr.h
fuzzdeveloper, line 1: Change instance vectors to uint32 rather than int (as that's the type used for the battleground instanceId's and many other id's in AC), same goes for BattlegroundData map (uint32 rather than int32_t), and also the instanceIds vector in RandomPlayerbotMgr::CheckBgQueue(). The latter being int triggers a compile time warning (which is what brought this to my attention).

You could probably go ahead and use uint32 for all the fields in the BattlegroundInfo struct too (not needed but no reason not to)

fuzzdeveloper (approved):
looks good

fuzzdeveloper:
@noisiver hey would you have any idea why macos build (and only macos build) of this is failing (when he hasn't touched classes in question or anything to do with SpellInfo)?

I ask because you seemed to be fixing a similar issue with [this commit](https://github.com/liyunfan1223/mod-playerbots/commit/74658389ecb2a4843a60fc68d45800da2cfd02e1) though that's already been merged to his branch.

noisiver:
The error they're getting is from the core and not this module. I don't know why though seeing as the other CIs aren't failing.

liyunfan1223 (approved):
I tested it in some normal cases, looks good to me. I think there are still some arena team creation and maintenance tasks that need to be improved after merging, especially on some long-running servers.

liyunfan1223:
It may be caused by CI cache. I think it has nothing to do with this PR. We can merge it first and take a look.

fuzzdeveloper:
> It may be caused by CI cache. I think it has nothing to do with this PR. We can merge it first and take a look.
Yeah I thought it might be something like that too, here's hoping that's the case!

PR #437, opened 2024-08-05
Better waypoint for reach combat and cast instant spell during movement

liyunfan1223:
- close https://github.com/liyunfan1223/mod-playerbots/issues/412
- Make accurate waypoints for follow, reach combat and move to rpg target.
- Enable instant spell to cast during movement.


liyunfan1223:
Hello @fuzzdeveloper! I add a param `exact_waypoint` in MovementAction::MoveTo. If the destination is confirmed to be good, you can set this parameter to true to disable splinepath. This can be used in battleground. I have now used this approach for most of movements, such as follow, reach combat and move to rpg target.

PR #445, opened 2024-08-06
Fix oil issue

atidot3:
https://github.com/liyunfan1223/mod-playerbots/issues/428
https://github.com/liyunfan1223/mod-playerbots/issues/441


src/PlayerbotAI.cpp
```diff
@@ -4405,25 +4405,22 @@ Item* PlayerbotAI::FindOilFor(Item* weapon) const
     if (pProto && (pProto->SubClass == ITEM_SUBCLASS_WEAPON_SWORD || pProto->SubClass == ITEM_SUBCLASS_WEAPON_STAFF ||
                    pProto->SubClass == ITEM_SUBCLASS_WEAPON_DAGGER))
     {
-        for (uint8 i = 0; i < std::size(uPriorizedWizardOilIds); ++i)
+        for (const auto& id : uPriorizedWizardOilIds)
         {
-            oil = FindConsumable(uPriorizedWizardOilIds[i]);
+            oil = FindConsumable(uPriorizedWizardOilIds[id]);
```
liyunfan1223: This seems even worse. Imagine uPrioritizedWizardOilIds[MINOR_WIZARD_OIL] being equivalent to uPrioritizedWizardOilIds[9731].

atidot3: yea

fuzzdeveloper:
There's a bit of redundancy in this solution:
```
            oil = FindConsumable(id);
            if (!oil)
                oil = FindConsumable(id);
            if (oil)
                return oil;
```
That FindConsumable method is not going to work better the second time, besides, you're not reproducing the (intended) behavior of the original method (which, don't get me wrong, is written like hell), where it alternates through both arrays, in different order for different weapon types. This is how it _seems_ to work:
image

I've never played a class which uses this stuff (so I've no idea why you'd want to priortise oils this way) but if you want to replacate this behavior, what I would do, is just store all oils in both arrays, one in correct order for swords/staffs/daggers, the other for maces, like this:
```
static const uint32 uOilsWithWizardPriorty[] = {
    MINOR_WIZARD_OIL, MINOR_MANA_OIL,
    LESSER_WIZARD_OIL, LESSER_MANA_OIL,
    BRILLIANT_WIZARD_OIL, BRILLIANT_MANA_OIL,
    WIZARD_OIL, SUPERIOR_MANA_OIL,
    SUPERIOR_WIZARD_OIL,
};

static const uint32 uOilsWithManaPriorty[] = {
    MINOR_MANA_OIL, MINOR_WIZARD_OIL,
    LESSER_MANA_OIL, LESSER_WIZARD_OIL,
    BRILLIANT_MANA_OIL, BRILLIANT_WIZARD_OIL,
    SUPERIOR_MANA_OIL, WIZARD_OIL,
    SUPERIOR_WIZARD_OIL,
};
```
Now they are ready to be iterated through in a sane manner (that wont make the next guy who reads the code have to spend several minutes figuring it out):
```
for (const auto& id : uOilsWithWizardPriorty)
{
    if (Item* oil = FindConsumable(id))
        return oil;
}
```

That code seems to be in the original mangos bot too btw, I'm surprised it's causing problems.

fuzzdeveloper:
There's a bit of redundancy here:
```
oil = FindConsumable(id);
if (!oil)
    oil = FindConsumable(id);
if (oil)
    return oil;
```
You're just making same FindConsumable call twice, the original code, if you read it carefully (and you _have_ to read it carefully because its written like hell) alternated through both arrays, in different orders for different weapon types like this:
image

I've never played a class that uses this stuff so I've no idea why you'd want to priortise the oils this way, but if you want to reproduce this behavior, I would do so by just making an array that's already in the correct order for swords/staffs/daggers, and another array that's in correct order for maces, like this:
```
static const uint32 uOilsWithWizardPriorty[] = {
    MINOR_WIZARD_OIL, MINOR_MANA_OIL,
    LESSER_WIZARD_OIL, LESSER_MANA_OIL,
    BRILLIANT_WIZARD_OIL, BRILLIANT_MANA_OIL,
    WIZARD_OIL, SUPERIOR_MANA_OIL,
    SUPERIOR_WIZARD_OIL,
};

static const uint32 uOilsWithManaPriorty[] = {
    MINOR_MANA_OIL, MINOR_WIZARD_OIL,
    LESSER_MANA_OIL, LESSER_WIZARD_OIL,
    BRILLIANT_MANA_OIL, BRILLIANT_WIZARD_OIL,
    SUPERIOR_MANA_OIL, WIZARD_OIL,
    SUPERIOR_WIZARD_OIL,
};
```

There's nothing else using those arrays so you can just replace them with the above, once you've done that you can write sane code to iterate through the arrays, like this:
```
    for (const auto& id : uOilsWithWizardPriorty)
    {
        for (Item* o = FindConsumable(id))
            return o;
    }
```

That's the loop for swords/staves/daggers, the mades one will iterate through other array (you can also remove the oil variable entirely and just return nullptr at the end if nothing is found.

Btw I actually wrote this entire comment twice (if you noticed) because I re-read the original code and thought I'd got it wrong (thought it finished the loop and return best oil found in either array) but then I read it again, nope, was correct. That code actually dates back to mangos bot so I'm surprised its causing issues (as horrible as it was to visually parse).

atidot3:
Maybe, but good call !

atidot3:
image
image

liyunfan1223 (approved):

PR #446, opened 2024-08-06
Many IOC fixes

fuzzdeveloper:
Ok this isn't quite done yet but I'd like to get this in before there's any more big refactors/reformats (like I just had to deal with). Limitations are:
- Bots are pretty much stuck in vehicles once they get in them
  - This is a pre-existing limitation due to the way code to get in/out of vehicle is triggered (code to get in triggered on timer, code to get out triggered randomly, code to get-in happens far more frequently therefore bots that exit vehicle will just get straight back in)
  - I plan to move the functions for entering/exiting vehicles somewhere else (where the decision to enter/exit can be made with with some kind of intelligence and knowledge of the battleground state)
- Bots are prevented from getting into canons (hardcoded limitation to avoid them jumping into defensive cannons immediately at start of game then never leaving)
- Bots cannot use the airship at all (similar reasons to above, and no they cant use it to parachute into base either at this stage, though that would be pretty awesome)
- Bots are prevented from using catapult (they don't know how to use it at all, they just drive it around)
- Bots are prevented from using Glaive Thrower's Blade Salvo ability (because it doesn't work properly, and after they use it they can never use any other abilities)
- Strategies are kind've basic
  - They basically just alternate grabbing bases until the bots in vehicles get the gate down, then they go all-in on the boss
  - I want to do the change to vehicle entry/exit before trying to make it any more advanced (as I think the way it works is actually interfering with the bots completing BG objectives)
- It's worth mentioning that the BG itself has a few issues (that are nothing to do with the bots) map/flag's don't update properly (eg: sometimes flags revert to blank neutral appearance when contested), siege-engine spawns when workshop captured (but is destroyed instantly so it has the correct impact gameplay-wise, but is kind've janky)

Fixes:
- Fixed bots not being able to revive in their own base GY
- Fixed bots not able to use flags in at all (they couldn't cap anything)
- Added paths within bases so bots could reach bosses and several other paths too
- Fixed horde bots not being able to tell that alliance gate is down (had wrong object id and stayed outside forever)
- Fixed alliance bots trying to shoot at side gate from front of horde base (another object id issue)
- Fixed bots not able to get in vehicles
- Fixed bots not able to drive vehicles
- Fixed bots in vehicles becoming unresponsive after a fight
  - In-combat status seems to never clear when in an IOC vehicle, I've confirmed it happens to player as well (which may be AC bug though it doesn't really effect player much, definitely effects bot behavior though so added workarounds for this)
- Reduced bot 'agro distance' when in vehicle so they can get to their objective
- Redid IOC selectObjective strat completely (they would wait for ALL bases to be capped before trying for boss which would never actually happen before reinforcements run out (and its pointless to wait because the boss is pretty easy)
- Several other fixes I'm probably forgetting

### DIFF
--- a/src/strategy/actions/BattleGroundTactics.cpp
+++ b/src/strategy/actions/BattleGroundTactics.cpp
@@ -66,17 +66,21 @@ Position const EY_WAITING_POS_ALLIANCE = {2526.020f, 1596.787f, 1270.127f, 3.14f
 Position const EY_FLAG_RETURN_POS_RETREAT_HORDE = {1885.529f, 1532.157f, 1200.635f, 0.0f};
 Position const EY_FLAG_RETURN_POS_RETREAT_ALLIANCE = {2452.253f, 1602.356f, 1203.617f, 0.0f};
 
-Position const IC_WAITING_POS_HORDE = {1166.322f, -762.402f, 48.628f, 3.14f};
-Position const IC_WAITING_POS_ALLIANCE = {387.893f, -833.384f, 48.714f, 6.28f};
+Position const IC_WAITING_POS_HORDE = {1166.322f, -762.402f, 48.628f};
+Position const IC_WEST_WAITING_POS_HORDE = {1217.666f, -685.449f, 48.915f};
+Position const IC_EAST_WAITING_POS_HORDE = {1219.068f, -838.791f, 48.916f};
 
-Position const IC_WEST_WAITING_POS_HORDE = {1217.666f, -685.449f, 48.915f, 1.54f};
-Position const IC_EAST_WAITING_POS_HORDE = {1219.068f, -838.791f, 48.916f, 4.72f};
-Position const IC_SIDE_WAITING_POS_ALLIANCE = {351.517f, -882.477f, 48.916f, 4.68f};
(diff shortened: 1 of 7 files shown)


fuzzdeveloper:
@liyunfan1223 I've noticed you changed movement code so that  MovementAction::WaitForReach() is no longer called (in most places movement methods anyway), and therefore botAI->SetNextCheckDelay() isn't being called either.

Is SetNextCheckDelay() no longer needed? Or is this an oversight?
- If not needed then SetNextCheckDelay() should be removed from MovementAction::JumpTo() (I just added static 1 second pause because wasn't sure how to calculate jump movement time but knew WaitForReach() wouldnt be right)
- If oversight then SetNextCheckDelay() should be added in 2 places in MovementAction::MoveTo() (3 places if I merge this PR first)

I wasn't sure when adding code for vehicle movement in this PR so left it out.

FizCap:
Bros a beast at coding, gosh damn

liyunfan1223:
I intentionally removed WaitForReach to support casting while moving. Now just setting last movement is sufficient. I've temporarily kept WaitForReach in MoveJump. I believe it's fine to remove it, but a check for IsWaitingForLastMove() is needed before moving.

PR #454, opened 2024-08-08
Fixed bots in EOTS alternating between mount/unmount near flag indefinitely

fuzzdeveloper:
Fixes this bug: https://youtu.be/_A6Z8wxUdgE

This bug was caused because I copied other code in BattleGroundTactics.cpp which erroneously uses sqrt on values returned by GetDistance() - result was that bots would dismount significantly further from flag than they should (then try mount again in a loop).

The use of sqrt is a relic from when this used to be a mangos bot - in mangos GetDistance() returns square of distance by default (this is done to skip sqrt operation when it is not needed - eg: when trying to find closest object and comparing several distances). In AC however GetDistance() GetExactDist() GetDistance2d() etc all return distance with sqrt operation already performed (so it is not correct to sqrt the returned value if accurate distance is desired). To avoid this problem in future I also searched for other erroneous usages of sqrt and and fixed them (except for a few places where code has been written around this problem and doesn't function correctly once fixed).

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -2566,7 +2566,7 @@ bool PlayerbotAI::CanCastSpell(uint32 spellid, float x, float y, float z, uint8 
 
     if (!itemTarget)
     {
-        if (sqrt(bot->GetDistance(x, y, z)) > sPlayerbotAIConfig->sightDistance)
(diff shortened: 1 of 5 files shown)


liyunfan1223:
Nice work! I can’t imagine it running normally for so long with this problem haha.

PR #471, opened 2024-08-13
Name Rework

CiMaWi:
1. Arena Team & Guild Names: Edited/deleted/replaced many names by hand, removing potentially offensive names, grammatical inconsistency, faction-specific names (since there is no faction categorization), non-standard spelling, etc. After the rework there are 100 arena team names per bracket & 400 guild names, which is fewer than before, but should be enough.
2. Replaced the playerbot name list with 100k names (some overlap), split into 5k names per race and gender, with human & undead sharing the same pool of double size. Module code has been edited to assign created bots a name specific to their race.
3. Minor: Random race is now guaranteed to be assigned, removing the miniscule chance of failure in 15 attempts.


noisiver:
The build is failing because Ubuntu 20.04 only comes with OpenSSL 1.1.1. It would need to manually compile OpenSSL 3.X in order to succeed and I don't know if that's something AzerothCore will do or not. If the 20.04 build is the only one that fails (which it is) we could probably merge it anyway.

liyunfan1223:
Looks good to me.

PR #477, opened 2024-08-14
General optimisation

atidot3:
Replaced alot of std::string copy by string_view and function prototype with references.
Rework helpers with modern c++ 20


liyunfan1223:
Looks good to me. Have you done any performance testing to see how much of a performance gain?

atidot3:
Its a grain of salt in the all process but it reduce allocation so its never bad :)

liyunfan1223:
I believe it's more than a grain of salt. String copy overhead (and string hash in something like map<std::string, ...>) is really terrible based on my performance test. But most of them can be optimized with a better coding approach.

I just feel that any modification should have a clear benefit, not just that way it looks good. Modification is always risky, and if there is no exact benefit, it should not be touched.

atidot3:
I mean based on the sources size of the module its a small amont of required changes, you cant test optimisation with only small amount of changes like this on few strings.

Its a long process as containers need to be changed to use also refs and views.

Small commit by small commit

liyunfan1223:
We are not talking about the same thing. You don't need to optimize everything, no. You only need to optimize the critical part, in other words performance bottleneck. If you don't even know where the bottleneck is, figure it out first.

atidot3:
image

Both run at 1000 bots for 10 minutes.
On the optimise side the cpu is at a lower usage and seem to cycle more per cycle on UpdateAIInternal.

PR #486, opened 2024-08-19
Improvement gear intialization & Disable ExtractAllItemIds for heavy regex

liyunfan1223:
- Add quest rewards to gear initialization
- Disable ExtractAllItemIds since heavy regex.


liyunfan1223:
Hello @atidot3. The regex you used in ChatHelper::ExtractAllItemIds is heavy, I'm going to temporarily disable it before optimizing. https://github.com/atidot3/mod-playerbots/blob/9c33837c8f15f94edce9a0e50e4ab27c0c2e883c/src/ChatHelper.cpp#L612

PR #487, opened 2024-08-19
Flame leviathan strategy

liyunfan1223:
1. Add Flame Leviathan strategy in Ulduar (normal mode only). When the the player is on vehicle, the bot will also enter other vehicles and take control.

2. Fixed an issue that duplicate logins that caused a crash. This should also resolve the issue with the incorrect mount of login. https://github.com/liyunfan1223/mod-playerbots/issues/255

3. Add `ApplyInstanceStrategies` option, allowing instance strategies to be automatically added when entering the instance.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -369,8 +369,8 @@ AiPlayerbot.SyncQuestForPlayer = 0
 #
 #
 
-# Bot can flee for enemy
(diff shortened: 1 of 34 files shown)


liyunfan1223:
Note: Also pull the latest core update. https://github.com/liyunfan1223/azerothcore-wotlk/commit/1bf88ad3a341ca1f232413515c487017e706035e

PR #489, opened 2024-08-21
Performance loading the botnames from a cache

hermensbas:
GenderRaceBotNames retrieved from a cache instead of using a DB call per name on a 100k record table 'playerbots_names'. 
The name fallback is still in place but had to replace to IN_USE mechanism with a inner join. The IN_USE enforced cache still need to make the IN_USE db call which took a huge performance hit. Before 1-2 seconds per account creation, now a few seconds for 100's of accounts. 

Tested with cache and without the cache enabled and with fallback mechanism active.

### DIFF
--- a/sql/characters/playerbots_names.sql
+++ b/sql/characters/playerbots_names.sql
@@ -100010,5 +100010,4 @@ INSERT INTO `playerbots_names` VALUES
 
 DELETE FROM `playerbots_names` WHERE LENGTH(`name`) > 12;
 ALTER TABLE `playerbots_names` MODIFY `name` varchar(12);
-
-ALTER TABLE `playerbots_names` ADD in_use BIT default 0;
+ALTER TABLE `playerbots_names` ADD UNIQUE INDEX name(name, gender);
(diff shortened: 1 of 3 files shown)


Rozza122:
> GenderRaceBotNames retrieved from a cache instead of using a DB call per name on a 100k record table 'playerbots_names'. The name fallback is still in place but had to replace to IN_USE mechanism with a inner join. The IN_USE enforced using the cache still to make a DB which took a huge performance hit. Before 1-2 seconds per account creation, now a few seconds for 100's of accounts.
> 
> Tested with cache and without the cache enabled with fallback mechanism active.

adding a new index to playerbots name increases speed slighly too:

time without key 337ms~, with key 58ms~.

Query ran:
SELECT n.name
FROM acore_characters.playerbots_names n 
LEFT OUTER JOIN acore_characters.characters c ON c.name = n.name 
WHERE c.guid IS NULL and n.gender = '8' 
ORDER BY RAND() LIMIT 1;

New Index: 
**ALTER TABLE acore_characters.playerbots_names 
  ADD UNIQUE INDEX name(name, gender);**

hermensbas:
Good point, just in case when the fallback is being used and tiny bit faster when loading the cache. Also noticed i didnt cleanup the in_use from the playerbots_names.sql.  Thanks :)

moncat2005:
@hermensbas 
How is the gender field defined for the playerbots_names table: 0-17
Odd numbers for women, even numbers for men?
How do I distinguish race? I want to translate to be localized.

PR #492, opened 2024-08-25
Updated Playerbots.cpp to allow bots to gain achievements

Dohtt:
This change should make it so that playerbots can gain achievements just like regular players.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -179,19 +179,16 @@ public:
     {
         if (sRandomPlayerbotMgr->IsRandomBot(player))
         {
-            return false;
+            return true;
         }
-        return true;
     }
 
     bool OnBeforeAchiComplete(Player* player, AchievementEntry const* /*achievement*/) override
     {
         if (sRandomPlayerbotMgr->IsRandomBot(player))
         {
-            return false;
+            return true;
         }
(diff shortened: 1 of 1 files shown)


tehho:
This should probably be done behind a feature flag. 
You need to keep the return true otherwise the function does not return anything for normal players.

Add a config value, suggestion AiPlayerbot.AllowAchievements, and default the config to 0, aka false. Then in the if bot, return is config set.

This should keep the default behavior and allow for enabling achievements

noisiver:
A few issues with this.
- How you're doing this is the worse way possible. Warnings all over the place when compiling.
- It should be configurable.
- You should check the achievements to make sure the rndbots don't get realm first achievements. It could be configurable but it should still be there.

Dohtt:
i'll leave the code part to people who can actually code, thanks for pointing out it won't work.

noisiver:
> i'll leave the code part to people who can actually code, thanks for pointing out it won't work.

That's not what I'm saying. I'm just saying that the code needs to be cleaned up and modified to include a few more things. If you want to enabled achievements on your end, just remove both of those functions instead of making them return true.

If you want this feature to be a part of the repo but aren't confident in doing it yourself I could sit down and add it as a configurable option.

PR #502, opened 2024-08-31
Improve druid cat spec

liyunfan1223:
Enable cat druid generation.


liyunfan1223:
Note: update playerbots.conf to enable cat druid generation.

PR #507, opened 2024-09-05
Equip init

liyunfan1223:
close https://github.com/liyunfan1223/mod-playerbots/issues/230


liyunfan1223:
https://github.com/liyunfan1223/mod-playerbots/issues/255

PR #522, opened 2024-09-12
Minor correction forgot to update the imperative default configs

hermensbas:
Minor correction for [commit](https://github.com/liyunfan1223/mod-playerbots/commit/1fbbdcd44b1301dbf7c5bab1ec6fc35190c110d2)


fuzzdeveloper:
MaxRandomBotInWorldTime and MaxRandomBotRandomizeTime are not consistent between conf.dist and defaults in code (they are both 1209600 in conf.dist and 12 * HOUR = 43200 in code, guessing the former is wrong)

hermensbas:
Not necessarily wrong but it does create fallback above the existing mechanisms for the rndbots.

PR #549, opened 2024-09-24
[WIP] Tank face and dps behind

liyunfan1223:
1. The main tank will not switch targets when there are multiple tanks in the group.
2. Melee DPS will move to the target's back (rear flank) when they are not positioned behind the target, gaining higher damage and avoiding cleave-like spells.
3. Tanks will reposition to ensure the current target does not face ranged players in the group (to counter breath and cleave-like abilities).
4. Fixed an issue where the bot would get stuck in target selection when the skull marker was placed on a friend unit.
5. Fixed an issue of the wrong knock back direction of bot.


jspsh:
什么时候能推送上去啊？大佬

PR #555, opened 2024-09-27
Teleport and performance optimizations

hermensbas:
The general idea is to supress the aggressive randomTeleport timings without losing functionality or bot behaviour. And thereby lowering the server resources even more so. Still testing, maxing out around 7000 bots atm due hitting my RAM limits. 

Using 4000 bots as baseline for testing:

image2
image

### DIFF
--- a/src/RandomPlayerbotMgr.cpp
+++ b/src/RandomPlayerbotMgr.cpp
@@ -1097,8 +1097,9 @@ bool RandomPlayerbotMgr::ProcessBot(uint32 bot)
                  player->GetLevel(), player->GetName().c_str());
         LogoutPlayerBot(botGUID);
         currentBots.remove(bot);
-        SetEventValue(bot, "logout", 1,
-                      urand(sPlayerbotAIConfig->minRandomBotInWorldTime, sPlayerbotAIConfig->maxRandomBotInWorldTime));
+        SetEventValue(bot, "logout", 1, urand(
+            sPlayerbotAIConfig->minRandomBotInWorldTime, 
+            sPlayerbotAIConfig->maxRandomBotInWorldTime));
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


hermensbas:
image

Just happy <:

hermensbas:
image

EricksOliveira:
Just a doubt. What is your machine configuration?

hermensbas:
> Just a doubt. What is your machine configuration?

An heavily undervolt 5700x (max 70 watts) 32 GB, vmware unbuntu server 24.04 minimaal, mysql 8.4.2, assigned 8 cores, 20GB to the VM.

[Config](https://github.com/hermensbas/azerothcore_installer/tree/main/config)

hermensbas:
> image

These numbers are coming from Marco_92, 7900x, 64GB mem. But on windows, which kinda soso performance wise. He is recompiling now with the second part of the fix, no idea what numbers he will be able to push out. For me (5700x, 32GB) its around 6000 bots with decent numbers.

hermensbas:
image
image

hermensbas:
Marco_92, amd 7900x, 64GB, Windows
image

hermensbas:
Marco_92 @ 12k bots
image

hermensbas:
Seems to good to be true, i've looked at bots how active they are and whatnot, cant find any issue's though. No laggs or delays.

hermensbas:
@Marco_92 has been testing the limits with these last couple of perf changes, laggy barely playable but playable. Obviously making the world feel alive is not equal the amount of bots in the world but mainly the behaviour of bots make the world feel alive. Even with 20k bots stormwind isnt as full as one would expect or even how busy it is with 5k bots with cmangos bots. But it doesnt hurt to optimize the process around it. The rest will hopefully follow overtime.
image

liyunfan1223:
As far as I know, there appears to be an issue with BotActiveAlone implementation. If this value is not set to 100, it possibly causes bots to only act when there are players within 300 yards nearby. This means the bot cannot grind effectively to level up. While this functionality is useful, it doesn't match the intended purpose of this configuration option. 

By the way, the issue should be in PlayerbotAI::AllowActive(). The diff used the maximum value since server startup instead of the maximum value within a time window. This could lead to the server being judged as always under high load.

https://github.com/liyunfan1223/mod-playerbots/blob/69c3568fb66959adaf671143fe5fbded6b12a859/src/PlayerbotAI.cpp#L4018

conf/playerbots.conf.dist
```diff
@@ -1460,7 +1460,7 @@ AiPlayerbot.RandombotsWalkingRPG.InDoors = 0
 
 # Specify percent of active bots
 # The default is 10. With 10% of all bots going active or inactive each minute.
-AiPlayerbot.BotActiveAlone = 100
+AiPlayerbot.BotActiveAlone = 90
```
liyunfan1223: Maybe we can keep this value default at 100 for now, and merge other optimization of teleport first.

hermensbas: I was kinda considering that as well, i will change it :)

hermensbas: ~~I did include a sync correction of the orginal value of 100 in 'src/PlayerbotAIConfig.cpp', the conf said 100 and code said 10. Or do you prefer that change in a seperate PR aswell?~~

Lets keep it clean, made another PR for the above :) So both values are set on 100.

hermensbas:
> As far as I know, there appears to be an issue with BotActiveAlone implementation. If this value is not set to 100, it possibly causes bots to only act when there are players within 300 yards nearby. This means the bot cannot grind effectively to level up. While this functionality is useful, it doesn't match the intended purpose of this configuration option.
> 
> By the way, the issue should be in PlayerbotAI::AllowActive(). The diff used the maximum value since server startup instead of the maximum value within a time window. This could lead to the server being judged as always under high load.
> 
> https://github.com/liyunfan1223/mod-playerbots/blob/69c3568fb66959adaf671143fe5fbded6b12a859/src/PlayerbotAI.cpp#L4018

Just noticed this msg, no idea how missed your wholw feedback reply. I will look into the allowActive, i did notice the function and me doubt where it should be added. But totally makes sense the intend fits alot better then in a method called processbot

PR #562, opened 2024-09-30
Modified the botActiveAlone autoscale functionality

hermensbas:
botActiveAlone had a related feature which autoscales the botActiveAlone percentages when the value is lower then 100 implicitily.

Modifed and made the functionality explicit through the conf file, awarely enabled by default. Enabled to by default because it simplifies the configuration with better and more stabilized performance out of the box. Providing the server breathing space instead of locking up the resources, while still being able to disable to feature and enforce a specific value instead.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -4095,8 +4095,6 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
             return true;
     }
 
-    uint32 maxDiff = sWorldUpdateTime.GetMaxUpdateTime();
-
     Group* group = bot->GetGroup();
     if (group)
     {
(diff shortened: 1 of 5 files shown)


src/PlayerbotAI.cpp
```diff
@@ -4144,35 +4142,31 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
...
-    {
-        if (Map* map = bot->GetMap())
-        {
-            if (map->GetEntry()->IsWorldMap())
+    // use the configured 'botActiveAlone 'percentages. Downscale the activity, within 
+    // reason ,with the intension to give the server some breathing space when needed.
+    uint32 mod = sPlayerbotAIConfig->botActiveAlone > 100 ? 100 : sPlayerbotAIConfig->botActiveAlone;
+    
+    if (sPlayerbotAIConfig->botActiveAloneAutoScale) {
+        uint32 maxDiff = sWorldUpdateTime.GetMaxUpdateTime();
```
liyunfan1223: It may be better to use `GetMaxUpdateTimeOfCurrentTable` to get the diff of the most recent period. GetMaxUpdateTime will record the highest diff, which usually occurs when the server starts (the bot login). The diff at this time generally does not last long.
```
uint32 maxDiff = sWorldUpdateTime.GetMaxUpdateTimeOfCurrentTable();
```

hermensbas: Thanks for the feedback, gonna have a try in the evening :)

hermensbas: Added some additional steps for more scaling granularity. I went for 
```
uint32 maxDiff = sWorldUpdateTime.GetAverageUpdateTime()
``` 
which gave a more stabile and predictable number since it provides the avg over the full spectrum. What do you think?

src/PlayerbotAI.cpp
```diff
@@ -4144,35 +4142,38 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
...
+            mod = (mod * 9) / 10;
+        if (maxDiff > 100)
+            mod = (mod * 8) / 10;
+        if (maxDiff > 120)
+            mod = (mod * 7) / 10;
+        if (maxDiff > 140)
+            mod = (mod * 6) / 10;
+        if (maxDiff > 160)
+            mod = (mod * 5) / 10;
+        if (maxDiff > 180)
```
liyunfan1223: `mod` here will be reduced repeatedly, is this intentional?

It seems that it should be judged from large to small, for example:
```
if (maxDiff > 180)
    mod = (mod * 3) / 10;
else if (maxDiff > 160)
    mod = (mod * 5) / 10;
else if
    ...
```

hermensbas: Yes and no, but now u mention it lets alter the original code abit more. Good catch. I didnt want to alter the original code to much.

PR #576, opened 2024-10-04
Healer attack (healer dps strategy) and healer enhancement

liyunfan1223:
`assist dps` strategy: healers cast damage spells if they have enough mana. The mana threshold depends on the power of the enemies (the stronger the enemy, the higher the threshold) to ensure that there's enough mana to heal.

The name is a bit confusing with `dps assist`. The strategy name may be modified further.


Dreathean:
Awesome additions! Maybe consider renaming to "healer dps" since it is a strat only for healers to dps? Agree there is confusion with "assist dps" and "dps assist"

PR #580, opened 2024-10-05
AllowActive prioritizing and bracket scaling based on bot activity 

hermensbas:
One of many to come.  

Found the issue together with @noisiver, basically backpressure of mapUpdate thread. The updateRequest are to many to process so the thread cant keep up and starts queueing them faster then processing. Thereby the memory goes up and up and up. Thereby bots actions where never executed or far to late as side-effects bots acting strange and or simple out of sync with the actual map.

@noisiver 
ac_profile

The main variables in this whole are part:

- CPU single core clock speed
- Amount of bots on the server
- The RandomBotUpdateInterval 
- The BotActiveAlone percentages and logic around it.
- Variables that effect the cluttering of bots and the amount of bots close to real players e.g.
          - AutoTeleportForLevel 
          - DisableRandomLevels
          - RandombotStartingLevel
          
          
Added/ported the two features to reduce the pressure backlog buildup of the MapUpdater thread.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -3926,7 +3926,10 @@ Player* PlayerbotAI::GetGroupMaster()
 
 uint32 PlayerbotAI::GetFixedBotNumer(BotTypeNumber typeNumber, uint32 maxNum, float cyclePerMin)
 {
-    uint32 randseed = rand32();                               // Seed random number
+    //deterministic seed 
+    uint8 seedNumber = uint8(typeNumber);
+    std::mt19937 rng(seedNumber);
+    uint32 randseed = rng();                                  // Seed random number
     uint32 randnum = bot->GetGUID().GetCounter() + randseed;  // Semi-random but fixed number for each bot.
 
     if (cyclePerMin > 0)
@@ -3936,8 +3939,7 @@ uint32 PlayerbotAI::GetFixedBotNumer(BotTypeNumber typeNumber, uint32 maxNum, fl
         randnum += cycle;                            // Make the random number cylce.
     }
 
-    randnum =
-        (randnum % (maxNum + 1));  // Loops the randomnumber at maxNum. Bassically removes all the numbers above 99.
+    randnum = (randnum % (maxNum + 1));  // Loops the randomnumber at maxNum. Bassically removes all the numbers above 99.
     return randnum;  // Now we have a number unique for each bot between 0 and maxNum that increases by cyclePerMin.
 }
 
@@ -4075,18 +4077,15 @@ inline bool HasRealPlayers(Map* map)
     return false;
 }
 
-bool PlayerbotAI::AllowActive(ActivityType activityType)
+ActivePiorityType PlayerbotAI::GetPriorityType(ActivityType activityType)
 {
-    // General exceptions
-    if (activityType == PACKET_ACTIVITY)
-        return true;
+    // First priority - priorities disabled or has player master. Always active.
+    if (HasRealPlayerMaster())
+        return ActivePiorityType::HAS_REAL_PLAYER_MASTER;
 
-    if (GetMaster())  // Has player master. Always active.
-    {
-        PlayerbotAI* masterBotAI = GET_PLAYERBOT_AI(GetMaster());
-        if (!masterBotAI || masterBotAI->IsRealPlayer())
-            return true;
-    }
+    // Self bot in a group with a bot master.
+    if (IsRealPlayer())
+        return ActivePiorityType::IS_REAL_PLAYER;
 
     Group* group = bot->GetGroup();
     if (group)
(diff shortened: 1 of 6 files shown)


liyunfan1223:
Thank you both for all the time you spent on this issue, valuable insights!

According to the test results, increasing `ReactDelay` should also significantly alleviate the thread accumulation issue, such as increasing it from 100 to 150 or 200. The side effect is that the interval of bots action will increase.

liyunfan1223:
What is strange is that the problem of high bot update overhead has always existed, but the issue of thread accumulation (or memory increasing) did not occur until recently. One possibility is that the overhead of some bots suddenly increases at a certain point. Once thread blocked, a vicious cycle will occur, and lags and memory consumption will continue to increase.

I feel it may be related to `DynamicDelay` option which introduced in https://github.com/liyunfan1223/mod-playerbots/pull/475/commits/2ddaae9ef05de5b2cd08a1590a44b57f25c52b68. This will cause the bot to incur different overheads in different situations (for example, in combat will be more than in rest). But I'm not sure yet and need further testing.

hermensbas:
Prolly in combination with some of performance tweaks i made on top of that.

hermensbas:
I think we have to step away from wanting all bots being active so they level as normal player. And lets the bots be active to the bare mininum as default config atleast. What do you think @liyunfan1223?

hermensbas:
I think it would be wise to step away from the idea of letting bots level as normal players, esp considering the AC core isnt multithreaded. And just let the bots became more active when there is a direct reason. And let the default value BotActiveAlone on 10%, atleast for now.

hermensbas:
Proof in the pudding, the users :)

hermensbas:
Last test results:

image

src/PlayerbotAI.cpp
```diff
@@ -4124,62 +4142,166 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
...
+    if (bot->IsBeingTeleported() || !bot->IsInWorld() || !HasRealPlayers(bot->GetMap()))
+        return ActivePiorityType::IN_INACTIVE_MAP;
+
+    // IN_ACTIVE_AREA
+    if (activityType == OUT_OF_PARTY_ACTIVITY || activityType == GRIND_ACTIVITY)
+    {
+        // Many bots nearby. Do not do heavy area checks.
         if (HasManyPlayersNearby())
-            return false;
+            ActivePiorityType::IN_ACTIVE_AREA;
```
liyunfan1223: Is this a typo error (missing `return`) and the function is missing a fallback return value?

src/PlayerbotAI.cpp
```diff
@@ -4124,62 +4142,166 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
...
+    }
+}
+
+// Returns the lower and upper bracket for bots to be active.
+// Ie. 10,20 means all bots in this bracket will be inactive below 10% activityMod, all bots in this bracket will be
+// active above 20% activityMod and scale between those values.
+std::pair<uint32, uint32> PlayerbotAI::GetPriorityBracket(ActivePiorityType type)
```
liyunfan1223, line 4190: The idea of ​​setting multiple active priorities is a good one, I'm just a little worried that it may make the server overhead unpredictable and make debugging more difficult. I'm not sure. We can merge it and see the effect to decide whether to keep it.

hermensbas: lets test abit more, and ye agree. I will the code but i will disable the functionality. Maybe its something we should consider after being stabile in those topics.

src/PlayerbotAI.cpp
hermensbas: Typo :)

hermensbas:
I've removed autoscaling based on latency entirely, never like it. We cant assume the backlog of an thread based on that information. I took the cmangos example where keep wihtin the border of the class and the knowlegde it has, by prioritizing and scaling based by there activity. I am getting stabile results. Its not perfect but alot better.

We possible can add an backlog hook in the core MapUpdater threading class or alike the obtain throughput data and finally update the acitivityPercentage based on that.  Cant shake the feeling there is still a memory leak aside the pressure buildup.

Eitherway test it, have a peek, but the solution feels more stabile and predictable. We might wanna tweak the scaling brackets. But prolly best we do that in another PR.

hermensbas:
Now they got more breathing space but aswell the prioritizing and what not, the bots behave much better. Give it a try set ur bots on 3k, 1-60, vanilla maps teleport and go to ogrimmar and go to pvp areas.

hermensbas:
Ive been testing with 10% activity and 100% activity configured, i am actually considering using 10% because the bots do what they suppose to be doing even with 10%. Interesting. Not talking about resource usage but just general behaviour seems more smooth somehow.

liyunfan1223:
This is awesome, please take a break and don't stress about it haha. Thank you for all your efforts!

PR #581, opened 2024-10-06
Partial memory mitigation issue quickfix for now

hermensbas:
This will mitigate the pressure by increasing the updateInterval, until more robust solution. See https://github.com/liyunfan1223/mod-playerbots/pull/580

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -728,7 +728,7 @@ AiPlayerbot.FastReactInBG = 1
 #
 
 # All In seconds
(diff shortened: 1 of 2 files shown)


zhangbo8418:
Isn't it normal for memory usage to increase? After all, the unmodified AzerothCore WotLK map fully loaded occupies over 9GB of memory.

noisiver:
The issue is that it increases endlessly. You will eventually run out of memory and the worldserver will either freeze or crash. It doesn't matter how much RAM you have, you **will** eventually run out. Fixing that issue is what's being worked on, and this is only the start.

PR #617, opened 2024-10-18
Update PlayerbotCommandServer.cpp

EricksOliveira:
Protection against race conditions: Added a global mutex (session_mutex) to synchronize access to the session() function when multiple threads are running. This avoids concurrency problems in environments with many simultaneous connections.

Thread pool usage: Replaced manual thread creation with boost::thread with a thread pool using boost::asio::thread_pool. The thread pool improves resource management and limits the number of simultaneous threads, optimizing performance.

Checking for socket errors: Added check to ensure the socket is not null or closed before trying to read it. Implemented detailed error handling for readings and writings, including detailed logs to facilitate debugging. Secure socket closure: Now sockets are correctly closed at the end of the session, even in the event of an error, preventing resource leaks.

### DIFF
--- a/src/PlayerbotCommandServer.cpp
+++ b/src/PlayerbotCommandServer.cpp
@@ -9,8 +9,10 @@
 #include <boost/bind.hpp>
 #include <boost/smart_ptr.hpp>
 #include <boost/thread/thread.hpp>
+#include <boost/asio/thread_pool.hpp>
 #include <cstdlib>
 #include <iostream>
+#include <mutex>
 
 #include "IoContext.h"
 #include "Playerbots.h"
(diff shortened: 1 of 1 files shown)


hermensbas:
Cant review the changes on itself without testing, which  i cant do atm. Having a mutex lock for threadsafety sounds like a good idea. Not entirely sure about having a threadpool amount hardcoded baked into the codebase. But maybe its better @liyunfan1223  to review.

liyunfan1223:
Thanks, looks good to me. I will give feedback after testing.

hermensbas:
Gonna test this soon

hermensbas:
Are you sure a threadpool model is good approach here? How are we gonna detect the threadpool is gonna be a bottleneck? Obviously a threadpool has a preference in most situations, but i am kinda having doubts in this situation.

Do you have any numbers in regard of current thread usage during a long session in difference situations such as alot of bots/players? Or better said how did you determine 4 is the correct sizing?

EricksOliveira:
You raise an important point! In a scenario with high bot/player load and prolonged sessions, the thread pool size can easily become a bottleneck if undersized, and the choice of a fixed size like "4" was more illustrative than calculated.

I don't have enough CPU to do these tests. Therefore, I am leaving some information on how to test better.

Measuring Session Time and CPU Usage: Collect data about the execution time and CPU usage of each session.
Identify whether there are blockages or spikes in usage in high traffic situations.

Latency and Thread Queue Monitoring: Observe the waiting time of sessions to acquire a thread.
If the queue grows or response latency increases significantly, it may be a sign that the pool is undersized.

Testing in High Load Environments: Simulate a large number of simultaneous bot connections to observe how the server responds and identify the saturation point.
With this, we can dynamically adjust the pool size according to the number of active bots.

hermensbas:
> You raise an important point! In a scenario with high bot/player load and prolonged sessions, the thread pool size can easily become a bottleneck if undersized, and the choice of a fixed size like "4" was more illustrative than calculated.
> 
> I don't have enough CPU to do these tests. Therefore, I am leaving some information on how to test better.
> 
> Measuring Session Time and CPU Usage: Collect data about the execution time and CPU usage of each session. Identify whether there are blockages or spikes in usage in high traffic situations.
> 
> Latency and Thread Queue Monitoring: Observe the waiting time of sessions to acquire a thread. If the queue grows or response latency increases significantly, it may be a sign that the pool is undersized.
> 
> Testing in High Load Environments: Simulate a large number of simultaneous bot connections to observe how the server responds and identify the saturation point. With this, we can dynamically adjust the pool size according to the number of active bots.

Those measurements can be rather complex though, not in monitoring but the possible configurarion options and usage variables atleast think so 🤔

Maybe its an idea to seperate mutex lock for the session from the threadpool impl? Also what came to mind was, wouldnt it make more sense to make session itself threadsafe?

liyunfan1223:
Wondering if there are any observable resource leaks or performance bottlenecks in actual runtime that can be fixed by these modifications?

hermensbas:
> Wondering if there are any observable resource leaks or performance bottlenecks in actual runtime that can be fixed by these modifications?

Thats the main reason i dont dare to hit merge, since i am not sure whats its trying to fix other then preventive measures based on ?

hermensbas:
I could merge it locally and tests i guess, although the threadpool is to risky in my opinion unless there is a direct reason for it.

PR #618, opened 2024-10-19
Armor token usage

avirar:
Added logic to LootRollAction to allow the bots to roll need/greed on armor tokens. The items are all class 15 subclass 0 which is MISC, JUNK. Luckily no other items in MISC, JUNK also have EPIC quality and AllowableClass bits so this seems like a safe approach.

### DIFF
--- a/src/strategy/actions/LootRollAction.cpp
+++ b/src/strategy/actions/LootRollAction.cpp
@@ -12,6 +12,20 @@
 #include "PlayerbotAIConfig.h"
 #include "Playerbots.h"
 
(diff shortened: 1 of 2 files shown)


src/strategy/actions/LootRollAction.cpp
```diff
@@ -132,10 +149,11 @@ bool LootRollAction::Execute(Event event)
     //             break;
     //     }
     // }
-
+
```
noisiver: There shouldn't be spaces here.

avirar: Removed spaces

PR #622, opened 2024-10-19
Preferred mount selection for bots

Bobblybook:
Added a new table to store user-specified mount ids.
Can specify flying and ground mounts for a character which will prioritise that mount. If multiple entries, a random entry will be selected from that list.

Currently no way to add entries other than manual DB editing, but can be developed into a whisper command pretty easily.

This should gracefully fail (if no entries exist for the character, or the db table does not exist) and fallback to the default random selection as usual.

### DIFF
--- a/sql/playerbots/base/playerbots_preferred_mounts.sql
+++ b/sql/playerbots/base/playerbots_preferred_mounts.sql
@@ -1,0 +1,10 @@
+DROP TABLE IF EXISTS `playerbots_preferred_mounts`;
+CREATE TABLE `playerbots_preferred_mounts` (
+  `id` INT(11) NOT NULL AUTO_INCREMENT,
+  `guid` INT(11) NOT NULL,
+  `type` TINYINT(3) NOT NULL COMMENT '0: Ground, 1: Flying',
+  `spellid` INT(11) NOT NULL,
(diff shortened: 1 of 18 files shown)


noisiver:
I think there should probably be some kind of command to add a mount to the preferred list in the future. As it stands, regular players can't decide what mounts they prefer the bots to use without access to the database.

I'm not saying it has to be included in this pull request, I'm not advanced enough to make that decision, but it should probably be possible eventually. That specific part should only work with altbots though. Regular players shouldn't be allowed to change the preferred mount(s) of rndbots.

Bobblybook:
> I think there should probably be some kind of command to add a mount to the preferred list in the future. As it stands, regular players can't decide what mounts they prefer the bots to use without access to the database.
> 
> I'm not saying it has to be included in this pull request, I'm not advanced enough to make that decision, but it should probably be possible eventually. That specific part should only work with altbots though. Regular players shouldn't be allowed to change the preferred mount(s) of rndbots.

Yeah agreed. I'm not sure what direction we want to go with this and I have a more important todo list currently (dungeons, warrior/shaman dual wield gear issues, rings/trinket equips) but I wanted to at least add this skeleton so it's possible to specify mounts if needed.

PR #638, opened 2024-10-24
Hack Fix Crash OnBotLogin

EricksOliveira:
Now prevents OnBotLogin Crash. However, to create Off char or the same bots via AddClass, Logout is necessary.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -176,7 +176,7 @@ void PlayerbotHolder::HandlePlayerBotLoginCallback(PlayerbotLoginQueryHolder con
         botSession->LogoutPlayer(true);
         delete botSession;
     }
-    botLoading.erase(holder.GetGuid());
+    //botLoading.erase(holder.GetGuid());
 }
 
 void PlayerbotHolder::UpdateSessions()


EricksOliveira:
So far it was the only way I found to prevent the Crash from happening again. I'm studying the code to try to add a fix.

src/PlayerbotMgr.cpp
zbhcn, line 1: May cause memory leaks

zbhcn (commented):
May cause memory leaks

Mr-Deadbeat:
have not experienced any on bot login crashes is it only related to the addclass functions ?

hermensbas:
converted to draft.

EricksOliveira:
AddClass and Alt bot. Crash occurs when you take the bot offline and then bring it online. Before the bot is 100% online, you relog and the server crashes.

noisiver:
> AddClass and Alt bot. Crash occurs when you take the bot offline and then bring it online. Before the bot is 100% online, you relog and the server crashes.

Never had that happen, ever.

EricksOliveira:
> > AddClass and Alt bot. Crash occurs when you take the bot offline and then bring it online. Before the bot is 100% online, you relog and the server crashes.
> 
> Never had that happen, ever.

https://youtu.be/IecCeDFsxMg

liyunfan1223:
The crash may be caused by repeatedly logging in and out within a short period of time. This modification actually disables a second login, but the cause of the crash is still unclear.

serverwar:
> The crash may be caused by repeatedly logging in and out within a short period of time. This modification actually disables a second login, but the cause of the crash is still unclear.

Ok, this fix didn't solve the crash, I'll explain here, every time you invite or remove when an error occurs for some reason it seems like you lose the connection with the character and then you relog and then generate the dc

serverwar:
> > AddClass and Alt bot. Crash occurs when you take the bot offline and then bring it online. Before the bot is 100% online, you relog and the server crashes.
> 
> Never had that happen, ever.

You can play and play, but at some point there will be an error with a character, and you relog the server will crash, now if you don't relog it won't crash, this makes the server crash 2 or 3 times a day, players use it a lot, so testing only with bots and with real players is different.

noisiver:
> > > AddClass and Alt bot. Crash occurs when you take the bot offline and then bring it online. Before the bot is 100% online, you relog and the server crashes.
> > 
> > 
> > Never had that happen, ever.
> 
> You can play and play, but at some point there will be an error with a character, and you relog the server will crash, now if you don't relog it won't crash, this makes the server crash 2 or 3 times a day, players use it a lot, so testing only with bots and with real players is different.

Again, I've never had that happen. I've relogged, frequently, with rndbots and altbots. Multiple times a day, a minimum of 5 a day, and it has never happened.

EricksOliveira:
> > > > AddClass e Alt bot. Ocorre travamento quando você coloca o bot offline e depois o coloca online. Antes que o bot esteja 100% online, você faz login novamente e o servidor trava.
> > > 
> > > 
> > > Isso nunca aconteceu, nunca.
> > 
> > 
> > Você pode jogar e jogar, mas em algum momento vai dar um erro com algum personagem, e você relogar o servidor vai cair, agora se você não relogar ele não vai cair, isso faz o servidor cair 2 ou 3 vezes por dia, os jogadores usam muito, então testar só com bots e com jogadores reais é diferente.
> 
> Novamente, isso nunca aconteceu comigo. Eu reloguei, frequentemente, com rndbots e altbots. Várias vezes por dia, no mínimo 5 por dia, e isso nunca aconteceu.

To be reproduced more easily. Use some AddClass bot in the group. Use the leave everyone off command, then use the leave everyone online command. Before everyone goes online, you log out. The crash occurs when relogging before the bot comes back online. It's a matter of seconds. However, it has to be quick and if possible it has to be in active InstantLogout.

hermensbas:
Cant reproduce

hermensbas:
closing it for now

Issue #544, opened 2024-09-21
Bots lose talents


Dohtt:
**Describe the bug**
playerbots lose their talent points after a period of time.

**To Reproduce**
1. Let bots level to level 10+
2. Watch them select talents or give them talents with a command
3. After a certain time they end up not having talents anymore.

**Expected behavior**
Bots should keep their talent points

**Screenshots**
![image](https://github.com/user-attachments/assets/315b3d22-441e-436b-b3eb-73de1b86b7df)

Dreathean:
Can confirm I’ve noticed the same issue, not sure when it first started occurring, I believe it was sometime before the most recent commits though.

EricksOliveira:
You cannot understand this error.
![Captura de tela 2024-09-21 123730](https://github.com/user-attachments/assets/05acf575-7af5-4bc9-a6c9-aafc02177f0f)

Dreathean:
Bad translation? Assuming you mean you aren’t experiencing this issue?

Dohtt:
Seeing as the provided picture still has the bots leveling: i did not have issues while they're leveling.
They just suddenly lost their talents on max level after being max level for a period of time.

EricksOliveira:
Can you provide me with the Bot's online time. so I can compare and test?

Dohtt:
The bot i saw was level 60 for 445.295 leveltime in my database.
They had talents before, but lost them on the way.

EricksOliveira:
I apologize, but I still haven't been able to identify this problem. Bots with 168h+ of play without any problems with talents.

Dreathean:
This was due to a bug with equipment persistence and was resolved with https://github.com/liyunfan1223/mod-playerbots/pull/778

PR #647, opened 2024-10-29
Rndbot talents reset bug

avirar:
Added check for bot persistence when bot->resetTalents is called in PlayerbotFactory::Randomize.

Enabled detection of 0/0/0 talents during PlayerbotFactory::Refresh. When a bot above the persistence level is found with unallocated talents (0/0/0), it will automatically select the optimal specialization by evaluating the cumulative gear score across all available specs and choosing the one with the highest alignment to its current equipment.

The 0/0/0 talent auto-fix is far from perfect as the scores aren't normalised across specs. It works well enough and ensures their gear/spec combination is viable.


avirar:
#544

hermensbas:
Isnt possible to reuse existing functions around autogear and init?

avirar:
Since the bug only appeards with equipmentPersistence I have made keeping the bots existing gear a priority. This isn't generating new gear for the bot, it's selecting a spec based on the gear they already have, and only if it detects 0/0/0 after a call to initTalentTrees. The bug that causes the wipe issue is resolved by putting resetTalents inside the if statement for equipmentPersistance, but that does not resolve the issue with bots that previously had their talents wiped since all of the calls to initTalentTree during regular bot activity require equipmentPersistence to be disabled to function.

hermensbas:
I cant help to wonder whether you are trying to solve an side-effect. The this states only occurs when bots reached maxed level not when leveling, also some are not able to reproduce.

Shouldnt we fix the core issue itself, why are bots at max level randomly losing talents? Or i am missing something?

src/factory/PlayerbotFactory.cpp
```diff
@@ -207,7 +207,10 @@ void PlayerbotFactory::Randomize(bool incremental)
     Prepare();
     LOG_DEBUG("playerbots", "Resetting player...");
     PerformanceMonitorOperation* pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Reset");
-    bot->resetTalents(true);
+    if (!sPlayerbotAIConfig->equipmentPersistence || bot->GetLevel() < sPlayerbotAIConfig->equipmentPersistenceLevel)
+    {
+        bot->resetTalents(true);
+    }
```
liyunfan1223, line 213: Modifying only these lines should solve this issue. Other modifications should be unnecessary.

PR #658, opened 2024-10-31
strategy about wotlk-fos boss bronjahm

LFTEC:
not perfect
1. playerbot may fall off the edge of the boss battle map.
2. The tank will first aggro the soul fragments, then run away with the boss.


hermensbas:
ive added @Bobblybook as reviewer

PR #672, opened 2024-11-03
Instance FOS strategy

LFTEC:


src/strategy/AiObjectContext.cpp
```diff
@@ -62,6 +62,7 @@ AiObjectContext::AiObjectContext(PlayerbotAI* botAI) : PlayerbotAIAware(botAI)
     actionContexts.Add(new WotlkDungeonOccActionContext());
     actionContexts.Add(new WotlkDungeonUPActionContext());
     actionContexts.Add(new WotlkDungeonCoSActionContext());
+    actionContexts.Add(new WotlkDungeonFosActionContext());
```
Bobblybook: It would be nice to fix capitalisation here to be consistent, "FoS"
But I can clean this up after PR if needed

src/strategy/dungeons/DungeonStrategyContext.h
```diff
@@ -14,6 +14,7 @@
 #include "wotlk/oculus/OculusStrategy.h"
 #include "wotlk/utgardepinnacle/UtgardePinnacleStrategy.h"
 #include "wotlk/cullingofstratholme/CullingOfStratholmeStrategy.h"
+#include "wotlk/forgeofsouls/FosStrategy.h"
```
Bobblybook: As above, would be nice to follow existing pattern and write this out as ForgeOfSouls.h.
There are many dungeon acronyms in the game and it will get messy very fast if we shorten them, especially for tracing back bugs, it's nice to have the full dungeon name listed in the header for clarity

src/strategy/dungeons/DungeonStrategyUtils.h
```diff
@@ -1,6 +1,6 @@
 #ifndef _PLAYERBOT_DUNGEONUTILS_H
 #define _PLAYERBOT_DUNGEONUTILS_H
-
+#include "Player.h"
```
Bobblybook: Probably not needed as Player.h is included in the calling headers, but this doesn't hurt to include here

src/strategy/dungeons/wotlk/WotlkDungeonActionContext.h
```diff
@@ -13,9 +13,9 @@
 #include "oculus/OculusActionContext.h"
 #include "utgardepinnacle/UtgardePinnacleActionContext.h"
 #include "cullingofstratholme/CullingOfStratholmeActionContext.h"
+#include "forgeofsouls/FosActionContext.h"
```
Bobblybook: As above, would be nice to use format 
`#include "forgeofsouls/ForgeOfSoulsActionContext.h"`

I have pre-written the templates and commented them out for this purpose

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
+#include "Playerbots.h"
+#include "FosActions.h"
+#include "FosStrategy.h"
+
+bool MoveFromBronjahmAction::Execute(Event event)
+{
+    Unit* boss = nullptr;
```
Bobblybook: This does not need to be pre-initialised, you can condense this to:
`boss = AI_VALUE2(Unit*, "find target", "bronjahm");`

AI_VALUE2 will return null pointer if target is not found

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    
+    float distance = bot->GetExactDist2d(boss->GetPosition());
+    float targetDis = 20.0f;
+    if (distance >= targetDis)
+        return false;
+    return MoveAway(boss, targetDis - distance);
+}
+
+bool MoveFromBronjahmAction::isUseful() { return bot->HasAura(SPELL_CORRUPT_SOUL); }
```
Bobblybook: There are multiple places you can do this sort of check - on the action itself (isUseful() or within the execute method itself), in a multiplier, or in the initial trigger. For this sort of check, it should be very easy to identify this condition in the trigger and perform this check there.

See notes in triggers.cpp below, then you can completely remove this isUseful() check

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+bool AttackCorruptedSoulFragmentAction::Execute(Event event)
+{
+    Unit* fragment = nullptr;
+
+    GuidVector targets = AI_VALUE(GuidVector, "possible targets no los");
+
+    for (auto i = targets.begin(); i != targets.end(); ++i)
```
Bobblybook: I've done target searching this way for a few bosses but I need to clean up that code a bit - 
This can be cleaned up and made for elegant to use references and iterating via:
`for (auto& target : targets)`

Just a bit easier to read, there isn't a need to use manual looping like this unless you need to reference the index of the entry.
Also for clarity, `i` is usually used as an integer iterator which represents the index of the current iteration of the loop. In this case it is actually being assigned the GUID of the element in the vector, not the index, so it should probably be named accordingly

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        Unit* unit = botAI->GetUnit(*i);
```
Bobblybook: As above, if switching this to a reference, this line should become:
`Unit* unit = botAI->GetUnit(target);`

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+        {
+            fragment = unit;
+            break;
+        }
+    }
+
+    if (!fragment)
+        return false;
+
+    if (botAI->IsDps(bot))
```
Bobblybook: This check should really be pulled out and checked early for performance reasons. This is a 'static' check and its result will not change depending on the combat circumstances - a dps will always be a dps for the entirety of the dungeon. There is no reason to do all the above target iteration and matching just to throw out the action at this point for non-dps.

You could move this to the top of the action, or preferred would be to move this to the trigger so the trigger never occurs for non-dps characters.

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        Unit* unit = botAI->GetUnit(*i);
+        if (unit && unit->GetEntry() == NPC_CORRUPTED_SOUL_FRAGMENT)
+        {
+            fragment = unit;
+            break;
+        }
+    }
+
+    if (!fragment)
```
Bobblybook: This null check can be combined with the remaining checks. There are a few layers of `return false` here that make it a bit difficult to see that this is just a basic null ptr check + attack command.

Something like:
```
if (fragment && AI_VALUE(Unit*, "current target") != fragment)
{
        return Attack(fragment);
} 

return false;
```
(and remove all checks below it)
Will cover the same logic and should be easier to read

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+            return false;
+
+        return Attack(fragment);
+    }    
+    else
+        return false;
+    
+}
+
+bool AttackCorruptedSoulFragmentAction::isUseful() { return botAI->IsDps(bot); }
```
Bobblybook: As above, should probably be moved to the trigger

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    return MoveTo(bot->GetMapId(), BRONJAHM_TANK_POSITION.GetPositionX(), BRONJAHM_TANK_POSITION.GetPositionY(),
+                  BRONJAHM_TANK_POSITION.GetPositionZ(), false, false, false, true, MovementPriority::MOVEMENT_COMBAT);
+}
+
+bool BronjahmTankPositionAction::isUseful() { return bot->GetExactDist2d(BRONJAHM_TANK_POSITION) > 5.0f; }
+
+bool BronjahmTankTargetAction::Execute(Event event)
+{
+    if (botAI->IsTank(bot))
```
Bobblybook: Can move to trigger

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        Unit* boss = AI_VALUE2(Unit*, "find target", "bronjahm");
```
Bobblybook: Probably want to have a nullptr check here, unsure what will happen if you try to Attack(nullptr), probably something bad.

Can do the same as above for the dps - 
```
if (boss && AI_VALUE(Unit*, "current target") != boss)
{
        return Attack(boss);
} 

return false;
```

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+        if (AI_VALUE(Unit*, "current target") == boss)
+            return false;
+
+        return Attack(boss);
+    }
+    else
+        return false;
+}
+
+bool BronjahmDpsPositionAction::Execute(Event event)
```
Bobblybook: I think this should be renamed - it's not so much dps positioning:
- It applies to all non-tank units (healer + dps)
- There is no fixed position to reach, it's more of a range/radius check

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "bronjahm");
+    if (!boss)
+        return false;
+
+    if (bot->getClass() == CLASS_HUNTER)
```
Bobblybook: Be careful with these sort of manual movements, avoid if possible.
If you need to use them, try and make use of min(x, y) to clamp the minimum movements.
Imagine a case where the hunter is at 11.0 unit range. He will move inwards 8 units and end up in melee range and start melee attacking. There needs to be some logic to retain a minimum range for hunters. I don't think this logic is set up correctly, this seems to be giving different classes different size move commands, what you really want to be doing is adjusting the final stopping distance per class instead.

Since the boss is being tanked in a static location, there's really no need to be doing these manual movements in small increments. Rather calculate the correct move distance in one go and perform a single move per character.

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
```diff
@@ -0,0 +1,99 @@
...
+        return Move(bot->GetAngle(boss), 5.0f);
+    }
+}
+
+bool BronjahmDpsPositionAction::isUseful()
+{
+    if (bot->GetExactDist2d(BRONJAHM_TANK_POSITION) <= 10.0f)
+        return false;
+
+    return botAI->IsDps(bot) || botAI->IsHeal(bot);
```
Bobblybook: Can be switched to IsTank() check and moved to Triggers

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.h
```diff
@@ -0,0 +1,55 @@
+#ifndef _PLAYERBOT_WOTLKDUNGEONFOSACTIONS_H
+#define _PLAYERBOT_WOTLKDUNGEONFOSACTIONS_H
+
+#include "Action.h"
+#include "AttackAction.h"
+#include "PlayerbotAI.h"
+#include "Playerbots.h"
+#include "FosTriggers.h"
+
+const Position BRONJAHM_TANK_POSITION = Position(5297.9204f, 2506.698f, 686.06793f);
```
Bobblybook: Just a small nitpick but there's not really a reason to use 5 significant figs for z coordinate here, and all three coords have different sigfigs.
I tend to stick to two or three digits only, but regardless of what you choose, I'd probably keep each value to the same length for consistency.

src/strategy/dungeons/wotlk/forgeofsouls/FosMultipliers.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "FosActions.h"
+
+
+float BronjahmMultiplier::GetValue(Action* action) {
+    Unit* boss = nullptr;
+    boss = AI_VALUE2(Unit *, "find target", "bronjahm");
+    if (!boss)
+        return 1.0f;
+    if (botAI->IsTank(bot))
+        if (dynamic_cast<TankAssistAction*>(action))
```
Bobblybook: Can be combined with above line to:
`if (botAI->IsTank(bot) && dynamic_cast<TankAssistAction*>(action))`

However I don't know that you need to check IsTank(), no other character should be performing TankAssistAction, so you can probably just check that directly. Either works

src/strategy/dungeons/wotlk/forgeofsouls/FosMultipliers.cpp
```diff
@@ -0,0 +1,55 @@
...
+            return 0.0f;
+
+    if (boss && boss->HasUnitState(UNIT_STATE_CASTING) && boss->FindCurrentSpellBySpellId(SPELL_CORRUPT_SOUL))
```
Bobblybook: No need to do nullptr check here, it's already been handled on line 13

src/strategy/dungeons/wotlk/forgeofsouls/FosMultipliers.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "FosActions.h"
+#include "GenericSpellActions.h"
+#include "ChooseTargetActions.h"
+#include "MovementActions.h"
+#include "FosTriggers.h"
+#include "FosActions.h"
+
+
+float BronjahmMultiplier::GetValue(Action* action) {
+    Unit* boss = nullptr;
```
Bobblybook: Don't need to pre-init pointer here:
`Unit* boss = AI_VALUE2(Unit*, "find target", "bronjahm");`

Will return nullptr if target not found

src/strategy/dungeons/wotlk/forgeofsouls/FosMultipliers.cpp
```diff
@@ -0,0 +1,55 @@
...
+        {
+            return 0.0f;
+        }
+    }
+    return 1.0f;
+}
+
+float AttackFragmentMultiplier::GetValue(Action* action)
+{
+    if (botAI->IsHeal(bot) && dynamic_cast<AttackCorruptedSoulFragmentAction*>(action))
```
Bobblybook: Logic should be moved to triggers, `AttackCorruptedSoulFragmentAction` should never be called for healers and you shouldn't need to suppress it in a multiplier

src/strategy/dungeons/wotlk/forgeofsouls/FosMultipliers.cpp
```diff
@@ -0,0 +1,55 @@
...
+    {
+        Unit* unit = botAI->GetUnit(*i);
+        if (unit && unit->GetEntry() == NPC_CORRUPTED_SOUL_FRAGMENT)
+        {
+            fragment = unit;
+            break;
+        }
+    }
+
+    if (fragment && dynamic_cast<BronjahmDpsPositionAction*>(action))
```
Bobblybook: What are the purposes of these two checks - it seems like you are suppressing these two behaviours based on fragments being alive?

src/strategy/dungeons/wotlk/forgeofsouls/FosMultipliers.h
```diff
@@ -0,0 +1,24 @@
...
+class BronjahmMultiplier : public Multiplier
+{
+    public:
+    BronjahmMultiplier(PlayerbotAI* ai) : Multiplier(ai, "bronjahm") {}
+
+    public:
+        virtual float GetValue(Action* action);
+};
+
+class AttackFragmentMultiplier : public Multiplier
```
Bobblybook: Should be removed, this logic can be baked into the action/trigger in a one or two-liner

src/strategy/dungeons/wotlk/forgeofsouls/FosStrategy.h
```diff
@@ -0,0 +1,16 @@
+#ifndef _PLAYERBOT_WOTLKDUNGEONFOSSTRATEGY_H
+#define _PLAYERBOT_WOTLKDUNGEONFOSSTRATEGY_H
+#include "Multiplier.h"
+#include "Strategy.h"
+
+class WotlkDungeonFosStrategy : public Strategy
+{
+public:
+    WotlkDungeonFosStrategy(PlayerbotAI* ai) : Strategy(ai) {}
+    std::string const getName() override { return "fos"; }
```
Bobblybook: Better to expand this out to "forge of souls" - it's almost impossible to understand what "fos" would mean in the strategy list when you print out all the existing strategies, unless you happen to remember the dungeon acronyms

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.cpp
```diff
@@ -0,0 +1,44 @@
+#include "FosTriggers.h"
+#include "Playerbots.h"
+#include "AiObject.h"
+#include "AiObjectContext.h"
+
+bool MoveFromBronjahmTrigger::IsActive()
+{
+    Unit* boss = nullptr;
+    boss = AI_VALUE2(Unit*, "find target", "bronjahm");
```
Bobblybook: Can be combined with above line

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.cpp
```diff
@@ -0,0 +1,44 @@
...
+    }
+    return false;
+}
+
+bool SwitchToSoulFragment::IsActive()
+{
+    Unit* fragment = nullptr;
+    GuidVector targets = AI_VALUE(GuidVector, "possible targets no los");
+
+    for (auto i = targets.begin(); i != targets.end(); ++i)
```
Bobblybook: Can be iterated as per other example

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "AiObject.h"
+#include "AiObjectContext.h"
+
+bool MoveFromBronjahmTrigger::IsActive()
+{
+    Unit* boss = nullptr;
+    boss = AI_VALUE2(Unit*, "find target", "bronjahm");
+    if (boss && boss->HasUnitState(UNIT_STATE_CASTING))
+    {
+        if (boss->FindCurrentSpellBySpellId(SPELL_CORRUPT_SOUL))
```
Bobblybook: Can be combined with the above check as there's no further logic forks needed, no need to add an extra depth here. We don't really even need to check both conditions, the spellId should be enough on it's own, but it's fine to check both.

Actually you may be able to replace this with a self-aura check as there is likely a debuff associated with the spell, have you had a look into that?

src/strategy/dungeons/wotlk/forgeofsouls/FosStrategy.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "FosMultipliers.h"
+
+void WotlkDungeonFosStrategy::InitTriggers(std::vector<TriggerNode*>& triggers) {
+    triggers.push_back(
+        new TriggerNode("move from bronjahm",
+                        NextAction::array(0, new NextAction("move from bronjahm", ACTION_MOVE + 5), nullptr)));
+    triggers.push_back(new TriggerNode(
+        "switch to soul fragment", NextAction::array(0, new NextAction("attack corrupted soul fragment", ACTION_RAID + 1), nullptr)));
+    triggers.push_back(new TriggerNode("bronjahm position",
+        NextAction::array(0, new NextAction("bronjahm tank position", ACTION_RAID + 1),
```
Bobblybook: These should probably not be queued like this, and tank target is not really a necessary action, you can do dps + tank + heal targeting in a single action

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.h
```diff
@@ -0,0 +1,40 @@
+#ifndef _PLAYERBOT_WOTLKDUNGEONFOSTRIGGERS_H
+#define _PLAYERBOT_WOTLKDUNGEONFOSTRIGGERS_H
+
+#include "Trigger.h"
+#include "PlayerbotAIConfig.h"
+#include "GenericTriggers.h"
+#include "DungeonStrategyUtils.h"
+
+enum FosIDs
```
Bobblybook: Should be expanded to avoid conflicts in the future as this is not in it's own namespace

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.cpp
```diff
@@ -0,0 +1,44 @@
...
+        {
+            return true;
+        }
+    }
+
+    return false;
+    
+}
+
+bool BronjahmPositionTrigger::IsActive()
```
Bobblybook: This check can be reduced to a single line:
`return bool(AI_VALUE2(Unit*, "find target", "bronjahm");`

Tank checks can also be added here, or this trigger scrapped entirely - this is not really a position trigger, it's just checking for the boss so it should probably be renamed and/or merged into something else

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.h
```diff
@@ -0,0 +1,40 @@
...
+{
+    // Boss1
```
Bobblybook: Needs to be updated to boss name

Bobblybook (changes requested):
Hi, thanks for working on this - have reviewed and added some notes to relevant sections

src/strategy/dungeons/wotlk/forgeofsouls/FosActions.cpp
LFTEC: i will check all isUseful

LFTEC:
@Bobblybook , I am very pleased to receive your feedback on the PR. Thank you very much for your suggestions, and I will make the necessary corrections based on them. Meanwhile, I have two questions, and if you have time, could you please answer them:
1. For the boss Bronjahm, all classes need to melee position themselves to deal damage, while any bot targeted by the boss should move away from the group. How can these two positioning actions be switched efficiently?
2. The tank assist strategy involves the tank generating threat on each NPC, but the adds spawned by this boss don't require tank intervention. Is there a more effective method aside from the tank assist strategy?

Bobblybook:
Great, thanks!

> 1. For the boss Bronjahm, all classes need to melee position themselves to deal damage, while any bot targeted by the boss should move away from the group. How can these two positioning actions be switched efficiently?

If all group members need to stack (including ranged, melee, tank, healer), probably the best logic to implement is the following. I typically try and implement the actions without many checks, you can assume that if the code execution reaches action::Execute(), then the action needs doing. You can implement filtering logic to prevent the actions at an earlier stage, eg. in the trigger, multiplier or IsUseful():
- Create a stacking action that will run the bot close to the boss (suppress this if bot is already suitably close, and handle exact distances for hunter vs. everyone else)
- Create a kiting action or similar that will run the bot away from the boss
- Create trigger for boss stacking action - this can be something general like "boss exists" if this behaviour needs to be active for the whole fight, otherwise use some logic to identify the phase where it is needed
- Create trigger for kiting based on the debuff or aura or something on the player that you can use to identify the target
- Nullify stacking trigger/behaviour for bots that are needing to kite away. This can be done either in the trigger for the stacking action, or in the multiplier. For this sort of thing, the multiplier is often the only way to do it because in the trigger, you only have access and visibility to the trigger and subsequent action you are about to raise - you can prevent your own action from going out, but you can't affect anything about the bot's other actions. If you are needing to also suppress other behaviours that you are not the cause of (eg. melee bot trying to reach the target to dps, or ranged dps trying to get away from the boss back to range, which are both part of the general bot AI) then you must do this in the multiplier. If you have to suppress actions like this, try and identify as specific of an action class as you can to suppress, to surgically remove certain behaviours, rather than disabling entire parent classes of actions. The more specific the better.
I can help you with this if you like, just let me know.

> 2. The tank assist strategy involves the tank generating threat on each NPC, but the adds spawned by this boss don't require tank intervention. Is there a more effective method aside from the tank assist strategy?
The TankAssist actions are the ones responsible for targeting adds and keeping aggro on alternate targets. If you want the tank to focus the boss and ignore adds then just nullifying TankAssist actions is probably the cleanest way to do it

LFTEC:
@Bobblybook,  I have fixed the issues and recorrect some parts of the code based on your suggestions. I have committed the latest code; please check it again and provide your valuable feedback.

src/strategy/dungeons/wotlk/forgeofsouls/ForgeOfSoulsActions.h
```diff
@@ -0,0 +1,37 @@
...
+};
+
+class AttackCorruptedSoulFragmentAction : public AttackAction
+{
+public:
+    AttackCorruptedSoulFragmentAction(PlayerbotAI* ai) : AttackAction(ai, "attack corrupted soul fragment") {}
+    bool Execute(Event event) override;
+};
+
+class BronjahmGroupPositionAction : public AttackAction
```
Bobblybook, line 26: Positioning action should probably be a MovementAction, see .cpp notes

src/strategy/dungeons/wotlk/forgeofsouls/ForgeOfSoulsActions.cpp
```diff
@@ -0,0 +1,87 @@
...
+        if (AI_VALUE2(bool, "has aggro", "current target"))
+            if (bot->GetExactDist2d(BRONJAHM_TANK_POSITION) > 5.0f)
+                return MoveTo(bot->GetMapId(), BRONJAHM_TANK_POSITION.GetPositionX(),
+                              BRONJAHM_TANK_POSITION.GetPositionY(), BRONJAHM_TANK_POSITION.GetPositionZ(), false,
+                              false, false, true, MovementPriority::MOVEMENT_NORMAL);
+            else
+                return Attack(boss);
+        else
+        {
+            return Attack(boss);
```
Bobblybook, line 62: This probably shouldn't be happening here, the tank will manage his target on his own, you do not want to be mixing attack commands in with the movement and positioning.
See IonarTankPositionAction for HoL, you should only need to move the tank and keep him there, the AI should do the rest: https://github.com/liyunfan1223/mod-playerbots/blob/7fa1ab36a3af189cb439d318fdf0fa67d849fe6e/src/strategy/dungeons/wotlk/hallsoflightning/HallsOfLightningActions.cpp#L111

Bobblybook (changes requested):
Thanks, all looks good except you are sending attack commands in a positioning action and I can see you've changed the parent class to AttackAction.. was there any reason for doing that? It shouldn't be necessary

src/strategy/dungeons/wotlk/forgeofsouls/FosTriggers.cpp
LFTEC: I tried using only the move command to position the tank to a specified location. However, when the adds appear, the tank moves to the adds to hold aggro and then moves back to the specified position. At this point, the boss follows the tank, reducing the distance between the adds and the boss. Therefore, I attempted to remove the tank assist strategy during the fight.
However, after removing the tank assist strategy, the tank just stands still without taking any action, so I used the attack command to make the tank attack only the boss. I know this isn't a perfect solution. Is there any other way to solve this issue?

Meanwhile, To solve the following issue, I combined the tanktarget and tankattack actions, which is why the move command is used in the AttackAction.

> These should probably not be queued like this, and tank target is not really a necessary action, you can do dps + tank + heal targeting in a single action
https://github.com/liyunfan1223/mod-playerbots/pull/672#discussion_r1827113840

src/strategy/dungeons/wotlk/forgeofsouls/ForgeOfSoulsActions.h
LFTEC: I tried using only the move command to position the tank to a specified location. However, when the adds appear, the tank moves to the adds to hold aggro and then moves back to the specified position. At this point, the boss follows the tank, reducing the distance between the adds and the boss. Therefore, I attempted to remove the tank assist strategy during the fight.
However, after removing the tank assist strategy, the tank just stands still without taking any action, so I used the attack command to make the tank attack only the boss. I know this isn't a perfect solution. Is there any other way to solve this issue?

Meanwhile, To solve the following issue, I combined the tanktarget and tankattack actions, which is why the move command is used in the AttackAction.

> These should probably not be queued like this, and tank target is not really a necessary action, you can do dps + tank + heal targeting in a single action
https://github.com/liyunfan1223/mod-playerbots/pull/672#discussion_r1827113840

Bobblybook: It sounds like there may be an error in your logic checking or multiplier if this is happening.
But assuming you've tested this, we can probably merge this and get people testing it and providing feedback.

I'll look into it once merged to see where we can correct the logic flow.

Thanks again for the PR!

Bobblybook (approved):
Needs some logic correction on tank positioning/attack targeting

Issue #649, opened 2024-10-29
Playerbots not levelling


Gherry777:
**Describe the bug**

Hello, I'm playing around with the module and I saw that it works perfectly while I'm in the zone, but as soon as I leave the zone, the bots stop levelling.

For example I created a new human character, I started playing in Northshire Abbey, made a bunch of level 1 bots and when I hit level 3, there were bots level 1, bots level 2 and few bots hit level 3 in a matter of minutes.
They were levelling just like me, doing their quests.

As soon as I left Nothshire Abbey, bots basically stopped being alive.
I see them on /who, but they're always the same level, so I imagine they're just doing nothing.
No levelling up in few hours.

**Commit hash**
The hash of the current commit.

**To Reproduce**
Just installed azerothcore-woltk and added playerbot and ah-bot modules

**Expected behavior**
I don't know if this is the intended behavior or not, I just assumed the bots would populate the world and keep doing their quests and levelling all the time.
If this is correct, feel free to just close the bug report.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Debian 12 bookworm
![bug-report-playerbot](https://github.com/user-attachments/assets/83355810-8af6-417b-8cd8-a4eebf1823fb)

lidstah:
Hi,

I've noticed the same behavior, but once the bots leave their starting zone (around level 5) they start leveling up as intended. You can notice it if you make a level 1 alt: all the starting zone bots will briefly appear as Away once you login before starting doing quests and killing mobs (which is quite fun to observe, by the way).

My two cents is that it is intended behavior as new players/alts will not be alone in their starting zone (until all bots hit level 5+ and start wandering the world and levelup normally).

hermensbas:
Its working as intended, the core and your hardware doesnt have the resources to have all bots active all the time. But that being that i am working on more workable solution.

hermensbas:
Try the latest version since the backport has been removed/rewritten. If it works lemme know so i can close this issue for now.

danielcoh92:
Hi,
This is what happens for me with the latest build (updated a few hours ago):
![image](https://github.com/user-attachments/assets/ce14fdf1-9f6b-448a-a4fb-7431b481c7fe)
As you can see, some bots are still afk.
I had a min bots of 500 and max bots of 1000 and they were ALL AFK.
After adjusting min=max, the result is in the screenshot I attached.

Looks like it's still not working properly?

hermensbas:
[...]

PR #678, opened 2024-11-04
[performance] bots scale idle fix

hermensbas:
Rebuild the scaling logic, tested with various hardware and config setups with rotationBots, without rotationBots, min/max equal and not equals. 10, 50, 2000, 4000, 5000 bots. 

1. idle bots 'fix'
2. gentle scaling when performance is on expected levels
3. hard scaling when performance isnt sufficient
4. init phase;  seperated allowedActive and login and first randomize and teleport boosters
5. init phase; added system call for expected init time of bots.
6. pushed the updateInterval back to 20 and compensated with booters to temporary lowering it for the init phase.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -728,7 +728,7 @@ AiPlayerbot.FastReactInBG = 1
 #
 
 # All In seconds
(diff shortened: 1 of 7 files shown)


hermensbas:
#649

PR #681, opened 2024-11-05
Update LfgAccept

EricksOliveira:
Fixed issue with Bots refusing to join DungeonFinder.

### DIFF
--- a/src/strategy/actions/LfgActions.cpp
+++ b/src/strategy/actions/LfgActions.cpp
@@ -188,7 +188,7 @@ bool LfgAcceptAction::Execute(Event event)
             LOG_INFO("playerbots", "Bot {} {}:{} <{}> is in combat and refuses LFG proposal {}",
                      bot->GetGUID().ToString().c_str(), bot->GetTeamId() == TEAM_ALLIANCE ? "A" : "H", bot->GetLevel(),
                      bot->GetName().c_str(), id);
-            sLFGMgr->UpdateProposal(id, bot->GetGUID(), false);
+            sLFGMgr->UpdateProposal(id, bot->GetGUID(), true);
             return true;
         }
 


hermensbas:
did you test it?

EricksOliveira:
Yes. works well for me.

mw108:
Thanks a lot for the fix!

But there are still some weird things going on.

Queue times are very long and when you get an invite 3 out of 5 times some bot doesn't confirm the invitation and the queue times out. And when they finally all accepted they instantly leave the group again:

image

EricksOliveira:
Good. I did some tests and my search in DF takes about 1 to 2 minutes. Then everyone accepts and teleports to the Dungeon normally. You have to check if your bots are AFK. I made more than 5 Dungeon Finder with online Bots only.

mw108:
How do I check if the bots are AFK? Bots shouldn't be AFK?

How do I check the build of the bots? Can I change the distribution of builds of the bots? Like 33% Tanks, 33% DDs, 33% Healers?

Why are the bots instantly leave the dungeon after they accepted? This is confusing.

`/who` lists only players of your own faction. From the screenshot above there were at least 10 bots eligible to be a tank. Difficulty in normal DF isn't of that high importance, so even DDs could queue up as tanks. What do you think?

hermensbas:
How to prevent bots going afk, they only go afk when your hardware resources or the pushing the limits of the amount of bots your setup can carry

Dreathean:
> How do I check the build of the bots? Can I change the distribution of builds of the bots? Like 33% Tanks, 33% DDs, 33% Healers?
> 
The bots are tanks/dd/healers based on spec. IE Holy priest will always have the healer role. You can change the distribution of specs in the config file.

EricksOliveira:
I'm making some improvements. I still found some problems, especially when searching for Specs. I believe that I will be posting some updates today for correction.

mw108:
> You can change the distribution of specs in the config file.

Where? I don't find anything relevant when searching for dist*, tank, prot*, heal*, role, etc.

blinkysc:
Just qued and 3 bots left from Azjol Nerub. Not the biggest issue becuase just look for more and end up getting replacements. Just wanted to let you know

EricksOliveira:
I already discovered the real problem. I'm working on fixing it.

mw108:
> I already discovered the real problem. I'm working on fixing it.

Thank you very much. Really appreciating your work and efforts you put into this project!

Dreathean:
You need to scroll down further right past where it lists the specs, to the probabilities - 
image

And I'm sure you already saw the specs themselves which line up with the probabilities below:
image

Dreathean:
Just to be clear, it's changing the probabilities of the specs directly and therefore the roles indirectly.

mw108:
> You need to scroll down further right past where it lists the specs, to the probabilities 

Cheers! Hidden in plain sight. 😊

mw108:
> I already discovered the real problem. I'm working on fixing it.

Any updates?

EricksOliveira:
> > Eu já descobri o problema real. Estou trabalhando para consertá-lo.
> 
> Alguma atualização?

Yes. https://github.com/liyunfan1223/mod-playerbots/pull/690

PR #683, opened 2024-11-05
Revert "disabled smartscale by default for now (#663)"

hermensbas:
Enables smartscale as default again, can be applied when proven/tested the current scaling is doing his job as intended.

I've run every possible test i can think of literally 100's of build and variants of config setups. But let the community decide, or atleast the owner.


hermensbas:
Ye was kinda having doubts either direction. Sounds like a good idea. I will close the the PR

PR #690, opened 2024-11-06
Fix LfgAction

EricksOliveira:
Fix Bot Leave PT in Join Dungeon.

### DIFF
--- a/src/strategy/actions/LfgActions.cpp
+++ b/src/strategy/actions/LfgActions.cpp
@@ -32,7 +32,7 @@ uint32 LfgJoinAction::GetRoles()
         case CLASS_DRUID:
             if (spec == 2)
                 return PLAYER_ROLE_HEALER;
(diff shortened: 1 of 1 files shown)


EricksOliveira:
> This will cause random robots to remain in the team. If the player quits, the robot cannot be teamed up with other players. This is not in line with normal player behavior.

Couldn't identify this problem. But I will carry out some more tests.

hermensbas:
Any progression or should this PR be on draft?

EricksOliveira:
Yes. I had a great recovery. However, I ask that you wait a little longer, as I am doing more tests regarding the specs and searches in the DG. More in relation to DG's Bot Kitar, it was resolved. I'm just improving the code. Time is too short for me to work constantly.

hermensbas:
hehe, yup time time. No hurry right? I will set on draft you can turn switch from draft once your rdy. So it doesnt get merged by mistake while ur not rdy.

PR #702, opened 2024-11-09
Do not disband alt bots group when master goes away

EricksOliveira:


hermensbas:
@noisiver whats your opinion on this? Shouldnt alt-bots group disband after the master leaves?

noisiver:
It's hard to say. If there are other players, should one of them become the new master or not? It depends a lot on the players and doesn't really have a one-size-fits-all solution. If there are no other players it makes no sense not to disband the group.

noisiver:
It shouldn't disband the group if the master actually is in the group though. I mean.. I log out with altbots in my party/raid all the time, I wouldn't want it to disband the group just because I go offline when they do. They won't log in again without me so as long as you're actually online (and in the group) when they are it shouldn't disband the group.

arcaneintellect:
I would love something like this implemented, bonus if its behind a config or something for those who want it to auto disband. But at this time I cant come up with a viable strategy for swapping altbots in the same dungeon/raid without them teleporting to the graveyard.

Dreathean:
Just to be clear, does "master goes away" refer only to logging out and staying in group, or to leaving group? Because if it's the former, I agree that as long as the master doesn't leave the group, then altbots should not auto-disband - for convenience like Revision said and for swapping purposes like arcaneintellect said. If the latter, then yeah they should disband.

EricksOliveira:
I need to read the code, I don't remember the problem anymore. lol

Grimfeather:
related to: https://github.com/liyunfan1223/mod-playerbots/issues/679

Just want to reaffirm the need to fix this issue.
It's a big problem. If you play with your own characters as alt bots it causes a lot of frustration.

When you are in a dungeon as a 5-man party, you and 4 alt bots, when you want to switch to one of the other 4 characters you will experience this problem. Your characters will get teleported to a graveyard.

It gets worse if you're in a raid with your own characters as alt bots. Especially if your characters are on different accounts.
Let's say you're raiding ZG with 19 alt bots. You get to a boss and like to switch to your druid healer on your 2nd account.
Major issue now, because your whole raid just disbanded.

Vortikai:
Seconded. I'd wish to help contribute but all I can do is say it's a much desired feature.

EricksOliveira:
Now it's optional.

Issue #713, opened 2024-11-14
Config "AiPlayerbot.RandomBotFixedLevel" not working


Regrad:
Bots continue to gain experience and level up when this option is enabled.

PR #714, opened 2024-11-14
Fixed issue with random bots not respecting RandomBotFixedLevel

avirar:
#713

Added check during random bot login for `RandomBotFixedLevel`, set `PLAYER_FLAGS_NO_XP_GAIN` if the config option is found, remove flag if it is not.

### DIFF
--- a/src/RandomPlayerbotMgr.cpp
+++ b/src/RandomPlayerbotMgr.cpp
@@ -2248,7 +2248,7 @@ void RandomPlayerbotMgr::OnBotLoginInternal(Player* const bot)
     {
         bot->SetPlayerFlag(PLAYER_FLAGS_NO_XP_GAIN);
     }
-    else if (!sPlayerbotAIConfig->randomBotFixedLevel)
+    else
     {
         bot->RemovePlayerFlag(PLAYER_FLAGS_NO_XP_GAIN);
     }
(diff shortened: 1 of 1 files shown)


src/RandomPlayerbotMgr.cpp
```diff
@@ -2243,6 +2243,15 @@ void RandomPlayerbotMgr::OnBotLoginInternal(Player* const bot)
 {
     LOG_INFO("playerbots", "{}/{} Bot {} logged in", playerBots.size(), sRandomPlayerbotMgr->GetMaxAllowedBotCount(),
              bot->GetName().c_str());
+    
+    if (sPlayerbotAIConfig->randomBotFixedLevel)
+    {
+        bot->SetPlayerFlag(PLAYER_FLAGS_NO_XP_GAIN);
+    }
+    else if (!sPlayerbotAIConfig->randomBotFixedLevel)
```
Bobblybook: Is there any reason to check the negation here? You're checking the inverse of the bool you just checked, an "else" should cover the same logic without an additional check unless I'm misunderstanding this block

avirar: Good catch, updated. I had just copied the existing code from PlayerbotMgr.cpp; Lines 552-559

PR #717, opened 2024-11-15
Allows the player to use a random bot as ATM

mw108:
Attempt to fix the issue that random bots refuse to just give the player money in a trade session.

### DIFF
--- a/src/strategy/actions/TradeStatusAction.cpp
+++ b/src/strategy/actions/TradeStatusAction.cpp
@@ -147,30 +147,30 @@ bool TradeStatusAction::CheckTrade()
     if (!bot->GetTradeData() || !trader->GetTradeData())
         return false;
 
+	bool isGivingItem = false;
+	for (uint32 slot = 0; slot < TRADE_SLOT_TRADED_COUNT; ++slot)
+	{
+		Item* item = bot->GetTradeData()->GetItem((TradeSlots)slot);
+		if (item)
+		{
+			isGivingItem = true;
+			break;
+		}
+	}
+
+	bool isGettingItem = false;
+	for (uint32 slot = 0; slot < TRADE_SLOT_TRADED_COUNT; ++slot)
+	{
+		Item* item = trader->GetTradeData()->GetItem((TradeSlots)slot);
+		if (item)
+		{
+			isGettingItem = true;
+			break;
+		}
+	}
+
     if (!botAI->HasActivePlayerMaster() && GET_PLAYERBOT_AI(bot->GetTrader()))
     {
-        bool isGivingItem = false;
-        for (uint32 slot = 0; slot < TRADE_SLOT_TRADED_COUNT; ++slot)
-        {
-            Item* item = bot->GetTradeData()->GetItem((TradeSlots)slot);
-            if (item)
-            {
-                isGivingItem = true;
-                break;
-            }
-        }
-
-        bool isGettingItem = false;
-        for (uint32 slot = 0; slot < TRADE_SLOT_TRADED_COUNT; ++slot)
-        {
-            Item* item = trader->GetTradeData()->GetItem((TradeSlots)slot);
-            if (item)
-            {
-                isGettingItem = true;
-                break;
-            }
-        }
-
         if (isGettingItem)
         {
             if (bot->GetGroup() && bot->GetGroup()->IsMember(bot->GetTrader()->GetGUID()) &&
@@ -182,10 +182,12 @@ bool TradeStatusAction::CheckTrade()
         }
         return isGettingItem;
     }
+	
     if (!bot->GetSession())
     {
         return false;
     }
+	
     uint32 accountId = bot->GetSession()->GetAccountId();
     if (!sPlayerbotAIConfig->IsInRandomAccountList(accountId))
     {
@@ -235,6 +237,17 @@ bool TradeStatusAction::CheckTrade()
 
     if (!botMoney && !playerMoney)
         return true;
+	
+	// Player just wants money from the bot
+    if ( ! isGettingItem && ! isGivingItem && botMoney > playerMoney )
+    {
+        std::ostringstream out;
+        out << "You are a thief!";
+        botAI->TellMaster( out );
+        botAI->PlaySound( TEXT_EMOTE_SIGH );
+        return true;
+    }
+	
 
     if (!botItemsMoney && !playerItemsMoney)
     {
--- a/src/strategy/actions/TradeStatusAction.cpp
+++ b/src/strategy/actions/TradeStatusAction.cpp
@@ -239,12 +239,12 @@ bool TradeStatusAction::CheckTrade()
         return true;
 	
 	// Player just wants money from the bot
-    if ( ! isGettingItem && ! isGivingItem && botMoney > playerMoney )
+    if (sPlayerbotAIConfig->allowRandomBotsAtm && !isGettingItem && !isGivingItem && botMoney > playerMoney)
     {
         std::ostringstream out;
         out << "You are a thief!";
-        botAI->TellMaster( out );
-        botAI->PlaySound( TEXT_EMOTE_SIGH );
+        botAI->TellMaster(out);
+        botAI->PlaySound(TEXT_EMOTE_SIGH);
         return true;
     }
 	
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -491,6 +491,10 @@ AiPlayerbot.RandomBotJoinLfg = 1
 # Need to reset rndbot after changing the setting (.playerbot rndbot reset)
 AiPlayerbot.DisableDeathKnightLogin = 0
 
+# Allow random bots to be used as ATM
+# Players can trade random bots and request an amount of gold from them via whisper command
+AiPlayerbot.AllowRandomBotsAtm = 0
+
 #
 #
 #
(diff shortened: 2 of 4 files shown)


mw108:
image

image

src/strategy/actions/TradeStatusAction.cpp
```diff
@@ -235,6 +237,17 @@ bool TradeStatusAction::CheckTrade()
 
     if (!botMoney && !playerMoney)
         return true;
+	
+	// Player just wants money from the bot
+    if ( ! isGettingItem && ! isGivingItem && botMoney > playerMoney )
```
noisiver: There shouldn't be all these spaces here
`if (!isGettingItem && !isGivingItem && botMoney > playerMoney)`

src/strategy/actions/TradeStatusAction.cpp
```diff
@@ -235,6 +237,17 @@ bool TradeStatusAction::CheckTrade()
...
+    {
+        std::ostringstream out;
+        out << "You are a thief!";
+        botAI->TellMaster( out );
+        botAI->PlaySound( TEXT_EMOTE_SIGH );
```
noisiver: Extra Spaces

src/strategy/actions/TradeStatusAction.cpp
```diff
@@ -235,6 +237,17 @@ bool TradeStatusAction::CheckTrade()
 
     if (!botMoney && !playerMoney)
         return true;
+	
+	// Player just wants money from the bot
+    if ( ! isGettingItem && ! isGivingItem && botMoney > playerMoney )
+    {
+        std::ostringstream out;
+        out << "You are a thief!";
+        botAI->TellMaster( out );
```
noisiver: Extra Spaces

noisiver:
This shouldn't be a default behavior, no. It's open to abuse since the random bots will be given random amount of gold. That would essentially be an endless supply of gold.

Configurable sounds like a reasonable thing if a thing like this is to be implemented.

PaoloBenigni:
I don't think this makes sense to implement. If you have a server where you play solo and want to give yourself money there are ways to do that without bot farming. If you have multiple players then this makes less sense to enable.

mw108:
In the Wiki there was no clear description that trading gold to a player is meant for altbots only. Therefore I assumed that this is a bug. But I understand the issue here. I agree that this shouldn't be the default behavior and configurable.

A scenario, where receiving gold from bots is desired, would be a server like SoloCraft where the aim is to clear the content mainly solo with the help of bots and where economy doesn't really matter.

PaoloBenigni:
Maybe the cmangos implementation documentation mentioned it.

Instead of farming bots why not just gm add a bunch of money?

mw108:
Yes, GM command to give you gold works for players with GM capabilities. I imagine a scenario for players without GM rights.

Dreathean:
We could certainly update the wiki to make it clear that it's intended for use with altbots specifically

hermensbas:
Why should this be a feature of bots? Providing players with gold? Sure there is trade function which can be abused, but that isnt the intend. I personally rather dont see configuration options or features like this.

liyunfan1223:
Agree with others, this shouldn't be the default behavior, but it's ok if you change it to a configurable option.

noisiver:
I don't think that's a good name for it.

You're allowing them to trade you gold, it's not a machine you go up to and insert your card into. 😆

mw108:
> You're allowing them to trade you gold, it's not a machine you go up to and insert your card into. 😆

I think that you can withdraw cash from an ATM with your card is a relict that will perish eventually and be replaced by biometrics or a chip implant or whatever. If cash still exists then. 😃 Also, in the end ATM is just an abbreviation.

noisiver:
I still think it's a bad name for it. I guess @hermensbas and @liyunfan1223 are best at making that judgement so hopefully they can chime in.

src/strategy/actions/TradeStatusAction.cpp
```diff
@@ -235,6 +237,17 @@ bool TradeStatusAction::CheckTrade()
 
     if (!botMoney && !playerMoney)
         return true;
+	
+	// Player just wants money from the bot
+    if (sPlayerbotAIConfig->allowRandomBotsAtm && !isGettingItem && !isGivingItem && botMoney > playerMoney)
```
noisiver, line 242: Isn't this the wrong way around? If the option is enabled it calls the player a thief, doesn't it?

mw108: The bot calls the player a thief because he's forced to give the player money.

noisiver: Yes but `(sPlayerbotAIConfig->allowRandomBotsAtm &&` <- if the option is **enabled** we call the player a thief. Shouldn't it be `(!sPlayerbotAIConfig->allowRandomBotsAtm &&`?

noisiver: Oh dear, I am not reading it right. You are correct, that's my bad. It accepts the trade, not deny it. Sorry. For some reason I thought it rejected the trade here but it doesn't.

mw108:
I've added an option to the config file which is disabled by default.

I'd like to apologize seeing the controversy this implementation sparked. It wasn't my intention to just implement a useless feature.

As I said, I was under the impression that this was a bug which was meant to work correctly due to the missing statement in the Wiki that trading gold to players is only meant for altbots.

You are - of course - not required to merge this PR. I think it is a nice addition though I would like to have for my server.

noisiver:
It's optional and disabled by default so I don't see why it shouldn't be merged.

PaoloBenigni:
I don't understand why it's a good thing to add at all.  I don't buy the logic of implementing a feature, even if disabled by default, if it doesn't add to the game.

Is farming bots for gold a reasonable feature to add? I can't see how this is a good thing in either solo or group play. It's easy enough to give a player a ton of gold without this.

noisiver:
> I don't understand why it's a good thing to add at all. I don't buy the logic of implementing a feature, even if disabled by default, if it doesn't add to the game.
> 
> Is farming bots for gold a reasonable feature to add? I can't see how this is a good thing in either solo or group play. It's easy enough to give a player a ton of gold without this.

- You can enable selfbot and use autogear to give yourself the best gear possible without having to actually do anything to get it
- You can use autogear with rndbots or altbots and use those to "boost" yourself through the content if the above is taking it too far
These points are the equipment equivalent of this but it's there still. A player also can't give themselves gold if they don't have access to the command.

This feature is not something I'd use myself, though I use other means of farming gold which could easily be replaced by GM commands, but if someone wants it it's there.

It's up to the other contributors to make the decision on wheather or not to merge this.

hermensbas:
> * ugh the content if the above

True but both are sides effects:

- self-bot intend i would suspect is there for debug purpose
- autogear prolly two reasons, bots dont always behave and collect the right gear while leveling themself, and if you dont let bots level natural.

While ATM functionality is basically a proxy cheat through the bots, it doesnt involve bots. If someone wants a gold cheat then either GM command it or make a module for it so none-GM player can spawn gold. But using playerbots as proxy cheat seems wrong imo. 

Surely we can add it by using with a config option but in that extend where to we draw the line? Imo honost opinion this should not be merged with all respect.

mw108:
> Is farming bots for gold a reasonable feature to add? I can't see how this is a good thing in either solo or group play. It's easy enough to give a player a ton of gold without this.

I think you're maybe overthinking it a bit too much. I mean you are completely right from the perspective of a public live server with many players and a player driven economy. Such a feature would be catastrophic. But from what I've seen there is no live WotLK private server that even runs mod-playerbots, which I don't understand honestly. There should be demand for it, don't you think? All I see are recommendations for SPP (Single Player Repacks).

Gold shouldn't be put too much weight on. I'm playing on a Vanilla server called "SoloCraft". It features bots you can add to your party or raid and play through the content on your own or with friends / guild mates. It is actually quite fun and the server nicely populated.

During leveling you always have a gold issue to buy skills and spells from class or profession trainers, especially you can learn all trade skills, not only 2. So people always ask around if somebody can send them some gold.

Once you hit raids, gold becomes completely obsolete. Because you can clear the raids with your bots, kick them from the party after each boss and loot 100-200g from the boss and then simply reinvite the bots and continue the raid. This means, during leveling you always have a personal gold scarcity and when you are 60 gold has no meaning anymore.

And it is not about farming bots for gold or change the game settings in a ways that will give you absurd amounts of gold. It is just about having gold on demand when you need it, for training skills and spells. At least that's what I use the feature for.

Dreathean:
I guess for my two cents, first I want to say that I appreciate anyone that takes the time to develop anything for this project, so thank you for what you’ve done. However I do agree with the others that there are already many options in place to freely give people gold and I don’t personally see the benefit of adding the option to take from all rndbots which are necessarily given unlimited gold. 

Yes you can always add the config like you have, but I guess there becomes the concern with bloat on a conf that is already quite large. 

The two main options that already exist are 1) increase gold drop rate in worldserver conf and 2) congratsonlevel module which you can adjust gold given to players based on level (ie 10 gold given for reaching level 20).

hermensbas:
@mw108 would you hate us if we decline the merge?

PaoloBenigni:
I do like adding in a response to the attempt to trade gold without anything.

mw108:
> @mw108 would you hate us if we decline the merge?

No, of course not. 😊 

> I do like adding in a response to the attempt to trade gold without anything.

There is a response. It says `There are no items to trade`

hermensbas:
We all ppl working and spending our precious spare time on this kinda things and we absolutely dont wanna discourage anyone. Maybe we can discuss these type of changes on beforehand when in doubt. We have a discord with a development channel :)

PR #745, opened 2024-11-29
Add a config option for dropping obsolete quests

noisiver:
This needs testing and I don't have the time to do it right now. Until that has been done it shouldn't be merged.

I decided not to add a print when the option is disabled since it runs every so often and it would most likely get very annoying.

This is helpful when completing quests for reputation and so on. The quests are often grey/gray when doing it and they'll constantly drop them unless it's possible to disable it.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -368,6 +368,10 @@ AiPlayerbot.SyncQuestWithPlayer = 1
 # Default: 0 (disabled)
 AiPlayerbot.SyncQuestForPlayer = 0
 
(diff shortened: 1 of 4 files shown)


hermensbas:
Does this still require additional testing?

noisiver:
I'll leave that for others to decide. I've tested it on my end and toggled off they don't drop grey/gray quests but they do if I enable it. I haven't tested every possible situation, nor could I, though.

hermensbas:
ive read the code, they isnt much that can go wrong tbh. esp considered the default mode.

PR #747, opened 2024-11-29
fix small typo and remove double logging

Raz0r1337:
Fix small Typo in playerbot.conf.

Appender.Playerbots=2,5,0,Playerbots.log,w
Logger.playerbots=5,Console Playerbots
These loggers already exist in the worldserver.conf.
These duplicate entries interfere with each other.


noisiver:
I personally think it makes more sense to have it in the module but not the core. However, if the module isn't present I imagine it could cause issues. At least if it existed only in the module it would reduce the core changes at least a tiny bit.

I'm sure someone with more knowledge on the matter can chime in.

Raz0r1337:
Since the modified core is not supposed to be used without the module and therefore both belong together, I don't see it that way. If the player bots only consisted of the module and you had to use the vanilla core, then I would agree with you.

I also made a pull request to the core that puts the logging in the right position in the config. If both are adopted then the module will be better integrated into the core.

https://github.com/liyunfan1223/azerothcore-wotlk/pull/29

noisiver:
If they're meant to be used together it makes more sense to keep it in the module rather than the core.

We've had many conversations about trying to migrate stuff out of the core because ideally it shouldn't require a custom fork.

Additional hooks used by the module that are added to the core aren't impossible to get merged into master but what we have now is 99.99% sure not to be merged into master, ever.

That is why I think @liyunfan1223 should be the deciding factor as they know the code better than anyone.

PR #753, opened 2024-12-01
Warlock soulstone action

kadeshar:


kadeshar:
In wotlk combat ress is limited? I thought it started in MoP.

noisiver:
I mean if combat ress (rebirth is the only one aside from soulstone) cancels soulstone it means you're losing a rebirth when you shouldn't. I'm only asking because I'm sure we all know how they like to throw those out whenever someone dies. If you don't know it's fine, I can test it, and it doesn't affect this PR; Just curiousity.

PR #759, opened 2024-12-03
[WIP] New rpg strategy

liyunfan1223:
I plan to implement a new rpg strategy for rndbots from scratch as the old rpg strategy has the following problems and is hard to modify.
- Bots are directly teleported to the mob's location instead of moving there by itself, lacking the ability to make long distances movement.
- Too less teleportation points causing the bot to be unevenly distributed on the map.
- Bots are overly attracted to friendly NPCs, causing most bots to just move between NPCs meaninglessly.
- Excessive overhead is incurred when selecting an rpg target.

Todo:
- [x] Now they can go to distant locations for grinding, and the distribution on the map is more even.
- [x] Multiple states are designed to control the behavior of bots to make them more purposeful, such as going to the grind location, returning to the innkeeper, exploring surrounding NPCs, etc.
- [x] Control the status of bots according to probability and simulate real servers. 
- [ ] Bots are now teleported to a innkeeper location, which means it always starts from a innkeeper.
- [ ] Set the proportion of bots in capital cities to simulate real servers.

Optional:
- [ ] Cancle teleport and make it entirely up to the bot to move and transport to the appropriate leveling location.
- [ ] Quest-related behaviors, such as accepting tasks, trying to complete some simple quests (such as kill mob, pick gameobject) and choosing approptiate rewards.
- [ ] More careful grind target selection strategies and group behavior to avoid death.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -29,6 +29,7 @@
 #include "MotionMaster.h"
 #include "MoveSpline.h"
 #include "MoveSplineInit.h"
+#include "NewRpgStrategy.h"
 #include "ObjectGuid.h"
 #include "PerformanceMonitor.h"
 #include "Player.h"
@@ -774,7 +775,8 @@ void PlayerbotAI::Reset(bool full)
     bot->GetMotionMaster()->Clear();
     // bot->CleanupAfterTaxiFlight();
     InterruptSpell();
-
+    rpgInfo = NewRpgInfo();
+    
     if (full)
     {
         for (uint8 i = 0; i < BOT_STATE_MAX; i++)
@@ -4137,16 +4139,16 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
 {
     // only keep updating till initializing time has completed,
     // which prevents unneeded expensive GameTime calls.
-    if (_isBotInitializing)
-    {
-        _isBotInitializing = GameTime::GetUptime().count() < sPlayerbotAIConfig->maxRandomBots * 0.12;
+    // if (_isBotInitializing)
+    // {
+    //     _isBotInitializing = GameTime::GetUptime().count() < sPlayerbotAIConfig->maxRandomBots * 0.12;
 
-        // no activity allowed during bot initialization
-        if (_isBotInitializing)
-        {
-            return false;
-        }
-    }
+    //     // no activity allowed during bot initialization
+    //     if (_isBotInitializing)
+    //     {
+    //         return false;
+    //     }
+    // }
 
     // General exceptions
     if (activityType == PACKET_ACTIVITY)
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -588,6 +588,11 @@ AiPlayerbot.RandomBotGroupNearby = 0
 # Default: 1 (enabled)
 AiPlayerbot.AutoDoQuests = 1
 
+# Random Bots will behave more like real players (exprimental)
+# This option will override AutoDoQuests
+# Default: 0 (disabled)
+AiPlayerbot.EnableNewRpgStrategy = 0
+
 # Quest items to leave (do not destroy)
 AiPlayerbot.RandomBotQuestItems = "6948,5175,5176,5177,5178,16309,12382,13704,11000"
 
--- a/src/AiFactory.cpp
+++ b/src/AiFactory.cpp
@@ -631,7 +631,11 @@ void AiFactory::AddDefaultNonCombatStrategies(Player* player, PlayerbotAI* const
             // nonCombatEngine->addStrategy("group");
             // nonCombatEngine->addStrategy("guild");
 
-            if (sPlayerbotAIConfig->autoDoQuests)
+            if (sPlayerbotAIConfig->enableNewRpgStrategy)
+            {
+                nonCombatEngine->addStrategy("new rpg", false);
+            }
+            else if (sPlayerbotAIConfig->autoDoQuests)
             {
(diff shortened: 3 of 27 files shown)


zbhcn:
I'm worried that the new code will cause a lot of crashes

Dreathean:
I'm not sure how else you expect progress to be made? It is not like Liyunfan is rushing this out, and it is posted here as a draft for anyone to test and analyze. And whenever it does get merged, you can always postpone updating.

zbhcn:
> I'm not sure how else you expect progress to be made? It is not like Liyunfan is rushing this out, and it is posted here as a draft for anyone to test and analyze. And whenever it does get merged, you can always postpone updating.

The truly useful test should be a server with more than 30 real players. Just because a single player server doesn't crash doesn't mean a multiplayer server won't crash

Dreathean:
I totally agree, if anyone with such a server is willing to test that, then that would be great. To my knowledge though, none of the current devs have such a server. So unless someone volunteers to test that on their server, well, they do the best they can.

zbhcn:
> I totally agree, if anyone with such a server is willing to test that, then that would be great. To my knowledge though, none of the current devs have such a server. So unless someone volunteers to test that on their server, well, they do the best they can.

I have such a server, but it's difficult for me to test it on the server because instability can cause players to protest and choose to leave

Dreathean:
Can definitely understand the challenges that can cause. My point still stands that you can wait for a while before accepting a big update like this. Also since this all pertains to the rpg strat, you should be able to just remove the rpg strat from rndbots and avoid it altogether.

noisiver:
> > I totally agree, if anyone with such a server is willing to test that, then that would be great. To my knowledge though, none of the current devs have such a server. So unless someone volunteers to test that on their server, well, they do the best they can.
> 
> I have such a server, but it's difficult for me to test it on the server because instability can cause players to protest and choose to leave 

I don't quite buy that. Any update can cause unexpected instability, even updates completely unrelated to playerbots. I think using it as an excuse is wrong. Instabilities usually ends in reverting code until the cause can be identified and resolved and using code from a PR is no different. 

I'm not saying you have to try this, or even that you should, but I don't think the reasoning is valid.

liyunfan1223:
It will be configurable. If you don't have a test condition, just disable it rather than making unfounded doubts.

zbhcn:
> It will be configurable. If you don't have a test condition, just disable it rather than making unfounded doubts.

Anyway, thank you for your hard work

liyunfan1223:
Thank you for your feedback. It seems that new rpg is not enabled as it's disabled by default. `AiPlayerbot.EnableNewRpgStrategy = 1` needs to be set in the config file.

btw I like this visual playermap, I will try it myself.

liyunfan1223:
I teste on my server and it works as expected. At the beginning, using teleport command, the bots start from the innkeeper.

e9c005b8b02c6f627142fe026bb9c9c
Then they start moving, about an hour later:
adc1fd44090624ee93da6233fb37724

The distribution should be much more even than before. The problem I noticed is that bots almost never go to areas where there is no innkeeper, such as Un'Goro. Perhaps in addition to innkeeper, flightmaster should also be considered.

xengine-qyt:
good !

hermensbas:
I love the idea only to use the initial teleport for certain spawns and let the bots travel. The teleported mechanisme currently used in the distribution of the bots. Basically keep teleporting aslong the bots are afk/idle until triggered to active e.g. due combat. Hence we see constant teleports, not a huge fan of that approach but idea kinda works regardless how long ur server is running.

But the bots traveling instead of teleporting nonstop would be amazing if the distribution stays in balance.

noisiver:
> You could consider using teleport as long the bots are still initializing and once done enable the the new rpg strategy? This might prevent bots cluttering while starting the server and a quicker bot distribution.

Without random levels, when the bots initialize they'll all be whatever level is set as the starting level. If all bots start at 1 or even 5 the place they need to be is in the starting zone for their race. You can't distribute that without affecting the "immersion" some expect (i.e only humans in the human starting zone) and even if you mess with that you're not going to change much if at all.

liyunfan1223:
After testing I belive it works as expected, so merged it first but disabled it by default. Even if new rpg is disabled, I have made some modifications to other parts, which may cause some issues.

src/PlayerbotAI.cpp
```diff
@@ -4137,16 +4139,16 @@ bool PlayerbotAI::AllowActive(ActivityType activityType)
 {
     // only keep updating till initializing time has completed,
     // which prevents unneeded expensive GameTime calls.
-    if (_isBotInitializing)
-    {
-        _isBotInitializing = GameTime::GetUptime().count() < sPlayerbotAIConfig->maxRandomBots * 0.12;
+    // if (_isBotInitializing)
```
nl-saw, line 4137: @liyunfan1223  @hermensbas 

Just wondering, was this code block commented out intentionally, or was it perhaps meant to be temporarily due to testing/debugging?

hermensbas: not sure, it shouldnt be disabled but maybe there is a reason for it? Only @liyunfan1223 can answer that

liyunfan1223: It's my fault that I didn't explain this change. I commented it because rndbots will not fight back when attacked by mobs during the initialization, so they will die a lot. I didn't experienced significant negative impact on performance on this modification.

This is just my opinion, do you think it needs to be kept?

hermensbas: True l, thats a known side-effect. Its alright to leave it off. I will see if i can optimize it to be less agressive.

Actually i might know a solution, which can offer the best of both worlds

PR #761, opened 2024-12-03
World buff

kadeshar:
- added spec configuration for world buffs
- added world buffs to configuration file
- fixed NeedWorldBuffTrigger

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1024,6 +1024,56 @@ AiPlayerbot.PremadeSpecLink.11.3.80 = -553202032322010053100030310511-205503012
 
 ###################################
 #                                 #
+# WORLD BUFFS					  #
+#                                 #
+###################################
+
+####################################################################################################
+# 
+#
+#
+
+# Applies a permanent buff to all bots when not in combat simulating flasks, food, rune etc.
+# WorldBuff.Faction.Class.Spec.MinLevel.MaxLevel
+
+AiPlayerbot.WorldBuff.0.1.0.80.80 = 53760,57358					#WARRIOR ARMS
+AiPlayerbot.WorldBuff.0.1.1.80.80 = 53760,57358					#WARRIOR FURY
+AiPlayerbot.WorldBuff.0.1.2.80.80 = 53758,57356					#WARRIOR PROTECTION
+AiPlayerbot.WorldBuff.0.2.0.80.80 = 60347,53749,57332			#PALADIN HOLY
+AiPlayerbot.WorldBuff.0.2.1.80.80 = 53758,57356					#PALADIN PROTECTION
+AiPlayerbot.WorldBuff.0.2.2.80.80 = 53760,57371					#PALADIN RETRIBUTION
+AiPlayerbot.WorldBuff.0.3.0.80.80 = 53760,57325					#HUNTER BEAST
+AiPlayerbot.WorldBuff.0.3.1.80.80 = 53760,57358					#HUNTER MARKSMANSHIP
+AiPlayerbot.WorldBuff.0.3.2.80.80 = 53760,57367					#HUNTER SURVIVAL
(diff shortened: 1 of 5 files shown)


dedahr:
Dunno if it's relevant but in patch is missing one DK spec **Double aura blood**

noisiver:
This is an amazing addition. I assume each comma-separated value is the spell id of a flask, elixir, food, etc and you can add as many as you want?

kadeshar:
1. About DK spec sadly in current implementation i only checking with spec tab have most points in tree (using currently existing method)
2. Yes, you can add your own spellId (aura id more precisly)

dedahr:
> 1. About DK spec sadly in current implementation i only checking with spec tab have most points in tree (using currently existing method).
> 

It will be fine as it is variant of blood for Boss tanking so Stoneblood flask is ok as it has more blood points just as blood spec.
So i presume this will be used no matter that exact spec is missing, right ?

kadeshar:
> > 1. About DK spec sadly in current implementation i only checking with spec tab have most points in tree (using currently existing method).
> 
> It will be fine as it is variant of blood for Boss tanking so Stoneblood flask is ok as it has more blood points just as blood spec. So i presume this will be used no matter that exact spec is missing in config, right ?

If second build have most points in blood tree then yes :)

Bobblybook:
> > > 1. About DK spec sadly in current implementation i only checking with spec tab have most points in tree (using currently existing method).
> > 
> > 
> > It will be fine as it is variant of blood for Boss tanking so Stoneblood flask is ok as it has more blood points just as blood spec. So i presume this will be used no matter that exact spec is missing in config, right ?
> 
> If second build have most points in blood tree then yes :)

Both of those blood specs are for tanking so should use the same consumables, shouldn't be an issue.
It gets more complicated if someone adds a blood dps spec, which do exist. You can still identify the role based on specific talent picks if you're crafty.

Same goes with Feral druid cat vs. bear spec. They are both deep feral tree specs, however you can use specific talents to identify the role.

I have a druid bot that has dual spec [Feral dps](https://www.wowhead.com/wotlk/talent-calc/druid/-543202132322010053120030310511-203503012_001ngz11nh221xjf31rfz41stk51rg1) and [Feral tank](https://www.wowhead.com/wotlk/talent-calc/druid/-503232132322010353120300313511-20350001_001ngv11ngx21zpv31rg241rg651rfz) and playerbots successfully identifies them as dps and tank roles when I switch, I'm guessing it's based off Thick Hide or something like that.

kadeshar:
> > > > 1. About DK spec sadly in current implementation i only checking with spec tab have most points in tree (using currently existing method).
> > > 
> > > 
> > > It will be fine as it is variant of blood for Boss tanking so Stoneblood flask is ok as it has more blood points just as blood spec. So i presume this will be used no matter that exact spec is missing in config, right ?
> > 
> > 
> > If second build have most points in blood tree then yes :)
> 
> Both of those blood specs are for tanking so should use the same consumables, shouldn't be an issue. It gets more complicated if someone adds a blood dps spec, which do exist. You can still identify the role based on specific talent picks if you're crafty.
> 
> Same goes with Feral druid cat vs. bear spec. They are both deep feral tree specs, however you can use specific talents to identify the role.
> 
> I have a druid bot that has dual spec [Feral dps](https://www.wowhead.com/wotlk/talent-calc/druid/-543202132322010053120030310511-203503012_001ngz11nh221xjf31rfz41stk51rg1) and [Feral tank](https://www.wowhead.com/wotlk/talent-calc/druid/-503232132322010353120300313511-20350001_001ngv11ngx21zpv31rg241rg651rfz) and playerbots successfully identifies them as dps and tank roles when I switch, I'm guessing it's based off Thick Hide or something like that.

This druid recognize its hardcoded. But I didnt thought about using "talents spec" command. This command somehow can match bot with build. I will think about that in next iteration also with lower than 80lvl buffs (such list exists in cmangos playerbots)
obraz
obraz
https://github.com/celguar/spp-classics-cmangos/blob/master/Settings/tbc/aiplayerbot.conf

PR #766, opened 2024-12-08
Make AutoScaleActivity further configurable

nl-saw:
These changes go further on the great work of @hermensbas and let you configure a floor-ceiling range DIFF-wise in which you'd like to have your server perform in respect to the amount of active bots. There's a 5-stage stepping calculated between these values, that logic pretty much stays the same.

The second addition is that one can now choose to compare against the AverageUpdateTime or MaxUpdateTime (spikes)
The benefit of comparing against MaxUpdateTime is that more efforts will be done to keep and try your server performance within set botActiveAloneDiffLimitCeiling.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1524,7 +1524,7 @@ AiPlayerbot.BotActiveAlone = 100
 AiPlayerbot.botActiveAloneSmartScale = 1
 AiPlayerbot.botActiveAloneSmartScaleWhenMinLevel = 1
 AiPlayerbot.botActiveAloneSmartScaleWhenMaxLevel = 80
-AiPlayerbot.botActiveAloneDiffMethod = 1
+AiPlayerbot.botActiveAloneDiffMethod = 2
 AiPlayerbot.botActiveAloneDiffLimitFloor = 30
 AiPlayerbot.botActiveAloneDiffLimitCeiling = 60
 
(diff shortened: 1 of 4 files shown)


hermensbas:
Lets have a talk about this fellow dutch man, since the mechanics aint working as you prolly expect. Upcoming week is gonna be tricky timewise, for now i would advice to wait merging this

hermensbas:
I dont have direct problem with implementation, well maybe. But ye lets have a talk about it, if you have patience due busy times atm.

I removed more complex mechanics aligned with what your're trying to archieve right now for various reasons.

The latency indicator is a very poor indicator when it comes to bot performance, esp with certain latency calculations. And activity percentages needs some explaining. But lets talk about it, in worse we both learn abit more :)

hermensbas:
Prolly discord is the best way, but lately less active due work. I am reserving some time upcoming tuesday :) Hopefully you have some time then aswell?

hermensbas:
there is

https://discord.gg/cuxXmAW3 

:)

hermensbas:
I will make a modified version of your version, so the current mechanism and configs will still be in place by default. You can either modify your MR or approve the other. 

That being said, if you wanna continue improving, hopefully, the mechanics around these type of mechanics a talk wouldnt hurt.

See https://github.com/liyunfan1223/mod-playerbots/pull/782

hermensbas:
I agree but it easier to show then explain, and test for matter. Hence the draft, eitherway it will be good to exchange ideas. Cuz there more optimizations which can be added at some point. I am not totally sure whats the best approach at this point, trying to define a goal or some goals would help in that.

hermensbas:
Good talk, i am finish the details tomorrow evening

PR #772, opened 2024-12-12
Prevent autogear on player alt bots

xSparky911x:


noisiver:
You're not preventing it only from altbots, you're preventing it from everyone.
```cpp
    Player* master = botAI->GetMaster();
    if (!master)
        return false;
```
That would return false on rndbots, would it not? It wouldn't even continue to the next if you implemented. It should also be configurable in my opinion. Most people don't want this limitation, I'm sure.

noisiver:
All I did to limit this for altbots on my end was this:

```cpp
if (!sRandomPlayerbotMgr->IsRandomBot(bot))
{
    return false;
}
```

Adding another check to that if to check the set config option would be easy enough, just like adding a message to it would be.

Keep in mind that I'm just sharing my opinion here, there doesn't need to be something complex and absolutely not a check for master, but I'm a low-level programmer so take it with a grain of salt.

xSparky911x:
> You're not preventing it only from altbots, you're preventing it from everyone.
> 
> ```c++
>     Player* master = botAI->GetMaster();
>     if (!master)
>         return false;
> ```
> 
> That would return false on rndbots, would it not? It wouldn't even continue to the next if you implemented. It should also be configurable in my opinion. Most people don't want this limitation, I'm sure.

I ran tests to make sure you are still able to run autogear on class bots and you are. Also the initialize command does not allow on alt bots but does on class bots and from what I could tell no configuration option for that so I followed that method. All I am doing here is making sure you can't cheat in gear by bypassing the restriction put in place by the init command and just use the autogear command

xSparky911x:
The !master part from what I understand was just to make sure they are your bots

Dreathean:
This should definitely have a config option like Revision said, a lot of people use it for their altbots for various reasons. 

But this is a nice addition for people who intend to use it like you with other people on their server.

noisiver:
I'd like to add that the option should be disabled by default to match the current behavior.

xSparky911x:
Ok will look at making that change. Keep in mind I am an even lower level programmer than he said lol. Little secret, AI helped me with this. Shouldn't the init command also be an option then for alt bots? Because currently from what I can see the init command is bot an option for player alt bots. It just says no. I'm not talking about random bots or class added bots by the way. Just player alts.

noisiver:
`addclass` actually uses rndbots so they're technically not any different from rndbots. They're just at the end of the list of rndbots so they aren't logged in automatically with the other rndbots.

conf/playerbots.conf.dist
```diff
@@ -1550,4 +1554,4 @@ AiPlayerbot.GuildTaskKillTaskDistance = 200
 AiPlayerbot.TargetPosRecalcDistance = 0.1
 
 # Allow bots to be summoned near innkeepers
-AiPlayerbot.SummonAtInnkeepersEnabled = 1
+AiPlayerbot.SummonAtInnkeepersEnabled = 1
```
noisiver: Don't remove the blank line

src/PlayerbotAIConfig.cpp
```diff
@@ -462,9 +462,11 @@ bool PlayerbotAIConfig::Initialize()
     addClassAccountPoolSize = sConfigMgr->GetOption<int32>("AiPlayerbot.AddClassAccountPoolSize", 50);
     maintenanceCommand = sConfigMgr->GetOption<int32>("AiPlayerbot.MaintenanceCommand", 1);
     autoGearCommand = sConfigMgr->GetOption<int32>("AiPlayerbot.AutoGearCommand", 1);
+    autoGearCommandAltBots = sConfigMgr->GetOption<int32>("AiPlayerbot.AutoGearCommandAltBots", 1);
     autoGearQualityLimit = sConfigMgr->GetOption<int32>("AiPlayerbot.AutoGearQualityLimit", 3);
     autoGearScoreLimit = sConfigMgr->GetOption<int32>("AiPlayerbot.AutoGearScoreLimit", 0);
 
+
```
noisiver: Remove the extra space (blank line) here

noisiver:
Aside from the two comments I made I don't see anything else that prevent this from being implemented. As long as it's tested and confirmed to be working as intended it should be good to go.

xSparky911x:
Interesting. I never though about the other players' alts and yes I would agree that it should be considered too. I have not tested using other players' alts as bots yet so can not confirm the broken config for that option but will look into it.

as per Revisions' suggestion I simplified it a bit by removing the original proposed changes

> ```c++
>     Player* master = botAI->GetMaster();
>     if (!master)
>         return false;
> ```

noisiver:
Adding additional checks for alts on other accounts are not needed in my opinion. You can't control alts from other accounts unless you're in the same guild and at that point they're either yours or you have full permission to use them anyway. Maybe it's just me.

noisiver:
> > Adding additional checks for alts on other accounts are not needed in my opinion. You can't control alts from other accounts unless you're in the same guild and at that point they're either yours or you have full permission to use them anyway. Maybe it's just me.
> 
> I'd say go for granular control. Let's consider the following - you and your friend and the alts are all in the same guild. You create a party, including some of your friend's alts for a raid and issue the autogear command. **Whoops**! Now you just geared the alt of your friend.

First of all, you are allowed to use them for a reason. In many cases they are actually your characters. Second, if you are actually raiding you shouldn't be anywhere near the autogear command. It completely defeats the purpose.

xSparky911x:
> > Adding additional checks for alts on other accounts are not needed in my opinion. You can't control alts from other accounts unless you're in the same guild and at that point they're either yours or you have full permission to use them anyway. Maybe it's just me.
> 
> I'd say go for granular control. Let's consider the following - you and your friend and the alts are all in the same guild. You create a party, including some of your friend's alts for a raid and issue the autogear command. **Whoops**! Now you just geared the alt of your friend.

Great point. Hence the whole reason I wanted to add this change was being able to cheat gear in with tis command. Never thought about the other player alts. Doesn't hurt to add the check and config

noisiver:
Yes, personal prefence is everything. If you want to add additional commands for it, do so. Just make sure they're matching the current behavior by default and that they work as intended.

serverwar:
> > > Adding additional checks for alts on other accounts are not needed in my opinion. You can't control alts from other accounts unless you're in the same guild and at that point they're either yours or you have full permission to use them anyway. Maybe it's just me.
> > 
> > 
> > I'd say go for granular control. Let's consider the following - you and your friend and the alts are all in the same guild. You create a party, including some of your friend's alts for a raid and issue the autogear command. **Whoops**! Now you just geared the alt of your friend.
> 
> Great point. Hence the whole reason I wanted to add this change was being able to cheat gear in with tis command. Never thought about the other player alts. Doesn't hurt to add the check and config

I was caught by surprise on my server players cheating unfortunately another issue command .playerbot rndbot reload is released to players without being GM if they start spamming it takes down the server

xSparky911x:
> # Allow login other players' characters as bots
> # Default: 0 (disabled)
> AiPlayerbot.AllowPlayerBots = 1
> ```

Can confirm this is indeed broken.

xSparky911x:
> I could not test the latter as it seems the following config is broken, and it currently never lets you login or control/whisper other persons alts as bots:
> 
> ```
> # Allow login other players' characters as bots
> # Default: 0 (disabled)
> AiPlayerbot.AllowPlayerBots = 1
> ```

Can confirm this is indeed broken

PR #774, opened 2024-12-13
Fix reload command security

nl-saw:
Security Fix: Require GM privilege to invoke the reload command, which is currently unprotected.


hermensbas:
Looks good for those who use live servers

Issue #344, opened 2024-07-15
(Re-)Initialitialized bots don't pick talents on EquipmentPersistence=1


HennyWilly:
**Describe the bug**
When initializing a bot while EquipmentPersistence=1 is set in the config file, the bot has no talent points selected after initialization.
Îf the bot had talent points assigned prior to (re-)initialization, those points were reset as well.
This also happens on newly created bots: They still have their initial gear equipped (as expected) but no talents were selected.

**Commit hash**
cabc09f

**To Reproduce**
1. Pick a manually created bot.
2. Initialize that bot (I used the english version of the unbot addon)
3. Inspect the bot.
4. See that all talent points are unassigned.

**Expected behavior**
The bot should have all talent points assigned after initialitaion.
I'm not sure if the talents should be kept or randomized in this case.
Maybe a new option (e.g. `TalentPersistence`) would be good...

**Desktop (please complete the following information):**
 - OS: Windows (run inside Docker)

**Additional context**
It should be noted that I set the max level of my server to 60 and my player character is level 60 as well.

These are all settings I changed compared to the default .dist file:
AiPlayerbot.MinRandomBots = 400
AiPlayerbot.MaxRandomBots = 500
AiPlayerbot.GroupInvitationPermission = 2
AiPlayerbot.SummonWhenGroup = 0
AiPlayerbot.AutoTrainSpells = free
AiPlayerbot.AutoPickReward = yes
AiPlayerbot.RandomBotMaxLevel = 60
AiPlayerbot.DisableRandomLevels = 1
AiPlayerbot.RandombotStartingLevel = 1
AiPlayerbot.RandomBotMaxLevelChance = 0
AiPlayerbot.LimitEnchantExpansion = 1
AiPlayerbot.LimitGearExpansion = 1
AiPlayerbot.EquipmentPersistence = 1
AiPlayerbot.EquipmentPersistenceLevel = 60
AiPlayerbot.RandomBotGroupNearby = 1
AiPlayerbot.AutoLearnQuestSpells = 1
AiPlayerbot.RandomBotJoinBG = 0
AiPlayerbot.MaxRandomRandomizeTime = 1209600
AiPlayerbot.RandomBotLoginAtStartup = 0

Dreathean:
[...]

PR #778, opened 2024-12-14
Resolved issues with rndbots and equipmentPersistence

avirar:
Bots with equipmentPersistenceLevel would not have talents or gear upon initialization. They would also lose their assigned talents during some calls to Randomize.

1. Check for persistence before calling resetTalents
2. Call InitTalentsTree and InitEquipment if not an incremental randomization

This should close two Issues:
#344 and #544

### DIFF
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -207,7 +207,10 @@ void PlayerbotFactory::Randomize(bool incremental)
     Prepare();
     LOG_DEBUG("playerbots", "Resetting player...");
     PerformanceMonitorOperation* pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Reset");
-    bot->resetTalents(true);
+    if (!sPlayerbotAIConfig->equipmentPersistence || bot->GetLevel() < sPlayerbotAIConfig->equipmentPersistenceLevel)
+    {
+        bot->resetTalents(true);
+    }
     // bot->SaveToDB(false, false);
     ClearSkills();
     // bot->SaveToDB(false, false);
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -270,7 +270,7 @@ void PlayerbotFactory::Randomize(bool incremental)
 
     pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Talents");
     LOG_DEBUG("playerbots", "Initializing talents...");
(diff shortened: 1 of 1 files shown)


src/factory/PlayerbotFactory.cpp
```diff
@@ -207,7 +207,10 @@ void PlayerbotFactory::Randomize(bool incremental)
     Prepare();
     LOG_DEBUG("playerbots", "Resetting player...");
     PerformanceMonitorOperation* pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Reset");
-    bot->resetTalents(true);
+    if (!sPlayerbotAIConfig->equipmentPersistence || level < sPlayerbotAIConfig->equipmentPersistenceLevel)
```
nl-saw, line 210: Shouldn't this be (goes for all 3 changes)

if (!sPlayerbotAIConfig->equipmentPersistence || (sPlayerbotAIConfig->equipmentPersistence && level < sPlayerbotAIConfig->equipmentPersistenceLevel))

As it looks to me that when you disable equipment persistence, but configure the persistence level in config to let's say 60, it seems it will skip bots from that level onwards.

avirar: They are functionally the same.

The condition: `if (!sPlayerbotAIConfig->equipmentPersistence || (sPlayerbotAIConfig->equipmentPersistence && level < sPlayerbotAIConfig->equipmentPersistenceLevel))` is functionally correct but redundant. 

The simplified version: `if (!sPlayerbotAIConfig->equipmentPersistence || level < sPlayerbotAIConfig->equipmentPersistenceLevel)` works the same because `!sPlayerbotAIConfig->equipmentPersistence` already covers cases where persistence is disabled, making the extra `sPlayerbotAIConfig->equipmentPersistence` check unnecessary.

I was thinking about adding `!incremental` in there too but opted against it for the moment. This is enough to solve the issues at hand.

nl-saw: > The simplified version: `if (!sPlayerbotAIConfig->equipmentPersistence || level < sPlayerbotAIConfig->equipmentPersistenceLevel)` works the same because `!sPlayerbotAIConfig->equipmentPersistence` already covers cases where persistence is disabled, making the extra `sPlayerbotAIConfig->equipmentPersistence` check unnecessary.

Ah yes ofc, you are right. I'll definitely try this out as I also run into the issues.

nl-saw: With this I haven't ran into bots having unspent talent points anymore.

PR #795, opened 2024-12-18
Ulduar-Razorscale strategy

avirar:
My first attempt at making a boss strategy, looking for feedback.

Triggers and associated actions:

1. Avoid Devouring Flame - All bots avoid it. Main tank (during boss air phase) moves a little over twice as far so melee have some space to attack adds.

2. Avoid Sentinel/Whirlwind - Ranged keep away from them all the times, melee only when they are casting Whirlwind. Main tank marks Sentinel with lowest HP with Skull RTI so they get nuked.

3. Ignore Razorscale when flying and alone - Place Moon RTI on Razorscale so the bots generally ignore her. Moon removed when she comes below Z threshold. This prevents the bots grouping directly under her location, at least at the start of the encounter, they still seem to do it when she transitions back to air. I'm almost certain there's a better way to do this. Bots will also move inside the arena, radius just covers the harpoons.

4. Harpoon available for use - Closest ranged DPS will move to and use any available harpoon. I think we'd check if any adds are alive and >50% before doing the last one, but it's not much of an issue as they usually are.

5. Boss grounded - Ranged bots MoveInside a large area behind the boss, they are able to out range the knock back so it's good to allow that. Melee do their thing behind boss naturally. Once boss is below 50% and permanently grounded the ranged bots will group closer to the centre of the arena.

5. Fused Armor - Bots set new main tank group flag when they get 2 stacks of fuse armor. The tank with the most HP that is not currently affected by Fuse Armor is chosen as the new main tank. This worked exactly as I hoped. Bots respect Main tank status so a bot freshly assigned as main tank will taunt and establish threat immediately and the old tank will let their threat drop so safe levels. Non-main tanks move near the main tank so they are ready.

### DIFF
--- a/src/strategy/raids/ulduar/RaidUlduarActionContext.h
+++ b/src/strategy/raids/ulduar/RaidUlduarActionContext.h
@@ -17,11 +17,25 @@ public:
     {
         creators["flame leviathan vehicle"] = &RaidUlduarActionContext::flame_leviathan_vehicle;
         creators["flame leviathan enter vehicle"] = &RaidUlduarActionContext::flame_leviathan_enter_vehicle;
+        creators["razorscale avoid devouring flames"] = &RaidUlduarActionContext::razorscale_avoid_devouring_flames;
+        creators["razorscale avoid sentinel"] = &RaidUlduarActionContext::razorscale_avoid_sentinel;
+        creators["razorscale ignore flying alone"] = &RaidUlduarActionContext::razorscale_ignore_flying_alone;
+        creators["razorscale avoid whirlwind"] = &RaidUlduarActionContext::razorscale_avoid_whirlwind;
+        creators["razorscale grounded"] = &RaidUlduarActionContext::razorscale_grounded;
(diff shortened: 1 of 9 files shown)


Dreathean:
Awesome work! Out of curiosity, how does the bots assigning main tank status work? Do you know if it takes into account a player-tank if the player has the raid assigned main-tank status?

avirar:
> Awesome work! Out of curiosity, how does the bots assigning main tank status work? Do you know if it takes into account a player-tank if the player has the raid assigned main-tank status?

Thank you very much! The bots don't check if tanks are human at the moment, but it's definitely something I want to implement. I have some time today, so I'll have a look :)

avirar:
1. Tank bots will reassign main tank from players when the players stacks get high.
2. Tank bots will continue to mark targets when not main tank.
3. Tank bots will prioritise stacking on the main tank during ground phase.

If you want to tank the boss first, just manually assign yourself as main tank before the fight or at least before the ground phase begins. There's an aggro wipe as she transitions so the bots may out threat you until you establish aggro properly (damn DKs).

PR #801, opened 2024-12-19
Only freeze bots during init when not in combat

hermensbas:
Only freeze bots during bot init phase and when not in combat. Which prevents unnecessary death, revive and teleporting rotations during that phase.


hermensbas:
> Thanks for also incorporating https://github.com/nl-saw/mod-playerbots/tree/tunebotperformance

Thanks for pointing it out, there might more tweaks laying around though. Same the path generation rule, feel free too experiment more.

PR #803, opened 2024-12-20
5-8 percent performance increase when when botActiveAlone is 100% and smartScale disabled

nl-saw:
Overall Diff-wise. By moving the when botActiveAlone is 100% and smartScale disabled return to the very top.


hermensbas:
Can someone else verify this please, cuz i am getting different results.

PR #805, opened 2024-12-21
Feature: Bots use flying mounts when player has flight form

nl-saw:
image

Added feature: make bots use flying mounts when player has flight form (if they can).

### DIFF
--- a/src/strategy/actions/CheckMountStateAction.cpp
+++ b/src/strategy/actions/CheckMountStateAction.cpp
@@ -71,14 +71,17 @@ bool CheckMountStateAction::Execute(Event event)
         if (!bot->GetGroup() || bot->GetGroup()->GetLeaderGUID() != master->GetGUID())
             return false;
 
+        auto masterInShapeshiftForm = master->GetShapeshiftForm();
+
         // bool farFromMaster = sServerFacade->GetDistance2d(bot, master) > sPlayerbotAIConfig->sightDistance;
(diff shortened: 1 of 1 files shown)


Raz0r1337:
I've tried it now and no i can't.
The Bots do not use their mounts when i am going to travelform only on fly/trvelform they are using thei fly mounts. ;-)
Can you maybe expand it to include this function?

noisiver:
The PR is to make them use their flying mount when the player uses flight form. It does that, so saying it doesn't work and then say it works is weird.

It could include them using regular mounts when the player uses travel form if @nl-saw wants to add it though, that makes sense.

Raz0r1337:
Please test it yourself first before you get involved! Then you would know what I mean.

The flight form and the travel form are two different things.

In the flight form it works, but in the travel form (which cannot fly) the bots do not use their mounts.

noisiver:
The pull request is about flight form. It's unrelated to travel form. Those are two completely different things and the intention of this pull request is **flight form**, NOT **travel form**. Just because flight form works doesn't automatically mean travel form works since they are **different things**.

Raz0r1337:
Read my first post carfully again.

I ask Clerly that: Can you maybe expand it to include this function?
And i answer on this PR because he send me a message on my issue that it works now.

Raz0r1337:
You are the best. Very thank you.

PR #806, opened 2024-12-21
Make bots use mounts when player is in travel form as well

nl-saw:
image

Addition to https://github.com/liyunfan1223/mod-playerbots/pull/805


noisiver:
@hermensbas We talked about this. Let the checks finish before merging, they are there for a reason.

PR #809, opened 2024-12-22
[Bugfix] reverted changes in need world buff trigger

kadeshar:
Reason:
https://github.com/liyunfan1223/mod-playerbots/issues/808

### DIFF
--- a/src/strategy/triggers/CureTriggers.cpp
+++ b/src/strategy/triggers/CureTriggers.cpp
@@ -25,8 +25,4 @@ bool PartyMemberNeedCureTrigger::IsActive()
     return target && target->IsInWorld();
 }
 
-bool NeedWorldBuffTrigger::IsActive()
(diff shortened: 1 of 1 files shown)


hermensbas:
didnt you test it?

kadeshar:
> didnt you test it?

Yes. But this previous version too and in windows environment both working.
I dont have linux arm development environment :(

hermensbas:
Strange it only crashes in certain environments. :/

noisiver:
> > didnt you test it?
> 
> Yes. But this previous version too and in windows environment both working. I dont have linux arm development environment :(

To be fair I switched to using my ARM setup only yesterday so I didn't catch it either.

I don't understand it either but at least it's fixed now. I've been running a few dungeons with worldbuff set and it works like a charm. Thank you for fixing this.

PR #816, opened 2024-12-24
Worldbuff feral druid bear/cat logic

avirar:
Added a fourth set of buffs for druids to separate cats from bears:

Detect if we are handling a druid with the feral spec tab, then check if they have Thick Hide rank 3 in their active spec. If they don't have the talent we use the fourth spec for druids in the config file. 

Tab/effectiveSpec remains unchanged in all other cases but we can easily add more class/spec combos.


kadeshar:
Nice addition. Do you know that death kinght need that too for blood spec?

avirar:
> Nice addition. Do you know that death kinght need that too for blood spec?

Thank you!

I'd read about it here or the discord, but I don't know the specifics of it, can you tell me which talent(s) we should check for, and the buffs we should apply?

blinkysc:
> > Nice addition. Do you know that death kinght need that too for blood spec?
> 
> Thank you!
> 
> I'd read about it here or the discord, but I don't know the specifics of it, can you tell me which talent(s) we should check for, and the buffs we should apply?

Think he might be talking about difference between dps/tank blood spec. Though blood dps is not very good; it was niche

Blood DPS
https://www.wowhead.com/wotlk/guide/classes/death-knight/blood/dps-overview-bis-gear-p3

Blood Tank
https://www.wowhead.com/wotlk/guide/classes/death-knight/blood/tank-overview-pve

kadeshar:
But is somehow definied in default config
obraz

avirar:
Done. Used Blade Barrier rank 5 as the check for tank/dps blood DKs.

PR #817, opened 2024-12-24
Poll results: Change default formation to Chaos

nl-saw:
image

This as it looks more natural than the default near.
A poll has been raised on Discord whether to change it or not.


hermensbas:
Ye agree on that as a default :)

PR #818, opened 2024-12-25
Add ratio config for alliance and horde rndbots

liyunfan1223:
Merge after https://github.com/liyunfan1223/azerothcore-wotlk/pull/32


hermensbas:
love it

PR #820, opened 2024-12-26
Make eligible bots cast Greater blessings for Might, Kings and Wisdom when in groups

nl-saw:
This will make **eligible** bots (when on or above the proper level) cast Greater blessings for Might, Kings and Wisdom.

Looking for testers and feedback.


noisiver:
I don't think the best solution is to swap minor for greater blessing as-is.

Just look at Paladin. Protection wants sanctuary, holy wants wisdom and retribution wants might/kings.

I think Ideally there should be some logic to determine if there are multiple characters of the same class in the group that need different blessings before choosing greater blessing.

noisiver:
For the time being replacing minor with greater blessing works fine. It's just something to add to in the future, some kind of logic there.

They do not use Sanctuary at all, you are correct.

PR #832, opened 2024-12-30
Move backwards when enemy too close

liyunfan1223:
When an enemy is too close, ranged bots move backward instead of turning around. This makes the bot move more like a real player.

A more useful future improvement would be to have tank move backwards to avoid leaving its back to the enemy.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -4115,11 +4115,16 @@ inline bool ZoneHasRealPlayers(Player* bot)
         return false;
     }
     
-    for (auto& player : sRandomPlayerbotMgr->GetPlayers())
+    for (Player* player : sRandomPlayerbotMgr->GetPlayers())
     {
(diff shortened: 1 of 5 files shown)


liyunfan1223:
PR affects all methods using `MoveAway` and may break other features. Mark draft until confirmed.

avirar:
Just tried it out. I really like the idea, but it was affecting movement away from hazards in dungeon/raid/avoid aoe situations. e.g for the Razorscale strat I used it for the flames and sentinels. Bots are too slow to move from the Sentinels and the melee group died when there was more than 1 sentinel doing whirlwind. It was great to see a warrior parrying the whirlwind as they backed away though!

Is it possible to make them strafe away so they retain full running speed while also ensuring they don't fully turn their back to the target?

liyunfan1223:
Agree, although implementing strafe away requires more effort. For now I added a parameter to MoveAway so that it will not be applied to all raid and dungeon strategies.

PR #834, opened 2024-12-31
[performance] smart scaling configurable

hermensbas:


hermensbas:
@nl-saw lemme know what you think

hermensbas:
For now happy new year :)

parazitenew:
On my side smart scaling is disabled and bots are set up active 100% with 1k bots, did not face lag issue.

hermensbas:
I will try to see if i can lower the default settings abit more, which values did you use @nl-saw and @davvvve1 ?

hermensbas:
@davvvve1 

The scaling is only applied when smartscale is enabled though. Which is set on disabled in your config. Why do you keep botAloneActive on the value 1? Why just not keep it on 0? 

You might wanna reconsider "AiPlayerbot.BotActiveAloneForceWhenInRadius” between the value of 150 and 300 if u want the bots around be more alive. Default sightDistance will enforce activity aswell which is 75 yards by default if not mistaken (see def config for that).

Although i appreciate you wanted to test it, but you didnt test the change :p

davvvve1:
> @davvvve1
> 
> The scaling is only applied when smartscale is enabled though. Which is set on disabled in your config. Why do you keep botAloneActive on the value 1? Why just not keep it on 0?
> 
> You might wanna reconsider "AiPlayerbot.BotActiveAloneForceWhenInRadius” between the value of 150 and 300 if u want the bots around be more alive. Default sightDistance will enforce activity aswell which is 75 yards by default if not mistaken (see def config for that).
> 
> Although i appreciate you wanted to test it, but you didnt test the change :p

Haha .  Thanks for the answer - I forgot to change my settings after i compiled . But why do i get better performance then ?

hermensbas:
The value 1 doesnt add anything, other then additional checks being executed. So rather keep is 0 when using 1. 

Autoscaling simply dynamically modifies the botAloneActive percentages based on the general latency you have. Also gives a little more headspace during temp performance loss of your wow server.

Basically you modify these values based on what type of behaviour u want. There is no wrong or right way, except keep botAloneActive on 1 seems abit pointless.

PR #838, opened 2025-01-03
Warrior strategy update

avirar:
1. Enabled Enraged regeneration for all specs.
2. Fixed issue where Rogues would constantly apply Expose Armor on targets that already had Sunder Armor.
3. Arms warriors use Retaliation when being melee'd by two or more creature.
4. Arms warriors use Shattering Throw on enemies with Divine Shield, Ice Block, or Hand of Protection.


Dreathean:
Awesome stuff! As a warrior tank I'm especially grateful for you fixing the rogues using expose armor when target was sundered, that was so frustrating at times

PR #840, opened 2025-01-03
Allow use of more portal game objects

avirar:
1. Added GAMEOBJECT_TYPE_GOOBER as a valid object type to check for portal spells
2. Added Eye of Eternity exit portal and Naxx Portals/Orbs

Get the bots within interaction range of the desired portal and type teleport in chat


avirar:
That moment when you debug the whole thing with a single !

Issue #693, opened 2024-11-06
Playerbots should always use the greater variant of a buff


Dohtt:
**Describe the bug**
Currently bots buff everyone individually, which leads to some trouble with timing of the buffs etc

**Expected behavior**
Bots should always cast the greater variant of a buff to make sure that everyone has the longest buff duration heading forward.

Dreathean:
Should be largely resolved with https://github.com/liyunfan1223/mod-playerbots/pull/1603

PR #844, opened 2025-01-04
Paladins use Greater Blessings

avirar:
Issue #693

1. Created bgreater nc strategy for paladins, set it as part of their default nc strategies
2. Used logic from PallyPower addon for BlessingManager. Also adjusted it so it uses paladins with Improved talent points where ever possible.
3. Druid Mark/Gift of the Wild while in a group
4. Priest Power Word: Fortitude/Prayer of Fortitude while in group
5. Mage Arcane Intellect/Arcane Brilliance while in group (also detects Dalaran Intellect/Brilliance as valid)

### DIFF
--- a/src/AiFactory.cpp
+++ b/src/AiFactory.cpp
@@ -530,7 +530,7 @@ void AiFactory::AddDefaultNonCombatStrategies(Player* player, PlayerbotAI* const
             else
                 nonCombatEngine->addStrategiesNoInit("dps assist", "bdps", "baoe", nullptr);
 
-            nonCombatEngine->addStrategiesNoInit("cure", nullptr);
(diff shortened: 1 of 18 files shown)


liyunfan1223:
Confirmed. We temporarily revert this, it does need more testing. https://github.com/liyunfan1223/mod-playerbots/pull/849

avirar:
Thanks for the feedback! I'll get that fixed up tomorrow

noisiver:
I don't see code for it so I have to ask. Does this not cover Divine Spirit/Prayer of Spirit?

noisiver:
I have seen them not use prayer of spirit while in a group, it hasn't always been used so I figured it should at least be looked at.

PR #846, opened 2025-01-04
Gunship fix

Noscopezz:
Bots will return to ship after killing mage


hermensbas:
Ive assigned to @noisiver since you seem to raid alot :)

noisiver:
I'm not raiding that one. I'm still on patch 3.0 so Naxxramas, EoE and OS only.

hermensbas:
In that case lets try and see

PR #850, opened 2025-01-04
Follow master instead of group master in formation

liyunfan1223:
Fixed follow issue when multiple players in one group. https://github.com/liyunfan1223/mod-playerbots/issues/843

### DIFF
--- a/src/strategy/values/Formations.cpp
+++ b/src/strategy/values/Formations.cpp
@@ -23,7 +23,7 @@ bool Formation::IsNullLocation(WorldLocation const& loc) { return IsSameLocation
 
 WorldLocation MoveAheadFormation::GetLocation()
 {
-    Player* master = botAI->GetGroupMaster();
+    Player* master = GetMaster();
     if (!master || master == bot)
         return WorldLocation();
 
@@ -110,7 +110,7 @@ public:
 
     WorldLocation GetLocationInternal() override
     {
-        Player* master = botAI->GetGroupMaster();
+        Player* master = GetMaster();
         if (!master)
             return WorldLocation();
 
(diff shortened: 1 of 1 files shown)


liyunfan1223:
It makes sense, although I feel like the master is always there at this time, just like the near formation and I want to keep it the same as before.

What I can think of is adding some distance/maps checks to select between master and group master, which should help with multiple players group.

Dreathean:
In my opinion the altbots corresponding to their specific master player should always follow their master, I don’t think they should fall under a new master if the original master leaves? That could get messy, no?

noisiver:
Wouldn't it make the most sense that the bots leave if their master (the one who "owns" them) leaves the group?

Dreathean:
I do think if you leave the group then your bots should leave as well, but if you simply logout while in a group they shouldn’t necessarily disband as long as you don’t leave the group. Then people like nl-saw could logout, switch characters, and then boot or login that other character as another altbot.

Dreathean:
I wonder how that impacts who the master technically is in that case though? Are they coded to strictly follow the original character that logged them in? The account? Kind of a complex scenario if you’re going for that it feels.

Dreathean:
Regardless, going back to the original point of this PR, I fully support and think that the altbots should follow their master instead of the group/raid master, that’s been an issue playing with my friend where all of his altbots follow me when we intentionally want them split between us.

liyunfan1223:
My point is that this PR only lets the bots follow the master (who adds them to the world) rather than the group leader, which is consistent with the original default formation. So I think it's ok to merge.

Actually I don't have many multiplayer experience, I mostly just play alone :(

PR #853, opened 2025-01-05
Feat: Mount level for bot configurable instead of hardcoded

valsan-azerty-boi:
### Feature
The level required for bot using mounts were hardcoded, I put them in conf to easily modify them
In conf I have put the previous hardcoded values as default values
- lvl 20 for slow ground mount
- lvl 40 for fast ground mount
- lvl 60 for slow fly mount
- lvl 70 for fast fly mount

### Testing
**Example in a battleground (bracket 30-39) with the slow ground mount configured at lvl 40 in conf file (instead of 20 by default)**
image
They walk instead of using the mount because they are not level 40

**Example in a battleground (bracket 40-49) with the slow ground mount configured at lvl 40 (instead of 20 by default) and fast ground mount at lvl 60 (instead of 40 by default) in conf file**
image
They correctly slow mount instead of fast mount cause they are level 40+ but not 60 :)

**Example in a battleground (bracket 60-69), same config again**
image
Now they fast mount instead cause they are level 60+

### Compilation OK
**Successfuly compiled and work on my server (Debian 11)**
- Azerothcore-wotlk (branch Playerbots)
- mod-playerbots
- mod-buff-command
- mod-bg-reward
- mod-server-auto-shutdown
- mod-assistant
- mod-individual-xp
- mod-ah-bot
- mod-autobalance
- mod-solocraft
- mod-ony-naxx-logout-teleport
- mod-individual-progression
- mod-pvp-titles
- Recache

### Other
If you use custom values, you need to execute this commands after changing config
.playerbot rndbot reset
.playerbot rndbot init

(Sorry for my bad english, I'm not native)

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -8,6 +8,7 @@
 #	PLAYERBOT SETTINGS
 #		GENERAL
 #		SUMMON OPTIONS
+#       MOUNT
 #		GEAR
 #		LOOTING
 #		TIMERS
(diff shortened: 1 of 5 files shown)


hermensbas:
I guess having this configurable doesnt hurt, although the core defaults, vendors are hardcoded aswell, if not mistaken. Nevertheless mod and whatnot can overwrite that behaviour. @noisiver what do you think, the change looks good to me.

noisiver:
Looks good to me. Trainer entries as well as items are set for the level too but that's really irrelevant in this case. It should be configurable for those of us who use some kind of progression that changes those levels.

liyunfan1223:
Looks good to me.

Regrad:
# Defines at what level a bot will naturally use its 150% fly mount
# note: was level 60 during WotLK, 70 during TBC
# default: 60
**AiPlayerbot.UseFlyMountAtMinLevel = 60;**
# Defines at what level a bot will naturally use its 280% fast fly mount
# note: was level 70 during WotLK and TBC
# default: 70
**AiPlayerbot.UseFastFlyMountAtMinLevel = 70;**

The semicolon is superfluous.

PR #854, opened 2025-01-06
Performance: Refactored Engine::DoNextAction

nl-saw:
Refactored Engine::DoNextAction so it is more performance-oriented. Mainly it applies multipliers early, which optimizes the action processing flow, and it exits the loop more efficiently when there’s no action to process.

I'd appreciate if some of you can test it to make sure there are no deficiencies in bot behavior.

No rush in merging :)

### DIFF
--- a/src/strategy/Engine.cpp
+++ b/src/strategy/Engine.cpp
@@ -146,167 +146,106 @@ bool Engine::DoNextAction(Unit* unit, uint32 depth, bool minimal)
 
     bool actionExecuted = false;
     ActionBasket* basket = nullptr;
-
     time_t currentTime = time(nullptr);
-    // aiObjectContext->Update();
+
+    // Update triggers and push default actions
     ProcessTriggers(minimal);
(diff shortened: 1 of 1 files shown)


blinkysc:
Putting on mine now

blinkysc:
It seems stable for me. I still get sharp spikes, with the max percentiles  reaching up to around 8,000 when using .server info, but this has been the status quo for quite some time.

blinkysc:
> > It seems stable for me. I still get sharp spikes, with the max percentiles reaching up to around 8,000 when using .server info, but this has been the status quo for quite some time.
> 
> Thanks for the feedback, this is just a minor performance tune and will not help much with (huge) spikes. Reason to test is to make sure the bot's behaviors are still intact.

Everything has been same so far for bots. I am using new RPG strat though too, though not sure if that would be a big diff

hermensbas:
i havent notice any strange behaviour and such although my experience in rather limited.

Issue #449, opened 2024-08-07
Pre-existing (player)characters Problem - Characters login as bots and log off immediately after


icemansparks:
**Describe the bug**
After staring up the Server and loggin into a character on a non admin account, the characters from the friendlist and guild are shown to be loggin in and immediately loggin out.

Settings from config:
```
# auto-login all player alts as bots on player login
AiPlayerbot.BotAutologin = 0

# Allow login other players' characters as bots Default: 0 (disabled)
AiPlayerbot.AllowPlayerBots = 0

# Allow/deny bots from your guild
AiPlayerbot.AllowGuildBots = 0
```

No error messages are shown in the server console, though.
The recurring message of "30 New bots" appears in the server console window instead.

**Commit hash**
264c533d1fd020234eb1c05005343d86fcca7a39

**To Reproduce**
1. startup auth and world server
2. login with non-admin account that has a guild with more player account characters in it (same as player account as well as from other players accounts) and also a friendlist with characters from player accounts that are not in the current guild.
3. Wait about a second or two
4. See "charactername has come online" and "charactername has gone offline" message in chat.

**Expected behavior**
With the options to deny bots from guild and disabled auto login, the alt characters of the player and other players accounts should not login at all automatically. 

**Desktop (please complete the following information):**
 - OS: Server compiled and running under windows 10

icemansparks:
Ok seems like I have found the cause for this issue and it is probably not that common.

**TL;DR:**
Custom SQL was used on server setup to add pre-existing player accounts to the account and character database.
But in `acore_playerbots.playerbots_random_bots` the value for "bot" is fixed on server stup and rfrncs bot charactr guids which ar in this case real characters instead.

**Long version:**
As far as I can tell, it is decided if a character is a bot / part of a bot account by looking up the value in `acore_playerbots.playerbots_random_bots`
The table is pre-populated with all the bots that have been created on first initialisation when starting up the server.

The problem in my case now seems to be, that the character IDs are used in this table in the "bot" column to link the character guid to the bot entry.
Since the values for the guids of the bots seem to be pre-populated (couldn't find the exact sql file yet), character guid values start at 1, just like the created character entries.

This is the breaking point for my setup as i added character and account entries for already existing user accounts with a custom SQL in the server sql directory (as supposed to).
These will be created before all modules, resulting in my character database having real user (player) characters inside them (guid 1-30 in this case).
[...]

Issue #815, opened 2024-12-23
Bots logging on/off


tugofpeas:
**Describe the bug**
Account bots, who are on friends' list will log on and immediately off about every 15 seconds. 

**Commit hash**
I'm currently on d9f9d980cff7b08864bf7c0bf4386d5aa2341971 but also observed it in  eef41ab3b713334584c3807e6198354714a9316b.

**To Reproduce**
1. Create an account character, enter the world. Log off.
2. Get on another character, add the other char to friends
3. Your chat window will start spamming that [Character] came online/ went offline with the frequency of about 15-20 seconds.

**Expected behavior**
Chars shouldn't be coming online randomly.

**Screenshots**
![image](https://github.com/user-attachments/assets/a20568e3-fece-437c-9c6b-c27d8726fa97)

**Desktop (please complete the following information):**
Windows 11, World of Warcraft 3.3.5a.12340 enUS client
AC Server - Debian 12

**Additional context**

Dreathean:
I'm not experiencing this and I'm updated up until maybe the last 1-2 PRs, what other modules are you running?

tugofpeas:
The only mods I have are:
mod_ah_bot
mod_playerbots

kaijikan:
> 1. Create an account character, enter the world. Log off.
> 2. Get on another character, add the other char to friends
> 3. Your chat window will start spamming that [Character] came online/ went offline with the frequency of about 15-20 seconds.
> 
> **Expected behavior** Chars shouldn't be coming online randomly.

I had a similar issue to this, and fixed it by wiping and regenerating the bots, which did by editing the playerbots.conf file

`.../azerothcore/env/dist/etc/modules/playerbots.conf`


Enable the Delete bots setting, and restart your server.
`AiPlayerbot.DeleteRandomBotAccounts = 1`

Once you server has completed loading, disable the setting once more, and restart your server
`AiPlayerbot.DeleteRandomBotAccounts = 0`

Now all the bots should re-generate, and you will not get them logging in and out en-mass

Hope this works for you too

tugofpeas:
@kaijikan Thanks, I'll try that.
In my case, this issue only occurs when I set AiPlayerbot.RandomBotAccountCount above 200 and the min and max bots are equal or more than 200. Feels like some sort of rounding issue or something. For example, when all 3 settings are at 300, the # of bots online will be 297, instead of 300. And the remaining 3 bots will keep on triggering the logging on/off endlessly.

[...]

PR #866, opened 2025-01-13
Fix bot account creation on DB with existing player accounts before mod init, refactor and improve readability of functions

icemansparks:
This pull request fixes a problem with the bot account creation in case of pre-existing player accounts / characters. 
closes: #449 
maybe also address: #815

refactors and optimizes the [RandomPlayerbotFactory.cpp] file to improve readability, maintainability, and testability. 
The changes include splitting the CreateRandomBot function into dedicated functions, adding error handling, improving logging, and optimizing the code.

### Changes Made
**1) Refactored CreateRandomBot Function:**
Split the CreateRandomBot function into dedicated functions:
- GenerateRandomGender: Generates a random gender.
- SelectRandomRace: Filters available races based on the alliance flag and selects a random race.
- GetRandomBotName: Creates or retrieves a random bot name from the name cache.
- CollectCustomizationOptions: Collects customization options (skin color, face, hair, facial hair) based on the race and gender.
- SelectRandomCustomizationOptions: Selects random customization options from the collected options.
- CreateAndInitializePlayer: Creates and initializes the player object with the generated name, race, class, gender, and customization options.

**2) Improved Error Handling and Logging:**
- Added error handling to ensure that any issues are reported.
- Improved logging to provide more detailed information about the process.

**3) Optimized Code:**
- Used std::move to avoid unnecessary copying of strings.
- Used std::uniform_int_distribution for better random number generation.


hermensbas:
I will test this once possible and completed.

icemansparks:
For testing I made sure that:
- I have exissting player / admin accounts 
- The existing accounts have characters on them

I tested:
- [x] Compiling
- [x] starting up the server with playerbots enabled
- [x] verified account, character and playerbots_random_bots entries are chreated with expected IDs and references
- [x] restarted with "delete bots" option to verify all bots get deleted and no player account or character is being removed
- [x] restarted again with bots active to verify bot generation

ScreenShot 2025-01-14_11 17 20@2x

One Problem I had: the first bot account seems to be skipped in character creation:
ScreenShot 2025-01-14_11 24 37@2x
I'll try to fix this today.

ScreenShot 2025-01-14_11 27 55@2x


I observed:
- Errors on restart after deting bots since apparently some entries for bots will not be deleted (like entries in character_action, pet_spell or the arena team association, possibly more as well - this should probably be an optimization in a separate issue)
- A weird error with the message that an accoutn cannot be delted because it isn't a bot account, while the stated account ID was a bot account and the account was deleted in the process as expected... I'll try to figure this out today as well.
`Deleting account accID: 816(19)...
Deleting characters for account accID: 817(20)...
Account ID 817 is not a bot account. Skipping character deletion.`

icemansparks:
I tested again and cannot reproduce the previuosly problem of characters being only created for the second and thrird bot account - all following initialisations and removals went as expected (as far as i can tell).

I think this might be ready to be review. :)

src/RandomPlayerbotFactory.cpp
```diff
@@ -369,139 +418,284 @@ std::string const RandomPlayerbotFactory::CreateRandomBotName(NameRaceAndGender
...
         sPlayerbotAIConfig->randomBotAccounts.push_back(accountId);
 
         uint32 count = AccountMgr::GetCharactersCount(accountId);
-        if (count >= 10)
+        if (count >= sPlayerbotAIConfig->maxAddedBots) // Check against the config value
         {
             continue;
         }
-        LOG_INFO("playerbots", "Creating random bot characters for account: [{}/{}]", accountNumber + 1,
+        LOG_INFO("playerbots", "Creating random bot characters for account: [{}/{}]", accountId,
```
liyunfan1223: `accountId` should be replaced with an incremental counter, We shouldn't print actually account id.

src/RandomPlayerbotFactory.cpp
```diff
@@ -369,139 +418,284 @@ std::string const RandomPlayerbotFactory::CreateRandomBotName(NameRaceAndGender
...
+    stmt->SetData(0, accountId);
+    PreparedQueryResult result = LoginDatabase.Query(stmt);
+    if (!result)
+        return;
+
+    Field* fields = result->Fetch();
+    std::string accountName = fields[0].Get<std::string>();
+
+    // Check if the account name starts with the configured bot account prefix
+    if (accountName.find(sPlayerbotAIConfig->randomBotAccountPrefix) != 0)
```
liyunfan1223: Should it be changed to `if (accountName.find(sPlayerbotAIConfig->randomBotAccountPrefix) == std::string::npos)` and abort delete at `DeleteRandomBotAccounts`?

When deleting accounts, the following info is printed incorrectly (they are bot accounts), and accounts are still deleted.
```
Deleting all random bot characters, 701 accounts collected...
Deleting characters for account accID: 304(1)...
Account ID 304 is not a bot account. Skipping character deletion.
Deleting account accID: 304(1)...
Deleting characters for account accID: 305(2)...
Account ID 305 is not a bot account. Skipping character deletion.
```

liyunfan1223 (changes requested):
Some log info and condition checks should be fixed.

Calling `AccountMgr::DeleteAccount` should also delete all characters under the account at the same time. So is `DeleteBotCharacter` redundant?

In addition, there is also a known issue that the database cannot be completely cleared when deleting the account, so sometimes we need scripts in https://github.com/liyunfan1223/mod-playerbots/wiki/Playerbot-Queries#clear-bots. It would be best if the scripts could be added to the deletion process, or if there were other methods to ensure a complete clearing of the database.

icemansparks:
> AccountMgr::DeleteAccount

I'll have a look at this and the requested changes as well - it seems like I can probably get rid of `DeleteBotCharacter` in the end.

src/RandomPlayerbotFactory.cpp
icemansparks: The Issue has been addressed in latest changes, we use a counter variable now.

src/RandomPlayerbotFactory.cpp
icemansparks: `DeleteRandomBotAccounts` has been removed as it was redundant. In addition, account name check has been adjusted to use Uppercase values. 
The function call has been adjusted to not check if the account is a randombot account or a user account as it will only delete randombot accounts which have been passed to it anyway. (the check happens before already).

icemansparks:
> https://github.com/liyunfan1223/mod-playerbots/wiki/Playerbot-Queries#clear-bots.

I also left a comment in the code to add a cleanup method which could handle removing of outdated botaccount reference data like mentioned in the link.

I don't have time in the next two weeks, but will probably have a look in february again.

icemansparks:
> > > https://github.com/liyunfan1223/mod-playerbots/wiki/Playerbot-Queries#clear-bots.
> > 
> > 
> > I also left a comment in the code to add a cleanup method which could handle removing of outdated botaccount reference data like mentioned in the link.
> 
> I think the following makes more sense:
> 
> USE `acore_playerbots`; DELETE FROM `acore_playerbots`;

I agree, but its not only the data in the dedicated table, but as mentioned by @liyunfan1223 before, there are sometimes problems with deleting botAccounts where some things don't get deleted properly. 
I observed this as well, when continuosly creating and deleting accounts - sometimes upon creating accounts in the same database again leads to error with pet_spell or arena team association as the entries are still rpesent and have not been deleted before when the account was deleted. (I can't tell yet if this is a general problem in the core, or just a problem of bot accounts)
However, as staed before, this should probably be a dedicated issue :)

src/RandomPlayerbotFactory.cpp
```diff
@@ -369,76 +420,207 @@ std::string const RandomPlayerbotFactory::CreateRandomBotName(NameRaceAndGender
...
+
+    totalBotAccounts = sPlayerbotAIConfig->randomBotAccountCount;
+    currentBotAccountCount = GetBotAccountIds().size();
+
+    // Get the next bot account counter
+    uint32 botAccountCounter = GetNextBotAccountCounter();
+
+    if (botAccountCounter == 0 || totalBotAccounts == 0)
+    {
+        LOG_ERROR("playerbots", "Failed to create bot accounts. Aborting random bot creation.");
```
liyunfan1223: When I delete bot accounts and recreate it, it stops here and print error log as `botAccountCounter` starts from 0. Have you tested this case?

icemansparks: Damn that looks like a dirty copy and paste error - I will have a look later on, sorry about that.

icemansparks: Fixed init of botAccountCounter return variable - we now have at least "1" to start the first bot account. I also did slight adjustments to the order of the checks and variable loading to  only load them after the check in case they're not needed before.

src/RandomPlayerbotFactory.cpp
```diff
@@ -448,61 +629,82 @@ void RandomPlayerbotFactory::CreateRandomBots()
...
+    uint32 currentAccountIndex = 0;
 
+    for (uint32 accountId : botAccountIds)
+    {
+        currentAccountIndex++;
         sPlayerbotAIConfig->randomBotAccounts.push_back(accountId);
 
         uint32 count = AccountMgr::GetCharactersCount(accountId);
-        if (count >= 10)
+        if (count >= sPlayerbotAIConfig->maxAddedBots) // Check against the config value
```
liyunfan1223: sPlayerbotAIConfig->maxAddedBots has nothing to do with the number of characters per account, it limits the number of times `.playerbot bot add` is used in game. The chars per account should not exceed 10 (the creation will actually fail), or other errors will occur.

src/RandomPlayerbotFactory.cpp
```diff
@@ -448,61 +629,82 @@ void RandomPlayerbotFactory::CreateRandomBots()
...
+ * @param accountName The name of the account to be created.
+ */
+void RandomPlayerbotFactory::CreateAccount(const std::string& accountName)
+{
+    std::string password = "";
+    if (sPlayerbotAIConfig->randomBotRandomPassword)
     {
-        /* wait for async accounts create to make character create correctly, same as account delete */
-        LOG_INFO("playerbots", "Waiting for {} accounts loading into database...", account_creation);
-        std::this_thread::sleep_for(10ms * sPlayerbotAIConfig->randomBotAccountCount);
```
liyunfan1223, line 473: The sleep between account creation and characters creation shouldn't be removed, otherwise chars cannot be created for all accounts.

liyunfan1223 (commented):
After testing, there are still some things that are not as expected. Account creation is a relatively fragile part, and I'm worried that this refactor will cause more unpredictable issues.

I re-read the issues you mentioned, and I think they have nothing to do with account creation. A simple way is to clear `acore_playerbots.playerbots_random_bots` by `.playerbot rndbot reset` command after the database is modified. The server will automatically repopulate this table after restart, and it should work normally. I plan to submit a PR to clear the add value in `playerbots_random_bots` every time server restarts, which should solve these issues.

src/RandomPlayerbotFactory.cpp
icemansparks: Reverted.

src/RandomPlayerbotFactory.cpp
icemansparks: I added the check back into the appropriate place and adjusted it to use the number of newly created bot accounts instead of always loading the total number from the config since we shoudl not need to wait for existing accounts but only for newly created ones. (The case that the config was changed from 10 Bot accounts to 100, and we need for 90 accounts to be created now)

icemansparks: Still i am not entirely sure now since I previously moved the waiting to CreateBotCharacters already - now it seems we have it duplicated?

liyunfan1223:
My thought is that non-bot characters will never be added to playerbots_random_bots. If it happens due to accidental reasons such as importing manual sqls, just clear playerbots_random_bots table. This table will be dynamically regenerated based on the current accounts and chars.

icemansparks:
> My thought is that non-bot characters will never be added to playerbots_random_bots. If it happens due to accidental reasons such as importing manual sqls, just clear playerbots_random_bots table. This table will be dynamically regenerated based on the current accounts and chars.
> 
> 
> 
> PR is here: https://github.com/liyunfan1223/mod-playerbots/pull/913

The issue was never caused by any manual interaction with the playerbots_random_bots table, but rather by the function that adds data to this table based on random bot account count while the account referenced in the entry is not the correct one (Caused by using loop variables as reference for the account in this table instead of using the actual account id).

As described in the original issue, this can be easily reproduced by creating a normal player / admin account before enabling playerbots.

I'll have a look at your PR as soon as I can and test this scenario with your changes as well.

liyunfan1223:
Can you point out where the loop variable is? If you are talking about the id column of playerbots_random_bots, this is just an auto-incremented primary key. It does not represent the account id. The only identifier recorded in the table is the guid of characters.

icemansparks:
I tried to point it out here: https://github.com/liyunfan1223/mod-playerbots/issues/449#issuecomment-2285734870 - but at the time i did not dive into the code that much yet.

I am trying to reproduce the behavior which i observed in august 2024, but can't get the same results. Since there are a ton of new commits since my original bug-report, it will be tedious to find out what went wrong on my end at this point.

As I could safely reproduce the error back then, i came to the conclusion that the problem did lie in the following loop
ScreenShot 2025-01-26_12 02 33@2x

Nonetheless, this screenshot is now taken from the commit I referenced in the original report and I cannot seem to find any problem with it... leaving me wondering how I ended up with players' character guids in playerbots_random_bots table in the first place...

I'll only be able to investigate this further next week though.

icemansparks:
I would not have put all the effort in this PR if there hadn't been (in my eyes) a serious problem when enabling playerbots / adding playerbots to an existing playerbase.

@liyunfan1223 honestly, I'll probably have to revisit the whole thing from scratch and if there is still something to find, I'll make a fresh PR with a smaller amount of changes in the upcoming weeks. 
This way we don't need to worry about too many adjustments in the account / character creation process at once.

Of course, if you don't see any more problems with the current PR, I am happy if the work doesn't go to waste - but if there is any uncertainty, lets close it for the better :)

Still, a bit refactoring of the creation process would probably be good to have it structured and more maintainable as it really took some time to get into all the things happening in this process and I think it could be easier with the use of smaller functions for some of the included tasks.

src/RandomPlayerbotFactory.cpp
```diff
@@ -370,76 +421,215 @@ std::string const RandomPlayerbotFactory::CreateRandomBotName(NameRaceAndGender
...
+ * highest bot account username and calculates the next available bot account number.
+ *
+ * This is important to ensure that new bot accounts are created with unique and sequential
+ * account numbers, avoiding conflicts with existing accounts.
+ *
+ * @return The next available bot account counter.
+ */
+uint32 RandomPlayerbotFactory::GetNextBotAccountCounter()
+{
+    QueryResult botAccountsResult = LoginDatabase.Query("SELECT username FROM account WHERE username LIKE '{}%%' ORDER BY username DESC LIMIT 1",
```
liyunfan1223, line 595: Order by on username is unreasonable, consider `rndbot99`>`rndbot300`.

liyunfan1223 (commented):
I appreciate your improvement to the readability of the code, but it actually introduces potential issues that, although a large part has been fixed, I don't think have been fully fixed yet.

We have spent a lot of effort, and my preference is to close this PR first. We need to clarify the purpose of the PR. If it is just for readability, we should not make numerous modifications to the logic without full assurance.

PR #874, opened 2025-01-15
Added ability to specify bot target for spell casting

achyles222:
1.	Added the ability to explicitly specify a target for spell casting using the " on PlayerName" syntax
2.	The code now parses command like "cast spell_name on PlayerName" where:
	• 	It extracts the player name after " on "
	•	Verifies if the spell exists without the target part
	•	Attempts to find the target player by name
	•	If found, sets that player as the spell target

[Video preview](https://www.youtube.com/live/mEs0fPy-_hU)

### DIFF
--- a/src/strategy/actions/CastCustomSpellAction.cpp
+++ b/src/strategy/actions/CastCustomSpellAction.cpp
@@ -54,6 +54,33 @@ bool CastCustomSpellAction::Execute(Event event)
         ltrim(text);
     }
 
+    uint32 spell = 0;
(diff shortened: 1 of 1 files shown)


hermensbas:
@Dreathean should we add the wiki docu first?

Dreathean:
Very cool and useful, thanks for implementing this! Agreed I'll get this added onto the wiki if the code looks good with you.

hermensbas:
code looks fine

PR #884, opened 2025-01-21
Improve ICC spread and fix Valithria trigger

liyunfan1223:
Hello @Noscopezz! First of all thank you for your fancy ICC strategy.

This PR makes some improvements to spread. It reduces the movement time required for spread and the number of lost spells for ranged caster, although it's still not perfect. I tested it in every situation I could think of and it works fine.

Another fix is ​​for Valithria trigger. It cannot be triggered now as `find target` is only effective for the hostile target.


Noscopezz:
Nice thanks, I was trying to fix valithria but couldnt figure out why it didnt work all of a sudden :D

PR #886, opened 2025-01-21
Tweaked Alterac Valley so bots kill final bosses

avirar:
1. Bots would only engage final boss when 4/4 towers were destroyed and they controlled the graveyard near the boss. Reduced tower requirement to any 2 of the towers. Bot still cap the GY before attempting to kill the boss.

2. Also adjusted the defender/attacker role chances so there are more attackers

I've seen them kill bosses in the last 3 out of 5 games. Stalemates still occur occasionally, but it seems much better than before.


avirar:
Adjusted some stuff for WSG too, support role bots will moveto their teams carrier if not in combat and >20yds away

PR #898, opened 2025-01-23
ICC A/H Raid buff

Noscopezz:
Fixed missing ICC buff for A and H

Buff will only be present when logged on and in ICC, once any bot or player leave ICC the buff is gone to prevent abuse.

This will make ICC easier now and with recent DPS update and movement improvement bots will now actully do decent dps and even greater healing.

Ally buff
https://www.wowhead.com/wotlk/spell=73828/strength-of-wrynn

Horde buff
https://www.wowhead.com/wotlk/spell=73822/hellscreams-warsong


liyunfan1223:
ICC buff will be automatically added when entering the instance on my end, even without this modification.
image

Noscopezz:
> ICC buff will be automatically added when entering the instance on my end, even without this modification. image

I will close this, since its a problem on my end, thanks for testing and verifying.

PR #901, opened 2025-01-24
Add undead-rogue to rndbots

jublinq:
Default race-class combination undead-rogue absent from rndbot creation pool in RandomPlayerbotFactory.cpp.

Added undead-rogue to rndbot combinations.


Vortikai:
Still missing blood elf rogues it seems. I don't know the systems behind all this, but it might be a good idea to check all race/class combos again?

jublinq:
Thanks @nl-saw 

@Vortikai I haven't experienced this, in fact have come across a number of BE rogues in my builds. I'm not very familiar with the code either but I think lines 83-86 controls for BE rogues where the build is at minimum at BC or later:
```
if (expansion >= EXPANSION_THE_BURNING_CRUSADE)
    {
        availableRaces[CLASS_ROGUE].push_back(RACE_BLOODELF);
    }
```

I did take a look through the other race-class combinations and didn't find other absences; wouldn't hurt for a pass from another set of eyes.

PR #913, opened 2025-01-26
Reset added rndbots on server restart to automatically apply configuration changes

liyunfan1223:
No longer need to manually execute the reset command after modifying configuration about added rndbots. It should also fix some issues that added rndbots does not match the database status. https://github.com/liyunfan1223/mod-playerbots/issues/449 

And some minor refactorings and optimizations.


liyunfan1223:
Hello @icemansparks. Can you give it a try to see if it solves the issue?

icemansparks:
> Hello @icemansparks. Can you give it a try to see if it solves the issue?

I could not find any issues with the changes, neither am i able to reproduce the problem (with the current commit) that I originally observed in august.

PR #914, opened 2025-01-26
Improve long distance movement in new rpg strategies

liyunfan1223:
Improve long distance movement in new rpg strategy and add a stuck check.

After some attempts I found that it is very difficult to ensure the success of long-distance movement on complex terrain without hard-coding waypoints. So when the bot is stuck for a long time, it will fallback to teleport to ensure the position is valid.


hermensbas:
Ive been playing that for a while as well, maybe we should considering using static waypoints but only when longtravel is been dectected. Did you get the boat travels to and such? I got so frustrated i ended up reverting all my changes haha

liyunfan1223:
Yeah it's a little frustrating haha. I have also considered transport like boat and elevator, which is technically feasible, but I haven't figured out what to do next.

Issue #894, opened 2025-01-23
(PVP\bgEye) Bots prioritize Flag over life


blinkysc:
**Bug Description**
Bots in Eye of Storm will hyper focus flag instead of just trying to have the upper hand before capture

**How To Reproduce Bug**
Join eye with bots

hermensbas:
https://github.com/liyunfan1223/mod-playerbots/pull/924

PR #924, opened 2025-01-28
Eye of the Storm changes

avirar:
1. Flag carriers will be defended by bots that are also trying to cap the flag.
2. Replaced Follow with MoveNear while not in combat.
3. Reduced action priority to ACTION_MOVE so bots can still perform some actions while running the flag.


avirar:
#894

blinkysc:
Seems bots, when capturing the flag, will still be hyper focused on flag and you can just cheese them by continuing to kill them all as they do so.

avirar:
I'll be away until Monday and won't be able to look at it further until then sorry, I'm not precious about it so make any changes you guys think would help :) Good luck, I'll check messages when i can

blinkysc:
Sounds good

hermensbas (commented):
Not really related but i like your using inline comments for your intensions.

PR #927, opened 2025-01-29
fix(BotAI/BattleGrounds) Bots AI improved based around capture points

blinkysc:
What it does
[x] Bots will now defend players/other bots with capture points (flags/and flag points ie AB)
[x] Bots will defend themselves if attacked while trying to capture

What it doesn't do 

Bots probably still need more work on general strategy but is a good start

Thanks
@avirar his recent PR sparked the idea; look forward to working together for more PR's
@nl-saw for testing with his bot config to verify no crashes

Probably deserves some more testing just in case... First iteration I wrote created crash so want to make sure its g2g...

### DIFF
--- a/src/strategy/actions/BattleGroundTactics.cpp
+++ b/src/strategy/actions/BattleGroundTactics.cpp
@@ -4097,8 +4097,51 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
     if (closeObjects.empty())
         return false;
 
+    // First check if there are enemies near flag we're trying to capture
+    GameObject* targetFlag = nullptr;
+    for (ObjectGuid const guid : closeObjects)
+    {
+        GameObject* go = botAI->GetGameObject(guid);
+        if (!go)
+            continue;
+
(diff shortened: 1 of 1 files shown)


Dreathean:
This is awesome, this applies to all BGs with capture points ie AB and AV+?  This was one of the last remaining major immersion breaking issues for these so really excited for this. Wish I could test in a decent timeframe but likely won’t be able to check it out until this weekend.

Going to go hand in hand very nicely with avirar’s PR!

blinkysc:
> This is awesome, this applies to all BGs with capture points ie AB and AV+? This was one of the last remaining major immersion breaking issues for these so really excited for this. Wish I could test in a decent timeframe but likely won’t be able to check it out until this weekend.
> 
> Going to go hand in hand very nicely with avirar’s PR!

Yeah works for all capture points (though if you notice bug do tell)

Dreathean:
Definitely, will do!

Dreathean:
@blinkysc just want to follow up on this and say this has been a tremendous improvement! Tested in both AB and AV, so much better and engaging around capture points. Sometimes they may take a second to react but I think that's more just related to their action updates. 

Thank you for helping make the BGs more awesome, they are in such a great state now after this and other BG-updates.

x92mko:
Thank you so much for this huge improvement!

PR #930, opened 2025-01-31
Improve login and init permission for rndbots and addclass bots

liyunfan1223:
This PR change the `.playerbot bot init` command permissions that may break the habits of existing players, so looking for some testing and feedback. It would be better to also reach a consensus on the definition of different types of bots (altbots, rndbots and addclass bots). 

After this modification, `.playerbot bot init` only takes effect for bots added by `.playerbot bot addclass` command, and is invalid for any bot added directly using character name. The intention of this modification is to prevent offline rndbots (after setting EnableRotation) from being modified by the init command. By design, rndbots should be independent and should not be arbitrarily modified by any players. 

Some players may be used to adding addclass bot to friends or guilds, then logging it through the name and using `init` command. This was allowed before, but is now prohibited. My thought is that this change would allow for a better distinction between rndbots and addclass bots (perhaps needing a better new name).

In addition, this PR removes some redundent configs and pulls up the bot permission check (adding a bot without permission will no longer cause it to briefly log in and out)

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -125,6 +125,9 @@ AiPlayerbot.GroupInvitationPermission = 1
 # auto-login all player alts as bots on player login
 AiPlayerbot.BotAutologin = 0
 
+# Allow/deny bots from your account
+AiPlayerbot.AllowAccountBots = 1
+
 # Allow/deny bots from your guild
 AiPlayerbot.AllowGuildBots = 1
 
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -125,10 +125,10 @@ AiPlayerbot.GroupInvitationPermission = 1
 # auto-login all player alts as bots on player login
 AiPlayerbot.BotAutologin = 0
 
-# Allow/deny bots from your account
+# Allow/deny bots from the player's account
 AiPlayerbot.AllowAccountBots = 1
 
-# Allow/deny bots from your guild
+# Allow/deny bots in the player's guild
 AiPlayerbot.AllowGuildBots = 1
 
 # Randombots will invite nearby bots to guilds
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -107,11 +107,8 @@ AiPlayerbot.DeleteRandomBotAccounts = 0
 #
 #
 
-# Maximum number of bots added by one account
+# The maximum number of bots that a player can control simultaneously
 AiPlayerbot.MaxAddedBots = 40
-
-# Maximum number of bots per class added by one account
-AiPlayerbot.MaxAddedBotsPerClass = 40
 
 # Enable/Disable create bot by addclass command (0 = GM only, 1 = enable)
 # default: 1 (enable)
@@ -127,10 +124,6 @@ AiPlayerbot.GroupInvitationPermission = 1
 
 # auto-login all player alts as bots on player login
 AiPlayerbot.BotAutologin = 0
-
-# Allow login other players' characters as bots
-# Default: 0 (disabled)
-AiPlayerbot.AllowPlayerBots = 0
 
 # Allow/deny bots from your guild
 AiPlayerbot.AllowGuildBots = 1
@@ -860,8 +853,8 @@ AiPlayerbot.FastReactInBG = 1
 AiPlayerbot.RandomBotUpdateInterval = 20
 AiPlayerbot.RandomBotCountChangeMinInterval = 1800
 AiPlayerbot.RandomBotCountChangeMaxInterval = 7200
-AiPlayerbot.MinRandomBotInWorldTime = 3600
-AiPlayerbot.MaxRandomBotInWorldTime = 1209600
+AiPlayerbot.MinRandomBotInWorldTime = 600
+AiPlayerbot.MaxRandomBotInWorldTime = 28800
 AiPlayerbot.MinRandomBotRandomizeTime = 7200
 AiPlayerbot.MaxRandomBotRandomizeTime = 1209600
 AiPlayerbot.RandomBotsPerInterval = 60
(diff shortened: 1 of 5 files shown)


zbhcn:
my suggestion is that rndbot should not automatically initialize after each restart, otherwise it may cause a crash. Luckily, I have playerbot database backup, otherwise, I wouldn't be able to recover from the catastrophic damage.

noisiver:
> my suggestion is that rndbot should not automatically initialize after each restart, otherwise it may cause a crash. Luckily, I have playerbot database backup, otherwise, I wouldn't be able to recover from the catastrophic damage.

Why does it crash on your end? Maybe that's something that should be looked at. I have never seen a crash on startup that was caused by the bots and even if something else caused a crash while the bots were initializing it just restarted and continued like normal with no loss of data.

mtm84:
It was my understanding that addclass chose a character of that class from a rndbot account?  Adding in different types of bots might be the best solution but would no doubt require a fair bit of simple yet tedious logic changes across the code base.
Personally I would like to see a separation between totally random bots that, say, randomize level+gear on login and that you would use with addclass, and more persistent "rpg" bots that act more like a normal player would and that normal players have much less "god" powers over.  I don't thinks that is strictly necessary however.

noisiver:
addclass grabs bots from the last X number of rndbot accounts (the X being specified in the playerbots config) so they aren't logged in as other rndbots and are kind of separate already. At least that's my understanding of the addclass command.

liyunfan1223:
> I am curious about this removal:
image 
What would the default be in that case? Because on or off makes a pretty significant difference for the server and I can see people using playerbots with or without altbots specifically.

I re-modified the configuration and now the altbots permission control will be based on the following, which should disambiguate. The old configuration `AllotPlayerbots` conflicts with `AllowGuildBots` and is therefore ambiguous.
```
# Allow/deny bots from your account
AiPlayerbot.AllowAccountBots = 1

# Allow/deny bots from your guild
AiPlayerbot.AllowGuildBots = 1
```

liyunfan1223:
I realize that addclass bots are generally not the best option for normal gameplay, and I actually use them mostly for debugging functions (like quickly setting up a raid to test ICC)

Dreathean:
> 
> I re-modified the configuration and now the altbots permission control will be based on the following, which should disambiguate. The old configuration `AllotPlayerbots` conflicts with `AllowGuildBots` and is therefore ambiguous.
> 
> ```
> # Allow/deny bots from your account
> AiPlayerbot.AllowAccountBots = 1
> 
> # Allow/deny bots from your guild
> AiPlayerbot.AllowGuildBots = 1
> ```
I like this, the original setup was a little redundant and I like that you can disable altbots coming from your own account versus strictly other accounts if I'm reading this right, I'm sure some people will appreciate that separation. Though I do wonder what would stop people from adding other real player's characters as bots if they are in the same guild? But I guess that would always have been an issue, it's not something I've messed with.

Dreathean:
> I realize that addclass bots are generally not the best option for normal gameplay, and I actually use them mostly for debugging functions (like quickly setting up a raid to test ICC)

I agree. I've been trying to think more about what they should be called, but it largely depends on what people agree their purpose should be for. What would you guys think about them being called and treated as "tempbots" for specific, temporary use? Though that would also probably require some further adjustments to them as well.

noisiver:
The issue is that a lot of people use them as a permanent thing. Being able to further restrict random bots could replace it though.

Options to stop them from changing their talents, glyphs, and make absolutely sure they never change their gear if equipment persistence is enabled could be some of those things. The last time I tried using random bots even if equipment persistence was activated they changed it when they decided to change their spec. You could tell them to change stuff using the talents spec command as well as maintenance and autogear but after they've picked a spec (or been told to pick one) they won't change it on their own. Again, it could and probably should be configurable.

Dreathean:
I feel like (could be wrong) most people using them like that only do so because you can’t get your altbots to grind on their own. If altbots could grind then you’d pretty much have all of that.

of course, if those were added to that type of bot like your saying, then perhaps that would be the optimal progression bot, but then again you still don’t have the ability to choose name/race/etc like you can with altbots unless you have database access, which would be an issue for any multiplayer server. Also I’m not sure how add class bots are handled when doing a full rndbot reset.

noisiver:
The people who use addclass probably don't want to use altbots because it takes longer to set up and more than 10 characters means you need more than 1 account and that means you need to form a guild to use them.

Vortikai:
> > I am curious about this removal:
> > image
> > What would the default be in that case? Because on or off makes a pretty significant difference for the server and I can see people using playerbots with or without altbots specifically.
> 
> I re-modified the configuration and now the altbots permission control will be based on the following, which should disambiguate. The old configuration `AllotPlayerbots` conflicts with `AllowGuildBots` and is therefore ambiguous.
> 
> ```
> # Allow/deny bots from your account
> AiPlayerbot.AllowAccountBots = 1
> 
> # Allow/deny bots from your guild
> AiPlayerbot.AllowGuildBots = 1
> ```

Might consider making the description to be something like `# Allow/deny bots from the player's account` and maybe `# Allow/deny bots from accounts in the player's guild`.

liyunfan1223:
Merge since it doesn't break existing functionality. Consider how to adapt the usage of addclass bots in another PR.

parazitenew:
As a player I noticed this change on the server i'm playing in. The problem that caused me is that I'm not able to log in the bots who are in my arena team anymore. The addclass using botname in not working anymore, also means when I start a raid, and then I want to finish it the day after, I can't recall the same bots with `addclass * ` with Unbot addon, and that makes me addclass new bots, since we are around 15 human players, doing that bring the ID raid issue, if we login new bots every day, most of them will have ID, and we can't use them. The init on altbot (the bots logged in with addclass) is important, cause to make a raid you need bots with your equivalent equipment level. The progression of bots is not persistent, so how could we do raid if init is disabled ?

liyunfan1223:
Reverted and fixed addclass bots related issues. https://github.com/liyunfan1223/mod-playerbots/pull/946

PR #961, opened 2025-02-10
German full translation

Raz0r1337:
With the power of ChatGPT i translated all chatter texts into German locale.
Inserted the German texts into text_loc3.
Everything else stay unchanged.

My test server has been running with the changes for five hours now with 2000 active bots.
There have been no crashes or errors.


Raz0r1337:
Second try.
I spent the whole day in fixing the files to generate me a new translated file which does change nothing but the text_loc3.
Now it should have no problems.

Raz0r1337:
I'm happy that I can help you.
I will continue to adapt and refine the translation in future, depending on how everything looks in the game. When enough has been collected, I will start the next Pull request.

PR #963, opened 2025-02-12
Add project announcement on server initialization and character login

liyunfan1223:
Add announcement on server initialization and character login.

For repack distribution and public server host, avoid unintentional violations of AGPLv3.0 license.

image

image

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -98,6 +98,17 @@ public:
             sPlayerbotsMgr->AddPlayerbotData(player, false);
             sRandomPlayerbotMgr->OnPlayerLogin(player);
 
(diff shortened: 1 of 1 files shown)


Dreathean:
Azerothcore does not directly allow you to change their message and you are violating their terms if you do so as well. This change is to be in line with Azerothcore.

liyunfan1223:
This is consistent with the terms of the license, the host should provide the source code to anyone who can connect to the server.

The only difference is that it does not need to be announced every time a player logs in and the source code only need to be provided on demand. Developers who are familiar with the terms usually also have the ability to modify the source code according to their own preference. So for now I want to keep it as is.

PR #964, opened 2025-02-12
Auto calculate RandomBotAccountCount by default

nl-saw:
I see too much confusion on the RandomBotAccountCount setting, and frankly it is not really required at all.

- Adjusted RandomBotAccountCount to be 0 by default in config, meaning auto calculation.
- Adjusted some defaults so they make sense, out-of-the-box.

### DIFF
--- a/src/RandomPlayerbotFactory.cpp
+++ b/src/RandomPlayerbotFactory.cpp
@@ -400,8 +400,8 @@ void RandomPlayerbotFactory::CreateRandomBots()
     {
         std::vector<uint32> botAccounts;
         std::vector<uint32> botFriends;
-
-        for (uint32 accountNumber = 0; accountNumber < sPlayerbotAIConfig->randomBotAccountCount; ++accountNumber)
+        uint32 totalAccountCount = (sPlayerbotAIConfig->maxRandomBots / 10) + sPlayerbotAIConfig->addClassAccountPoolSize;
+        for (uint32 accountNumber = 0; accountNumber < totalAccountCount; ++accountNumber)
         {
             std::ostringstream out;
             out << sPlayerbotAIConfig->randomBotAccountPrefix << accountNumber;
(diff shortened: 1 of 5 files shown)


noisiver:
You can't make such an assumption every time. If expansion is not set to WotLK you have to deal with the fact that each account only holds 9 characters, not 10.

src/RandomPlayerbotFactory.cpp
```diff
@@ -440,11 +441,11 @@ void RandomPlayerbotFactory::CreateRandomBots()
 
     LOG_INFO("playerbots", "Creating random bot accounts...");
     std::unordered_map<NameRaceAndGender, std::vector<std::string>> nameCache;
-    uint32 totalAccCount = sPlayerbotAIConfig->randomBotAccountCount;
+    uint32 totalAccountCount = (sPlayerbotAIConfig->maxRandomBots / 10) + sPlayerbotAIConfig->addClassAccountPoolSize;
```
icemansparks: I could be wrong but since its integer division, decimals will be cut off here, this could potentially lead to "not enough accounts" for creating the desired maxRandomBots number. (Not sure how addClassAccountPoolSize is used in the code though)

555 / 10 = 55 -> accounts for only 550 characters

rounding up or adding 1 account could prevent this?

nl-saw: Ah yes I missed the +1 on the end, thx

hermensbas:
confirmed

liyunfan1223:
Nice work. To add, some configurations affect the calculation of accounts num:

1. If `AiPlayerbot.DisableDeathKnightLogin = 1`, each account can only have 9 characters, similar to the WotLK expansion setting.
2. The percentage ratio of horde and alliance can reduce the number of characters per account.
For example, the following configuration will reduce the available characters per account by half as only horde characeters are available:
```
# Percentage ratio of Alliance and Horde  
AiPlayerbot.RandomBotAllianceRatio = 0  
AiPlayerbot.RandomBotHordeRatio = 100  
```

liyunfan1223:
When creating an account, each account always has 10 characters (one for each class, randomly set to alliance or horde). The advantage of this is that no matter what the configuration is, at least half of the roles are available and do not need to be recreated when the configuration is modified.

liyunfan1223:
Noticing that the modification to RotationPoolSize is different from its original intention. RotationPoolSize should be set greater than MaxRandomBots (number of `MaxRandomBots` bots are randomly online among total `RotationPoolSize` bots). I'll refine the ambiguity on this later.

PR #969, opened 2025-02-13
Re-instate random bots guilds

nl-saw:
I believe this should have been enabled again at some point.

https://github.com/liyunfan1223/mod-playerbots/issues/968


noisiver:
Can the bots actually invite members after it has been created now? I know the issue before was that after the guild were formed nothing more happened so there were just a bunch of guilds with 10 members in them.

liyunfan1223:
It's fine to enable it again, but there are many issues to work around in future. I will make a PR for addclass bots as they can not be summoned by addclass while the bots are in a guild.

In addition, if all rndbots have already joined the guild, it will be troublesome for players to register the guild. Although `guild leave` command can achieve this, it is not very friendly now. Perhaps there should also be a configurable proportion of guildless rndbots left so they can join player guilds. In my opinion, the most difficult thing is to reach a consensus, but subsequent adjustments can be made through feedback.

PR #975, opened 2025-02-15
Only invoke InitGuild when randomBotGuildCount > 0

nl-saw:
Only invoke InitGuild(); when sPlayerbotAIConfig->randomBotGuildCount > 0 to avoid continues "No random guilds available" if disabled.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -132,6 +132,12 @@ AiPlayerbot.AllowAccountBots = 1
 # Allow/deny bots in the player's guild
 AiPlayerbot.AllowGuildBots = 1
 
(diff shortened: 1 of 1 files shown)


kadeshar:
1. Such log should be inside method
2. randomBotGuildCount entry is depraceted. If you want to use, better move that somewhere else in config file
3. Also if we restoring randomBotGuildCount we should take into consideration in BuyPetitionAction::canBuyPetition method to dont make too much guild by bots. Currently such condition dont exists.

kadeshar:
1. Should because for reusability such thing should be in one place. Dont have to because most likely it will be never reused :P 
3. I know. I only highlight that to not forget :)

PR #977, opened 2025-02-16
Hodir tactics

kadeshar:
- Added tactic for Hodir Biting Cold ability
- Added tactic for Hodir Flash Freeze


zhangbo8418:
/modules/mod-playerbots/src/strategy/raids/ulduar/RaidUlduarActions.cpp:1177:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
 1177 | }

zhangbo8418:
/modules/mod-playerbots/src/strategy/raids/ulduar/RaidUlduarTriggers.cpp:265:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
  265 | }
      | ^


 //Snowpacked Icicle Target
bool HodirNearSnowpackedIcicleTrigger::IsActive()
{
    // Check boss and it is alive
    Unit* boss = AI_VALUE2(Unit*, "find target", "hodir");
    if (!boss || !boss->IsAlive())
    {
        return false;
    }

    // Find the nearest Snowpacked Icicle Target
    Creature* target = bot->FindNearestCreature(33174, 100.0f);
    return target != nullptr;  // Is this okay?
}

PR #979, opened 2025-02-16
Fixraidbuff

shenkai4257:
fix palading greaterblessing(on inraidgroup)


shenkai4257:
ok,I'll fix it

avirar:
I kinda gave up on my implementation, it mostly works, but i feel it is pretty janky so I haven't made a PR of it. Mixed level groups were the real trouble. It doesn't use the most resources according to pmon anymore at least!

https://github.com/avirar/mod-playerbots/tree/paladinblessingsv3

PR #980, opened 2025-02-17
Fix stuck on knockback, enhance movement & flee and trap weave strats

liyunfan1223:
- Fix bots sometimes stuck in air during knockback.
- Hunter and mage now use their spells (disengage and blink) to flee when the enemy is too close. Hunters will use traps before that.
- Added the "trap weave" strategy for hunter to increase DPS (significant for survival hunt). It is disabled by default for now to avoid conflicts with other strats.
- Adjusted action queue in AI engine to prevent actions from being saved in the queue. Fixing the issue of continuous unnecessary movement.


liyunfan1223:
Sure, this is a nice improvement.

PR #991, opened 2025-02-20
Revert expireActionTime

nl-saw:
Fix for https://github.com/liyunfan1223/mod-playerbots/issues/989 - Bots no longer use Flightmaster.


It0w:
I've done a few quests and used the flight master a couple times. It seems to be working again.

I'm not sure if it's related to bb729e3, but the bot no longer gets stuck in the loot animation.

liyunfan1223:
Agree that we should revert it for now. I'll look into this later.

endymion2k14:
I had issues with my bots not auto equipping better gear anymore. I think this revert also fixed that.

Im not 100% certain.

PR #999, opened 2025-02-21
This might fix the possible race condition

nl-saw:
Order Reversal: In both combat and non-combat paths, sLFGMgr->UpdateProposal() is called before clearing the "lfg proposal" ID. This prevents a new proposal from overwriting the ID before it's processed.

Clear ID in Combat Case: The combat case now clears the proposal ID after updating, avoiding reprocessing the same proposal.

Quickly tested, still works ok.

If OK, should be merged before https://github.com/liyunfan1223/mod-playerbots/pull/997

### DIFF
--- a/src/strategy/actions/LfgActions.cpp
+++ b/src/strategy/actions/LfgActions.cpp
@@ -226,6 +226,7 @@ bool LfgAcceptAction::Execute(Event event)
         {
             sRandomPlayerbotMgr->Refresh(bot);
             botAI->ResetStrategies();
+            // bot->TeleportToHomebind();
         }
 
         botAI->Reset();


noisiver:
I assumed this was an issue on the side of AzerothCore and not playerbots. He wrote this in the AzerothCore discord server after all.
{ACCCBECA-3BAD-4974-978B-D38C25DDA434}

noisiver:
Oh. I'm sorry, I misunderstood the title and didn't properly look at what the PR does.

liyunfan1223:
Thanks for following this issue. The one I intend to fix is that `sLFGMgr->UpdateProposal` is not thread-safe, and bots on different maps can do this action in parallel, causing a race condition on sLFGMgr. So we may be looking at different issues, but it would also be great if this PR solves another issue of reprocessing the same proposal.

To explain further, the race condition will trigger a crash although not frequently. I've only encountered this once, since it was running in a release build, I didn't have a crash log, just the runtime frame when it crashed.

image

The solution I think is to remove the direct call to sLFGMgr->UpdateProposal and replace it by simulating a client packet. Similar to https://github.com/liyunfan1223/mod-playerbots/pull/799. That's why I leave a todo mark here.

hermensbas:
> Thanks for following this issue. The one I intend to fix is that `sLFGMgr->UpdateProposal` is not thread-safe, and bots on different maps can do this action in parallel, causing a race condition on sLFGMgr. So we may be looking at different issues, but it would also be great if this PR solves another issue of reprocessing the same proposal.
> 
> To explain further, the race condition will trigger a crash although not frequently. I've only encountered this once, since it was running in a release build, I didn't have a crash log, just the runtime frame when it crashed.
> 
> image
> 
> The solution I think is to remove the direct call to sLFGMgr->UpdateProposal and replace it by simulating a client packet. Similar to #799. That's why I leave a todo mark here.

How does this compare to mutex lock for threadsafety? I am not totally familiar with the client package queque.

PR #1000, opened 2025-02-21
Fix blessings overlapping each other

shenkai4257:
Fixed the issue where Greater Blessing and Blessing were overlapping each other


src/strategy/druid/DruidTriggers.cpp
```diff
@@ -17,6 +17,18 @@ bool MarkOfTheWildTrigger::IsActive()
     return BuffTrigger::IsActive() && !botAI->HasAura("gift of the wild", GetTarget());
 }
 
+bool GiftOfTheWildOnPartyTrigger::IsActive()
+{
```
liyunfan1223, line 21: Add a target check to prevent crash as the target can be pet. https://github.com/liyunfan1223/mod-playerbots/pull/570

```
Unit* target = GetTarget();
if (!target || !target->IsPlayer())
    return false;
```

src/strategy/mage/MageTriggers.cpp
```diff
@@ -18,6 +18,19 @@ bool ArcaneIntellectTrigger::IsActive()
     return BuffTrigger::IsActive() && !botAI->HasAura("arcane brilliance", GetTarget());
 }
 
+bool ArcaneBrillianceOnPartyTrigger::IsActive()
+{
```
liyunfan1223, line 22: Add a target check

src/strategy/paladin/PaladinTriggers.h
```diff
@@ -210,33 +213,132 @@ class SacredShieldOnMainTankTrigger : public BuffOnMainTankTrigger
...
     {
+        static PaladinSelectionGroupManager instance;
+        return instance;
     }
+
+    PaladinSelectionGroup* GetPaladinSelectionGroup(Group* group);
+
+private:
+    std::unordered_map<Group*, std::unique_ptr<PaladinSelectionGroup>> groupCache;
+    mutable std::mutex mutex_;
```
liyunfan1223, line 306: Is singleton and global mutex necessary? It would be best to avoid such an implementation here.

liyunfan1223 (changes requested):

src/strategy/druid/DruidTriggers.cpp
shenkai4257: ok，it's done

src/strategy/mage/MageTriggers.cpp
shenkai4257: ok ，fix it

PR #1008, opened 2025-02-23
Fix chat and beautify server promo box

nl-saw:
- Fixes the issue bots not responding to group chat anymore
- Beautified the promotional box for Playerbots


sudlud:
Looks like I messed it up a bit in #1004.

Why are you adding hooks like "before update"? I didn't see it in the playerbotscript, what am I missing

sudlud:
Also I thought that the order of hooks listed is irrelevant actually, does it have an effect?

sudlud:
alright, thanks for the feedback.

especially any `UPDATE` related hooks should not be enabled if unused, these probably have the most impact on peformance as they're called on each player update for each player / bot.

after looking through the playerscript again, it's using a total of 8 hooks, so these can also be removed imo:

```
PLAYERHOOK_ON_CHAT_WITH_GUILD
PLAYERHOOK_ON_CHAT_WITH_RECEIVER

PLAYERHOOK_CAN_PLAYER_USE_CHANNEL_CHAT
PLAYERHOOK_CAN_PLAYER_USE_GROUP_CHAT
PLAYERHOOK_CAN_PLAYER_USE_CHAT
PLAYERHOOK_CAN_PLAYER_USE_GUILD_CHAT
```

My goal was to have the enabled hook list to be 1:1 with the actually used hooks. (which i messed up obviously in my last PR by mixing up OnChat with CanChat hooks)

sudlud (approved):

PR #1035, opened 2025-02-27
Improve XP rate config

liyunfan1223:
Rename from `KillXPRate` to `PlayerbotsXPRate`.

Affect all xp including kills and quests. Allows to be set to a float number less than 1.0.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -369,9 +369,9 @@ AiPlayerbot.AggroDistance = 22
 #
 #
 
-# Set kill XP rate for bots (default: 1)
-# Server XP Rate * AiPlayerbot.KillXPRate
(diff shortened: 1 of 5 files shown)


Dreathean:
Man I was just wondering about this if it was an int or float after reading the quest PR (which is incredibly exciting by the way!), nice to get confirmation and excellent to see it changed from int to float!

Dreathean:
As it stands bots level way slower regardless, will be interesting to see how much adding quests will impact their progress, it's really nice to allow people to fine-tune the rate in the event that, say, .8 or 1.5 rate is the perfect spot instead of 1 or 2.

DustinHen:
This is a wonderful change!

hermensbas:
And this is the correct way to influence leveling speed not the bot acitivity which has a different purpose. 🥰

PR #1036, opened 2025-02-27
ChatHelper link localization

liyunfan1223:
Get name and title based on default dbc locale. Tested on English and Chinese locale.

Previous localization changes resulted in some crashes. Although I'm handling them carefully in this PR to avoid possible crash situations, it's still nice to do some testing, especially under different DBC.Locale configurations.

https://github.com/liyunfan1223/mod-playerbots/issues/1022

### DIFF
--- a/src/ChatHelper.cpp
+++ b/src/ChatHelper.cpp
@@ -6,6 +6,9 @@
 #include "ChatHelper.h"
 
 #include "AiFactory.h"
+#include "Common.h"
+#include "ItemTemplate.h"
+#include "ObjectMgr.h"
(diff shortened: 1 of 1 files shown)


src/ChatHelper.cpp
```diff
@@ -370,11 +384,18 @@ std::string const ChatHelper::FormatItem(ItemTemplate const* proto, uint32 count
     char color[32];
```
zbhcn, line 384: if (!proto || !proto->Quality || proto->Quality > 7)
    {
        return "未知物品";  // 或者抛出异常
    }

PR #1037, opened 2025-02-27
Update GrindTargetValue.cpp - Added safety check

nl-saw:
Crash occurred due to a segmentation fault (SIGSEGV) in the Creature::GetAggroRange function. The function is being called on a null object.

### DIFF
--- a/src/strategy/values/GrindTargetValue.cpp
+++ b/src/strategy/values/GrindTargetValue.cpp
@@ -116,9 +116,13 @@ Unit* GrindTargetValue::FindTargetForGrinding(uint32 assistCount)
             botAI->rpgInfo.status == RPG_GO_INNKEEPER ||
             botAI->rpgInfo.status == RPG_DO_QUEST;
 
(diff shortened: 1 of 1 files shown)


noisiver:
Sure, I'll give it a go. I don't understand why it only happened on Linux though but it doesn't matter if this fixes it.

noisiver:
Yeah maybe but on Linux it *always* happens and always within the first 4 minutes. I'm adding this code now and will be testing. I'll let you know in a few minutes.

noisiver:
9 minutes uptime and counting, no crash yet. I'll leave it running for at least 30 minutes to make sure, because it did always crash within the few minutes before; 100% of the time.

noisiver:
32 minutes and no crash. It seems to be good now.

liyunfan1223:
Thanks! Fixed in https://github.com/liyunfan1223/mod-playerbots/pull/1034/commits/10e90cd7dbb8aca0f54de7a678bdfc8281ba80aa. I remove !unit check as it has been checked.

PR #1040, opened 2025-02-28
dynamically load database names for cleanup queries

icemansparks:
Recent changes in #1032 used fixed database names instead of referencing the ones from worldserver.conf - this PR should fix this by using variable database names.

@blinkysc please have a look in case you are already working on more improvements to the cleanup.

There may be ways to improve this / go a different way by replicating the dB name config in the playerbot config (let the user add the dBnames there as well) and loading it from the config file (This would reduce database operations).

### DIFF
--- a/src/RandomPlayerbotFactory.cpp
+++ b/src/RandomPlayerbotFactory.cpp
@@ -278,7 +278,7 @@ std::string const RandomPlayerbotFactory::CreateRandomBotName(NameRaceAndGender 
         {
             break;
         }
-        
+
         Field* fields = result->Fetch();
         botName = fields[0].Get<std::string>();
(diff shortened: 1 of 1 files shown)


noisiver:
Perhaps you could test this but doesn't `LoginDatabase.GetConnectionInfo()->database` and `CharacterDatabase.GetConnectionInfo()->database` get the names of the databases?

icemansparks:
> Perhaps you could properly test this but doesn't `LoginDatabase.GetConnectionInfo()->database` and `CharacterDatabase.GetConnectionInfo()->database` get the names of the databases?

Thanks! I didn't know that was available. I tried to use CharacterDatabase.GetDatabaseName() before, but found that it is private and cannot be accessed from the module side.

PR #1042, opened 2025-02-28
Fix/Feature: Bots can use travel/flight Shapeshift where appropriate

nl-saw:
image
image

**Improvements:**
- Druid bots will now use their Shapeshift forms (Travel if player/master is in travel shape, (Swift) Flight when master is in flight shape or on a flying mount)
- Minor improvements to the code in general
- Rearranged some code blocks (better logical order / readability)
- Files' build warnings resolved

### DIFF
--- a/src/strategy/actions/CheckMountStateAction.cpp
+++ b/src/strategy/actions/CheckMountStateAction.cpp
@@ -228,8 +228,11 @@ void CheckMountStateAction::Dismount()
     bot->GetSession()->HandleCancelMountAuraOpcode(emptyPacket);
 }
 
-bool CheckMountStateAction::TryForms(Player* /*master*/, int32 masterMountType, int32 masterSpeed) const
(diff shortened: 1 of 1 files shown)


noisiver:
For some reason if I go from travel form to a regular mount they won't mount unless they catch up to me, or rather unless they stop (usually by catching up to me). I don't know if that's a pre-existing issue though. The same for me going flight form while on a ground mount (and they are too) they remain on the ground mount instead of switching to a flying mount or flight form. Might be unrelated to this PR though but I might as well mention it.
{E345010A-7770-4DAC-9D1C-0C883E569833}

noisiver:
I must say it looks so amazing when they all sit down when I do, including in flight form. Beautiful. No other issues so far and I've been doing all I can think of right now over and over and over.

noisiver:
I couldn't find any issues. They removed form when I did and properly switched between travel and flight form when I did.

PR #1052, opened 2025-03-03
Fix server crash when bots apply strategies on a custom naxx raid

valsan-azerty-boi:
Related to issue https://github.com/liyunfan1223/mod-playerbots/issues/530

If you use _**mod-playerbots**_ with a module such as _**mod-individual-progression**_ (or other modules bringing a custom or vanilla version of Naxxramas into the game), applying Naxx WotLK strategies for bots on modified versions of Naxx with different scripts or bosses sometimes causes strange behavior with bots or even server crashes (especially on Heigan).

This PR is intended to prevent this kind of crash or strange behavior. You don't have to set ApplyInstanceStrategies = 0 anymore in conf file to prevent crash.

**Tested on Naxx Vanilla raid:** The server no longer crashes at certain strategic moments. The entire raid can be done with the bots without having to deactivate their strategy. 

**Tested on Naxx 10 (Wotlk):** No changes or side effects, the bots continue to behave as before without problem. PR does not negatively impact Naxx WotLK content with playerbots.


Dreathean:
This is very nice! Question, do they still use strats then or are they deactivated for the vanilla version?

noisiver:
This is a great update, however I think the SQL changes should be in a file in the updates folder rather than the base folder. It's to make sure the changes are always applied. Base files aren't (always) re-applied since the tables already exist.

valsan-azerty-boi:
> This is very nice! Question, do they still use strats then or are they deactivated for the vanilla version?
No, they often go tunnel vision/dps without applying the strategy (On grobulus, distant/heal try to move correctly but not the tank lol). I think someone will have to code vanilla naxx for bots strategies separately, but until it exists at least it doesn't crash the server anymore.

> This is a great update, however I think the SQL changes should be in a file in the updates folder rather than the base folder. It's to make sure the changes are always applied. Base files aren't (always) re-applied since the tables already exist.
I go look at that !

Dreathean:
Got it, yeah I figured that would be the case but either way it is great to prevent the crashes, thank you for this!

noisiver:
The vanilla strats are very similar to the wotlk ones for pretty much all bosses so it shouldn't be too hard to add those at least. It's great that it doesn't crash the server anymore!

valsan-azerty-boi:
> The vanilla strats are very similar to the wotlk ones for pretty much all bosses so it shouldn't be too hard to add those at least.

I'm too much of an average developer to do that right now lol. For now I need to learn a little more about how acore and playerbots work.

noisiver:
I'm not saying that you should or have to, just that when someone decides to do it it will at least not be a complete rewrite of the strategies. This alone is fantastic, no more crashes!

songjunst:
Although the server no longer crashes, there are no policies in the replica, and the robots are very chaotic.

valsan-azerty-boi:
> Although the server no longer crashes, there are no policies in the replica, and the robots are very chaotic.

Yes someone need to create new strategies for custom naxx/naxx vanilla

noisiver:
It's not that simple though. Seeing as the bots use the boss scripts you would have to use a different set of scripts and include those as headers just like the original strategies. I suppose you could rewrite the strategies and not use the boss scripts but I don't know if that's at all possible which is why it was done in the first place.

PR #1055, opened 2025-03-04
Rogue bots can unlock items in their bags and in the trade window

avirar:
1. Added handling for SMSG_TRADE_STATUS_EXTENDED packets to trigger when items are placed into the Will Not Be Traded slot of the trade window. Only contains logic for lockboxes at the moment, but can be expanded. Rogues with Pick Lock will automatically perform the "unlocked traded items" action when a locked box is in the correct slot.
2. CastSpell(uint32 spellId, Unit* target, Item* itemTarget)
   - Handle SPELL_EFFECT_OPEN_LOCK spells explicitly since it can be cast on Objects, Items, and Trade Window Items
   - During SPELL_EFFECT_OPEN_LOCK block: Assign the correct target using SetItemTarget and SetTradeItemTarget
   - Added comprehensive output for CastSpell failure when the bot has the "debug spell" non-combat strategy
4. Added chat commands: 
   - "unlock items" the bot will attempt to unlock any items in their bags
   - "unlock traded items" will attempt to unlock any item in the Will Not Be Traded slot
6. Allow trade with all group members
7. Perform unlock and open item actions when "item push result" is triggered so bots automatically open and unlock looted items when possible.

Known issue: Trade window must be closed when performing "unlock items" on items in the bots inventory. This is due to Pick Lock being intercepted in CastSpell and having the ItemTarget assigned based on the trade window being present. Simply close the trade window when asking the bot to unlock items in their own bags to prevent the issue. I'll have a look at resolving that tomorrow.

Final thoughts: Control auto-open/unlock with config settings or strategies?

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -3179,6 +3179,12 @@ bool PlayerbotAI::CastSpell(uint32 spellId, Unit* target, Item* itemTarget)
         else if (itemTarget)
         {
             targets.SetItemTarget(itemTarget);
+            Player* trader = bot->GetTrader();
+            if (trader)
+            {
+                targets.SetUnitTarget(trader);
+                faceTo = trader;
+            }
         }
         else
         {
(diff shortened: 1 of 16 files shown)


Dreathean:
This is a really cool addition, thank you Avirar!

Curious what others think, but I personally don't feel that the auto-unlock feature needs to be added to the config as most users are small scale private servers either solo or family/friends, and for the few large scale servers utilizing playerbots they can make their own adjustments if they don't want this feature enabled.

avirar:
Sorry it took a while to get back to this, I've created helpers in PlayerbotAI.cpp and refactored the actions.

PR #1070, opened 2025-03-10
add ability to add or remove character with same name as account

xSparky911x:
This fixes an issue with adding bot characters with the same name as an account name. the new commands are as follows:

`.playerbots bot add character <character name>` or
`.playerbots bot add account <account name>`

To remove, the commands are similar:

`.playerbots bot remove character <character name>` or
`.playerbots bot remove account <account name>` or
`.playerbots bot remove *` (to remove all bots in group)

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -816,17 +816,34 @@ std::vector<std::string> PlayerbotHolder::HandlePlayerbotCommand(char const* arg
 
     if (!*args)
     {
-        messages.push_back("usage: list/reload/tweak/self or add/init/remove PLAYERNAME\n");
-        messages.push_back("usage: addclass CLASSNAME");
+        messages.push_back("usage: bot add character/account NAME or bot remove character/account NAME");
+        messages.push_back("usage: bot addclass CLASSNAME");
+        messages.push_back("usage: list/reload/tweak/self");
         return messages;
     }
 
     char* cmd = strtok((char*)args, " ");
-    char* charname = strtok(nullptr, " ");
     if (!cmd)
     {
-        messages.push_back("usage: list/reload/tweak/self or add/init/remove PLAYERNAME or addclass CLASSNAME");
+        messages.push_back("usage: bot add character/account NAME or bot remove character/account NAME");
+        messages.push_back("usage: bot addclass CLASSNAME");
+        messages.push_back("usage: list/reload/tweak/self");
         return messages;
+    }
+
+    // Handle the "bot" command prefix if it exists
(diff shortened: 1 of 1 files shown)


xSparky911x:
Ok this seems to be working properly and WAY less things changed. Forgive me I am no where near a coder (but can grasp some things) and used the assistance of AI but didn't babysit enough before submitting last pull request. (learning as I go). But this tested well in my test server, no issues compiling.

Dreathean:
Wouldn't impact me either way since it boils down to a one-click macro to summon my account(s) of altbots regardless, but ultimately given the number of issues and complaints we've seen on here and discord with people having the issue of account=character name I see the usefulness of splitting the two.

PaoloBenigni:
Would it be possible to maintain the original functionality while adding the more specific?

xSparky911x:
I guess we could go with something like 
`.playerbots bot add <name>` 
and then if they want to do the whole account have 
`.playerbots bot add account <account name>`

Or actuallt maybe something like 
`.playerbots bot addacount <account name>`

That way the original functionality still stays the same (except the issue of same account name as character name) and the command structure doesn't grow any longer like nl-saw pointed out earlier

xSparky911x:
Ok so this fix made it so that adding a character the traditional way does not add an entire account. If an account is wanted there is a different command to accomplish that. The way I have it setup now is
 `.playerbots bot add <character name>` (so the original command)
`.playerbots bot addaccount <account name>` (this way its just as short of a command as before)

`.playerbots bot remove *` will still remove all bots in the party as well

xSparky911x:
never tried `.playerbots bot add *`. Wasn't aware that was a thing but can go test it. The point I was making with it "just as short" was it wasn't more arguments for example `.playerbots bot add account <account name>` vs `.playerbots bot addaccount <account name>` Still 3 arguments as before (just a longer word is all)

xSparky911x:
Even before my fix here `.playerbots bot add *` does not work. Tried both versions, My fix (on my test server) and without my fix (on my live server)

noisiver:
It works fine, I use it **all** the time. I never leave the group so it's the only command I use to bring my bots online.

xSparky911x:
Oh ok. I was out of a group. let me try getting them all online first, then put them offline then try the command without leaving the group

PaoloBenigni:
For me it works both in group or out of group

xSparky911x:
`.playerbots bot add *` works with my fix being in the group. As I said I could not get it working for me in either version being out of the group theough

liyunfan1223:
Tested and works fine to me.

PR #1098, opened 2025-03-20
ICC - Implement Val'kyr grabbing bots and dropping

mostlynick3:
Adjusts the ICC scripting to permit bots to be grabbed by valkyrs, and if thrown over the edge, drop to their death

### DIFF
--- a/src/strategy/raids/icecrown/RaidIccActions.cpp
+++ b/src/strategy/raids/icecrown/RaidIccActions.cpp
@@ -3206,8 +3206,35 @@ bool IccLichKingWinterAction::Execute(Event event)
 
 bool IccLichKingAddsAction::Execute(Event event)
 {
+    if (bot->HasAura(68985) || !bot->IsAlive()) // Don't process actions if bot is picked up by Val'kyr or is dead
+        return false;
+    
     Unit* boss = AI_VALUE2(Unit*, "find target", "the lich king");
     Unit* spiritWarden = AI_VALUE2(Unit*, "find target", "spirit warden");
+    
+	if (bot->HasAura(30440)) // Random aura tracking whether bot has fallen off edge / been thrown by Val'kyr
+	{
+		if(bot->GetPositionZ() > 779.0f)
+			return JumpTo(bot->GetMapId(), bot->GetPositionX(), bot->GetPositionY(), 740.01f);
+		else
+		{
+			bot->KillPlayer(); // If bot has jumped past the kill Z (780), kill
+			return true;
+		}
+	}
+    
+    if (boss && (boss->GetHealthPct() < 69) && (boss->GetHealthPct() > 39)) // If boss is in p2, check if bot has been thrown off platform
+    {
+       float dx = bot->GetPositionX() - 503.0f;
+       float dy = bot->GetPositionY() - (-2124.0f);
+       float distance = sqrt(dx*dx + dy*dy); // Calculate distance from the center of the platform
+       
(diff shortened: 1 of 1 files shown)


src/strategy/raids/icecrown/RaidIccActions.cpp
```diff
@@ -3206,8 +3206,35 @@ bool IccLichKingWinterAction::Execute(Event event)
...
+		if(bot->GetPositionZ() > 779.0f)
+			return JumpTo(bot->GetMapId(), bot->GetPositionX(), bot->GetPositionY(), 740.01f);
+		else
+		{
+			bot->KillPlayer(); // If bot has jumped past the kill Z (780), kill
+			return true;
+		}
+	}
+    
+    if (boss && (boss->GetHealthPct() < 69) && (boss->GetHealthPct() > 39)) // If boss is in p2, check if bot has been thrown off platform
```
nl-saw: Small note: The parentheses around the conditions are technically unnecessary in this case because the logical operators (&& and < >) have a clear precedence:

```if (boss && boss->GetHealthPct() < 69 && boss->GetHealthPct() > 39)```

mostlynick3: Changed, thanks!

Noscopezz:
Hello, glad to see you posting PR, howerver upon testing your changes, they broke the winter phase and made the bots go under the ice platform and stay frozen in place.
    On another note I dont think we should force bots to jump to their death, if it can be detected that they were indeed grabbed by valkyr we should only bypass the bypass that makes them teleport up on z axis. If this bypass bothers you, you can simply comment out bypass 3212-3215 in code for your server.

//temp solution for bots going underground due to buggy ice platfroms and adds that go underground
    if (abs(bot->GetPositionZ() - 840.857f) > 1.0f)
        return bot->TeleportTo(bot->GetMapId(), bot->GetPositionX(),
                          bot->GetPositionY(), 840.857f, bot->GetOrientation());

    Any bot that is outside walkable platforms will simply fall and die if there is no bypass active, I have made that bypass becouse the whole LK platform was a mess for bots to navigate around and they were falling down. In return that caused them to "cheat" valkyrs.
    I think its best to leave valkyrs as is until we figure out how to fix that issue without breaking other things.

Noscopezz:
Upon further testing, by simply disabling bypass in original code valkyrs will work as they should normaly. 

I would guess that new pathing in core and LoS checks before moving fixed the problem with bots falling thru floor on lk map. I will post PR with bypass commented out, so that if someone prefer the way that it is now they can simply remove comments and play as it was.

Noscopezz:
It was update prior to that, in my testing mobs and bots didnt go underground.

For me freeze happened before they even reached platforms in winter phase.

Lets wait for more people to test this, it could be a problem on my end only.

mostlynick3:
I managed to replicate the issue shortly after my last comments here. Not sure how I managed to get it right... But testing with a full group led to most of my group dropping to the abyss. I've fool proofed it some more now to ensure they don't jump to their deaths while remorseless winter is being cast, and perhaps most notably, I've made the jump only start if bots have a Z > 844 (as valks drop players as Z = 845).

It may work fine with just removing the Z 840 teleport, but I'm keeping this updated just in case it comes in handy at a later point :)

Noscopezz:
Awesome, thats even better.

Noscopezz:
After testing latest PR everything is working as it should. This is safe to merge. Thank you for your contribution and time.

PR #1106, opened 2025-03-21
Fix frost nova crash

liyunfan1223:
https://github.com/liyunfan1223/mod-playerbots/issues/1105


Regrad:
I add it to the server, we are waiting for a day, and it is being tested. While I sent another bug there. Thank you for your work!

PR #1107, opened 2025-03-21
Test potential solution for reported crash on lfgteleport

nl-saw:


Regrad:
I made changes to the server, and testing is underway.

PR #1109, opened 2025-03-22
[New RPG] Make a full check before questgiver choose such as reputation

liyunfan1223:
https://github.com/liyunfan1223/mod-playerbots/issues/1094


blinkysc:
Adding to my server now to run 24/7

liyunfan1223:
This PR is to solve the issue that bots gather to some quest givers with quests that cannot be accepted. Some rewarded quest checks are good, but may not be directly relevant.

blinkysc:
Been running since checkout, seems to be good

PR #1114, opened 2025-03-24
Fix PlayerbotAI Crashes in Trigger Processing

blinkysc:
# Fix Crash in Engine::ProcessTriggers

## Changes
- Added null pointer checks throughout the trigger processing
- Added bounds checking when accessing trigger collection
- Added exception handling for trigger check operations
- Improved validation before accessing trigger properties

## Accomplishments
- Prevents crashes when accessing invalid pointers in trigger nodes
- Handles cases where triggers might be modified during processing
- Gracefully continues execution when trigger checks fail
- Ensures performance monitoring is properly cleaned up even during errors

This fix addresses the specific crashes observed in `Engine::ProcessTriggers` by validating all object access before use, making the code more resilient to unexpected states.

### DIFF
--- a/src/strategy/Engine.cpp
+++ b/src/strategy/Engine.cpp
@@ -432,59 +432,122 @@ bool Engine::HasStrategy(std::string const name) { return strategies.find(name) 
 void Engine::ProcessTriggers(bool minimal)
 {
     std::unordered_map<Trigger*, Event> fires;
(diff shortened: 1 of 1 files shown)


notOrrytrout:
Does this work? Seems like a small change..

notOrrytrout:
Nvm, I see it has been marked draft.

PR #1115, opened 2025-03-24
Chat: Fix index out of bounds crash in case verb_pos equals zero

EricksOliveira:
Prevents invalid accesses to the word[] array, preventing possible crashes.

Improves the message selection logic in bot chat.

Fixes an error in the switch-case structure, ensuring that all possible messages are used correctly.


liyunfan1223:
Nice catch. But even after the fix this code still looks very fragile.

PR #1118, opened 2025-03-24
feat(concurrency): Add thread-safe playerbot update tracking

blinkysc:
# Add Thread-Safe Playerbot Updates

## Summary
Implemented thread-safe concurrent processing for playerbot updates to prevent race conditions when multiple threads update the same player bots simultaneously.

## Changes
- Added `PlayerbotsUpdateTracker` class with atomic update flags
- Modified player update logic to use atomic flag acquisition
- Added non-blocking skip mechanism when a player is already being updated by another thread

## Impact
This change eliminates potential crashes and data corruption in high-concurrency environments while maintaining performance by using non-blocking operations.

## NOTE: FOR TESTING ONLY
This implementation is currently intended for testing purposes and should not be merged to until thoroughly validated.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -30,6 +30,10 @@
 #include "cs_playerbots.h"
 #include "cmath"
 
+// Initialize static members for PlayerbotsUpdateTracker
+std::unordered_map<uint64, std::atomic<bool>> PlayerbotsUpdateTracker::s_playerUpdateFlags;
+std::mutex PlayerbotsUpdateTracker::s_mapMutex;
+
 class PlayerbotsDatabaseScript : public DatabaseScript
 {
 public:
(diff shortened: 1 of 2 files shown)


Regrad:
Is this PR compatible with https://github.com/liyunfan1223/mod-playerbots/pull/1113, or is it an alternative implementation?

walkline:
Before reviewing the implementation, did I understand correctly that you assumed a single player can be updated multiple times and on different threads during the world update tick?

blinkysc:
> Before reviewing the implementation, did I understand correctly that you assumed a single player can be updated multiple times and on different threads during the world update tick?

A single playerbot update it seems so, it seems there is a race condition in how the playerbot AI works within the map update threads. Seems every crash comes from 

```c++
// Originally line 130
if (PlayerbotAI* botAI = GET_PLAYERBOT_AI(player))
{
    botAI->UpdateAI(diff);
}
```

gutoukuaiken:
Remember me? I'm the one whose hard drive broke down. Haha, I replaced the hard drive and increased the memory capacity, but the server still crashed. Although it wasn't as severe as when the hard drive broke down, it still crashed frequently. Although the reasons were different each time, they were all related to the data obtained by playerbotAI, such as map data, aura data, and some strange errors. I have changed the code and will see how it works. Here are my recent logs.
[gdb.txt](https://github.com/user-attachments/files/19450795/gdb.txt)

Regrad:
With both of these fixes (this and https://github.com/liyunfan1223/mod-playerbots/pull/1113), the server is much more stable.

walkline:
> > Before reviewing the implementation, did I understand correctly that you assumed a single player can be updated multiple times and on different threads during the world update tick?
> 
> A single playerbot update it seems so in map threads, it seems there is a race condition in how the playerbot AI works within the map update threads. Seems every crash (ACCESS VIOLATION) comes from
> 
> ```c++
> void OnPlayerAfterUpdate(Player* player, uint32 diff) override
> {
>      if (PlayerbotAI* botAI = GET_PLAYERBOT_AI(player))
>      {
>          botAI->UpdateAI(diff);
>      }
>     
>      if (PlayerbotMgr* playerbotMgr = GET_PLAYERBOT_MGR(player))
>      {
>          playerbotMgr->UpdateAI(diff);
>      }
> }
> ```
> 
> EDIT: Read your discord too =)

Most likely the issue is in PlayerbotsMgr. In some cases, we are calling `PlayerbotsMgr::AddPlayerbotData` from Map thread https://github.com/liyunfan1223/mod-playerbots/blob/feb7bb8a8b9aa98165f825ea330883f9cb936f9e/src/PlayerbotMgr.cpp#L121-L123.
Since multiple map threads may exist, this results in concurrent writes without synchronization mechanisms, along with concurrent read/write issues.

I would move the playerbot login code to the world thread. Also, ensure that the removal code is executed from the world thread as well.

Regrad:
There is a problem. The server now sometimes freezes, and crashes occur without crash reports.

blinkysc:
> Since multiple map threads may exist, this results in concurrent writes without synchronization mechanisms, along with concurrent read/write issues.
> I would move the playerbot login code to the world thread. Also, ensure that the removal code is executed from the world thread as well.

Awesome thank you @walkline! I'll see what I can do; no promises lol

gutoukuaiken:
> There is a problem. The server now sometimes freezes, and crashes occur without crash reports.

It's hard to say for sure. One possibility is that the code still needs optimization. Another possibility is that it's a hardware issue. Your CPU has a large number of threads but the frequency is too low. This could be the cause, as I have experienced the same thing before. I replaced the hardware with better ones and the frequency at which the server froze decreased significantly.

hermensbas:
With all respect but is addind mutex locks the way to go? Aint there any local threads usage or thread process queueing?

Issue #1121, opened 2025-03-25
Setting to disable bot purchases


Regrad:
**Is your feature request related to a problem? Please describe.**
Yes. Players use trading to make endless deposits. I described it.

**Describe the solution you'd like**
Please make an option that allows bots to no longer buy anything from the player.

**Describe alternatives you've considered**
Nope

**Additional context**
Nope

PaoloBenigni:
Isn't this an open issue already?

Regrad:
> Isn't this an open issue already?

No. This is a suggestion to add functionality, not a problem.

PaoloBenigni:
Duplicate of https://github.com/liyunfan1223/mod-playerbots/issues/1089#issue-2927773506

Regrad:
> Duplicate of [#1089 (comment)](https://github.com/liyunfan1223/mod-playerbots/issues/1089#issue-2927773506)

There is a bug report, and here is a request to add functionality. Fixing a bug does not imply adding functionality, and adding functionality does not imply fixing a bug. They are not the same thing.

Dreathean:
I would say they are pretty distinct enough. For the original issue, the issue is that bots will buy an item for more than they sell back to you, resulting in an easy loop to take all of a bot's gold. Fixing that is separate from disabling bots buying from you altogether.

Dreathean:
Should be added now as of https://github.com/liyunfan1223/mod-playerbots/pull/1132, if any issues arise related to this config please open a new issue.

Regrad:
Unfortunately, the RND-called bots continue to trade.

PR #1132, opened 2025-03-27
Add config option for RND bot trading.

PaoloBenigni:
Added config option to disable trading.
Related to issue. #1121

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -544,6 +544,8 @@ AiPlayerbot.DisableDeathKnightLogin = 0
 # Default: 0
 AiPlayerbot.LimitTalentsExpansion = 0
 
+# Allow random bots to trade
+AiPlayerbot.EnableRandomBotTrading= 1
 #
 #
 #
--- a/src/strategy/actions/TradeStatusAction.cpp
+++ b/src/strategy/actions/TradeStatusAction.cpp
@@ -32,6 +32,12 @@ bool TradeStatusAction::Execute(Event event)
         return false;
     }
 
+    if (!sPlayerbotAIConfig->enableRandomBotTrading && sRandomPlayerbotMgr->IsRandomBot(bot))
+    {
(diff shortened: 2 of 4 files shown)


conf/playerbots.conf.dist
```diff
@@ -544,6 +544,8 @@ AiPlayerbot.DisableDeathKnightLogin = 0
 # Default: 0
 AiPlayerbot.LimitTalentsExpansion = 0
 
+# Allow random bots to trade
+AiPlayerbot.EnableRandomBotTrading= 1
```
nl-saw: Space please

Dreathean:
I personally don't think it should be necessary for altbots, you'll likely always need to be able to trade with your altbots for various reasons and currently there isn't significant room for gold exploits with them to my knowledge. 

I guess people could set x amount to grinding but that's just something for large server owners to deal with imo. Even if you disabled trading, players would be able to simply log onto those altbots and mail if it really came down to it.

src/strategy/actions/TradeStatusAction.cpp
```diff
@@ -32,6 +32,12 @@ bool TradeStatusAction::Execute(Event event)
         return false;
     }
 
+    if (!sPlayerbotAIConfig->enableRandomBotTrading && sRandomPlayerbotMgr->IsRandomBot(bot))
+    {
+        bot->Whisper("Trading is distabled", LANG_UNIVERSAL, trader);
```
DustinHendrickson: Little typo hiding here "distabled"

Regrad:
Unfortunately, the RND-called bots continue to trade.

PaoloBenigni:
You set the config parameter to 0?

Regrad:
> You set the config parameter to 0?

Sure

Regrad:
I also opened another request for functionality, explaining my previous request a little better. Thank you for your hard work!

PaoloBenigni:
When you say summoned ones continue to do so, what exactly do you mean? 
A random bot that is in your group will not trade with you?

PR #1135, opened 2025-03-27
This update adds null pointer checks and extra validations to the Att…

EricksOliveira:
…ackAnythingAction::isUseful() method, preventing invalid accesses that could cause crashes.

Checking bots and botAI before method calls.

Validating targets and confirming if they are still in the world (IsInWorld()).

Adding debug logs (LOG_DEBUG) to make it easier to identify issues.


hermensbas:
ofc, didnt notice the review pending ❤️

PR #1140, opened 2025-03-28
Set Lockpicking Skill

nl-saw:
On init and on maintenance


noisiver:
Should this also be added to the trainer action? That's what the maintenance command uses iirc. Like in https://github.com/liyunfan1223/mod-playerbots/blob/917c96a28161dc357809f82e0849f820fe9a6e4d/src/strategy/actions/TrainerAction.cpp#L158

noisiver:
Yes but if you use maintenance on altbots (for example) the lockpicking skill will not be covered by it. I think it should be included in that, but that's just my opinion and someone else should weigh in.

PR #1142, opened 2025-03-29
Crash fix: Bots inviting teleported players.

EricksOliveira:


src/strategy/actions/InviteToGroupAction.cpp
```diff
@@ -53,6 +53,9 @@ bool InviteNearbyToGroupAction::Execute(Event event)
         if (!player)
             continue;
 
+        if (!player->GetMapId() != bot->GetMapId())
```
BoiseComputer, line 56: **if (player->GetMapId() != bot->GetMapId()) continue;**
This condition checks if the player's map ID is not equal to the bot's map ID.
If they are on different maps, it will skip the current iteration and continue to the next player.

**if (player->GetMapId() == bot->GetMapId()) continue;**
This condition checks if the player's map ID is equal to the bot's map ID.
If they are on the same map, it will skip the current iteration and continue to the next player.

BoiseComputer: Which one are we going for here?

PR #1143, opened 2025-03-29
Fix multithread issue on LFG and group leave

liyunfan1223:
Replace direct thread-unsafe funtion calls with simulated client packets.

These thread-unsafe funcs will be executed serially in WorldSessionMgr::UpdateSession instead of multi-threaded execution in Map::Update.

Some other thread unsafe crashes can also be solved in this way. The identifiers in Opcodes.cpp can help distinguish which methods are thread-safe or thread-unsafe (PROCESS_THREADSAFE or PROCESS_THREADUNSAFE)

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -374,7 +374,9 @@ void PlayerbotAI::UpdateAIGroupMembership()
             PlayerbotAI* leaderAI = GET_PLAYERBOT_AI(leader);
             if (leaderAI && !leaderAI->IsRealPlayer())
             {
-                bot->RemoveFromGroup();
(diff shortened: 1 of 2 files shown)


blinkysc:
Checking out now

blinkysc:
Seems to be working fine for me... Though I only have me and my kids lol.

hermensbas:
the actions list, hard reference:
https://github.com/azerothcore/azerothcore-wotlk/blob/master/src/server/game/Server/Protocol/Opcodes.cpp

serverwar:
Server was online for 9 hours and when it went down there were no blank logs, much more stable

PR #1171, opened 2025-04-06
Travel: Fixed crash when trying to travel in maps with no nodes

EricksOliveira:


liyunfan1223:
Some safe checks would be nice anyway, but as far as I know this part code should not be executed for now.

hermensbas:
Could not hurt, even if not used at this point

PR #1173, opened 2025-04-06
Attempt to fix crash on _checkcontact

liyunfan1223:
Try to avoid crash on `_checkcontact` by solving race condition on login.

There's no guarantee it works as I can't reproduce the crash. Some testing is necessary.


serverwar:
I will apply it and if it solves it I will return with ok, chat there that reverts the error reduction by 95% but it still happens

serverwar:
So far zero crash, more than 12 hours without crash checkcontact

serverwar:
Almost 24 hours without any crashes.

PR #1182, opened 2025-04-08
Feat/onyxia raid strategy init

jl178:
## Add Playerbot Strategy Support for Onyxia's Lair

Hi everyone - This is my first PR & naturally I haven't created a strategy before. Criticism and feedback welcome.

### Summary

This PR introduces support for a new **strategy for Onyxia's Lair**. Bots can handle **some** encounter mechanics including:
1. breath avoidance
2. fireball splash spreading
3. tail swipe positioning

#### Why?
While running Onyxia with a full level 60 party, the raid was consistently wiping due to fireballs and standing in incorrect locations (in breath when ony is flying and not to the sides of boss as healer/dps)

---

#### Strategy Framework
- Added `RaidOnyxiaStrategy` implementation with relevant triggers and actions.
- Registered `onyxia` in `RaidStrategyContext`, allowing strategy loading based on Onyxia map id (249).

#### Behavior Changes
- **Actions Added**:
  - `ony move to side`: Avoids Onyxia’s front and tail cone effects.
  - `ony move to safe zone`: Reacts to breath casting by repositioning.
  - `ony spread out`: Prevents splash damage if targeted by Fireball.
  
- **Triggers Added**:
  - `ony near tail`: Activated when the bot is behind/in front of Onyxia and isn't a tank.
  - `ony deep breath warning`: Detects breath casts using specific spell IDs and moves to sides.
  - `ony fireball splash incoming`: Triggered by casting of Fireball while near others.

#### Context Integration
- `RaidOnyxiaActionContext` and `RaidOnyxiaTriggerContext` registered in:
  - `AiObjectContext.cpp`
  - `RaidStrategyContext.h`

### Extra Details

- Breath spell detection uses hardcoded classic spell IDs (e.g., `18351`, `18576`, etc.). Validated from Azerothcore scripts.
- Side movement and safe zone logic calculated using boss orientation and positioning.
- Fireball target detection ensures only the bot receiving the spell reacts.

### Testing

- Bots correctly yell and reposition based on Onyxia mechanics [Removed in PR - Used for Debug].  
- Strategy auto-loads when entering Onyxia's Lair.

### DIFF
--- a/src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
+++ b/src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
@@ -1,0 +1,22 @@
+
+// RaidOnyxiaStrategy.h
+#ifndef _PLAYERBOT_RAIDONYXIASTRATEGY_H_
+#define _PLAYERBOT_RAIDONYXIASTRATEGY_H_
+
+#include "Strategy.h"
+
+class RaidOnyxiaStrategy : public Strategy
+{
+public:
+    RaidOnyxiaStrategy(PlayerbotAI* ai) : Strategy(ai) {}
+
+    const std::string& getName() const;
+
+    void InitTriggers(std::vector<TriggerNode*>& triggers) override;
+    void InitMultipliers(std::vector<Multiplier*>& multipliers) override;
+
+private:
+    static const std::string name;
+};
+
+#endif
--- a/src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
+++ b/src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
@@ -10,7 +10,7 @@ class RaidOnyxiaStrategy : public Strategy
 public:
     RaidOnyxiaStrategy(PlayerbotAI* ai) : Strategy(ai) {}
 
-    std::string const getName() override { return "raid onyxia"; }
+    std::string const getName() override { return "onyxia"; }
 
     void InitTriggers(std::vector<TriggerNode*>& triggers) override;
     void InitMultipliers(std::vector<Multiplier*>& multipliers) override;
--- a/src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
+++ b/src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
@@ -10,13 +10,10 @@ class RaidOnyxiaStrategy : public Strategy
 public:
     RaidOnyxiaStrategy(PlayerbotAI* ai) : Strategy(ai) {}
 
-    const std::string& getName() const;
+    std::string const getName() override { return "raid onyxia"; }
 
     void InitTriggers(std::vector<TriggerNode*>& triggers) override;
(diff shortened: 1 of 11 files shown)


src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
```diff
@@ -0,0 +1,19 @@
...
+#define _PLAYERBOT_RAIDONYXIASTRATEGY_H_
+
+#include "Strategy.h"
+
+class RaidOnyxiaStrategy : public Strategy
+{
+public:
+    RaidOnyxiaStrategy(PlayerbotAI* ai) : Strategy(ai) {}
+
+    std::string const getName() override { return "raid onyxia"; }
```
Celandriel: I think this would be better just as `onyxia`

PaoloBenigni:
Tested 2x on the lv 60 version using Individual Progression. Works well!!

src/strategy/raids/onyxia/RaidOnyxiaStrategy.h
jl178: Fixed

valsan-azerty-boi:
excellent !!

liyunfan1223:
Tested on 25-man level 80, generally it works well without any crash and boss spells seem to be recognized correctly.

Just recording some issues: some manual commands (e.g. attack, summon and disperse) are still needed to take down the boss. The strategy does not ensure that the bots survive by breathing, and they sometimes move into the air in response to breathing.

jl178:
> Tested on 25-man level 80, generally it works well without any crash and boss spells seem to be recognized correctly.
> 
> Just recording some issues: some manual commands (e.g. attack, summon and disperse) are still needed to take down the boss. The strategy does not ensure that the bots survive by breathing, and bots sometimes move into the air in response to breathing. It would be better if adds in the flight phase could be attacked with priority.

Thank you for the feedback and great catches - I see the issue already with the breath mechanic and I will work on both these recommendations in the next week! Thanks!

jl178:
@liyunfan1223 - I have updated the PR with the following changes:

1. Took existing spell IDs for breath and mapped to locations on the Ony map, which bots will move to depending on the spell being cast. I took this route as the simple approach to avoid breaths based on the image below. Bots no longer fly in the air, as they are not taking Ony's Z position incorrectly
image
2. I have added an avoid mechanic to the whelp egg locations, in an attempt to prevent bots from lingering by eggs and spawning more whelps. Again, this is hardcoded for simplicity - LMK if you think it should be dynamic somehow
3. I have added a new Action/Trigger for whelp spawns - When Ony takes to the air, bots will focus on the whelps. However, I have noticed they sometimes get stuck, constantly switching between Ony and the Whelps. I would greatly appreciate if you could take a look at the logic & make sure I am not missing anything (I tried to mimic existing encounter logic, so I am not sure where it is going wrong).

liyunfan1223:
The breathing mechanism has been significantly improved and 25-man onyxia is doable without manual commands.

PR #1190, opened 2025-04-10
Loot packet fix

liyunfan1223:
close https://github.com/liyunfan1223/mod-playerbots/issues/1187

### DIFF
--- a/src/strategy/actions/LootAction.cpp
+++ b/src/strategy/actions/LootAction.cpp
@@ -29,9 +29,10 @@ bool LootAction::Execute(Event event)
 
     if (!prevLoot.IsEmpty() && prevLoot.guid != lootObject.guid)
     {
-        WorldPacket packet(CMSG_LOOT_RELEASE, 8);
-        packet << prevLoot.guid;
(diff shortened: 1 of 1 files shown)


hermensbas:
Should all the bypasseses of queueing (bot->GetSession()->Handle...(package) replaced with queueing? or whats the general idea?

liyunfan1223:
Not all of them have to be replaced, only the handle functions that are not thread-safe, typical the LFG related functions. At the same time, since the thread model of azerothcore ensures that the same map is single-threaded, this should reduce the number of functions that need to be replaced.

For now I replace them mainly based on the crash log. A major unresolved issue is multi-threaded bots logout, which may not be simply replaced by queuing packages.

hermensbas:
i am going through the Regard and Serverwar crash logs aswell. The logout itself prolly not issue but code that has some defense checks which isnt sufficient to have 100% certainty the player/bot is still there. 

So i can imagine some can be fixed by going through world thread queue instead map thread, some mutex im edge cases or even error handling?

PR #1194, opened 2025-04-11
InviteToGroupAction: minor correction

hermensbas:
Do not invite if that nearest player is not in the same map. 

Current/old code basically translated into:
```
if ((player->GetMapId() == 0 ? true : false) != bot->GetMapId())
    continue;
```


hermensbas:
didnt notice #1154 (closed)

PR #1216, opened 2025-04-19
ICC Sindragosa and LK Improve

Noscopezz:
Sindragosa

Improved tomb los for frost bombs. Bots will now mark tomb with moon rti to avoid killing it too soon. (If they still attack the tomb, simply reset and after bombs are over, simply attack the tomb to kill or mark with a skull.)

Main Tank will reset mystic buffet stacks now to avoid tank swapping since it's really unreliable which will make sindy less frustrating to kill (other bots will need to do mechanics properly by hiding behind ice tombs).

LK

Necrotic plague action

Bots will now check before moving to the Shambling if they are in front of the Shambling to avoid getting one shoted by shockwave.
Improved multiplier, which will now properly handle dispelling of necrotic plague.

Winter phase action

Bots will properly move behind adds to avoid getting one shoted by shamblings or raging spirits.
If there is a hunter in the group, hunter will focus on spheres, if not, any ranged DPS will focus spheres.

Adds action

Bots will now be in proper positions during the 1st phase, and when the off tank is done collecting shamblings, it will move away from the group.

Bots (non-tanks) will move behind the shambling if they are in front of it to avoid getting one shoted in the 1st phase.

Restored action for checking if bots are at ground level, since they were still sometimes glitching through the floor (valkrys will be able to carry bots and drop them off the edge if not killed in time).

Improved defile calculations for 10 man and 25 man raids since defile was growing differently for 10 and 25, which would make bots in 25 man move very far away from somewhat small defile puddle.

Bots will now properly CC any valkyr that wasn't CC'd, and they won't freak out anymore if multiple valkyrs are present. (They can't tell which valkyr is holding bots/ players, so we should command an attack on valkyrs that are holding someone.)

Bots will now run away from Vile Spirits if they are targeted by them during the last phase.


src/strategy/raids/icecrown/RaidIccTriggers.cpp
```diff
@@ -851,6 +851,10 @@ bool IccSindragosaMainTankMysticBuffetTrigger::IsActive()
     if (!boss)
         return false;
 
+    Aura* aura = botAI->GetAura("mystic buffet", bot, false, false);
```
blinkysc, line 854: Maybe after merge we open issue saying there is currently cheat for mystic buffets for tanks?

Noscopezz: Thats a good idea. I will open issue with current cheats/bypasses/bugs and ways to fix them if possible with commands.

PR #1260, opened 2025-05-03
Prevents crash when logging in bots whose master disconnected before login

EricksOliveira:
This PR adds a check to the HandlePlayerBotLoginCallback function to ensure that the master player is still online before completing the bot login. If the master is no longer present, the bot is safely logged out, preventing invalid pointer access and related crashes. This change resolves a common race condition where the master disconnects shortly after adding a bot to the group.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -98,24 +98,20 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
     Player* masterPlayer = masterSession ? masterSession->GetPlayer() : nullptr;
 
     bool isRndbot = !masterAccountId;
-
-    std::ostringstream out;
     bool allowed = true;
-    time_t currentTime = time(nullptr);
+	std::ostringstream out;
 
-    if (!isRndbot)
-    {
-        time_t& lastAction = lastBotActionTime[masterAccountId];
-        if (lastAction && (currentTime - lastAction) < 5)
-        {
-            allowed = false;
-            out << "Wait " << (5 - (currentTime - lastAction)) << "s to call the next bot";
-        }
-        else
-        {
-            lastAction = currentTime;
-        }
-    }
+	if(!isRndbot)
+	{
+		time_t CurrentTime = time(nullptr);
+		if (DelayBot && (CurrentTime - DelayBot) <= 10)
+		{
+			allowed = false;
+			out << "Wait 10s to call the next bot";
+		}
+
+		DelayBot = CurrentTime;
+	}
 
     bool sameAccount = sPlayerbotAIConfig->allowAccountBots && accountId == masterAccountId;
     Guild* guild = masterPlayer ? sGuildMgr->GetGuildById(masterPlayer->GetGuildId()) : nullptr;
@@ -153,13 +149,6 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
         }
         return;
     }
-
-    // 5 second artificial delay for manual bots (not rndbot)
-    if (!isRndbot)
-    {
-        std::this_thread::sleep_for(std::chrono::seconds(5));
-    }
-
     std::shared_ptr<PlayerbotLoginQueryHolder> holder =
         std::make_shared<PlayerbotLoginQueryHolder>(this, masterAccountId, accountId, playerGuid);
     if (!holder->Initialize())
@@ -318,28 +307,28 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid, uint32 masterAccountId)
 {
     if (Player* bot = GetPlayerBot(guid))
     {
-        time_t currentTime = time(nullptr);
+        //Checks if the bot is in a group and removes it to avoid a possible crash
+		if (ObjectGuid groupId = sCharacterCache->GetCharacterGroupGuidByGuid(guid))
+		{
+			if (Group* group = sGroupMgr->GetGroupByGUID(groupId.GetCounter()))
+			{
+				group->RemoveMember(guid);
+				group = nullptr;
+			}
+		}
 
-        if (masterAccountId)
-        {
-            time_t& lastAction = lastBotActionTime[masterAccountId];
-            if (lastAction && (currentTime - lastAction) < 5)
-                return;
-            else
-                lastAction = currentTime;
-        }
+		uint32 accountId = sCharacterCache->GetCharacterAccountIdByGuid(guid);
+		
+		if(accountId == masterAccountId)
+		{
+			time_t CurrentTime = time(nullptr);
+			if (DelayBot && (CurrentTime - DelayBot) <= 10)
+			{
+				return;
+			}
 
-        // 5 second artificial delay
-        std::this_thread::sleep_for(std::chrono::seconds(5));
-
-        // Remove from group if necessary
-        if (ObjectGuid groupId = sCharacterCache->GetCharacterGroupGuidByGuid(guid))
-        {
-            if (Group* group = sGroupMgr->GetGroupByGUID(groupId.GetCounter()))
-            {
-                group->RemoveMember(guid);
-            }
-        }
+			DelayBot = CurrentTime;
+		}
 
         PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
         if (!botAI)
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -98,13 +98,26 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
     Player* masterPlayer = masterSession ? masterSession->GetPlayer() : nullptr;
 
     bool isRndbot = !masterAccountId;
+    bool allowed = true;
+	std::ostringstream out;
+
+	if(!isRndbot)
+	{
+		time_t CurrentTime = time(nullptr);
+		if (DelayBot && (CurrentTime - DelayBot) <= 10)
+		{
+			allowed = false;
+			out << "Wait 10s to call the next bot";
+		}
+
+		DelayBot = CurrentTime;
+	}
+
     bool sameAccount = sPlayerbotAIConfig->allowAccountBots && accountId == masterAccountId;
     Guild* guild = masterPlayer ? sGuildMgr->GetGuildById(masterPlayer->GetGuildId()) : nullptr;
     bool sameGuild = sPlayerbotAIConfig->allowGuildBots && guild && guild->GetMember(playerGuid); 
     bool addClassBot = sRandomPlayerbotMgr->IsAddclassBot(playerGuid.GetCounter());
 
-    bool allowed = true;
-    std::ostringstream out;
     std::string botName;
     sCharacterCache->GetCharacterNameByGuid(playerGuid, botName);
     if (!isRndbot && !sameAccount && !sameGuild && !addClassBot)
@@ -290,10 +303,33 @@ void PlayerbotMgr::CancelLogout()
     }
 }
 
-void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
+void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid, uint32 masterAccountId)
 {
     if (Player* bot = GetPlayerBot(guid))
     {
+        //Checks if the bot is in a group and removes it to avoid a possible crash
+		if (ObjectGuid groupId = sCharacterCache->GetCharacterGroupGuidByGuid(guid))
+		{
+			if (Group* group = sGroupMgr->GetGroupByGUID(groupId.GetCounter()))
+			{
+				group->RemoveMember(guid);
+				group = nullptr;
+			}
+		}
+
+		uint32 accountId = sCharacterCache->GetCharacterAccountIdByGuid(guid);
+		
+		if(accountId == masterAccountId)
+		{
+			time_t CurrentTime = time(nullptr);
+			if (DelayBot && (CurrentTime - DelayBot) <= 10)
+			{
+				return;
+			}
+
+			DelayBot = CurrentTime;
+		}
+
         PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
         if (!botAI)
             return;
@@ -689,7 +725,7 @@ std::string const PlayerbotHolder::ProcessBotCommand(std::string const cmd, Obje
         if (!GetPlayerBot(guid))
             return "not your bot";
 
-        LogoutPlayerBot(guid);
+        LogoutPlayerBot(guid, masterAccountId);
         return "ok";
     }
 
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -98,20 +98,24 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
     Player* masterPlayer = masterSession ? masterSession->GetPlayer() : nullptr;
 
     bool isRndbot = !masterAccountId;
+
+    std::ostringstream out;
     bool allowed = true;
-	std::ostringstream out;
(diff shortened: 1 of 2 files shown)


Regrad:
I'm asking you to add and change a few things, please.
1) The timer should not be restarted every time a bot is called, if the bot is not summoned. (Pressing the button too early resets the timer without calling the bot)
2) Add a timer of at least 5 seconds for initializing the bot (my request is that by pressing raid initialization 100 times, you can completely suspend the server).
3) Make the same timer for sending the bot offline for 5 seconds.

And I think a 5-second bot summon timer should be enough.

And a question. Is this a personal CD or a server CD? That is, if two players want to summon a bot at the same time, the server may still crash if it is a personal CD.

hermensbas:
I wanna investigate the real problem, but i simply dont have enough time on my hands. Its a mitigation path, not entirely sure we should implement this. But even if so as temporary solution. My suggestion would be is to make the delay timings configurable and not hardcorded. Nevertheless amazing work on the trying to figure out what the issue's are.

whether this should be merged i leave the owner.

EricksOliveira:
I will work to make the code cleaner and more reusable. I just ask for a little more time.

serverwar:
> Quero investigar o problema real, mas simplesmente não tenho tempo suficiente. Essa abordagem é um caminho de mitigação, não tenho certeza se devemos implementá-la. Mas, mesmo que seja, é uma solução temporária. Minha sugestão seria tornar os tempos de atraso configuráveis ​​e não fixos. Mesmo assim, um trabalho incrível tentando descobrir quais são os problemas, juntamente com...@Regrade@serverwar
> 
> Deixo a decisão sobre se isso deve ser mesclado para o proprietário.



> I wanna investigate the real problem, but i simply dont have enough time on my hands. This approach is a mitigation path, not entirely sure we should implement this. But even if so as temporary solution. My suggestion would be is to make the delay timings configurable and not hardcorded. Nevertheless amazing work on the trying to figure out what the issue's are together with @Regrad and @serverwar
> 
> Whether this should be merged i leave to the owner.

It shouldn't be merged yet, because I'm testing it and on Monday we'll improve it. In my case, crashes dropped a lot. I managed to stop crashes that don't happen anymore. There's still one that happens, I'm trying to find out. I agree that it's not the best way. I don't have another way. This was the one that gave me a break.

serverwar:
When you turn off the bots, they leave instantly. Ideally, they should log out after a pause.

noisiver:
Looking at the code, does this mean that whenever the bots log out they will leave the group? If so, that's terrible and not something that should be enforced like this.

serverwar:
> Looking at the code, does this mean that whenever the bots log out they will leave the group? If so, that's terrible and not something that should be enforced like this.

Yes, to avoid a master failure, you create a raid with alt, the leader logs out and when someone else logs in, the raid crashes.

noisiver:
> > Looking at the code, does this mean that whenever the bots log out they will leave the group? If so, that's terrible and not something that should be enforced like this.
> 
> Yes, to avoid a master failure, you create a raid with alt, the leader logs out and when someone else logs in, the raid crashes.

The changes to this project should not be tailored to only suit the larger population servers. I, like the majority of users, don't want to reform my group whenver I relog and the majority of us don't see crashes like yours.

It should be configurable if anything, where the default option is the current behavior.

serverwar:
Don't merge this project, leave it there for those who really need to use and test it, it's frustrating for us to crash every 10 minutes.

src/PlayerbotMgr.cpp
```diff
@@ -98,13 +98,26 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
     Player* masterPlayer = masterSession ? masterSession->GetPlayer() : nullptr;
 
     bool isRndbot = !masterAccountId;
+    bool allowed = true;
+	std::ostringstream out;
+
+	if(!isRndbot)
```
kadeshar: That is temporary workaround for servers which have problem with multithreading. In most common aproach that will be problematic. For me that should be configurable.

src/PlayerbotMgr.cpp
```diff
@@ -98,13 +98,26 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
...
+	{
+		time_t CurrentTime = time(nullptr);
+		if (DelayBot && (CurrentTime - DelayBot) <= 10)
```
kadeshar: For me that 10 sec delay should be configurable. Or if you have some measurements which proof that 10 sec is perfectly suitable for workaround please provide them

src/PlayerbotMgr.cpp
```diff
@@ -290,10 +303,33 @@ void PlayerbotMgr::CancelLogout()
...
     {
+        //Checks if the bot is in a group and removes it to avoid a possible crash
+		if (ObjectGuid groupId = sCharacterCache->GetCharacterGroupGuidByGuid(guid))
+		{
+			if (Group* group = sGroupMgr->GetGroupByGUID(groupId.GetCounter()))
+			{
+				group->RemoveMember(guid);
+				group = nullptr;
+			}
+		}
```
kadeshar: Something like this dont work?
if (bot->GetGroup())
    bot->RemoveFromGroup();

Currently GetGroup is called in line 337.

Ofc personally i think that should be configurable

src/PlayerbotMgr.h
```diff
@@ -30,7 +30,7 @@ class PlayerbotHolder : public PlayerbotAIBase
     void AddPlayerBot(ObjectGuid guid, uint32 masterAccountId);
     void HandlePlayerBotLoginCallback(PlayerbotLoginQueryHolder const& holder);
 
-    void LogoutPlayerBot(ObjectGuid guid);
+    void LogoutPlayerBot(ObjectGuid guid, uint32 masterAccountId = 0);
```
kadeshar: Why 0 not nullptr?

src/PlayerbotMgr.cpp
```diff
@@ -290,10 +303,33 @@ void PlayerbotMgr::CancelLogout()
...
+		
+		if(accountId == masterAccountId)
+		{
+			time_t CurrentTime = time(nullptr);
+			if (DelayBot && (CurrentTime - DelayBot) <= 10)
+			{
+				return;
+			}
+
+			DelayBot = CurrentTime;
```
kadeshar: Repeated code. Maybe that can be implemented as method

EricksOliveira:
> > > Olhando o código, isso significa que sempre que os bots fizerem logout, eles sairão do grupo? Se sim, isso é terrível e não é algo que deva ser imposto dessa forma.
> > 
> > 
> > Sim, para evitar uma falha do mestre, você cria um ataque com alt, o líder sai do jogo e quando outra pessoa entra, o ataque trava.
> 
> As mudanças neste projeto não devem ser adaptadas apenas para servidores com maior população. Eu, como a maioria dos usuários, não quero reformar meu grupo sempre que faço login novamente, e a maioria de nós não vê travamentos como o seu.
> 
> Deve ser configurável, se possível, onde a opção padrão é o comportamento atual.

**This PR is not for that. Even because I already made an optional PR for this function https://github.com/liyunfan1223/mod-playerbots/pull/702. This PR is only to prevent the Bot from logging into the game and not having a Leader present, the server Crash, with that it can only be called every 10s, to prevent many bots from being called and the group leader logging out without the bot completing the login.**

Regrad:
With the latest update of this PR, the server began to hang. It seems that some of the actions, such as entering the bot or initializing it, completely stop the server. And until the timer runs out (5 seconds), the server just hangs.

noisiver:
I don't quite understand. Does that mean that there will be a 10 second delay between each bot logging in? If so, that sounds terrible. That would mean it takes roughly 3 minutes to log in my 39 altbots.

serverwar:
> I don't quite understand. Does that mean that there will be a 10 second delay between each bot logging in? If so, that sounds terrible. That would mean it takes roughly 3 minutes to log in my 39 altbots.
This is for large servers, we can't just think about small or solo servers and we don't have time to think about it, we all have a social life and not everyone wants to work for small servers for the most part, Small servers, as you say, don't crash, I just used your words in the opposite way.

noisiver:
> This is for large servers, we can't just think about small or solo servers and we don't have time to think about it, we all have a social life and not everyone wants to work for small servers for the most part, Small servers, as you say, don't crash, I just used your words in the opposite way.

Gimping the server for us that enjoy it as it is is **not** a solution. Right now logging in my 39 altbots is instant, literally, and adding a bunch of delay to that will ruin a part of the enjoyment of the bots. There is simply no way I am alone in feeling that way either.

As I have said in the past, the main focus of this project is a small to tiny population. You can't ruin the process for the majority (the 99%) just to suit the 1%. A configurable thing to add a delay would be fine in my eyes, as long as it's **disabled** by default, but enforcing one is just wrong.

Dreathean:
I would have to imagine that if this were implemented it would be strictly as a config so that those with small servers can safely ignore it. There shouldn't be any issue there.

This main focus of the project without a doubt has always been and still is to support solo/small servers, but that also doesn't mean people can't help support large servers if they so desire.

kadeshar:
I have same point of view and included that in PR review.

JuniorMarioT:
> > This is for large servers, we can't just think about small or solo servers and we don't have time to think about it, we all have a social life and not everyone wants to work for small servers for the most part, Small servers, as you say, don't crash, I just used your words in the opposite way.
> 
> Gimping the server for us that enjoy it as it is is **not** a solution. Right now logging in my 39 altbots is instant, literally, and adding a bunch of delay to that will ruin a part of the enjoyment of the bots. There is simply no way I am alone in feeling that way either.
> 
> As I have said in the past, the main focus of this project is a small to tiny population. You can't ruin the process for the majority (the 99%) just to suit the 1%. A configurable thing to add a delay would be fine in my eyes, as long as it's **disabled** by default, but enforcing one is just wrong.

It needs to be a config option imo, as i thought the whole idea to have playerbots was for very small solo servers, i actually don`t see the point in a "large" server having playerbots to begin with as they are there to fill the gaps that real players would take.

serverwar:
> > > This is for large servers, we can't just think about small or solo servers and we don't have time to think about it, we all have a social life and not everyone wants to work for small servers for the most part, Small servers, as you say, don't crash, I just used your words in the opposite way.
> > 
> > 
> > Gimping the server for us that enjoy it as it is is **not** a solution. Right now logging in my 39 altbots is instant, literally, and adding a bunch of delay to that will ruin a part of the enjoyment of the bots. There is simply no way I am alone in feeling that way either.
> > As I have said in the past, the main focus of this project is a small to tiny population. You can't ruin the process for the majority (the 99%) just to suit the 1%. A configurable thing to add a delay would be fine in my eyes, as long as it's **disabled** by default, but enforcing one is just wrong.
> 
> It needs to be a config option imo, as i thought the whole idea to have playerbots was for very small solo servers, i actually don`t see the point in a "large" server having playerbots to begin with as they are there to fill the gaps that real players would take.

The idea of ​​using a module is to offer players gameplay with or without friends, imagine a guy who only has 5 am to play, at that time servers are empty, there will be no groups for icc, df or bg generally populated servers have a balance problem between ally and horde, or even a df always depending on tank and healer.

serverwar:
> > > > This is for large servers, we can't just think about small or solo servers and we don't have time to think about it, we all have a social life and not everyone wants to work for small servers for the most part, Small servers, as you say, don't crash, I just used your words in the opposite way.
> > > 
> > > 
> > > Gimping the server for us that enjoy it as it is is **not** a solution. Right now logging in my 39 altbots is instant, literally, and adding a bunch of delay to that will ruin a part of the enjoyment of the bots. There is simply no way I am alone in feeling that way either.
> > > As I have said in the past, the main focus of this project is a small to tiny population. You can't ruin the process for the majority (the 99%) just to suit the 1%. A configurable thing to add a delay would be fine in my eyes, as long as it's **disabled** by default, but enforcing one is just wrong.
> > 
> > 
> > It needs to be a config option imo, as i thought the whole idea to have playerbots was for very small solo servers, i actually don`t see the point in a "large" server having playerbots to begin with as they are there to fill the gaps that real players would take.
> 
> The idea of ​​using a module is to offer players gameplay with or without friends, imagine a guy who only has 5 am to play, at that time servers are empty, there will be no groups for icc, df or bg generally populated servers have a balance problem between ally and horde, or even a df always depending on tank and healer.

The system is good for both, those who want to play solo and those who want to do something more to integrate with real players. As the community is made up of both, I don't see this difference, especially because we help each other improve with reports and this improves even for those who play solo, I test raids and let you know how it went.

noisiver:
> > > > > This is for large servers, we can't just think about small or solo servers and we don't have time to think about it, we all have a social life and not everyone wants to work for small servers for the most part, Small servers, as you say, don't crash, I just used your words in the opposite way.
> > > > 
> > > > 
> > > > Gimping the server for us that enjoy it as it is is **not** a solution. Right now logging in my 39 altbots is instant, literally, and adding a bunch of delay to that will ruin a part of the enjoyment of the bots. There is simply no way I am alone in feeling that way either.
> > > > As I have said in the past, the main focus of this project is a small to tiny population. You can't ruin the process for the majority (the 99%) just to suit the 1%. A configurable thing to add a delay would be fine in my eyes, as long as it's **disabled** by default, but enforcing one is just wrong.
> > > 
> > > 
> > > It needs to be a config option imo, as i thought the whole idea to have playerbots was for very small solo servers, i actually don`t see the point in a "large" server having playerbots to begin with as they are there to fill the gaps that real players would take.
> > 
> > 
> > The idea of ​​using a module is to offer players gameplay with or without friends, imagine a guy who only has 5 am to play, at that time servers are empty, there will be no groups for icc, df or bg generally populated servers have a balance problem between ally and horde, or even a df always depending on tank and healer.
> 
> The system is good for both, those who want to play solo and those who want to do something more to integrate with real players. As the community is made up of both, I don't see this difference, especially because we help each other improve with reports and this improves even for those who play solo, I test raids and let you know how it went.
> 
> The system was made for those who need it, those who don't need it just don't apply it and continue their game without a CD

Adding a delay is not good for both. It might fit with a larger playerbase, like in your case, but it only affects the solo to tiny population in a negative way.

`those who don't need it just don't apply it` - It's enforced. You can't toggle it off as it is. With the majority not wanting a delay, it should be the larger servers that should add it on their end (or toggle it on if it's an option, making it disabled by default).

I'm sorry to say that I get the feeling that you're only looking at things from your perspective, and it really feels like you're thinking your experience matter more than other users.

Dreathean:
I think there was some confusion there possibly due to translation - I believe his first line regarding "the system" is referring to the playerbots module in general, not this PR specifically. 

I also believe that everyone is on the same page that this sort of implementation cannot proceed without being a config that is set to disabled by default so as not to disrupt any solo/small servers.

Regrad:
> I think there was some confusion there possibly due to translation - I believe his first line regarding "the system" is referring to the playerbots module in general, not this PR specifically.
> 
> I also believe that everyone is on the same page that this sort of implementation cannot proceed without being a config that is set to disabled by default so as not to disrupt any solo/small servers.

I think this is a temporary solution until the real cause of the crashes is fixed. And if so, the configuration is needed, because not everyone needs this fix.

EricksOliveira:
I don't have time to release an update for this fix without the waiting time. However, I am busy with work.

Regrad:
I will check this PR, but we still have a problem in which the player can press "Initialize" and freeze the server indefinitely.
Plus, we still have a server crash during a quick click on "Go offline" in turn to a large number of bots in the raid, which I even recorded on video. Many thanks to EricksOliveira for working on this fix, I hope it will at least partially solve the problems with crashes.

Regrad:
Crashes still exist, but they seem to have become less frequent.

EricksOliveira:
Thanks for the feedback. First update is only for the bot login. I will still make a correction for the logout.

serverwar:
I really gave up using alt chars, low dc frequency, I have nothing more to report, it's all there

hermensbas:
> I really gave up using alt chars, low dc frequency, I have nothing more to report, it's all there

In context semi public server you mean?

PR #1263, opened 2025-05-04
Comment fix

NoxMax:
Reflecting changes made a while back about changing the default active bots from 10% to 100% 9fb2b51f169ddb9ab7c7a463f85cafe1da33c56c


hermensbas:
Not 100% correct but good enough for now

PR #1266, opened 2025-05-05
Fix inventory clear on init quests

liyunfan1223:
InitQuests no longer clear inventory but instead destroy the corresponding reward items. 

Fix issues that sometimes cause inventory to be incorrectly cleared, (e.g. maintenance https://github.com/liyunfan1223/mod-playerbots/issues/1244).

### DIFF
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -2773,7 +2773,7 @@ void PlayerbotFactory::AddPrevQuests(uint32 questId, std::list<uint32>& questIds
     }
 }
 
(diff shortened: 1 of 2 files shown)


brighton-chi:
Awesome, this fixed the issue for me! Is there any way to explain in a simple way why this issue was impacting only a few players such as myself, and why it did not affect bots below level 60?

liyunfan1223:
Inventory is cleared on `InitAttunementQuests` (initialization of tbc dungeon unlocking quests). https://github.com/liyunfan1223/mod-playerbots/issues/1014

In short, the issue only occurs during the first maintenance, or when the bot level is between 60 to 69. In addition, the issue will not occur if all unlocking quests have been completed.

PR #1267, opened 2025-05-05
feature - (#1100) Linking of "trusted" accounts to allow altbot-control apart from own account or guild

### LINKED ISSUE #932: Idea for Bots on other Accounts
The only method to summon and control bots from another account is to use guilds right?

extremely complicated to set up.
Many many steps and you have to be online with two accounts to invite them in the guild.

A method to link a bunch of accounts together will be nice. So you make one account as the master. This account is for your player characters. And then you create an account named like yours with _bots at the end only for your bots.
Now the request link feature comes in. You link the accounts together and now you are able to use all Chars on the bot account as your Playerbots.

With this it should be very easy to make and manage bots in a separate bot account. Especially for new gamers.

I find it very unpleasant to have players and bots on the same account. You have less space for characters and even less for the bots. Sure you can make more accounts but this is unpleasant to.

If any of you find this idea good and can code, I will love you if you implement such a feature. ❤️

### LINKED ISSUE #1100: Feature request / Idea - Account linking for altbot control (only control altbots from accounts you established a link with)
When running a server with multiple players who are also in guilds with one another, there is no solution to allow altbot usage for specific users/accounts only. 

The config currently allows you to:
a) Allow/deny bots from the player's account
AiPlayerbot.AllowAccountBots = 1
OR / AND
b) Allow/deny bots in the player's guild
AiPlayerbot.AllowGuildBots = 1

Since players are using guilds together, this opens up access to characters of other players' accounts. The only alternative would be to be restricted to only characters of your current account.

A possible solution could be the addition of chat based commands to allow the player to set-up links between accounts, making them "trusted" which will in return allow them to use characters from this account based on a config option: AllowTrustedAccountBots

Idea outline:
- Add a chat command to allow players to set up a securityKey that will let them (or other "trusted" players - like spouses) link this account to another. (e.g. `.playerbots account set securityKey=someKey`)
- Add a chat command to establish a link to another account (e.g. `.playerbots account link $accountname $securityKey`) - The account with name=`accountname` can now be used for altbots on the current account
- Add a chat command to see linked accounts (e.g. `.playerbots account linkedAccounts`) which will print all linked accounts in the chat
- Add a chat command to revoke access to the current account (unlink account) (e.g. `.playerbots account unlink $accountname`) - unlink can be called from both accounts and will revoke the access to the accounts characters

Additional adjustments needed:
- Adjustments will be necessary in the logic where the mod is currently using the beforementioned config options.

- writing a dedicated module as a companion module to mod-playerbots (since the functionality would anyhow require adjustments in mod-playerbots, this seems overcomplicated)
- using a dedicated guild per player (this would work as each player would only have access to characters from his guild, but would completely cancel out the possibility of guild-play and communication with other players and would feel more like a restriction)
-  relying on "common sense" that other players will not use your character in some malicious activities or tinker with it in any way (not a particularly strong point in online gaming 🙈 ) - despite for players who you specifically trust to "handle your character with care".

**Additional context**
This is an early-state idea - if people find it interesting we can find a good way for implementation or I may even start to work on a PR myself, once the concept is "approved" (still learning though).

icemansparks:
# This PR introduces linking of accounts as an alternative to existing measures for allowing Alt-Bot control.
This PR implements and closes #1100  as mentioned in comments, also closes #932 

As an alternative to controlling the characters from the own account or guild member accounts, this allows to link up different accounts together to allow control of the characters as altbots.

Players will be able to define a security key. Once _Player A_ has set up a security key, he can share it with "trusted individuals" (or use it to link two of his own accounts). _Player B_ will now be able to link his account with the account of Player A, allowing vice versa control of each others characters as altbots.

This can be used as addition to the current options, or separately and therefore has no effect on previous altbot control options.

## The following chat commands have been added:

1. Define a **securityKey** for the current account (the key will be overridden when defined multiple times - no need to print the current key at any given point)
`.playerbots account setKey somePlayerDefinedSecurityKeyValue`

2. Link an account by using its **securityKey**
`.playerbots account link ACCOUNTNAME somePlayerDefinedSecurityKeyValue`

3. Shows a list of accounts, which are currently linked with this account
`.playerbots account linkedAccounts`

4. Remove a link to a linked account (this will remove the link for both sides, ensuring that access to the characters can be canceled from both sides anytime)
`.playerbots account unlink ACCOUNTNAME`

## The PR introduces two new tables in the playerbots database to:
- store account security keys
- store account linking info

## What is not handled with this change?
- When an account is unlinked, currently active altbots are not updated / logged out specifically (if this is not handled in general for example in case of config updates, I at least did not add any specific handling for that).


Raz0r1337:
That implements this too.

https://github.com/liyunfan1223/mod-playerbots/issues/932

Thank you for that, you are my hero. 👍

icemansparks:
Not sure how to create a PR to the wiki, so here are the suggested changes to the ``Playerbot-Commands.md``:
`## Account linking for Altbot control

As an alternative to controlling the characters from your own account or guild member accounts, this allows to link up different accounts together to allow control of the characters as Altbots. This is based on the ``AiPlayerbot.AllowTrustedAccountBots = 1`` option in the playerbots config file.

command | action
:---|:---
``.playerbots account setKey somePlayerDefinedSecurityKeyValue`` | Define a securityKey for the current account (the key will be overridden when defined multiple times)
``.playerbots account link ACCOUNTNAME somePlayerDefinedSecurityKeyValue`` | Link an account by using its securityKey
``.playerbots account linkedAccounts`` | Shows a list of accounts, which are currently linked with this account
``.playerbots account unlink ACCOUNTNAME`` | Remove a linked account (this will remove the link for both sides, ensuring that access to the characters can be canceled from both sides anytime)`

Dreathean:
This is awesome! And thank you for including the wiki addition, will gladly get that entered once this is merged.

hermensbas:
I will add it to the wiki

PR #1279, opened 2025-05-09
Fixed a typo in filename and renamed table for consistency

noisiver:
- All tables in the database start with `playerbots_` so this should too.
- Fixed a typo in the filename for the new table

### DIFF
--- a/data/sql/playerbots/updates/db_playerbots/2025_05_09_00.sql
+++ b/data/sql/playerbots/updates/db_playerbots/2025_05_09_00.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `playerbot_account_keys` RENAME `playerbots_account_keys`;
+ALTER TABLE `playerbot_account_links` RENAME `playerbots_account_links`;
--- a/data/sql/playerbots/updates/db_playerbots/2025_05_09_00.sql
+++ b/data/sql/playerbots/updates/db_playerbots/2025_05_09_00.sql
@@ -1,2 +1,15 @@
-ALTER TABLE `playerbot_account_keys` RENAME `playerbots_account_keys`;
-ALTER TABLE `playerbot_account_links` RENAME `playerbots_account_links`;
+DROP TABLE IF EXISTS `playerbot_account_keys`;
+CREATE TABLE IF NOT EXISTS `playerbots_account_keys` (
+    `account_id` INT PRIMARY KEY,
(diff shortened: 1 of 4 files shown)


data/sql/playerbots/updates/db_playerbots/2025_05_09_00.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `playerbot_account_keys` RENAME `playerbots_account_keys`;
+ALTER TABLE `playerbot_account_links` RENAME `playerbots_account_links`;
```
nl-saw: I'm concerned this will throw errors on fresh installations as then these tables do no exist. Perhaps it might be best to just DROP TABLE IF EXISTS on these 2, despite the potential data-loss.

noisiver: Maybe I shouldn't alter the original SQL files other than the rename so by default it imports the original names and then change them in the update? Would that potentially work?

nl-saw: Yeah that would work, yet given the short live-span of this functionality, it might be cleaner to solve this in the base files as you have done and just keep the update to delete the old ones if they exist. The latter is just my personal preference.

Edit: Coming to think of it, the correct tables then also should be created if not exists inside the update file to not break current installations. Leaving the decision up to you.

noisiver: Check out the changes I made to the update file. I don't know the best way to handle it but now it'll drop tables with the old name if they exist and create tables with the new names if they don't exist already, so at least there's some logic to it. Of course it does mean data loss as you say but it's a one-off (hopefully).

nl-saw: Yeah, this looks good to me, and indeed unfortunate, yet I don't see major ramifications there and a one-off indeed.

Raz0r1337:
After applying this update, all saved keys and account links were deleted.

noisiver:
That was known and considered an acceptable loss since it'll only ever happen once. If you have a backup you could restore those tables and rename them yourself.

Raz0r1337:
It wasn't a big deal, I just thought it should be reported. 😉 I was able to reconnect everything quickly.

PR #1281, opened 2025-05-10
feat(playerbots): staggered taxi take-off for bots

### LINKED ISSUE #1017: Bots use Flight master nearly the same time
This is not a bug or anything like that, more of a minor inconvenience. My wife and I noticed that the bots were all using the flight master at nearly the same time.
We had the idea that it would greatly improve the immersion if there was some kind of timer that would cause the bots to fly one after the other, so that a line with different distances was formed.

If any of you like this idea and want to and have the time to implement it, we would be very happy about it. You fly quite a lot in this game and then look at the "flight formation" and the environments. 🤣

Type1Error:
Adds four new configurable settings to playerbots.conf:

- AiPlayerbot.BotTaxiDelayMinMs:   Min random delay before the 1st follower bot clicks the flight-master
- AiPlayerbot.BotTaxiDelayMaxMs:   Upper bound for the overall taxi delay window – larger spreads big raids
- AiPlayerbot.BotTaxiGapMs:        Fixed gap added per group-slot so bots never take off together
- AiPlayerbot.BotTaxiGapJitterMs:  Extra small randomness added to each gap so launches don’t look robotic

These options allow server owners to fine-tune how bots queue up and take off from flight masters, making their behavior appear more natural.

Closes #1017 : Bots use Flight master nearly the same time.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -18,6 +18,7 @@
 #        COMBAT
 #        CHEATS
 #        SPELLS
+#        FLIGHTPATH
 #    PLAYERBOT RNDBOT SPECIFIC SETTINGS
 #        GENERAL
 #        LEVELS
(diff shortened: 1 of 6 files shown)


hermensbas:
nice

Raz0r1337:
Love it, thank you very much for implementing this.

hermensbas:
> Thanks for the PR, I have one concern reading the random timing being 5 seconds max, whereas this perhaps might interfere in combination with the expireActionTime configuration (which is also 5 seconds by default), but it is something that can be tested.

mmm interesting, i havent tested that. Nor researched the code for that particular use case. I will try and this this evening or atleast verify code wise.

PR #1290, opened 2025-05-13
paladin and feral cat offheal strat

NotCronky:
co +offheal to either ret pally or cat druid will now change from dps to offheal, the bots will prioritise dps but heal when needed


noisiver:
The paladin off-heal headers are spelled with a lowercase H but the include has it set to uppercase. Rename the header to proceed with this.

noisiver:
I think it would make sense, for consistency, to either rename the druid files to OffhealFeralStrategy or to rename the paladin files to OffHealPaladinStrategy.

Issue #941, opened 2025-02-03
[feature request] Create Raid Composition


blinkysc:
Was thinking maybe could create a new playerbot command that would create a bot account with 25 bots that compose a solid raid composition. Or maybe it could be 40 for us boomers? 

Maybe "switches" that define level starts for the grinders that like leveling with their bots, or if people want to just cheat and start raiding?

Maybe logic that just selects based on filling all raid buffs? Or maybe better hardcoded per faction?

blinkysc:
![Image](https://github.com/user-attachments/assets/631459b1-5aec-4052-8ecf-7e5f2175c266)

Here was a LOD raid comp my group once used... Discord is cleaned out lol

kadeshar:
I think making support in multiple places is unneccesary.
Multibot should handle that like you suggest https://github.com/Macx-Lio/MultiBot/issues/11

blinkysc:
I think you would probably need the underlining functionality in player bot and for multi bot the “front end” for it (so without having to do commands). But maybe there’s just a way with multibot.

kadeshar:
In TBC i used Method Raid Tools for auto-invite prepared composition. Sadly this not exists for wotlk client.

PR #1291, opened 2025-05-13
CMaNGOS Playerbots "lfg" command implemented

brian8544:
This PR implements the "LFG" command functionality for CMaNGOS Playerbots.

The command allows for the automatic creation of raid groups based on the player's input and the bot's allowed roles set within the script. It supports automated group formation based on parameters such as group size and roles (tank, healer, DPS).

While I know there are existing addons for this, this should also make addon development easier, since all they need to do is call a /say with the group size (`/say /1 lfg 10` -- `/say /1 lfg 25`. So this could be seen as benificial for everyone, as it has also been referenced in a GitHub issue here: #941

Before merging, please verify that no other issues arise. I’m not claiming to be highly experienced, so please be cautious. Fortunately, there should be plenty of experienced module/core developers available to assist with this.


brian8544:
- Tank and healer roles are _not_ being filled consistently. Sometimes it works, sometimes it doesn't. This should be checked. 
- Implement a role switch command (e.g., Retribution Paladin switching to Protection) would be a great feature to have along this feature as well.
- Group search _continues_ even after the group has been filled, causing unnecessary strain on the server. You can test this by leaving your group - it will instantly assign you to another one. I'm not sure why this happens, as it doesn't occur on CMaNGOS.

PR #1294, opened 2025-05-13
Clarify EnableNewRpgStrategy relation to RandomBotTeleLowerLevel and RandomBotTeleHigherLevel, and making them more intuitive

NoxMax:
**Changes:**
1. Clarifying in conf.dist that EnableNewRpgStrategy RandomBotTeleLowerLevel and RandomBotTeleHigherLevel (it would instead use the ranges in [PrepareZone2LevelBracket](https://github.com/liyunfan1223/mod-playerbots/blob/f17ec36cdecc908b7b51a3b161a1136462ba6348/src/RandomPlayerbotMgr.cpp#L1494))

2. Clarifying in conf.dist what RandomBotTeleLowerLevel and RandomBotTeleHigherLevel do.

3. Switching around the names of the variables to make them actually intuitive to their function. Example, consider Borean Tundra, a zone with creatures from level 68 to 72. With these default values
```
AiPlayerbot.RandomBotTeleLowerLevel = 1
AiPlayerbot.RandomBotTeleHigherLevel = 3
```
now the zone is open to teleporting bots from level 67 (68-1) to 75 (72+3), whereas before the numbers were switched to do the same function.

~~4. Capitalized every mention of the word "default" in playerbots.conf.dist that should've been capitalized, because it bothered me that some weren't -_-~~ Addressed by https://github.com/liyunfan1223/mod-playerbots/pull/1296

**Closes issue:** https://github.com/liyunfan1223/mod-playerbots/issues/1288

**Next step:** 
As stated, RandomBotTeleLowerLevel and RandomBotTeleHigherLevel are overridden by EnableNewRpgStrategy anyway, and since EnableNewRpgStrategy is enabled by default now, I am thinking of making the zones brackets in [PrepareZone2LevelBracket](https://github.com/liyunfan1223/mod-playerbots/blob/f17ec36cdecc908b7b51a3b161a1136462ba6348/src/RandomPlayerbotMgr.cpp#L1494) user facing, so they can be updated in the conf if someone wants to change them. Would like thoughts and opinions on that.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -718,7 +718,7 @@ AiPlayerbot.RandomBotGroupNearby = 0
 AiPlayerbot.AutoDoQuests = 1
 
 # Random Bots will behave more like real players (exprimental)
-# This option will override AutoDoQuests
+# This option will override AutoDoQuests, RandomBotTeleLowerLevel, and RandomBotTeleHigherLevel
 # Default: 1 (enabled)
 AiPlayerbot.EnableNewRpgStrategy = 1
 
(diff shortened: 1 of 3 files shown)


hermensbas:
Not entirely sure why someone wants to chance the level teleport brackets, but it doesnt hurt to make it configurable either

PR #1298, opened 2025-05-15
druid cat and ret pally offheal strat

NotCronky:
co +offheal to either cat druid or ret pally will disable dps mode and enable offheal, the bots will now focus damage but heal when necessary.


hermensbas:
if merged, we might wanna add the new option as in the wiki.

PR #1300, opened 2025-05-15
Use instalt logout instead of waiting

EricksOliveira:
The main goal of this change is to improve the efficiency and responsiveness of bots, especially in situations where immediate logout is necessary, such as: 

Quick bot restarts. Avoiding inconsistent states when switching characters or restarting sessions. 

Reducing wait times in automated tests or management scripts.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -330,6 +330,9 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
         // check for instant logout
         bool logout = botWorldSessionPtr->ShouldLogOut(time(nullptr));
 
+        // make instant logout for now
+        logout = true;
+
         if (masterWorldSessionPtr && masterWorldSessionPtr->ShouldLogOut(time(nullptr)))
             logout = true;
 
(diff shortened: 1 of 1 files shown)


hermensbas:
Can you explain? Because the code has not changed. You changed the order, and added additional safety check, which shouldnt happen but it cant hurt either. But seems a bit random changes.

EricksOliveira:
This update can prevent crashes like boot login and logout. "If I found a Player* with the guid, and it is not nullptr, I return it. Otherwise, I return nullptr"

EricksOliveira:
Thanks for the detailed feedback! You're right about the functional equivalence — using nullptr instead of 0 is more of a style/modern (C++11+) improvement, with no direct impact on stability. The redundant check it->second ? it->second : nullptr is also unnecessary, and your suggestion to simplify it to return (it == playerBots.end()) ? nullptr : it->second; is valid and cleaner. I totally agree. About logout = true, I don't really have a specific crash tied to it — this change was driven by local testing with bots hanging in a logged out state, but I admit I don't have any concrete logs or crash reports to justify it. I can review and remove it if there's no clear benefit. Thanks for the review — I'll review those parts to keep the PR clean and purposeful.

hermensbas:
Thanks for trying, but in this case there is no obviously point merging this. Closing it for now to keep PR a bit more clean, when your ready you can open the PR again.

PR #1301, opened 2025-05-16
Additional improvements to readability and accessibility of config

brighton-chi:
A few more tweaks, largely based on input from members on Discord, to try to clarify some of the more obscure but also useful features (e.g., worldbuffs).


NoxMax (commented):
Thanks for the additional updates. The only thing I have a comment about is this:

"# Additionally, the numbers after AiPlayerbot.WorldBuff can be customized and represent Faction.Class.Spec.MinLevel.MaxLevel (but the edit must be made in both the .conf and .conf.dist file to work)"

I will be testing changing the min/max level and would like someone else to test it as well. Just to confirm it's safe before we merge something telling the user they can go tinkering with .dist.

As a general note about what you're doing, I would like to say that sometimes when people add/change a function that requires adding/changing things to .dist files, some of the documentation they make is... presumptive. They assume a setting obvious to them will be obvious to other users. So thank you again for helping make the documentation clearer for everyone👍

brighton-chi:
> Thanks for the additional updates. The only thing I have a comment about is this:
> 
> 
> 
> "# Additionally, the numbers after AiPlayerbot.WorldBuff can be customized and represent Faction.Class.Spec.MinLevel.MaxLevel (but the edit must be made in both the .conf and .conf.dist file to work)"
> 
> 
> 
> I will be testing changing the min/max level and would like someone else to test it as well. Just to confirm it's safe before we merge something telling the user they can go tinkering with .dist.
> 
> 
> 
> As a general note about what you're doing, I would like to say that sometimes when people add/change a function that requires adding/changing things to .dist files, some of the documentation they make is... presumptive. They assume a setting obvious to them will be obvious to other users. So thank you again for helping make the documentation clearer for everyone👍 

Curse of knowledge, right? 

Re: the worldbuffs, I've never come across another situation in this or any other mod where editing the .dist was needed, but it seemed on Discord that this was the only way to edit anything other than the spell IDs (and it was the only was I was able to make it work). As a layperson, I had thought the .conf essentially was intended to function as an editable version of the .dist. I don't know if there's a practical way to implement this without needing to get into the .dist; this seems unusual.

NoxMax (approved):
Tested it. As far as intent, it works. Editing all figures in both conf and dist to .50.70 for example, made sure that nc +worldbuff would only activate the buff on bots from level 50 to 70


> As a layperson, I had thought the .conf essentially was intended to function as an editable version of the .dist.

That is because we're talking here about editing things before the equal sign, while the code is expecting only things after the equal sign to be edited. So my recommendation right now is this:

Have the comment say this:

```
# Applies a permanent buff to all bots simulating effects of spells, flasks, food, runes, etc.
# Requires sending the command "nc +worldbuff" in chat to a bot (or a group of bots) to enable
# Numbers after the equal signs are spell IDs and may be customized
# See Randombots Default Talent Specs for more info on each spec; they are listed in that section by the names in the parentheticals (e.g., arms pve, fury pve)
```

Leave out any mention of editing values before = or the dist file for now. Afterall, it would get overwritten the next time the user updates PB.

Then the next step is to make a new PR editing the code to properly parse level parameters, so it can accept lines this way:

`AiPlayerbot.WorldBuff.0.1.0 = 80,80,53760,57358 `

From the top of my head, I am guessing this needs editing PlayerbotAIConfig.cpp and PlayerbotAIConfig.h. If you don't know how to go about that yet, hey, neither did I a couple months ago. So consider it a learning opportunity to see how the sausage is made in the backend. A good place to start if you haven't checkout, is just searching for "[worldbuff](https://github.com/search?q=repo%3Aliyunfan1223%2Fmod-playerbots%20worldbuff&type=code)" and see what's going on in the background. 

The idea of configuring the bot levels for world buffs is pretty good, and I hope you implement it. Good luck.

brighton-chi:
Sounds good, I'll plan on updating the config later today. I agree that the better approach is to wait for the code to be revised to make any mention to changing the levels, factions, etc. It seems that editing the .dist is a bandaid approach.

NoxMax:
Any particular reason you want to make the faction ID also modifiable? I can't think of reason to make only alliance/horde bots buffable. So maybe keep that one as 0 for both. But if you can think of a good scenario to make that part of the parameter also configurable, go for it.

brighton-chi:
> Any particular reason you want to make the faction ID also modifiable? I can't think of reason to make only alliance/horde bots buffable. So maybe keep that one as 0 for both. But if you can think of a good scenario to make that part of the parameter also configurable, go for it.

Not really, I just presumed that was intended since it was one of the parameters that was specifically called out. I personally don't have any need for it, though I suppose I could think of a scenario in which one might (perhaps if you were playing Horde and Alliance characters at the same time and had them at different levels of content).

PR #1306, opened 2025-05-17
Translated Chinese documents in playerbot text files

sd3420625:
Added simplified Chinese translations to ai_playerbot_texts.sql and 2024_08_07_00.sql


liyunfan1223:
Nice, tested and works great.

liyunfan1223:
By the way, I'm not sure if this is the standard way to update in sql, perhaps someone is familiar with this? (At least the update can be automatically applied for me)

sd3420625:
Thank you for helping me fix the bug, because I don't know how to incrementally update the database content. Thank you again.

Issue #1299, opened 2025-05-15
Bug with AllowTrustedAccountBots


Regrad:
**Bug Description**
If you give your character to a friend via the account link, and then log in , the bot continues to control the character of the logged-in player. If a friend calls the bot offline during this process, the client crashes.

**Commit Hash**
4f26a8a09b8832a2d129a020daa096238dd8f84d

**How To Reproduce Bug**
Described above.

**Expected Behavior**
When a player enters a character that is currently working as a bot, the bot should be disabled.

**Screenshots**
No

**Modules**
Transmog

**Playerbot Settings**
AiPlayerbot.AllowTrustedAccountBots = 1

**System**
OS: Windows 10
Hardware: Xeon 22 cores, 64 RAM

**Additional Context**
No

Regrad:
And such a player also has two pets called up. From the bot and from the player.

Regrad:
![Image](https://github.com/user-attachments/assets/7913085c-f7d3-48e0-a43c-499c9941ffdd)

Regrad:
Not only that, the player breaks down and refuses to leave the world.

Regrad:
And so you can breed pets endlessly.

![Image](https://github.com/user-attachments/assets/e55fdfb9-176e-443d-aa02-03e516450bc1)

icemansparks:
The issue seems to be not specific to linked accounts, the same also can happen with just the guildbots option active. 
A player logging onto a character should disconnect the character first if it is used as an altbot at this point - no question.
This is how it is handled for normal player logins in the core if you log in a second time. 

I did not touch any of that logic, so I can't tell why this doesn't apply to altbots in this case. 🤔

Dreathean:
Yup I can second @icemansparks, this is not specifically a bug with AllowTrustedAccountBots and it is definitely an issue for anyone logging into an altbot while the bot is active on that character.

Regrad:
This problem can cause the server to crash under certain conditions.

icemansparks:
The relevant part to handle that seems to be in the core in `src/server/game/Server/WorldSessionMgr.cpp`

A specific handling for kicking bot sessions would need to be implemented here, I suppose?
I am, however, not 100% sure if I found the correct part for this - I did not dive into core functionality yet.

`void WorldSessionMgr::AddSession_(WorldSession* session)
{
    ASSERT(session);

    // kick existing session with same account (if any)
    // if character on old session is being loaded, then return
[...]

PR #1317, opened 2025-05-19
(#1299) Implement AccountScript override to make sure altbots are logged off when a player logs onto this account.

icemansparks:
# Overview
This PR improves the handling of altbot logouts when a real player logs into an account that has characters currently online as altbots. It ensures that all altbots (which have been manually added by some player) are properly logged out before the real player session takes over, preventing session conflicts and "ghost" bots.

closes: #1299 

# Key Changes

## New Helper Function:
Added `LogoutAltBot(ObjectGuid guid)` to robustly log out a bot by GUID.

- Tries to log out via the bot's master [PlayerbotMgr]if possible.
- Falls back to forcibly kicking the session if the bot is not tracked in any manager.

## Account Login Hook:
The PlayerbotsAccountScript::OnAccountLogin hook now calls `LogoutAltBot` for each character of the logging-in account, ensuring all bots are logged out before the player logs in.

## Efficiency:
Instead of looping over all players, the code uses the bot's AI to directly find its master and the correct PlayerbotMgr.

## Fallback Safety:
If a bot is not tracked in any manager but is still in the world, its session is forcibly kicked to guarantee cleanup.

# Why?
- Fixes issues where bots added via .playerbot bot add were not being logged out on account login.
- Prevents session conflicts and ensures a clean handover from bot to real player.

# Testing
- Log in to an account with characters currently online as bots (both random and manually added).
- Verify that all such bots are logged out before the real player enters the world.
- Check server logs for confirmation of logout actions.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -704,8 +704,13 @@ std::string const PlayerbotHolder::ProcessBotCommand(std::string const cmd, Obje
 
     if (cmd == "add" || cmd == "addaccount" || cmd == "login")
     {
-        if (ObjectAccessor::FindPlayer(guid))
-            return "player already logged in";
+        if (Player* player = ObjectAccessor::FindPlayer(guid))
+        {
+            if (!player->GetSession()->IsBot())
+                return "Cannot add bot: character is currently being played by a real player.";
+            else
+                return "player already logged in as bot";
+        }
 
         // For addaccount command, verify it's an account name
         if (cmd == "addaccount")
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -417,6 +417,9 @@ void LogoutAltBot(ObjectGuid guid)
                     // Remove from master's PlayerbotMgr map
                     mgr->RemoveFromPlayerbotsMap(guid);
 
+                    // Set AI master pointer to player
+                    botAI->SetMaster(player);
+
                     return;
                 }
             }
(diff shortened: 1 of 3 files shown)


noisiver:
I just wanted to make sure I understood what it does, it wasn't a comment hinting at it being a bad implementation or anything. I can't make that call, this is fine with me but others may disagree.

icemansparks:
> I just wanted to make sure I understood what it does, it wasn't a comment hinting at it being a bad implementation or anything. I can't make that call, this is fine with me but others may disagree.

I am not able to get it to work with `OnPlayerLogin` anyway so far 🤔 
I'll try this evening again but if someone has a suggestion, I'll happily take it.

icemansparks:
Progress Update: 
The bot is now logged out when the real player logs onto this specific character. After that, the logged-in character is no longer following and adhering to commands from the former bot owner.

HOWEVER, if the previous bot owner tries to send commands to the "former" bot, like `.playerbots bot remove xxx` or `.playerbots bot add xxx` to add him again, the connection seems to still exist somehow, leading to a server crash without logs.

A possible origin could be, that there are still remnants of the former session that have now been transferred to the user. OR the connection between bot and owner is not removed (if someone can tell me how I can verify this last bit, that would be great).

icemansparks:
Another update, still trying to find the issue...

If I understand correctly, `PlayerbotHolder::HandlePlayerbotCommand` which handles bot commands, doesn't have a check if the character that is being added as a bot (regardless of `addaccount` or `add` is used) is currently being online and being played by a real player.
Therefore, using `.playerbots bot add XXX` still results in the character trying to be added even though a real player is playing right now - which results in the crash.

icemansparks:
I hit a wall with this... not having enough time in the last months to follow up on a deeper level. 
I'll close this PR for now with the hope of having time in the future to come back to it.

PR #1328, opened 2025-05-22
Autogear override items

kadeshar:
- Added autogear action possibility to override items without ilvl condition

Example:
AiPlayerbot.AutoGearQualityLimit = 4
AiPlayerbot.AutoGearScoreLimit = 213
AiPlayerbot.RandomGearLoweringChance = 0.5

obraz

### DIFF
--- a/src/strategy/actions/TrainerAction.cpp
+++ b/src/strategy/actions/TrainerAction.cpp
@@ -227,7 +227,7 @@ bool AutoGearAction::Execute(Event event)
                     : PlayerbotFactory::CalcMixedGearScore(sPlayerbotAIConfig->autoGearScoreLimit,
                                                            sPlayerbotAIConfig->autoGearQualityLimit);
     PlayerbotFactory factory(bot, bot->GetLevel(), sPlayerbotAIConfig->autoGearQualityLimit, gs);
-    factory.InitEquipment(true);
+    factory.InitEquipment(true, true);
     factory.InitAmmo();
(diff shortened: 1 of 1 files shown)


liyunfan1223:
I would feel this is a bit strange, the second parameter of `InitEquipment` (`bool second_chance`) is orginally used to help the bot generate better equipment, such as avoiding stats overflow and incomplete itemsets. As a side effect it will destroy all old items before generating (I guess this solves the overridng problem), but it's not actually for this.

kadeshar:
Currently autogear dont gear up if current item is better than choosen by method. That means when you change settings for example from epic to rare and use autogear that will not work.

liyunfan1223:
My idea (at least original idea) is that autogear is an incremental update and if the current item is better, it should not be replaced. The situation you mentioned can be solved by unequipping and destroying the current items before autogear. Of course there can be a better way.

kadeshar:
Then maybe i will add parameter for that?

liyunfan1223:
If you or anyone else feels it's necessary to add a parameter, I'm open to it.

kadeshar:
That was simple workaround to linked issue. More welcome is parameter where autogear command create gear with same ilvl like master :) But i think even that simple parameter only extend configurability :)

Similar problem i saw in glyphs. When you change spec (and also glyphs) autogear dont generate glyphs for new spec. 
But for that is workaround with commands "remove glyph" + "autogear"

PR #1329, opened 2025-05-22
Added casting Blessing of Sanctuary on tank

kadeshar:
When 2 paladins have setted blessing on BoM then first will cast BoS on tank and second will cast BoM

obraz

Resolve: 
https://github.com/liyunfan1223/mod-playerbots/issues/1326
https://github.com/liyunfan1223/mod-playerbots/issues/676

### DIFF
--- a/src/strategy/paladin/PaladinActions.cpp
+++ b/src/strategy/paladin/PaladinActions.cpp
@@ -12,13 +12,23 @@
 #include "PlayerbotFactory.h"
 #include "Playerbots.h"
 #include "SharedDefines.h"
+#include "../../../../../src/server/scripts/Spells/spell_generic.cpp"
 
-inline std::string const GetActualBlessingOfMight(Unit* target)
+inline std::string const GetActualBlessingOfMight(Unit* target, PlayerbotAI* botAI, Player* bot)
 {
-    if (!target->ToPlayer())
+    Player* targetPlayer = target->ToPlayer();
+
+    if (!targetPlayer)
     {
         return "blessing of might";
     }
+
+    if (targetPlayer->HasTankSpec() && !targetPlayer->HasAura(SPELL_BLESSING_OF_SANCTUARY) &&
+        bot->HasSpell(SPELL_BLESSING_OF_SANCTUARY))
+    {
+        return "blessing of sanctuary";
+    }
+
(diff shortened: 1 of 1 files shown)


brighton-chi:
Great, thank you. Does it also work to directly set a paladin's strat to be to cast blessing of sanctuary? Or does it work only with two paladins set to might?

kadeshar:
First paladin in group which setted "bdps" strategy will not cast BoM on tank target but BoS. On other targets will cast BoM (or BoW if dont have benefit from attack). You dont need second paladin. But second paladin additionally add this tank target BoM.

noisiver:
It doesn't work for me either. I keep removing the blessing from my protection paladin but he only uses Blessing of Kings on himself and the other tank. If I make him cast Blessing of Sanctuary on himself he just casts Blessing of Kings again and overwrites it.

noisiver:
If I use `nc +bdps` he casts Blessing of Sanctuary and then 5 seconds later Blessing of Might and it stays like that.

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -12,13 +12,23 @@
...
 {
-    if (!target->ToPlayer())
+    Player* targetPlayer = target->ToPlayer();
+
+    if (!targetPlayer)
     {
         return "blessing of might";
     }
+
+    if (targetPlayer->HasTankSpec() && !targetPlayer->HasAura(SPELL_BLESSING_OF_SANCTUARY) &&
```
noisiver, line 26: This is why it doesn't work properly. If the user has the aura he reverts to might (when targeting a protection paladin). It shouldn't check if the target has the aura because it doesn't for the other spells.

Edit: Removing the HasAura call it constantly casts Blessing of Sanctuary. I guess a check is used elsewhere after this function is called.

kadeshar:
Yesterday i make raid with that and worked. Blessing of Sanctury have only protection paladin? When i change on him stat on "bdps" next blessing refreshment give him BoS.

noisiver:
With `nc +bdps` it does, but 5 seconds after casting Blessing of Sanctuary on himself he casts Blessing of Might on himself and overwrites Sanctuary.

noisiver:
To simplify things when showing this I used selfbot since it does exactly the same thing.

https://www.youtube.com/watch?v=BEJVgBcOIeU

PR #1338, opened 2025-05-24
Active loot roll

liyunfan1223:
Let the bot do loot roll actively. They used to wait for their master to do loot roll before reacting.

This is an immersion improvement and a preparation for a possible feature that random bots form groups without real player.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -157,7 +157,7 @@ PlayerbotAI::PlayerbotAI(Player* bot)
     masterIncomingPacketHandlers.AddHandler(CMSG_GAMEOBJ_USE, "use game object");
     masterIncomingPacketHandlers.AddHandler(CMSG_AREATRIGGER, "area trigger");
     // masterIncomingPacketHandlers.AddHandler(CMSG_GAMEOBJ_USE, "use game object");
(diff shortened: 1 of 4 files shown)


Mr-Deadbeat:
> Let the bot do loot roll actively. They used to wait for their master to do loot roll before reacting.
> 
> This is an immersion improvement and a preparation for a possible feature that random bots form groups without real player.

just a thought atm the bots equip whatever they deem as an upgrade for them meaning plate users will probably roll for cloth if it is considered a upgrade for the bot so might be good to tweak the bots so they use their appropriate gear for example i've seen paladin Tanks in full cloth on my server not a very efficient tank if you know what i mean and other classes have the same issue they use gear more suitable for other classes

if a party consists of a mixed group of plate, cloth, leather users they should only roll for items they have the most use of and what they benefit the most from considering their role just a thought ;)

kadeshar:
Make issue with specific situation. There working based on stats weight. If cloth item had for example stamina and old plate didnt then its possible that cloth item was better even for tank.

PaoloBenigni:
The reason not to limit by specific material is that in some cases using 'softer' material gear is actually advantageous. Like ret pally using leather gear in classic.

liyunfan1223:
I can get the point. For example, a hunter should have priority over a rogue or warrior for ranged weapons. It requires a rule, but the rule may be a little complicated.

PR #1345, opened 2025-05-27
Blessing of Sanctuary improvements

kadeshar:
- Added Blessing of Sanctuary to paladin health buff strategy
- Removed Blessing of Sanctuary from paladin dps buff strategy
- Added health buff strategy as default for protection paladin


kadeshar:
Test: https://www.youtube.com/watch?v=8TeGnLqlWZ4

nl-saw:
Apologies for not elaborating, I haven't had time to test, yet my fear is that it might affect how it distributes blessings when in a group with mixed level chars ( < 60 and 80 pala's for instance). Previous changes here led to undesired behavior in those cases and they kept blessing/overriding each other.

kadeshar:
Yeap i know. Now i tested overlapping other blessings :). And im using standard implementation (used in other blessings) instead of custom.

PR #1363, opened 2025-06-06
Fixed opening trade window while using DBM or Questie addon

kadeshar:


hermensbas:
Without knowing the specifics, even though the add-ons are kinda static phase till a certain extend, its a variable factor from the outer domain we dont control so configration option should be wise.

kadeshar:
Hmmm ofc i can add that to config but still most probably that require debugging to detect that prefix

kadeshar:
I added prefixes to config

Jellypowered:
Including #include <PlayerbotAIConfig.cpp>
prevents compiling at least on Ubuntu. 

The error messages state basically multiple lines already defined. 

I'm not smart enough to fix it properly, but what I did was 


```c++
//#include <PlayerbotAIConfig.cpp>

// Local declaration of this function, to avoid circular dependencies.
template <class T>
void TempLoadListString(std::string const value, T& list)
{
    std::vector<std::string> strings = split(value, ',');
    for (std::vector<std::string>::iterator i = strings.begin(); i != strings.end(); i++)
    {
        std::string const string = *i;
        if (string.empty())
            continue;

        list.push_back(string);
    }
}
``` 

And the other LoadListString to
> TempLoadListString<std::vector<std::string>>(sPlayerbotAIConfig->tradeActionExcludedPrefixes, excludedPrefixes);

hermensbas:
@kadeshar pls correct. I cant do it myself atm since i am on my mobile. Anyway:

playerbotAIconfig.cpp:
```
remove:
tradeActionExcludedPrefixes...

remove:
 std::vector<std::string> excludedPrefixes;
    LoadListString<std::vector<std::string>>(sPlayerbotAIConfig->tradeActionExcludedPrefixes, excludedPrefixes);

add:
LoadListString<std::vector<std::string>>(sConfigMgr->GetOption<std::string>(
    "AiPlayerbot.TradeActionExcludedPrefixes", ""),
   TradeActionExcludedPrefixes);
```

**playerbotAIconfig.h:**
```
remove:
std::string tradeActionExcludedPrefixes

add:
std::vector<std::string> TradeActionExcludedPrefixes;
```

**TradeAction.cpp:**
```
remove:
#include <PlayerbotAIConfig.cpp>

remove:
loadListString logic

add:
#include "PlayerbotAIConfig.h"

modify for loop:
for (const auto& prefix : sPlayerbotAIConfig->TradeActionExcludedPrefixes)
```

Royel-Payne:
Heads up — this PR introduces multiple-definition linker errors due to #include <PlayerbotAIConfig.cpp> in TradeAction.cpp.
A revert was already created as PR #1367, but it's not yet merged into master, so anyone pulling the latest will still hit this error.
Recommend merging the revert soon or addressing the duplication properly.

kadeshar:
Now i will be 2 weeks afk. Strange that checks doesnt detected that issue. On windows compiling normally. After back i will try fix it and test with someone with ubuntu development environment

hermensbas:
The pipeline builds were cancelled for some reason, hence unnoticed. Minor fix though, aside from that i should've seen the impl error on review. But easy fix though. I am abroad myself and dont have my working lappy configured for this project i am afraid.

nl-saw:
> The pipeline builds were cancelled for some reason

Just to elaborate on that; I've made some changes to the build pipelines as they were many times running obsolete jobs. The original PR does indicate it did not compile after the 2nd commit: https://github.com/liyunfan1223/mod-playerbots/pull/1363/commits

PR #1375, opened 2025-06-12
AutoScaleActivity refactoring

NoxMax:
## Current State:
SmartScale observes the max latency of the server. If it is >= SmartScaleDiffLimitfloor, it begins applying scaling of bots with a switch case of 90%, 70%, 50%, 30%, 10%, or 0%, to keep server max latency in check with SmartScaleDiffLimitCeiling.

## Changes Proposed:
Instead of a static case of only 6 possibilities of percentages, this refractor applies a linear percent of scaling, that varies to any percentage of active bots, depending on server max latency relation to floor/ceiling latency.

## Reason For Change:
The issue of the current system might be less obvious for those running only few hundred bots, or if their server machine is highly capable and doesn't even need SmartScale. However, if you are running with several thousand bots, the difference between 50% and 30% of active might be too great.

30% can be easily handled by your server, while 50% pushes max latency over SmartScaleDiffLimitCeiling. My testing has shown that this can result in sudden fluctuations in number of active bots, which on average cause a needless decrease in the percentage of active bots.

By making the scaling linear with a dynamic option of activity percentage, I found that on average there is greatly more active bots, without a sacrifice in server lag.

## Test Parameters:
AiPlayerbot.MinRandomBots = 7500
AiPlayerbot.MaxRandomBots = 7500
AiPlayerbot.botActiveAloneSmartScaleDiffLimitfloor = 150
AiPlayerbot.botActiveAloneSmartScaleDiffLimitCeiling = 650

With switch case scaling:
~1200 - 1500 active bots 
~200ms mean server latency

With linear scaling:
~1800 - 2000 active bots
~200ms mean server latency

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -482,6 +482,15 @@ AiPlayerbot.AutoGearCommandAltBots = 1
 AiPlayerbot.AutoGearQualityLimit = 3
 
 # Equipment item level (not gearscore) limitation for autogear command (0 = no limit)
+# Classic
+# Max iLVL Tier 1 = 66 | Tier 2 = 76 | Tier 2.5 = 81 | Tier 3 = 99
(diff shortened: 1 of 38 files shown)


hermensbas:
I will look into reviewing this when i can for now converted to draft

PR #1389, opened 2025-06-21
Fix various hunter bugs (Amended w/ Nullptr check)

Jellypowered:
Addresses oversight of not having a nullptr check in my PR #1377 

Odd that it never crashed for me, but I also had bot login's disabled so that would make sense, it wasn't a big enough sample size to check for bugs.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -717,8 +717,8 @@ AiPlayerbot.BotActiveAloneForceWhenInGuild = 1
 # The default is 1. When enabled (smart) scales the 'BotActiveAlone' value.
 # (The scaling will be overruled by the BotActiveAloneForceWhen...rules)
 #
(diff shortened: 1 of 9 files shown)


Jellypowered:
3000 bots online, 335 of which are hunters. 11 mins and counting with no crashes.  

Working as intended now. Again, apologies for the snafu. 

Edit: Over an Hour of active playtime
AC> AzerothCore rev. dea3399d8300 2025-06-01 11:00:27 -0500 (Personalization branch) (Unix, RelWithDebInfo, Static)
Connected players: 1. Characters in world: 3010.
Connection peak: 1.
Server uptime: 1 hour(s) 10 minute(s) 34 second(s)
Update time diff: 18ms. Last 500 diffs summary:
|- Mean: 37ms
|- Median: 29ms
|- Percentiles (95, 99, max): 68ms, 77ms, 99ms

No crashing :)

PR #1402, opened 2025-06-28
Warlock Soulstone Fixes/Pet Changes

ThePenguinMan96:
Hello Playerbots community,

I released a PR last night that changed quite a few things, and I got some feedback from the community. Here is a patch to fix some issues that arose.

srt/AiFactory.cpp: Removed the pet-based non-combat strategies entirely. **Why the heck would I do that?!** There is a bug that came to my attention when it comes to adding multiple non-combat strategies from the same header, aka GenericWarlockNonCombatStrategies.h It inherently adds "nc" as a non-combat strategy, and we were asking it to also add a pet that correlates with a spec ("imp" for destruction, "felhunter" for affliction, "felguard" for demonology). When we add ANOTHER non-combat strategy, such as "ss self", or another one of the soulstone non-combat strategies, one of the strategies "falls off"! I implore y'all to recreate the bug for yourself. I received a few discord messages of people saying "when I add the soulstone strategy, my pet strategy goes away". I spent the last 5 hours trying to figure out why, with no such luck. I am a simple thinker, so I ended up migrating all of the summoning logic to the GenericWarlockNonCombatStrategy.cpp, where it checks the player's spec, then summons accordingly. After I made this change, I actually noticed that the Warlock ends up summoning the demon quicker for some reason... Couldn't tell ya why. **TLDR: Changed pet summoning from separate strategies into triggers -> actions based on spec.**

src/strategy/warlock/GenericWarlockNonCombatStrategy.cpp: Added summoning logic based on tab (spec).

src/strategy/warlock/GenericWarlockNonCombatStrategy.h: Gutted pet strategies out of this entirely.

src/strategy/warlock/WarlockActions.cpp: So I was made aware that bots were attempting to soulstone the same healer, or tank. I looked over the code, and the logic for checking for the soulstone aura was already implemented - the issue was, multiple warlocks began using the item at the exact same time. The soulstone resurrection aura was not on the healer when they started, so of course they went to town in all happily casting soulstone on the same damn healer. **The change I am making is to implement something called a "soulstone reservation"**, which means the bot "reserves the right" to soulstone the target for 2 seconds, so other bots can't attempt to soulstone the same target. I also added **line of sight and 30m range requirements**, so they aren't attempting to soulstone something out of range or LoS. I tested this and it seemed to work, but _please_ test this to confirm!

src/strategy/warlock/WarlockAiObjectContext.cpp: Gutted pet strategies out of this entirely.

Last but not least, I wanted to mention something. Tonight I tried to remove the DpsWarlockStrategy.cpp and .h, and clean up their usage in WarlockAiObjectContext, but for some reason the bots performed CONSIDERABLY worse. I have no idea why. On Noth the Plaguebringer, with the DpsWarlockStrategy they were pulling 3.8k-4.7k with ilevel 187 blues. Without, they were pulling 2.9k-3.9k. Same thing on target dummies - they were doing less damage, and they weren't prioritizing the DoTs as well. I will have to look into this further, but for the meantime, the DpsWarlockStrategy.cpp and .h will stay.

P.S. thank all of you for the continued support on this pet project of mine. I highly appreciate the praise, advice, and suggestions.

### DIFF
--- a/src/AiFactory.cpp
+++ b/src/AiFactory.cpp
@@ -594,18 +594,7 @@ void AiFactory::AddDefaultNonCombatStrategies(Player* player, PlayerbotAI* const
                 nonCombatEngine->addStrategy("dps assist", false);
             break;
         case CLASS_WARLOCK:
-            if (tab == WARLOCK_TAB_AFFLICATION)
-            {
-                nonCombatEngine->addStrategiesNoInit("felhunter", nullptr);
-            }
-            else if (tab == WARLOCK_TAB_DEMONOLOGY)
-            {
-                nonCombatEngine->addStrategiesNoInit("felguard", nullptr);
-            }
-            else if (tab == WARLOCK_TAB_DESTRUCTION)
-            {
-                nonCombatEngine->addStrategiesNoInit("imp", nullptr);
-            }
+            // Pet selection is now handled in GenericWarlockNonCombatStrategy triggers
             nonCombatEngine->addStrategiesNoInit("dps assist", nullptr);
             break;
         case CLASS_DEATH_KNIGHT:
--- a/src/strategy/warlock/WarlockActions.cpp
+++ b/src/strategy/warlock/WarlockActions.cpp
@@ -15,6 +15,7 @@
 #include "Playerbots.h"
 #include "ServerFacade.h"
 #include "Unit.h"
+#include "Timer.h"
 
 // Checks if the bot has less than 32 soul shards, and if so, allows casting Drain Soul
 bool CastDrainSoulAction::isUseful() { return AI_VALUE2(uint32, "item count", "soul shard") < 32; }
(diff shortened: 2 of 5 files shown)


src/AiFactory.cpp
```diff
@@ -594,18 +594,7 @@ void AiFactory::AddDefaultNonCombatStrategies(Player* player, PlayerbotAI* const
...
-            }
-            else if (tab == WARLOCK_TAB_DEMONOLOGY)
-            {
-                nonCombatEngine->addStrategiesNoInit("felguard", nullptr);
-            }
-            else if (tab == WARLOCK_TAB_DESTRUCTION)
-            {
-                nonCombatEngine->addStrategiesNoInit("imp", nullptr);
-            }
+            // Pet selection is now handled in GenericWarlockNonCombatStrategy triggers
```
kadeshar, line 597: Its deletion explanation not code comment

kadeshar:
Im not sure that time-based reservation system is good direction. Every warlock should calculate his target based on some condition for example order in group. If all warlock have setted soulstone on tanks then based on warlock order in group there put soulstone in order:
- main tank
- assist tank (in their order)
- tank in group order

About DpsWarlockStrategy that means in there spec specific strategy something is wrong implemented. They should have same or better result.

src/strategy/warlock/WarlockActions.cpp
```diff
@@ -134,9 +135,27 @@ bool UseSoulstoneSelfAction::Execute(Event event)
     return UseItem(items[0], ObjectGuid::Empty, nullptr, bot);
 }
 
+// Reservation map for soulstone targets (GUID -> reservation expiry in ms)
+static std::unordered_map<ObjectGuid, uint32> soulstoneReservations;
```
liyunfan1223, line 141: Just a reminder: `soulstoneReservations` appear to be shared by all bots and can be modified concurrently. Is there a multithreading risk here that could cause a crash?

ThePenguinMan96: Good point - will try and redesign this so they aren't able to all modify soulstoneReservations concurrently. 

I wonder if it's because the reservation is in the action vs the trigger? I will look into this and reupload. It's quite the puzzle getting this soulstone thing smoothed out.

ThePenguinMan96:
> Im not sure that time-based reservation system is good direction. Every warlock should calculate his target based on some condition for example order in group. If all warlock have setted soulstone on tanks then based on warlock order in group there put soulstone in order:
> 
> * main tank
> * assist tank (in their order)
> * tank in group order
> 
> About DpsWarlockStrategy that means in there spec specific strategy something is wrong implemented. They should have same or better result.

I just updated WarlockActions.cpp to check for the main tank before any other tank.

src/strategy/warlock/WarlockActions.cpp
ThePenguinMan96: Reuploaded WarlockActions.cpp with a mutex - lets see what happens.

PR #1405, opened 2025-06-29
Warlock Soulstone fixes/pet changes/DPS strategy cleanup

ThePenguinMan96:
This is a fixed re-upload of my last PR, fixing some issues. Read below for a list of changes:

src/AiFactory.cpp: **Old Info** Removed the pet-based non-combat strategies entirely. Why the heck would I do that?! There is a bug that came to my attention when it comes to adding multiple non-combat strategies from the same header, aka GenericWarlockNonCombatStrategies.h It inherently adds "nc" as a non-combat strategy, and we were asking it to also add a pet that correlates with a spec ("imp" for destruction, "felhunter" for affliction, "felguard" for demonology). When we add ANOTHER non-combat strategy, such as "ss self", or another one of the soulstone non-combat strategies, one of the strategies "falls off"! I implore y'all to recreate the bug for yourself. I received a few discord messages of people saying "when I add the soulstone strategy, my pet strategy goes away". I spent the last 5 hours trying to figure out why, with no such luck. I am a simple thinker, so I ended up migrating all of the summoning logic to the GenericWarlockNonCombatStrategy.cpp, where it checks the player's spec, then summons accordingly. After I made this change, I actually noticed that the Warlock ends up summoning the demon quicker for some reason... Couldn't tell ya why. TLDR: Changed pet summoning from separate strategies into triggers -> actions based on spec. **New Info - also removed an unnecessary comment line.**

src\strategy\warlock\AfflictionWarlockStrategy.h: Added a crucial flag that tells the ai that this is considered a DPS strategy.

src\strategy\warlock\DemonologyWarlockStrategy.h: Added a crucial flag that tells the ai that this is considered a DPS strategy.

src\strategy\warlock\DestructionWarlockStrategy.h: Added a crucial flag that tells the ai that this is considered a DPS strategy.

src\strategy\warlock\DpsWarlockStrategy.cpp and .h: These were completely unused assets as of the last PR, and I now know why the other dps strategies were underperforming by themselves (see WarlockAiObjectContext below, and the header strategies above). Those issues have been fixed, and I tested the warlocks dps without these two files - it's better than before. With itemlevel 200 blues, the warlocks were performing 3.7k-4.2k single target dps on the 80 training dummy.

src\strategy\warlock\GenericWarlockNonCombatStrategy.cpp: Added summoning logic based on tab (spec).

src\strategy\warlock\GenericWarlockNonCombatStrategy.h: Gutted pet strategies out of this entirely.

src\strategy\warlock\WarlockActions.cpp: **Old Info** So I was made aware that bots were attempting to soulstone the same healer, or tank. I looked over the code, and the logic for checking for the soulstone aura was already implemented - the issue was, multiple warlocks began using the item at the exact same time. The soulstone resurrection aura was not on the healer when they started, so of course they went to town in all happily casting soulstone on the same damn healer. The change I am making is to implement something called a "soulstone reservation", which means the bot "reserves the right" to soulstone the target for 2 seconds, so other bots can't attempt to soulstone the same target. I also added line of sight and 30m range requirements, so they aren't attempting to soulstone something out of range or LoS. I tested this and it seemed to work, but please test this to confirm! 
**New Info**It was made to my attention that coding within this file is not safe, due to an issue with the soulstone reservation. See explanation: "std::unordered_map is not thread-safe. If two bots (or threads) call CleanupSoulstoneReservations() or assign to soulstoneReservations[...] at the same time, you get a data race. Data races on C++ containers lead to undefined behavior: crashes, memory corruption, subtle bugs. These actions can and will happen concurrently if bots act in parallel (which is likely in a bot framework). Simultaneous modifications (insert/erase) from different threads. One thread iterating (in CleanupSoulstoneReservations) while another thread inserts/erases. The map can become corrupted, causing random crashes, assertion failures, or even infinite loops." So this aims to be fixed with a mutex.
**"What is a Mutex?** A mutex (short for "mutual exclusion") is a synchronization primitive used in programming to ensure that only one thread at a time can access a particular section of code or data. It acts like a lock: When a thread locks a mutex, other threads must wait until it's unlocked before they can proceed. This prevents data races and ensures data is not corrupted by concurrent access." "By locking a mutex around every access to the map, you guarantee only one thread can read/write it at once." 
I also added logic in the tank soulstone action to first soulstone the main tank, then soulstone other tanks in the group.

src/strategy/warlock/WarlockAiObjectContext.cpp: **Old Info** Gutted pet strategies out of this entirely. **New Info** I found out why the Affliction, Demonology, and Destruction specs by themselves weren't doing too hot - they were registered under the WarlockStrategyFactoryInternal() creators prior, and now they are correctly registered under the  Warlock**Combat**StrategyFactoryInternal(). There is a higher sense of urgency in the bot while in combat to call upon the actions in the combat strategies vs the normal factory.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1954,9 +1954,15 @@ AiPlayerbot.AllowedLogFiles = ""
 
 ####################################################################################################
 # A list of gameObject GUID's that are not allowed for bots to interact with.
-# Example: 176213 = Blood of Heroes
-# Example: 17155 = Defias Gunpowder
-AiPlayerbot.DisallowedGameObjects = 176213,17155
+#
+AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,17155
+#
+# List of GUID's:
+# QuestItems:
+# 176213 = Blood of Heroes, 17155 = Defias Gunpowder, 2656 = Waterlogged Envelope
+# Chests:
+# Large Solid Chest = 74448, Box of Assorted Parts = 19020, Food Crate = 3719, Water Barrel = 3658, Barrel of Milk = 3705, Barrel of sweet Nectar = 3706, Tattered Chest = 105579, Large bettered Chest = 75293
+# Feel free to edit and help to complete.
 #
 ####################################################################################################
 


kadeshar:
Why pet summoning strategies are deleted? They shouldnt. Master should have posibility to override default pet.

ThePenguinMan96:
> Why pet summoning strategies are deleted? They shouldnt. Master should have posibility to override default pet.

The master does have the ability to override by telling the warlock to cast summon succubus, or whatever pet they want. This is functionally the same as before, except without relying on a non-combat strategy. 

There is currently an overflow error with warlocks non-combat strategies if they have a pet strategy and a soulstone strategy at the same time. You can test it yourself by having a pet strategy assigned, and a soulstone strategy assigned, then using a command like summon, or tele into a raid like naxx or an instance like utgarde pinnacle. The pet strategy will "fall off", causing the warlock not to resummon the pet. 

This fixes that problem.

ThePenguinMan96:
> Why pet summoning strategies are deleted? They shouldnt. Master should have posibility to override default pet.

Would you like me to try and fix this so the pet strategies are still in-tact as non-combat strategies? I can give it another go.

kadeshar:
Yes, pet strategies should exists. Cast command its not strategy. Require manual handling every death.
Maybe some additional "pet auto" strategy should be default and handling which pet should be summoned depends on content (pvp, raid, dungeon, leveling) or spec. Im not warlock specialist :(

kadeshar:
I found this issue. I this such "pet auto" resolve also that issue https://github.com/liyunfan1223/mod-playerbots/issues/675

PR #1408, opened 2025-06-29
Added chat command to wipe group

kadeshar:


brighton-chi:
Hah, nice. Definitely will be a useful one for me.

Raz0r1337:
Why should I wipe my own group?
Could not .die or .kill (idk how the exact GM command is) to that thing?

brighton-chi:
> Why should I wipe my own group?
> Could not .die or .kill (idk how the exact GM command is) to that thing?

You can, but you have to kill members off one by one that way. This command is great because you may find yourself in a position with half the raid left and you clearly will not be able to complete the fight--in reality when this happens, GM just tells everybody to go intentionally kill themselves. This is a good equivalent imo. Can save a lot of time.

kadeshar:
Not all players have gm rights even on small servers. And even that killing 25 or 40man group using that is annoying. Especial that sometimes ".die" command is locking and i have to wait some time to wait on next usage but that problem is maybe only because my client or something. Also more people asked about that feature.

Raz0r1337:
That wasn't meant to be a border or anything like that. I just didn't understand why I should kill my group. 😅 But your explanations helped me, thanks. 😉
It's true that not every player has a GM, of course; I hadn't considered that. 😅

But I can't confirm that the .die command is banned or restricted. About a week ago, I was bored on my test server, so I created a .die macro and roamed the world. I spammed the command hundreds of times, and it worked. 😉

Tierisch:
Wouldn't just setting them to passive mid-fight accomplish the same thing as an actual raid leader telling the group to wipe? Why does it need to be forced?

brighton-chi:
> Wouldn't just setting them to passive mid-fight accomplish the same thing as an actual raid leader telling the group to wipe? Why does it need to be forced?

Sure, but then you gotta wait for the boss to kill everybody. I don't see what the issue with introducing this command is--it seems to do nothing harmful for somebody who finds no use in it.

ThePenguinMan96:
Speaking from past raid leading experiences, I feel like this would be a good thing. Over the course of a night of raiding, this could save quite a bit of time, and allow for more attempts - especially when you are pushing difficult content and know a wipe is inevitable (I'm looking at you, Yogg-Saron). I find myself using my .die macro individually on my whole raid pretty often, and this is a pleasant quality of life upgrade.

PR #1409, opened 2025-06-30
Corrections to Vanilla max iLevels

brighton-chi:
I noticed the max iLevels in the config are not correct with respect to Vanilla so this should now have the correct iLevels.


ThePenguinMan96:
Can confirm, these new values are the correct ones!

noisiver:
Well I'm glad you went with actual 3.3.5 item levels and not classic since that doesn't match.

PR #1410, opened 2025-06-30
Fixed bug with not respecting InstantFlightPaths config

kadeshar:
Resolve: https://github.com/liyunfan1223/mod-playerbots/issues/1407

### DIFF
--- a/src/strategy/actions/TaxiAction.cpp
+++ b/src/strategy/actions/TaxiAction.cpp
@@ -73,7 +73,7 @@ bool TaxiAction::Execute(Event event)
                 {
                     if (Creature* npcPtr = ObjectAccessor::GetCreature(*bot, npcGuid))
                         if (!movement.taxiNodes.empty())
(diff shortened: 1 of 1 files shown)


HennyWilly:
There is one combination that doesn't work as intended:
If "InstantFlightPaths" is set to "2" but instant flight is toggled off at the flight master, the bots teleport anyway.

The rest looks good.

kadeshar:
> There is one combination that doesn't work as intended: If "InstantFlightPaths" is set to "2" but instant flight is toggled off at the flight master, the bots teleport anyway.
> 
> The rest looks good.

I think that toggle is personal not server-wide and bots dont toggle that.

HennyWilly:
Makes sense. If there's no trivial way to check the master's toggle state, then I'm fine with the current solution.

kadeshar:
This PR only solves config problem. It not implements master toggle state check.

PR #1413, opened 2025-07-02
Fix and harden PlayerbotAI::GetEquipGearScore calculation

Wishmaster117:
What changed

uint32 PlayerbotAI::GetEquipGearScore(Player* player)
was completely rewritten to address two long-standing issues and to make the implementation more robust.

Old behaviour	New behaviour
Double-counts item-level when a true two-handed weapon is equipped in the main-hand slot (iLvl was added once for the main-hand and once again for the off-hand).	Off-hand slot is ignored when a 2-H is equipped and the character does not have Titan Grip.
Uses a fixed divisor EQUIPMENT_SLOT_END - 2, so empty equipment slots still increase the average denominator → under-estimated gear scores and a potential divide-by-zero if every counted slot was empty.	Only increments count for slots that actually contain a valid item; returns 0 if nothing is equipped, eliminating the divide-by-zero case.

Additional improvements

Explicitly excludes Body and Tabard slots instead of relying on a magic constant.

Detects two-handed weapons via proto->InventoryType == INVTYPE_2HWEAPON.

Properly supports Warriors with Titan Grip (SPELL_TITAN_GRIP) who may wield two two-handed weapons.

Returns sum / count only when count > 0, otherwise 0.

Readability: early continue statements make the logic clearer and O(19) performance cost is negligible.

Why this matters
Produces an accurate average item-level for player-bots, leading to better difficulty scaling and gear-based logic elsewhere in the module.

Removes a silent bug that could inflate scores for two-hand users or crash when no valid gear was equipped.

Keeps behaviour consistent across all classes and special cases without introducing new dependencies.

No new headers or external dependencies were introduced; only the constant SPELL_TITAN_GRIP must resolve to the passive Titan Grip aura ID (46917 in WotLK DB).

This PR is self-contained and does not affect any other subsystem.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -4430,9 +4430,49 @@ void PlayerbotAI::RemoveShapeshift()
     // RemoveAura("tree of life");
 }
 
+// NOTE : function rewritten – we ignore Body & Tabard, and we do **not**
+// double–count the main-hand when the character is using a 2-hand weapon
+// (unless they have Titan Grip).
+
+uint32 PlayerbotAI::GetEquipGearScore(Player* player)
+{
+    uint32 sum   = 0;      // total of all considered item-levels
+    uint8  count = 0;      // number of slots that really contributed
+    bool   twoHandMain = false;
+
+    for (uint8 slot = EQUIPMENT_SLOT_START; slot < EQUIPMENT_SLOT_END; ++slot)
+    {
+        if (slot == EQUIPMENT_SLOT_BODY || slot == EQUIPMENT_SLOT_TABARD)
+            continue;                                   // these slots never count
+
+        Item* item = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot);
+        if (!item)
+            continue;
+
+        ItemTemplate const* proto = item->GetTemplate();
+        if (!proto)
+            continue;
+
+        if (slot == EQUIPMENT_SLOT_MAINHAND)
+            twoHandMain = (proto->InventoryType == INVTYPE_2HWEAPON);
+
+        // If a genuine 2-H weapon is equipped in main-hand **and**
+        // the character does *not* have Titan Grip, we skip the off-hand slot.
+        if (slot == EQUIPMENT_SLOT_OFFHAND &&
+            twoHandMain &&
+            !player->HasAura(SPELL_TITAN_GRIP))
+            continue;
+
+        sum   += proto->ItemLevel;
+        ++count;
+    }
+
+    return count ? sum / count : 0;
+}
+
 // NOTE : function rewritten as flags "withBags" and "withBank" not used, and _fillGearScoreData sometimes attribute
 // one-hand/2H Weapon in wrong slots 
-uint32 PlayerbotAI::GetEquipGearScore(Player* player)
+/*uint32 PlayerbotAI::GetEquipGearScore(Player* player)
 {
     // This function aims to calculate the equipped gear score
  
@@ -4458,7 +4498,7 @@ uint32 PlayerbotAI::GetEquipGearScore(Player* player)
 
     uint32 gs = uint32(sum / count);
     return gs;
-}
+}*/
 
 /*uint32 PlayerbotAI::GetEquipGearScore(Player* player, bool withBags, bool withBank)
 {


kadeshar:
Great, i though about that :)
Could you cover also proper counting when both hands are empty? https://wowpedia.fandom.com/wiki/API_GetAverageItemLevel#Details

Wishmaster117:
> Great, i though about that :) Could you cover also proper counting when both hands are empty? https://wowpedia.fandom.com/wiki/API_GetAverageItemLevel#Details

The code already handle cases where the player has no weapon equipped:

uint32 sum   = 0;
uint8  count = 0;
bool   twoHandMain = false;
…
return count ? sum / count : 0;   // returns 0 if nothing was counted

Main-hand empty

twoHandMain remains false.

The test that might skip the off-hand slot

slot == EQUIPMENT_SLOT_OFFHAND &&
twoHandMain &&
!player->HasAura(SPELL_TITAN_GRIP)

never fires.

The off-hand is included only if an item is actually equipped there.


Both main-hand and off-hand empty

Neither slot increases sum or count.

Since count is 0, the ternary count ? sum / count : 0 returns 0.

That is the expected behavior: a character with no relevant gear has a gear score of 0.

No gear at all (e.g. a brand-new character)

Likewise, count stays 0, so the function safely returns 0 and never divides by zero.

In short, whether one weapon is missing or both are missing, the function:

1. Never triggers an exception or a divide-by-zero.

2. Returns a logical gear score—either the average of the items actually equipped, or 0 if nothing should be counted.

No additional handling is required for these cases.

kadeshar:
@Wishmaster117 
Lines

Item* item = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot);
        if (!item)
            continue;
            
dont skip counting for every empty slot (should only skip when in certain situations)?

You cancelled this PR. You will make new one?

Wishmaster117:
> @Wishmaster117 Lines
> 
> Item* item = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot); if (!item) continue;
> 
> dont skip counting for every empty slot (should only skip when in certain situations)?
> 
> You cancelled this PR. You will make new one?

My bad i have accidently deleted my correction branch, have to make another PR.

To answer your question:

Here’s the idea behind those two lines:

```
Item* item = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot);
if (!item)
    continue;           // skip: nothing equipped in that slot
```

It won’t stop the whole process—continue only skips that one iteration of the for-loop.
Execution immediately jumps to the next slot index and keeps evaluating the rest of the equipment slots. So:

Empty slot found → we ignore it for the average.

Loop continues → the following slots (rings, trinkets, etc.) are still examined and, if they contain an item, their ItemLevel is added to the total.

The only way the counting could end early is if the loop condition itself fails (i.e. slot >= EQUIPMENT_SLOT_END), which is independent of whether any slot was empty. In short, the continue does not break or terminate the loop; it merely moves on to the next slot.

Why we do skip every empty slot
“Equipped‐only” average.
GearScore (in most add-ons and the original Blizzard formula) is meant to reflect the quality of the items the character is actually wearing, not to penalise missing pieces.
If there is no item, there is no ItemLevel to add to the sum, and we also don’t want that slot to enlarge the divisor.

Consistent with the count logic.
We increment count only when we successfully add that slot’s ItemLevel:

```
sum   += proto->ItemLevel;
++count;
```
Letting an empty slot fall through would inflate count while sum stays the same, artificially lowering the average. Skipping keeps the ratio coherent.

Edge cases stay safe.

A fresh character with no gear at all yields count == 0, and we return 0 via

`return count ? sum / count : 0;.`

Partially geared characters get an average of the pieces they do wear.

When do we not skip a slot?
The only intentional exception is the off-hand slot when a genuine 2-hand weapon is in the main hand and the player lacks Titan Grip. In that case we also continue, but the rule is driven by gameplay logic (a 2-H weapon already accounts for both hands).

Current implementation follows the “equipped-only” convention.

kadeshar:
I linked official blizzard formula and missing gear is also counting https://wowpedia.fandom.com/wiki/API_GetAverageItemLevel#Details. 
Same with elvui 
obraz

Gearscore works different and shouldnt be use as example for this method. Its counting sum of all statistics. Doesnt care about items amount.

EDIT:
Proof:
Equipped one item 226 ilvl. Average ilvl is 14 because 226/16=14.125
obraz
obraz

Wishmaster117:
> Gearscore works different and shouldnt be use as example for this method. Its counting sum of all statistics. Doesnt care about items amount.

Blizzard’s official “average item level” API (GetAverageItemLevel) always divides by a fixed slot-count (16 or 17), so every empty slot implicitly contributes a “0 iLvl.” The Lua snippet you pasted mirrors that approach: it starts with local items = 16, subtracts one when a true 2-hander removes the off-hand, and divides by that fixed count even if several slots are blank.

Our C++ helper for bots deliberately chose a different rule: it ignores empty slots when computing the average, because for balance-logic we don’t want a half-naked bot to be judged harsher than one wearing low-level greys. Both approaches are technically fine, they just serve different purposes:

Aspect -> Denominator
Blizzard API / Lua (your screenshot) : Always 16 (or 17 with Titan Grip)
My PlayerbotAI::GetEquipGearScore : Only the slots that actually have items

Aspect -> Empty slot effect
Blizzard API / Lua (your screenshot) : Counts as iLvl 0 → lowers average
My PlayerbotAI::GetEquipGearScore : Ignored → average reflects what’s equipped

Aspect -> Goal
Blizzard API / Lua (your screenshot) : Show true combat readiness of a player (missing gear hurts)
My PlayerbotAI::GetEquipGearScore : Estimate how strong this bot already is, to decide upgrades/encounters

If you’d rather mimic Blizzard exactly, you have to revert to a fixed-slot divisor (16/17) and leave the continue alone so that empty slots still enter the average as 0.

Wishmaster117:
> I linked official blizzard formula and missing gear is also counting 
> 
> Gearscore works different and shouldnt be use as example for this method. Its counting sum of all statistics. Doesnt care about items amount.
> 
> EDIT: Proof: Equipped one item 226 ilvl. Average ilvl is 14 because 226/16=14.125 obraz, we can add a second
function that follows the 16/17-slot formula verbatim.

Why my bot helper doesn’t follow that rule:

Bot logic (upgrade & encounter decisions)	-> Judge “quality of the items that exist”, not how many are missing	-> Dynamic: count only filled slots

Example:
A fresh level-80 bot wearing just two 200-iLvl heirlooms

Blizzard AIL -> (200 + 200) / 16 ≈ 25	 -> Looks terrible—even worse than a level-60’s gear
Bot helper-> (200 + 200) / 2 = 200	 -> Says “the gear it has is 200-quality, but many slots are still empty”—that’s exactly what the upgrade routine needs to know.

With my rule the gearing AI says “fill empty slots with greens/blues” instead of “replace heirlooms because average iLvl is low.”

Feel free to adjust the wording, but that’s the crux: different metric for a different task, and both can coexist if needed.

kadeshar:
Your goal is wrong. Bots decide about their upgrades based on stat weights not this method (you can check action EquipUpgradeAction).
This method is used for command "who" which only determine bot combat power and more of that is used in addon multibot

Wishmaster117:
> Your goal is wrong. Bots decide about their upgrades based on stat weights not this method (you can check action EquipUpgradeAction).
> This method is used for command "who" which only determine bot combat power and more of that is used in addon multibot

You’re right – I misunderstood the call-site.
GetEquipGearScore is used purely for displaying the bot’s combat power (who list + MultiBot), not for the upgrade algorithm.
I’ll revert to the official Blizzard AIL formula:
sum of equipped item-levels / 16 (or 17 with Titan Grip), so missing gear lowers the shown average just like in the client API.
Thanks for catching that!


```
// Official‐style Average Item Level for display / Who command
uint32 PlayerbotAI::GetEquipGearScore(Player* player)
{
    const uint8 kTotalSlots = 16;                      // default divisor
    const bool  hasTitanGrip =
        player->HasAura(SPELL_TITAN_GRIP);

    uint32 sum = 0;
    uint8  divisor = kTotalSlots;

    bool mainHas2H = false;

    for (uint8 slot = EQUIPMENT_SLOT_START; slot < EQUIPMENT_SLOT_END; ++slot)
    {
        if (slot == EQUIPMENT_SLOT_BODY || slot == EQUIPMENT_SLOT_TABARD)
            continue;                                  // never counted in AIL

        Item* item = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot);
        if (!item)
            continue;                                  // empty slot ⇒ contributes 0

        ItemTemplate const* proto = item->GetTemplate();
        if (!proto)
            continue;

        if (slot == EQUIPMENT_SLOT_MAINHAND)
            mainHas2H = (proto->InventoryType == INVTYPE_2HWEAPON);

        // Off-hand with real 2-H in main and no Titan Grip still counts as
        // an *empty* slot, so we keep the divisor unchanged and add 0 here.
        if (slot == EQUIPMENT_SLOT_OFFHAND &&
            mainHas2H && !hasTitanGrip)
            continue;

        sum += proto->ItemLevel;
    }

    // If Titan Grip is active and two genuine 2-H weapons are equipped,
    // we use 17 slots just like the client API.
    if (hasTitanGrip && mainHas2H)
        ++divisor;

    return divisor ? sum / divisor : 0;
}
```

Now even with zero items equipped we still loop through all slots, leave sum == 0, and divide by divisor (16) → returns 0.
So an empty inventory is handled correctly.

What do you think?

Wishmaster117:
I will make another one

Issue #534, opened 2024-09-16
Bots hover way behind the team in battlegrounds like AV - AV Reputation Trinket Issue


Dohtt:
**Describe the bug**
5-10 bots per game decide not to participate and try to defend, while nothing has been taken.

**Commit hash**
bfbc451

**To Reproduce**
1. Join AV
2. After the first exchange bots will run all the way back to the main base, while nothing is taken on that side of the map

**Expected behavior**
Bots should not try to defend objectives that arn't taken by the enemy team, they should move forward and take more objectives or defend objectives they're currently taken OR try to defend objectives that the enemy is taking from them.

**Screenshots**
![image](https://github.com/user-attachments/assets/8a0f27be-1d58-4210-8bb1-8e8c2f72917a)

Dohtt:
Solved

This was caused by the AV reputation trinket that some bots had equipped. They instantly use trinkets on conflict, so they teleported back there and got confused.

Dreathean:
If anything, may want to keep this open as this should still probably be addressed in some way

Dreathean:
@Dohtt have you still noticed them using the trinket causing this issue? I know some improvements to trinket usage have been made, but not sure if this in particular has been resolved yet or not

Dohtt:
Initial test shows them no longer using the trinket. Will try it out more, but seems to be functioning as intended.

PR #1418, opened 2025-07-03
Bots Bg Tactics rewrite (WSG, AB, AV, EYE)

Boxhead78:
This rewrites the Battleground-Tactics for bots completely in the following Battlegrounds
- Warsong Gulch
- Arathi Basin
- Alterac Valley
- Eye of the Storm

Isle of Conquest isn't touched but also needs a lot of work.

When I first played Warsong Gulch I noticed many battleground-tactic bugs and how inefficient they were - especially in Warsong Gulch and Alterac Valley. This PR aims to resolve that, additionally with Arathi Basin and Eye.
I'm going to try to list the changes as best as possible:

**BattleGroundTactics.cpp:**
- WSG: Got rid of Rym's very old waypoint system and implemented WSG with the new waypoint system.
- EYE: Made sure bots stay at their objective until they've captured it.
- Bots are now able to camp at the enemy's graveyard if they themselves are in lead and depending on which battleground they play.
- Improved how bots decide to target flag carriers or when to target nearby players/bots.
- Fixed bots often getting stuck on Alterac Valley.
- Added several strategies for the bots to choose from in WSG, AB, AV and EYE (all bots will have a random strategy on battleground start). Example: Those different strategies allow to have fast matches on AV but also the slow zerging matches.
- Changed waypoints for bots in WSG and AV
- Other changes which I can't think of right now and of course many bug fixes

**BattlegroundStrategy.cpp**
- Refactored it and added ACTION_BG
- Bots bg objective now recalculates on:
1.) Death
2.) Depending on Battleground after a timer expired (but only if a bot's current objective is interrupted for example by an enemy)
This change is supposed to make the bots react better to different situations which can occur on a battleground, for example the enemy takes a flag or a point gets captured.

**Known Issues**
- EYE: Bots instant capture flag.
- WSG: Sometimes alliance bots get stuck at alliance graveyard.

**Notes**
- I've developed this a few months ago, invested quite some weeks into it and also did a lot of testing. I run this on my 24/7 server, allowing the bots to queue for battlegrounds without a player and didn't receive any crashes or issues while doing it.
- My core and mod-playerbots have many other changes because I'm retroporting lots of content from Cataclysm which **shouldn't** have impact on this PR but due to the complexity of a wow-server this can't be excluded. This PR of course only comes with changes to the bots.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -29,6 +29,7 @@
 #include "ScriptMgr.h"
 #include "cs_playerbots.h"
 #include "cmath"
+#include "BattleGroundTactics.h"
 
 class PlayerbotsDatabaseScript : public DatabaseScript
 {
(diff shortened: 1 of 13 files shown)


brighton-chi:
This sounds great. By the following:

"Added several strategies for the bots to choose from in WSG, AB, AV and EYE (all bots will have a random strategy on battleground start). Example: Those different strategies allow to have fast matches on AV but also the slow zerging matches."

Is this randomized on a whole team or an individual bot basis? And can they be commanded to change strategies?

kadeshar:
@Boxhead78  Wow, that is huge contribution for playerbots pvp. Thanks a lot. Personally i cant even think where to start to check that PR :) I will mark that PR to test. Maybe someone will try new pvp :)

Boxhead78:
> This sounds great. By the following:
> 
> "Added several strategies for the bots to choose from in WSG, AB, AV and EYE (all bots will have a random strategy on battleground start). Example: Those different strategies allow to have fast matches on AV but also the slow zerging matches."
> 
> Is this randomized on a whole team or an individual bot basis? And can they be commanded to change strategies?

This is randomized on a whole team until the battleground is over, each new battleground the strategies are randomly selected for the current battleground.

I've added no command for it. Certainly a command could be implemented in the future.

kadeshar:
Do you think that resolving that issue https://github.com/liyunfan1223/mod-playerbots/issues/534 ?

Boxhead78:
> Do you think that resolving that issue #534 ?

Yes! Now bots should always either a) try to defend objectives that are being taken from them, or b) attack enemy bases.

In Alterac Valley: If both sides happen to get the defensive strategy, they will automatically switch to the aggressive strategy to avoid having the majority of bots just sitting in their base.

Another note for Alterac Valley: Bots currently don't or just on accident take the mines. Their logic for the mines is currently disabled because they would get stuck due to pathfinding issues (especially on the horde side).

Dreathean:
Perhaps we can also close out https://github.com/liyunfan1223/mod-playerbots/issues/682 with this as well?

PR #1420, opened 2025-07-03
Revert GetEquipGearScore to Blizzard’s fixed‐slot average item level formula

Wishmaster117:
Description:
This patch restores the “Average Item Level” calculation in PlayerbotAI::GetEquipGearScore to match Blizzard’s official API, using a fixed divisor of 16 slots (17 when Titan Grip is active). Previously, we experimented with a dynamic slot count that only divided by truly equipped slots, but that breaks the expected behavior in the .who command and the MultiBot addon (which both rely on Blizzard’s definition: missing gear should lower the displayed average).

Changes:

Use a constant divisor of 16 (body and tabard excluded).

Skip off-hand slot when a genuine 2H weapon is in main-hand and Titan Grip is not present (Blizzard uses 16 slots in that case).

Increment divisor to 17 only if Titan Grip is active and a 2H weapon is equipped in main hand, matching the client.

Ensure empty slots contribute zero rather than prematurely ending the loop.

Why:

The .who command and the MultiBot addon display combat power based on Blizzard’s Average Item Level, which always divides by a fixed slot count, not by the number of filled slots.

Keeping this function aligned with the API prevents misleadingly high “ilvl” readings for bots with missing gear.

Thanks:
Huge thanks to kadeshar for catching this mistake and clarifying that my dynamic‐slot approach was incorrect for display purposes!

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -4434,35 +4434,39 @@ void PlayerbotAI::RemoveShapeshift()
 // https://wowpedia.fandom.com/wiki/API_GetAverageItemLevel
 uint32 PlayerbotAI::GetEquipGearScore(Player* player)
 {
-    constexpr uint8 BASE_SLOTS = 17;          // everything except Body & Tabard
-    uint32  sumLevel   = 0;
-    uint8   divisor    = BASE_SLOTS;
+    constexpr uint8 TOTAL_SLOTS = 17;                      // every slot except Body & Tabard
+    uint32 sumLevel = 0;
 
-    Item* main  = player->GetItemByPos(INVENTORY_SLOT_BAG_0, EQUIPMENT_SLOT_MAINHAND);
-    Item* off   = player->GetItemByPos(INVENTORY_SLOT_BAG_0, EQUIPMENT_SLOT_OFFHAND);
+    /* ---------- 0.  Detect “ignore off-hand” situations --------- */
+    Item* main = player->GetItemByPos(INVENTORY_SLOT_BAG_0, EQUIPMENT_SLOT_MAINHAND);
+    Item* off  = player->GetItemByPos(INVENTORY_SLOT_BAG_0, EQUIPMENT_SLOT_OFFHAND);
 
-    bool twoHandMain = false;
+    bool ignoreOffhand = false;                           // true → divisor = 16
     if (main)
-        twoHandMain = (main->GetTemplate()->InventoryType == INVTYPE_2HWEAPON);
+    {
+        bool twoHand = (main->GetTemplate()->InventoryType == INVTYPE_2HWEAPON);
+        if (twoHand && !player->HasAura(SPELL_TITAN_GRIP))
+            ignoreOffhand = true;                         // classic 2-hander
+    }
+    else if (!off)                                        // both hands empty
+        ignoreOffhand = true;
 
-    /* ---------- 1.  Add up item-levels ---------- */
+    /* ---------- 1.  Sum up item-levels -------------------------- */
     for (uint8 slot = EQUIPMENT_SLOT_START; slot < EQUIPMENT_SLOT_END; ++slot)
     {
         if (slot == EQUIPMENT_SLOT_BODY || slot == EQUIPMENT_SLOT_TABARD)
-            continue;
+            continue;                                     // Blizzard never counts these
+
+        if (ignoreOffhand && slot == EQUIPMENT_SLOT_OFFHAND)
+            continue;                                     // skip off-hand in 2-H case
 
         if (Item* it = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot))
-            sumLevel += it->GetTemplate()->ItemLevel;
+            sumLevel += it->GetTemplate()->ItemLevel;     // missing items add 0
     }
 
-    /* ---------- 2.  Adjust divisor -------------- */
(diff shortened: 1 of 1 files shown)


src/PlayerbotAI.cpp
```diff
@@ -4430,9 +4430,54 @@ void PlayerbotAI::RemoveShapeshift()
...
+            mainHas2H && !hasTitanGrip)
+            continue;
+
+        sum += proto->ItemLevel;
+    }
+
+    // If Titan Grip is active and two genuine 2-H weapons are equipped,
+    // we use 17 slots just like the client API.
+    if (hasTitanGrip && mainHas2H)
+        ++divisor;
```
kadeshar: This doesnt solve all situations.

17 with both hands holding items (in your code that is 16)
17 with a single one-hand item (or a single two-handed item with Titan's Grip) (in your code for single one-hand is 16)
16 with a two-handed item equipped (and no Titan's Grip)
16 with both hands empty

Wishmaster117: Hi 👋
First of all, huge thanks to @kadeshar for catching the previous mistake and for the detailed review — it helped a lot to nail down the exact edge-cases. 🙏

Here is the re-implements PlayerbotAI::GetEquipGearScore so that it strictly follows Blizzard’s GetAverageItemLevel formula.

Handles every main-/off-hand combination correctly, choosing the same divisor (16 or 17) that the live client uses.


Two 1-H weapons or Titan Grip + 2-H	- slots counted= 17	Divisor= 17	 Example (ilvl = 226): 226 ÷ 17 ≈ 13.29
One 1-H weapon, off-hand empty - slots counted= 17	Divisor= 17	 Example (ilvl = 226): 226 ÷ 17 ≈ 13.29
True 2-H weapon, no TG - slots counted= 16	Divisor= 16	(ilvl = 226): 226 ÷ 16 ≈ 14.12
Both hands empty - slots counted= 16	Divisor= 16

No more double-counting the main-hand when a 2-H weapon is equipped.

Skips BODY & TABARD slots, as the API does.

Returns 0 gracefully if every equipment slot is empty.

Why it matters
GetEquipGearScore is consumed by:

* the .who command (combat-power display),

* the Multibot addon,

* various heuristics that compare bot strength.

With the old logic, bots running odd gear setups (e.g. one empty hand) were undervalued, causing misleading power estimates and potentially wrong upgrade decisions downstream.

Testing
Single 2-H (ilvl 226) – client shows Average ilvl ≈ 14, new function returns 14.

Dual 1-H (ilvl 200 + 200) – client shows ≈ 23, function returns 23.

No weapons – returns 0.

Verified on GCC 11, Clang 14 and MSVC 19. No warnings.

The loop now:

Ignores BODY & TABARD.

Detects a 2-H main-hand.

Skips off-hand only in that precise case (and only if Titan Grip aura is absent).

Computes sum / count once at the end.

No other logic or public interface changes.

src/PlayerbotAI.cpp
```diff
@@ -4430,9 +4430,44 @@ void PlayerbotAI::RemoveShapeshift()
...
+        twoHandMain = (main->GetTemplate()->InventoryType == INVTYPE_2HWEAPON);
+
+    /* ---------- 1.  Add up item-levels ---------- */
+    for (uint8 slot = EQUIPMENT_SLOT_START; slot < EQUIPMENT_SLOT_END; ++slot)
+    {
+        if (slot == EQUIPMENT_SLOT_BODY || slot == EQUIPMENT_SLOT_TABARD)
+            continue;
+
+        if (Item* it = player->GetItemByPos(INVENTORY_SLOT_BAG_0, slot))
+            sumLevel += it->GetTemplate()->ItemLevel;
```
kadeshar: You dont have to protect situation when it is nullptr?

src/PlayerbotAI.cpp
```diff
@@ -4430,9 +4430,44 @@ void PlayerbotAI::RemoveShapeshift()
...
+    }
+
+    /* ---------- 2.  Adjust divisor -------------- */
+    if ((twoHandMain && !player->HasAura(SPELL_TITAN_GRIP)) ||  // real 2-H weapon
+        (!main && !off))                                        // both hands empty
+    {
+        divisor = BASE_SLOTS - 1;           // use 16
+    }
+
+    return divisor ? sumLevel / divisor : 0;
```
kadeshar: divisor ? 
that is unneccesary. Its always is 17 or 16

kadeshar (changes requested):

src/PlayerbotAI.cpp
Wishmaster117: We still need a null-check when we dereference the pointer inside the loop—otherwise we’d crash on an empty slot. The check is now the single if (Item* it = …) shown above (no extra nesting).

src/PlayerbotAI.cpp
Wishmaster117: Agreed. With the logic above divisor is guaranteed to be 16 or 17, so we removed the ternary and return sumLevel / divisor directly.

Wishmaster117: Two-hander, no Titan Grip	-> 16	-> 226 / 16 ≈ 14.1
Both hands empty	-> 16	-> 0
One-hand or two-hand + Titan Grip	-> 17	-> 226 / 17 ≈ 13.3
Both hands filled	 -> 17	-> normal calculation

kadeshar (approved):

PR #1422, opened 2025-07-04
Warlock Ranged Designation/DPS Strategy Cleanup

ThePenguinMan96:
Hello community,

This PR focuses on 4 things:

Recognizing the Warlock as a "ranged" bot, so they will follow ranged commands and strategies, in GenericWarlockStrategy.h: 

uint32 GetType() const override { return CombatStrategy::GetType() | **STRATEGY_TYPE_RANGED** | STRATEGY_TYPE_DPS; }

Cleanup/deletion of the DpsWarlockStrategy.cpp and .h (no longer used or referenced anywhere)

Fixes soulstone logic so multiple Warlocks don't soulstone the same target, and don't try to soulstone a target that is too far away or out of line of sight (WarlockActions.cpp):

float distance = sServerFacade->GetDistance2d(bot, member);
                    if (**distance < 30.0f** && bot->**IsWithinLOSInMap(member)**)      

**Moved the summoning strategies from the "Buff" area of the WarlockAiObjectContext to the "general" strategies, and it works now with the soulstones! This was the main issue - only 1 noncombat buff strategy can be active at a time, and that is why I was having the error of the pet strategy disappearing - because there was already a soulstone buff strategy active. I reverted the summoning logic to the toggleable non-combat strategies "imp", "voidwalker", "succubus", "felhunter",  and "felguard".**


kadeshar:
Still that PR introduces pet type strategy deletion

ThePenguinMan96:
> Still that PR introduces pet type strategy deletion

Changed it so now the pet strategies are working with the soulstone strategies! They were accidently under the buff strategies, and there can only be one active buff non-combat strategy. Moved the pet strategies up to the warlockfactoryinternal strategies and it works perfectly now.

PR #1426, opened 2025-07-06
Fix: DeleteRandomBotAccounts sometimes leaves orphans. Also fix accumulating orphan pet data in DB

NoxMax:
## Current State:
This PR fixes two issues. The main issue is DeleteRandomBotAccounts sometimes (but not always) leaves behind orphaned data. This creates a major issue where on following server restart and accounts creation, all the necessary accounts are created, but the server mistakenly thinks the accounts already have characters.

The culprit seems to be 
`CharacterDatabase.Execute("DELETE FROM characters WHERE account IN (SELECT id FROM " + loginDBName + ".account WHERE username LIKE '{}%%')",`
not having enough time to execute fully, before dependant executions start.

Second issue: If you check pet_aura, pet_spell, pet_spell_cooldown in your DB, there's a good chance that you will find far too many rows for bot pets, than bot pets actually exist. A while back I had noticed that my pet_spell table had 2.7 MILLION entries after running my server on and off for several months. The humongous data was caused by orphan data from bot pets (summons of all kinds) that don't even exist in character_pet anymore.

## Proposed Changes:
For DeleteRandomBotAccounts, the fix is making sure characters have have enough time to be deleting before moving forward, and also all the data changes in the tables are committed and flushed before stopping world server.

For the orphan pet data, we can just clean them up once per server start, for pets that are no longer in the DB.

## Test parameters:
The bug with DeleteRandomBotAccounts appeared about 1 out of 4 times when testing with the original code. After fix, I tested **20 consecutive times** with DeleteRandomBotAccounts rinse and repeats (yes it took a while to do and check each) and not a single one of them had less than expected number of bots.

MaxRandomBots = 500
RandomBotAccountCount = 0
AddClassAccountPoolSize = 10

Expected results:
61 accounts (the code currently has a +1 account offset)
610 bots
Reduction in number of entries in the tables pet_aura, pet_spell, pet_spell_cooldown

---
Everybody hates orphans, data orphans...

### DIFF
--- a/src/RandomPlayerbotFactory.cpp
+++ b/src/RandomPlayerbotFactory.cpp
@@ -486,6 +486,13 @@ void RandomPlayerbotFactory::CreateRandomBots()
         CharacterDatabase.Execute("DELETE FROM characters WHERE account IN (SELECT id FROM " + loginDBName + ".account WHERE username LIKE '{}%%')",
             sPlayerbotAIConfig->randomBotAccountPrefix.c_str());
 
+        // Wait for the characters to be deleted before proceeding to dependent deletes
+        while (CharacterDatabase.QueueSize())
+        {
+            std::this_thread::sleep_for(1s);
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));    // Extra 100ms fixed delay for safety.
+
         // Clean up orphaned entries in playerbots_guild_tasks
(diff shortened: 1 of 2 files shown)


kadeshar:
Your issue :D https://github.com/liyunfan1223/mod-playerbots/issues/1336

NoxMax:
Thanks for linking the issue @kadeshar. I genuinely forgot that I posted that issue back in May. I thought I only discussed it on Discord 😂

PR #1431, opened 2025-07-08
Minor Battleground tactics improvements

Boxhead78:
- Fixes bots often ignoring contested nodes in Arathi Basin
- Replaces magic number with VMAP_INVALID_HEIGHT_VALUE

### DIFF
--- a/src/strategy/actions/BattleGroundTactics.cpp
+++ b/src/strategy/actions/BattleGroundTactics.cpp
@@ -22,6 +22,7 @@
 #include "BattlegroundSA.h"
 #include "BattlegroundWS.h"
 #include "Event.h"
+#include "IVMapMgr.h"
 #include "Playerbots.h"
 #include "PositionValue.h"
 #include "PvpTriggers.h"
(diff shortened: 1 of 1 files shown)


brighton-chi:
I did one WSG earlier, and the entire team camped the opposing graveyard, included the flag bearer. It seems to me that there should be some sort of override so that the team will actually cap the flag in that situation.

Boxhead78:
> I did one WSG earlier, and the entire team camped the opposing graveyard, included the flag bearer. It seems to me that there should be some sort of override so that the team will actually cap the flag in that situation.

The graveyard camping will happen if one team is far in the lead while the other team hasn't captured a flag. (Example score 2 : 0) and they stop camp if the other team captures a flag (Example score: 2 : 1).

My ideas would be to add the following:
Additionally to the current logic have the bots camp only a) if 5 minutes of the bg remains and b) if camping is active, have a chance for the bots to still capture the flag / follow they're normal strategy.

1. Can you confirm to me that the entire team only camped when they were far in lead?
2. Do you have changed settings in battlegrounds in the worldserver config?

brighton-chi:
Yep, my team was up 2-0 at the time.

In the worldserver config, I have changed the wait time for pre mades to make it shorter. I also restored some stuff about AB and AV to vanilla values (like no resource limit for AV) but haven't messed with WSG.

PR #1434, opened 2025-07-11
Fix: Properly track RNDbot and AddClass accounts, and login faction balance issue

NoxMax:
## Current State: 
Account Assigment: The current implementation uses a positional-based system to differentiate between RNDbot accounts (the ones that login normally) and AddClass bot accounts. During server initialization, accounts are created based on RandomBotAccountCount configuration or automatic calculation. AddClass accounts are determined by their position in the account list - specifically, the last N accounts where N = AddClassAccountPoolSize. This positional logic is used throughout the codebase when determining if a bot belongs to AddClass or RNDbot category.

Bot Login: You ask for 1000 bots, and have set in the conf the default faction ratio of 50/50. So the system goes "you need a thousand, equal ratio, here's a 100 +1 offset accounts"
But while those 101 accounts have 1010 bots, they probably don't have 505/505 A/H. The might have 520/490 A/H, due to how bot generation works in CreateRandomBot. So you get an error about not having enough accounts.
## Changes Proposed:
Here I introduce an explicit account type tracking system using a new database table. playerbots_account_type tracks each account with three possible types:

1. Type 0: Unassigned (available for assignment)
2. Type 1: RNDbot account
3. Type 2: AddClass account

Account assignment logic:
On server start, ALL randombot accounts in the database are cataloged. RNDbot accounts are assigned from the lowest position upward based on MaxRandomBots needs. AddClass accounts are assigned from the highest position downward based on AddClassAccountPoolSize.

Once assigned, account types 1 and 2 are permanent and never change or decrease (they can only increase if there's a need for it) with the exception of when MaxRandomBots or AddClassAccountPoolSize are set to 0. Then and only then are their associated accounts types set to 0 (unassigned).

Updated bot selection:
Server now uses filtered lists (rndBotTypeAccounts, addClassTypeAccounts) instead of positional calculations. Account type checks query the database table instead of relying on position.

Updated bot login:
AddRandomBots has been refactored to login all the Alliance bots that it should, then full whatever demand is left by logging in Horde bots. If demand is not fulfilled, it goes back and logs in some more Alliance bots. This can result in faction ratios that deviate from the configured ratios, by perhaps up to 2%, but you will get all the bots that you have asked for.

However, if after all that the server can't login the requested bots, then RandomBotAccountCount is manually set and is actually lower than it should. You will get an error about how many more accounts are needed.

## Reason For Change:
Several months ago when automatic RandomBotAccountCount was implemented, @liyunfan1223 [identified a problem with how accounts are assigned](https://github.com/liyunfan1223/mod-playerbots/blob/00cc2468f13261d616a058a4fde6536766aaeb41/src/RandomPlayerbotMgr.cpp#L1843): 
```
// @FIXME: Modifying RandomBotAccountCount may cause the original addclass bots to be converted into rndbots,
// which needs to be fixed by separating the two accounts in implementation
```
I came across this issue when I was trying to implement another feature for login filtering of bots according to their level. There I found out that as RandomBotAccountCount changed (particularly if it decreased) RNDbots can change to AddClass bots and vice versa.

And not only that, RandomBotAccountCount changes can cause some accounts to be a mix of RNDbots and AddClass bots. This PR aims to stop such arbitrary mixes and changes to bots.

Furthermore, the change can be of help in future development:
- You want accounts that are specifically for RNDbots? Here's rndBotTypeAccounts for you. 
- Is there a third type of random bot accounts introduced and there's concern about mixing with other account? Well just assign it type 3 and you can call on them when you need.

For AddRandomBots, the reason for change is as explained: It demanded perfect faction ratio representation, even if that perfect ratio didn't exist in the DB. So now it can have a leeway of few percentages if it needs to. If you still want a perfectly represented ratio though, set MaxRandomBots to +10% higher than usual, reboot the server and wait for all account assigments to be done. Then lower MaxRandomBots back to the usual value you had. This will make sure you have more than enough rndBotTypeAccounts for a perfect ratio. Otherwise being off by 1 or 2% is fine.

Other Changes: 
- With RandomBotAccountCount = 0, it originally created + 1 account offset. I removed this for now as the new system uses proper rounding up if the calculated number of needed accounts is not an integer.
- Minor cleanup to RandomPlayerbotMgr.h

## Testing:
I ran tests under many different parameters, and it seems to behave as expected. Testing paraments included:

- RandomBotAccountCount = 0 and RandomBotAccountCount = N (N being a manual number that is >= what's needed)
- With RandomBotAccountCount = 0, changed values of MaxRandomBots and AddClassAccountPoolSize up and down to make sure it behaves correctly.
- Seeing how account assignments behave when there is not enough type 1 or 2 accounts. It should first check for type 0 accounts, then create new accounts accounts if needed.
- Checked that DeleteRandomBotAccounts cleared playerbots_account_type, and new accounts were assigned correctly on reboot.
- Checked that both EnablePeriodicOnlineOffline and changes to the Alliance/Horde ratio work correctly with the account assignment feature.

**Important**: Since this PR is modifying accounts behavior and introducing a new table to the DB, I would appreciate it if at least a couple of other people poked and prodded with tests for issues, before potentially merging.

---

Credit to @noisiver for suggesting the idea of marking accounts in the DB.

**PS to maintainers**: If merged, please add the following [to the WiKi](https://github.com/liyunfan1223/mod-playerbots/wiki/Playerbot-Queries) under acore_playerbots:
```
DELETE FROM `playerbots_account_type`
```
Also update the wiki to mention AddClass. In the syntax for ".playerbots bot addclass [className]" if not all the classes are listed, please make sure to at least mention that the syntax for death knight is "dk"


hermensbas:
Cant test yet, on vacation. If someone else can, would be great otherwise the ppl have to be patient. ;)

liyunfan1223:
While testing I found that some AddClass bots didn't seem to be recognized correctly when the server was first started (not automatically initialized to the same level as the player when added), but this problem was solved after I restarted the server. Frankly it's not a big deal, but it would be nice if it could be explained.

NoxMax:
I will be looking into it. For now I just resolved branch conflicts from the recent flight PR.

Side question about code style, given a single statement conditionals such as:
```
if (condition)
    statement;
```
I typically leave it as-is, unless there is code below it that can be mistaken as part of the statement, so I write it as
```
if (condition) {
    statement;
}
```
but I noticed in your recent PR you have edited several single liners as
```
if (condition)
{
    statement;
}
```
Would you prefer that to be the standard code style, that curly braces are always in their own lines?

noisiver:
For single-line ifs it's ok to skip the curly brackets but it's not the end of the world if you use brackets. For multi-line (or if you decide to add brackets to a one liner) the curly bracket should be on a new line instead of after the parenthesis.

NoxMax:
@liyunfan1223 could you please elaborate on the unrecognized bots. Did they not respond in-game, or were they not even marked according to the terminal?

At boot there would have been a log
> Account type assignment complete: X RNDbot accounts, Y AddClass accounts, Z unassigned
> 10*Y characters collected for addclass command from Y AddClass accounts.    

Do you recall if it gave correct values?

liyunfan1223:
> @liyunfan1223 could you please elaborate on the unrecognized bots. Did they not respond in-game, or were they not even marked according to the terminal?
> 
> At boot there would have been a log
> 
> > Account type assignment complete: X RNDbot accounts, Y AddClass accounts, Z unassigned
> > 10*Y characters collected for addclass command from Y AddClass accounts.
> 
> Do you recall if it gave correct values?

They respond fine except that they are not initialized to the same level as player. I haven't checked the terminal yet and can't test it again for now. If I or someone else confirms it, it will be fine.

About the code style: @noisiver gave the answer, the code is formatted by clang-format, and the rules are defined in the config file https://github.com/liyunfan1223/mod-playerbots/blob/master/.clang-format. We don't really emphasize code formatting right now, but using it as a standard when conflicts arise between devs. I only use code formatting when necessary.

NoxMax:
Got it. Probably your terminal would show everything is fine, but if in game you added bots and they didn't all init to your level, it might be that addclassCache is getting called before addClassTypeAccounts is fully populated (the latter being empty only on first run)

Will check if that's the issue in the logic when I get home.

NoxMax:
Still looking into it. Tried several things based on the premiss that class cache is getting called to early, but I still got the issue described. And yeah it only happens one time, when the server boots with playerbots_account_type empty. During that boot, called addClass bots don't get reinitlized to player level. On subsequent server boots though, everything works fine.

Will continue to look more into what's causing this.

NoxMax:
Fixed. PlayerbotMgr.cpp needed a slight update for the new system. All should be done now.

liyunfan1223:
Now it works great in all my test scenarios, thanks for your efforts!

PR #1445, opened 2025-07-15
Update RPG strategy to support travel flight and make rpg status weight configurable

liyunfan1223:
```
# Control probability weights for RPG status of bots. Takes effect only when the status meets its premise.​
# Sum of weights need not be 100. Set to 0 to disable the status.
#
# WanderRandom  (Default: 15  Move randomly nearby to find and kill mobs)
# WanderNpc     (Default: 20  Randomly interact with nearby NPCs​)
# GoGrind       (Default: 15  Go to nearby level-appropriate locations to grind for killing mobs)
# GoCamp        (Default: 10  Return to a nearby camp depending on innkeeper/flightmaster)​
# DoQuest       (Default: 60  Select quest from the quest log and head to the location to attempt completion​)
# TravelFlight  (Default: 15  Go to the nearest flightmaster and fly to a level-appropriate area)
# Rest          (Default: 5   Take a break for a while and do nothing)
AiPlayerbot.RpgStatusProbWeight.WanderRandom = 15
AiPlayerbot.RpgStatusProbWeight.WanderNpc = 20
AiPlayerbot.RpgStatusProbWeight.GoGrind = 15
AiPlayerbot.RpgStatusProbWeight.GoCamp = 10
AiPlayerbot.RpgStatusProbWeight.DoQuest = 60
AiPlayerbot.RpgStatusProbWeight.TravelFlight = 15
AiPlayerbot.RpgStatusProbWeight.Rest = 5
```

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -32,6 +32,7 @@
 #        ACTIVITIES
 #        SPELLS
 #        STRATEGIES
+#        RPG STRATEGY
 #        TELEPORTS
 #        BATTLEGROUND & ARENA & PVP
 #        INTERVALS
@@ -756,11 +757,6 @@ AiPlayerbot.RandomBotGroupNearby = 0
 # Default: 1 (enabled)
 AiPlayerbot.AutoDoQuests = 1
 
-# Randombots will behave more like real players (experimental)
-# This option will override AiPlayerbot.AutoDoQuests, RandomBotTeleLowerLevel, and RandomBotTeleHigherLevel
-# Default: 1 (enabled)
-AiPlayerbot.EnableNewRpgStrategy = 1
-
 # Quest items to keep in bots' inventories (do not destroy)
 AiPlayerbot.RandomBotQuestItems = "5175,5176,5177,5178,6948,11000,12382,13704,16309"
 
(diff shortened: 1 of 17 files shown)


hermensbas:
oooeh thats actually really nice!

NoxMax:
Given that this puts grinding into RPG strat, would there be anymore point of...

```
# Chance bot chooses RPG (Teleport to random camp for their level) instead of grinding
AiPlayerbot.RandomBotRpgChance = 0.20
```

liyunfan1223:
Marked as draft for some performance checks.

Currently travel flight relies on the taxi cheat to work, so make sure `AiPlayerbot.BotCheats = "taxi"`.

liyunfan1223:
> Given that this puts grinding into RPG strat, would there be anymore point of...
> 
> ```
> # Chance bot chooses RPG (Teleport to random camp for their level) instead of grinding
> AiPlayerbot.RandomBotRpgChance = 0.20
> ```

Nice catch, I'll check it out. It looks like this config is no longer used.

NoxMax:
Also need a check about how it interacts with ChooseTravelTargetAction.cpp

liyunfan1223:
ChooseTravelTargetAction.cpp and RandomBotRpgChance are now dead code, they came from cmangos bots rpg strategy and have been deprecated as they didn't work well on this branch, but some code and settings have not been cleaned up yet. Cleaning up dead code can be an important but complex task that I think should be done in another place.

liyunfan1223:
No noticeable performance degradation was observed, so it is now merged.

DustinHendrickson:
Awesome update! 

I ran into the bots in Stormwind arrivals getting stuck in the air on dismount from the Gryphon.

image

Raz0r1337:
@DustinHendrickson 
I've had this exact problem before, even without this new feature, when my bots followed me while I was flying. This issue has nothing to do with this or the RPG Strategy. But maybe someone will find a solution for it. It's annoying. 😅

NoxMax:
I have also found bots mysteriously in midair before, without anything related to flight. Observation I have seen is them walking a flat road, and then start climbing an invisible hill, then climbing down. Oddly enough, I also observed it in Stormwind. Might be related to MaxMovementSearchTime and related mechanisms.

DustinHendrickson:
> @DustinHendrickson I've had this exact problem before, even without this new feature, when my bots followed me while I was flying. This issue has nothing to do with this or the RPG Strategy. But maybe someone will find a solution for it. It's annoying. 😅

This is specific to the new flight path drops though, not just random Z height issues. Reported because it's consistent at this spot when they "land". They begin to drop, then jump up to their height till they move and then come back down.

Raz0r1337:
The RPG strategy update only causes more bots to have this problem because RND bots now automatically use flight. Perhaps because they don't move for a while after the flight is over. My ALT bots and RND bots in my groups have always behaved exactly this way. As soon as they move, they fall to the ground, and everything goes back to normal.

noisiver:
If bots don't move, they don't fall. The same can be said for any spell that raises a bot into the air. If they aren't moving they won't fall back down. If they were affected by gravity in some way it wouldn't happen and it's likely you wouldn't notice it.

Raz0r1337:
All that would have to be done is to make the bots take a step forward quick after dismounting and the problem after flying would be solved.

noisiver:
The ideal solution would be to make them respect gravity.

Raz0r1337:
> All that would have to be done is to make the bots take a step forward quick after dismounting and the problem after flying would be solved.

Hotfix 🤣
until they respect gravity.

Raz0r1337:
> The ideal solution would be to make them respect gravity.

But how?

PR #1446, opened 2025-07-16
Add ability to configure bot level range for worldbuff strategy

brighton-chi:
I've proposed some changes to allow setting a global minimum and maximum level in the config for bots that can apply worldbuffs with the nc +worldbuff strategy. I decided to implement it on a global basis instead of spec-by-spec or class-by-class because I don't see much use for levels to be tailored on such a granular level, and thus I thought a global setting would be preferable due to its user-friendliness. I also removed references to faction from the code and config because it didn't seem to be doing anything of value and also didn't seem worthwhile to adapt into a configurable feature like bot level.

I've kept the default minimum/maximum level at 80 and also kept the default spells the same as what existed before. My personal preference is to set default at 0 (so no level limitation) and then have no default spells, but I'd like to get some input on the approach there.

This is my first PR working with the code, and I have no programming experience, so feedback is very much welcome.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1521,43 +1521,48 @@ AiPlayerbot.PremadeSpecLink.11.6.80 = 05320021--230033312031500531353013251
 #
 #
 
-# Applies a permanent buff to all bots simulating effects of spells, flasks, food, runes, etc.
+# Applies a permanent buff to bots simulating effects of spells, flasks, food, runes, etc.
 # Requires sending the command "nc +worldbuff" in chat to a bot (or a group of bots) to enable
+
+# Minimum and maximum levels for bots that will apply buffs with the worldbuff strategy (enter 0 for no level limit)
+AiPlayerbot.WorldBuffMinLevel = 80
+AiPlayerbot.WorldBuffMaxLevel = 80
+
 # Numbers after the equal signs are spell IDs and may be customized
 # See Randombots Default Talent Specs for more info on each spec; they are listed in that section by the names in the parentheticals (e.g., arms pve, fury pve)
 
-AiPlayerbot.WorldBuff.0.1.0.80.80 = 53760,57358                    #WARRIOR ARMS (arms pve)
-AiPlayerbot.WorldBuff.0.1.1.80.80 = 53760,57358                    #WARRIOR FURY (fury pve)
-AiPlayerbot.WorldBuff.0.1.2.80.80 = 53758,57356                    #WARRIOR PROTECTION (prot pve)
-AiPlayerbot.WorldBuff.0.2.0.80.80 = 60347,53749,57332              #PALADIN HOLY (holy pve)
-AiPlayerbot.WorldBuff.0.2.1.80.80 = 53758,57356                    #PALADIN PROTECTION (prot pve)
-AiPlayerbot.WorldBuff.0.2.2.80.80 = 53760,57371                    #PALADIN RETRIBUTION (ret pve)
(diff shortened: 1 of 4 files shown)


noisiver:
Personally I don't think this is a good idea.

The current system allows you to add world buffs for multiple ranges, like one set for level 60, one for 70 and one for 80. After adding them you don't have to do anything else, but merging this PR means you'll have to remember or store each level range and change the config if you want to switch.

I'd like more input on the matter because it's merely my personal opinion though.

brighton-chi:
Are you able to modify the levels to which worldbuffs apply without modifying the .dist? That is the only way I can get it to work, and if that is the case, I don't think we should consider it to be a truly configurable option. That being said, I think it should be possible to make it so that the levels are configurable while also retaining the ability to allow different levels to have different sets of buffs simultaneously, and I could see if I can figure that out. I liked the concept of a single global switch for the level because it is very straightforward and I think would generally suit peoples' needs, but your point makes sense as some people might play vanilla/TBC/WotLK at the same time.

Another option would be to modify the default .dist so that it has three lists of classes/specs/buffs for levels 60, 70, and 80 (basically take what is there and duplicate it twice, with modifications to the min/max level fields to 60 and 70 for the duplicated lists). That wouldn't allow more granular level customization, but probably those three levels should be generally sufficient. That would also be the easiest change to make since it wouldn't require any modification to the code (though I would suggest modifying the code anyway to remove all the references to faction, which I don't think serve any useful purpose).

Dreathean:
I think the second option you mentioned probably makes the most sense to have coverage on all 3 main endgame levels. Only "issue" is that it takes up a lot more space but I can't really think of a better way to handle 3 different sets of endgame buffs.

noisiver:
I think if at least 60, 70 and 80 were covered (that's the last level of each expansion after all) that would be ideal. The faction is required for the level 60 bracket since Warchief's Blessing is Horde only.

The level 80 bracket were added with the feature, I assume the intention was that someone could add more level brackets to the module if desired. I don't know why no one has, or why I haven't submitted the level 60 and 70 alternatives from my own fork.

In case you are interested, the level 60 variant I made is [here](https://github.com/noisiver/mod-playerbots/blob/381e7224c2d0f5cf5bda37ef44e4c42838cb69a8/conf/playerbots.conf.dist#L1171) and the level 70 variant is [here](https://github.com/noisiver/mod-playerbots/blob/381e7224c2d0f5cf5bda37ef44e4c42838cb69a8/conf/playerbots.conf.dist#L1230). I know they are not ideal but if nothing else it could provide a base to add to the module config and others can submit changes if they wish.

brighton-chi:
Okay, so I think the gist is the result should be (1) level and faction can be configured in the .conf, (2) the format should remain the same as it is right now, and (3) by default, we will include three sets of specs in the .dist, one for level 60, one for level 70, and one for level 80. Or do we want to do four sets, with Alliance 60, Horde 60, Neutral 70, Neutral 80?

I'm happy to prepopulate level 60. What I use is quite similar to Revision's, except I include normal buffs too. But I think for a basic list, we probably keep it shorter, something like: dragonslayer, warchief's blessing (Horde only), spirit of zandalar, spec-based flask, DM buffs (AP and spell crit only for applicable specs), songflower. Maybe DMF too?

kadeshar:
Personally im think that static config is not solution. 
We should fix loading config not existing in .dist. Thats all.
That allows you cover more complex configuration like
obraz

brighton-chi:
Sounds good. After I figure out the code (assuming I do), I'll need to think about how to set up the defaults in the .dist and the comments. But I think this is a feature for more advanced users anyway so it's OK if it's a little more complicated. I'm going to close this PR and open a new one once it's ready.

PR #1456, opened 2025-07-21
Make world buff strategy conditions configurable

brighton-chi:
Phew, I had a lot of difficulty getting Azerothcore to allow keys in the .conf without being in the .dist for this purpose. The result of this PR is not reliant on the .dist. You can enter in any number of combinations of action, class, spec, minlevel, maxlevel, and buffs in the .conf. All the .dist really needs is AiPlayerbot.WorldBuffMatrix = " ". 

While this PR should function, one weakness is that every single entry (meaning each particular combination of faction, class, spec, minlevel, maxlevel, and buffs) is on the same line. I could not get a multi-line variant to work, unless the .dist included a blank entry for every single entry in the .conf. Since we don't know how many entries a player wants, it would have to be some arbitrary number. Basically, you'd have lines in the .dist that read WorldBuffEntry1 = , WorldBuffEntry2 = , etc., up to whatever number you wanted to cap the maximum allowable entries at. I didn't think this was a better approach but am open to thoughts.

There is also the question as to what entries to include in the .dist. I included only the same level 80 cross-faction list that is included by default in the current .dist. It could be nice to include defaults for levels 60 and 70 also, but I wanted to avoid cluttering up the .dist too much. Open to feedback here as well.

### DIFF
--- a/src/AiFactory.cpp
+++ b/src/AiFactory.cpp
@@ -88,6 +88,9 @@ uint8 AiFactory::GetPlayerSpecTab(Player* bot)
             case CLASS_PRIEST:
                 tab = PRIEST_TAB_HOLY;
                 break;
+            case CLASS_WARLOCK:
+                tab = WARLOCK_TAB_DEMONOLOGY;
+                break;
         }
 
         return tab;
--- a/src/LootObjectStack.cpp
+++ b/src/LootObjectStack.cpp
@@ -6,6 +6,8 @@
 #include "LootObjectStack.h"
 
 #include "LootMgr.h"
+#include "Object.h"
+#include "ObjectAccessor.h"
 #include "Playerbots.h"
 #include "Unit.h"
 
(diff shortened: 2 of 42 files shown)


Dreathean:
Looks good to me, bit messier looking due to the constraints you mentioned, but the added configurability without having to also edit the dist is really nice, and saves space at that.

kadeshar:
Yeap for me also that looks better in term of adding custom values like vanilla worldbuffs

brighton-chi:
The cleanest is if the .dist has no default buffs included. It works fine that way--it can be completely blank. Currently, the prior level 80 buffs are included in the .dist, but the only benefit of that is convenience for the player. Like my personal list in my conf contains only level 60 buffs. So the question is do we want a clean .dist or do we think it's helpful to include a default setup (whether it be what it currently is or something else).

Dreathean:
I think it’s very helpful to keep the current default setup, there’s already going to be people asking questions about it even with the default present.

brighton-chi:
Sounds good to me.

PR #1462, opened 2025-07-23
Engine optimization for better performance and mem usage 

liyunfan1223:
Optimized some key parts and bots engine, trying to improve performance and memory usage. In conclusion I observed about 10% memory reduction, and the server can hold 30% more random bots with the same level of latency. Metrics may vary depending on configuration and hardware, such as whether SmartScale and RandomBotTalk are enabled, so it's better to do some additional testing.

In short, the optimization mainly reduces the number of unordered_map accesses from the engine (merging multiple strategy/trigger/action/value maps into one) to optimize performance, and sets AiObjectContext to shared, so that it does not need to be reinitialized for each bot to reduce memory usage.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -32,6 +32,7 @@
 #        ACTIVITIES
 #        SPELLS
 #        STRATEGIES
+#        RPG STRATEGY
 #        TELEPORTS
 #        BATTLEGROUND & ARENA & PVP
 #        INTERVALS
@@ -756,11 +757,6 @@ AiPlayerbot.RandomBotGroupNearby = 0
 # Default: 1 (enabled)
 AiPlayerbot.AutoDoQuests = 1
 
-# Randombots will behave more like real players (experimental)
-# This option will override AiPlayerbot.AutoDoQuests, RandomBotTeleLowerLevel, and RandomBotTeleHigherLevel
-# Default: 1 (enabled)
-AiPlayerbot.EnableNewRpgStrategy = 1
-
 # Quest items to keep in bots' inventories (do not destroy)
 AiPlayerbot.RandomBotQuestItems = "5175,5176,5177,5178,6948,11000,12382,13704,16309"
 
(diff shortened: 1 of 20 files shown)


NoxMax:
30% more ACTIVE random bots with the same level of latency? 

Update: So it doesn't affect number of active bots, but yes I got a good amount of memory savings. Also confirmed @yqsy2025  observation. Crash when AutoDoQuests = 0, but regardless of EnableNewRpgStrategy value (1 or 0). Crash happens shortly after Playerbots begins loading.

Initially I thought that had something to do with the earlier rpg status weight PR, but went back and AutoDoQuests = 0 is okay over there.

yqsy2025:
When AiPlayerbot.AutoDoQuests = 0 or AiPlayerbot.EnableNewRpgStrategy = 0 then crash

liyunfan1223:
> 30% more ACTIVE random bots with the same level of latency?
> 
> Update: So it doesn't affect number of active bots, but yes I got a good amount of memory savings. Also confirmed @yqsy2025 observation. Crash when AutoDoQuests = 0, but regardless of EnableNewRpgStrategy value (1 or 0). Crash happens shortly after Playerbots begins loading.
> 
> Initially I thought that had something to do with the earlier rpg status weight PR, but went back and AutoDoQuests = 0 is okay over there.

I tested with SmartScale disabled, so yes. Maybe 30% is a bit exaggerated, but it should effect the number of active bots, you can also post some metrics.

Crash when AutoDoQuests = 0 is fixed now.

NoxMax:
Okay so I am running with 7500 bots, smartscale floor/ceiling 150/650. ~4.5/32GB usage when AC is not running. Values below are after about 20 minutes of running the server.

Before this PR, I got ~16.5GB usage with ~1800 active bots.
In the test I did earlier with this PR, I got ~13.2GB usage. with  ~1900 active bots.
In the most recent test, I didn't get as much of a drop in RAM, only down to ~15.5GB, but active bots was consistently around 2300. So 30% might not be such an exaggeration. Not sure if there was some fluke in the first test, but hopefully others give this a test and see.

Also AutoDoQuests is confirmed fixed.

Issue #1439, opened 2025-07-14
Warlock Bot Excessively Collects Soul Shards, Causing Inventory Spam


moncat2005:
**Description**
The default behavior of altbot causes Warlock bots to repeatedly collect Soul Shards, triggering constant "Inventory full" warnings.

**Commit Hash**
[bc5d602](https://gitee.com/vob2005/mod-playerbots/commit/bc5d602326ffb1a6bb2a275f2659b081da29b5fa)

**Suggestion**
Could a configurable item limit (e.g., max stack/cap for specific item types like Soul Shards) be implemented to prevent this issue?

**System**
OS: [Linux]
Hardware: 8vcpu 12G

mtm84:
There was just a new commit that should handle this issue with a more reasonable number.  Not configurable but that could be a reasonable possibility.

[fa79fff](https://github.com/liyunfan1223/mod-playerbots/commit/fa79fff4f4003c7e6f2a4680d5696ff1abf5e6e2)

ThePenguinMan96:
> **Description** The default behavior of altbot causes Warlock bots to repeatedly collect Soul Shards, triggering constant "Inventory full" warnings.
> 
> **Commit Hash** [bc5d602](https://gitee.com/vob2005/mod-playerbots/commit/bc5d602326ffb1a6bb2a275f2659b081da29b5fa)
> 
> **Suggestion** Could a configurable item limit (e.g., max stack/cap for specific item types like Soul Shards) be implemented to prevent this issue?
> 
> **System** OS: [Linux] Hardware: 8vcpu 12G

Yeah, a few players made me aware of the issue so I created the PR that was just merged earlier today. If you update to the latest build the issue should be resolved. The cap is set to 5, so soul shards don't clog the bots inventory.

moncat2005:
After updating to the latest fixed [96cc0da](https://github.com/liyunfan1223/mod-playerbots/commit/96cc0daea6eaa91782ec9b5e34766705dd533529), player testing confirms Warlock bots continue accumulating Soul Shards beyond the intended limit of 5.
<img width="458" height="373" alt="Image" src="https://github.com/user-attachments/assets/4498c8fa-7f9c-4e82-b7da-949831b1787e" />

ThePenguinMan96:
[...]

PR #1463, opened 2025-07-23
Warlock Soul Shard/Soulstone ID conversion

ThePenguinMan96:
Hello everyone,

I'm still working on fixing issue #1439 , and I have a theory - it could be because the clients that are having the issue are not english, and the code currently checks for strings "soul shard" and "soulstone". This PR converts the check over to the item IDs, so it should work regardless of what client is being used. I tested it myself with the english client and there is functionally no difference than before - but I hope it solves the issue #1439 for the non-english clients and community.

### DIFF
--- a/src/strategy/warlock/WarlockActions.cpp
+++ b/src/strategy/warlock/WarlockActions.cpp
@@ -145,9 +145,51 @@ bool CreateSoulShardAction::isUseful()
     uint32 currentShards = bot->GetItemCount(ITEM_SOUL_SHARD, false);  // false = only bags
     const uint32 SHARD_CAP = 6;                                    // adjust as needed
 
-    return currentShards < SHARD_CAP;
+    // Only allow if under cap AND there is space for a new shard
+    ItemPosCountVec dest;
+    uint32 count = 1;
+    bool hasSpace = (bot->CanStoreNewItem(NULL_BAG, NULL_SLOT, dest, ITEM_SOUL_SHARD, count) == EQUIP_ERR_OK);
+
+    return (currentShards < SHARD_CAP) && hasSpace;
 }
 
+bool CastCreateSoulstoneAction::isUseful()
+{
+    Player* bot = botAI->GetBot();
+    if (!bot)
+        return false;
+
+    // List of all Soulstone item IDs
+    static const std::vector<uint32> soulstoneIds = {
+        5232,   // Minor Soulstone
+        16892,  // Lesser Soulstone
+        16893,  // Soulstone
+        16895,  // Greater Soulstone
+        16896,  // Major Soulstone
+        22116,  // Master Soulstone
+        36895   // Demonic Soulstone
+    };
+
+    // Check if the bot already has any soulstone
+    uint32 currentSoulstones = 0;
+    for (uint32 id : soulstoneIds)
+        currentSoulstones += bot->GetItemCount(id, false);  // false = only bags
+
+    // Allow only if the bot has no soulstone AND there is space for one
+    ItemPosCountVec dest;
+    uint32 count = 1;
+    bool hasSpace = false;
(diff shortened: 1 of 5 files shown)


src/strategy/warlock/WarlockActions.cpp
```diff
@@ -145,9 +145,51 @@ bool CreateSoulShardAction::isUseful()
...
+        16895,  // Greater Soulstone
+        16896,  // Major Soulstone
+        22116,  // Master Soulstone
+        36895   // Demonic Soulstone
+    };
+
+    // Check if the bot already has any soulstone
+    uint32 currentSoulstones = 0;
+    for (uint32 id : soulstoneIds)
+        currentSoulstones += bot->GetItemCount(id, false);  // false = only bags
```
kadeshar: You can return false if GetItemCount return more then 0. Counting soulstone and checking equipment fulness is redundant in such situation

src/strategy/warlock/WarlockActions.cpp
```diff
@@ -145,9 +145,51 @@ bool CreateSoulShardAction::isUseful()
...
+
+    // Allow only if the bot has no soulstone AND there is space for one
+    ItemPosCountVec dest;
+    uint32 count = 1;
+    bool hasSpace = false;
+    for (uint32 id : soulstoneIds)
+    {
+        if (bot->CanStoreNewItem(NULL_BAG, NULL_SLOT, dest, id, count) == EQUIP_ERR_OK)
```
kadeshar: This loop is unnecessary. Only one CanStoreNewItem have to pass because there is condition that isUseful return false when there is any soulstone in equipment.

src/strategy/warlock/WarlockTriggers.cpp
```diff
@@ -16,6 +42,12 @@ bool WarlockConjuredItemTrigger::IsActive()
     return ItemCountTrigger::IsActive() && AI_VALUE2(uint32, "item count", "soul shard") > 0;
 }
 
+bool OutOfSoulShardsTrigger::IsActive() { return GetSoulShardCount(botAI->GetBot()) == 0; }
+
+bool TooManySoulShardsTrigger::IsActive() { return GetSoulShardCount(botAI->GetBot()) >= 6; }
+
+bool HasSoulstoneTrigger::IsActive() { return GetSoulstoneCount(botAI->GetBot()) == 0; }
```
kadeshar: Should be > 0

kadeshar (changes requested):

src/strategy/warlock/WarlockTriggers.cpp
ThePenguinMan96: If you look at the hassoulstone trigger registration in warlockaiobjectcontext, it's referenced as "no soulstone". See below:

creators["no soulstone"] = &WarlockTriggerFactoryInternal::HasSoulstone;
static Trigger* HasSoulstone(PlayerbotAI* botAI) { return new HasSoulstoneTrigger(botAI); }

And in the genericwarlocknoncombatstrategy:

triggers.push_back(new TriggerNode("no soulstone", NextAction::array(0, new NextAction("create soulstone", 25.0f), nullptr)));

So having this value ==0 is actually correct. If it was > 0, the "no soulstone" trigger wouldn't fire to create one. 

The wording is quite wierd - and unchanged in my warlock overhaul. I might consider changing it after these warlock bugs are tackled.

kadeshar: Then naming is wrong. Should be OutOfSoulstoneTrigger or something.

ThePenguinMan96: I will update this.

kadeshar (approved):

PR #1466, opened 2025-07-25
Add /w botname "glyphs" and  "glyph equip" commands

Wishmaster117:
Description:
This PR enhances Playerbots with two new chat commands and a persistence workflow for custom glyph sets:

`/w glyphs`

Lists the bot’s currently equipped glyphs in the client locale, using item links for proper translations.

Implements TellGlyphsAction and wiring in ChatTriggerContext / ChatActionContext.

`/w glyph equip GlyphID1 GlyphID2 GlyphID3 GlyphID4 GlyphID5 GlyphID6`

Accepts up to 6 glyph item IDs (any order), auto-splits into 3 majors & 3 minors, checks duplicates and class/race requirements.

Applies each glyph’s aura and saves it via bot->SetGlyph(...).

Persistent custom glyph set

On successful glyph equip, sets an AI flag custom_glyphs = true.

On server restart or relog, if all 6 slots are occupied and custom_glyphs == true, skips applying the preset from playerbot.conf.

Whispering glyph equip again overwrites the set and re-sets the flag.

Auto-reapply on spec change

ChangeTalentsAction clears custom_glyphs before applying a new spec.

PlayerbotFactory::InitGlyphs(false) now early-exits only when custom_glyphs == true, otherwise applies the preset glyphs for the new spec.

Debug logging

INFO-level logs when the flag toggles ON/OFF and prints equipped item IDs.

Implementation Details:

New Actions:

TellGlyphsAction in src/strategy/actions/TellGlyphsAction.*

EquipGlyphsAction in src/strategy/actions/EquipGlyphsAction.*

Chat Wiring:

Added triggers & creators for "glyphs" and "glyph equip" in ChatTriggerContext.h, ChatActionContext.h, and ChatCommandHandlerStrategy.cpp.

AI Value:

Registered custom_glyphs in ValueContext as a ManualSetValue (default false).

Factory Change:

PlayerbotFactory::InitGlyphs now checks custom_glyphs via ai->GetAiObjectContext() and skips preset if true.

Spec Change Hook:

ChangeTalentsAction clears the custom_glyphs flag before switching specs.

Testing Guide:

List Equipped Glyphs :

/w glyphs
→ Bot replies with localized, colored glyph names.

Equip Custom Glyphs :

/w glyph equip 40913 43331 40906 43335 43674 45623
→ Bot applies glyphs, replies “Glyphs updated.”, and console logs IDs + “Flag set to ON”.

Persistence :

Relog or server restart → custom glyphs remain if all slots filled.

Spec Change :

/w talents spec healpve
→ Bot logs “Flag set to OFF” once, then applies preset glyphs for the new spec.

Notes:

Backward-compatible: bots that don’t use glyph equip continue using presets from playerbot.conf.

No changes to existing talent or gear workflows.

Please review and let me know of any feedback or adjustments!

Here is an example of the implementation I made with this change in the MultiBot addon (i'me waiting from this merge to make the PR).

image

image

image

BOT Level 15 Two slots:
image

BOT Level 30 Three Slots:
image

BOT Level 50 Four Slots:
image

BOT Level 70 Five Slots:
image

BOT Level 80 Six Slots:
image
 
image

And a Custom talents interface too:

image

### DIFF
--- a/src/strategy/actions/EquipGlyphsAction.cpp
+++ b/src/strategy/actions/EquipGlyphsAction.cpp
@@ -106,7 +106,7 @@ bool EquipGlyphsAction::Execute(Event event)
     std::vector<GlyphInfo const*> glyphs;
     if (!CollectGlyphs(itemIds, glyphs))
     {
-        botAI->TellMaster("glyph equip : liste invalide (IDs items glyphes de ta classe, max 6).");
+        botAI->TellMaster("Usage: glyph equip <6 glyph item IDs> (3 major, 3 minor).");
         return false;
     }
 
--- a/src/strategy/actions/EquipGlyphsAction.cpp
+++ b/src/strategy/actions/EquipGlyphsAction.cpp
@@ -1,0 +1,159 @@
+/*
+ * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license, you may redistribute it
+ * and/or modify it under version 2 of the License, or (at your option), any later version.
+ */
+
+#include "EquipGlyphsAction.h"
+
+#include "Playerbots.h"
+#include "ObjectMgr.h"
+#include "SpellMgr.h"
+#include "DBCStores.h"
+#include "AiObjectContext.h"
+#include "Log.h"  
+
+#include <unordered_map>
+#include <sstream>
+#include <unordered_set>
+
+namespace
+{
+    // itemId -> GlyphInfo
+    std::unordered_map<uint32, EquipGlyphsAction::GlyphInfo> s_GlyphCache;
+}
+
+void EquipGlyphsAction::BuildGlyphCache()
+{
+    if (!s_GlyphCache.empty())
+        return;
+
+    ItemTemplateContainer const* store = sObjectMgr->GetItemTemplateStore();
+
+    for (auto const& kv : *store)
+    {
+        uint32 itemId = kv.first;
+        ItemTemplate const* proto = &kv.second;
+        if (!proto || proto->Class != ITEM_CLASS_GLYPH)
+            continue;
+
+        // inspect item spell
+        for (uint32 i = 0; i < MAX_ITEM_PROTO_SPELLS; ++i)
+        {
+            uint32 spellId = proto->Spells[i].SpellId;
+            if (!spellId) continue;
+
+            SpellInfo const* si = sSpellMgr->GetSpellInfo(spellId);
+            if (!si)         continue;
+
+            for (uint8 eff = 0; eff <= EFFECT_2; ++eff)
+            {
+                if (si->Effects[eff].Effect != SPELL_EFFECT_APPLY_GLYPH)
+                    continue;
+
+                uint32 glyphId = si->Effects[eff].MiscValue;
+                if (!glyphId) continue;
+
+                if (auto const* gp = sGlyphPropertiesStore.LookupEntry(glyphId))
+                    s_GlyphCache[itemId] = {gp, proto};
+            }
+        }
+    }
+}
+
+EquipGlyphsAction::GlyphInfo const* EquipGlyphsAction::GetGlyphInfo(uint32 itemId)
+{
+    BuildGlyphCache();
+    auto it = s_GlyphCache.find(itemId);
+    return (it == s_GlyphCache.end()) ? nullptr : &it->second;
+}
+
+/// -----------------------------------------------------------------
+///  Validation and collect
+/// -----------------------------------------------------------------
+bool EquipGlyphsAction::CollectGlyphs(std::vector<uint32> const& itemIds,
+                                      std::vector<GlyphInfo const*>& out) const
+{
+    std::unordered_set<uint32> seen;
+
+    for (uint32 itemId : itemIds)
+    {
+        if (!seen.insert(itemId).second)
+            return false;                                 // double
+
+        auto const* info = GetGlyphInfo(itemId);
+        if (!info)                                        // no good glyph
+            return false;
+
+        // check class by AllowableClass
+        if ((info->proto->AllowableClass & bot->getClassMask()) == 0)
+            return false;
+
+        out.push_back(info);
+    }
+    return out.size() <= 6 && !out.empty();
+}
+
+/// -----------------------------------------------------------------
+///  Action
+/// -----------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


src/strategy/actions/EquipGlyphsAction.cpp
```diff
@@ -0,0 +1,159 @@
...
+{
+    // 1) parse IDs
+    std::vector<uint32> itemIds;
+    std::istringstream iss(event.getParam());
+    for (uint32 id; iss >> id; ) itemIds.push_back(id);
+
+    std::vector<GlyphInfo const*> glyphs;
+    if (!CollectGlyphs(itemIds, glyphs))
+    {
+        botAI->TellMaster("glyph equip : liste invalide (IDs items glyphes de ta classe, max 6).");
```
kadeshar: Fix message

src/strategy/actions/EquipGlyphsAction.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    // 2) prepare a empty slots table ?
+    bool used[6] = {false,false,false,false,false,false};
+
+    // 3) for each glyph, find the first available and compatible socket
+    for (auto const* g : glyphs)
+    {
+        bool placed = false;
+
+        for (uint8 i = 0; i < MAX_GLYPH_SLOT_INDEX; ++i)
```
kadeshar, line 121: Im not sure about this for. That shouldnt be calculated from current bot level?

src/strategy/triggers/ChatTriggerContext.h
```diff
@@ -134,6 +134,8 @@ class ChatTriggerContext : public NamedObjectContext<Trigger>
         creators["qi"] = &ChatTriggerContext::qi;
         creators["wipe"] = &ChatTriggerContext::wipe;
         creators["pet"] = &ChatTriggerContext::pet;
+		creators["glyphs"] = &ChatTriggerContext::glyphs; // Added for custom Glyphs
+		creators["glyph equip"] = &ChatTriggerContext::glyph_equip; // Added for custom Glyphs
```
kadeshar: Github shows some strange tabs and in other code places

src/strategy/actions/EquipGlyphsAction.cpp
Wishmaster117: My implementation already respects the level-based glyph limits, so for exemple a level-15 bot can never receive more than its two unlocked sockets.

At every iteration we ask the core which glyph socket is actually available:

```
uint32 slotId = bot->GetGlyphSlot(i);
auto const* gs = sGlyphSlotStore.LookupEntry(slotId);
if (!gs)                      // slotId == 0  →  socket still locked for this level
    continue;                 // we skip it
```
Locked sockets return slotId == 0.
LookupEntry(0) yields nullptr, so the loop immediately skips that index.

Type-safety: For open sockets we also test
gs->TypeFlags == g->prop->TypeFlags to ensure major glyphs go only into major slots and minors into minor slots.

Result for exemple:

Level 15 → only slots 0 & 1 are open → the loop ever considers max 2 sockets.

Level 30 → slots 0 1 3 are open → max 3 sockets, etc.

If we run out of compatible, unlocked sockets, placed stays false and we abort with
"Not enough empty sockets for all glyphs."

So a level-15 bot can equip at most two glyphs; any extra IDs in the command will make the action fail gracefully.
No possibility to overfill low-level bots.

The key is the double loop:

```
for (auto const* g : glyphs)                  // every requested glyph
{
    bool placed = false;

    for (uint8 i = 0; i &lt; MAX_GLYPH_SLOT_INDEX; ++i)   // iterate over the 6 sockets
    {
        if (used[i])                           // already filled during this command
            continue;

        uint32 slotId  = bot-&gt;GetGlyphSlot(i);         // critical call
        auto const* gs = sGlyphSlotStore.LookupEntry(slotId);

        // skip if the socket is locked   OR   type (major/minor) does not match
        if (!gs || gs-&gt;TypeFlags != g-&gt;prop-&gt;TypeFlags)
            continue;

        /* remove current aura (if any), cast new one, SetGlyph, mark used[i] */
    }
}
```
1. `bot-&gt;GetGlyphSlot(i)` hides locked sockets
At level 15 only sockets 0 and 1 are unlocked; every other index returns 0.

LookupEntry(0) ⇒ nullptr, so the continue immediately skips that socket.

The loop therefore never tries to write into a socket that the bot has not unlocked yet.

2. Major / Minor check
Even for an unlocked socket we compare
gs-&gt;TypeFlags (1 = major, 2 = minor) with g-&gt;prop-&gt;TypeFlags.

If they differ we skip the slot.

As soon as they match we remove the old aura, cast the new glyph spell and call SetGlyph.

3. used[6] array
Prevents placing two glyphs in the same slot during the same command; we set used[i] = true after a successful placement.

4. “Not enough sockets” protection
If, after scanning all unlocked sockets, no compatible place was found, placed stays false and we abort:
```
botAI-&gt;TellMaster(&quot;Not enough empty sockets for all glyphs.&quot;);
return false;
```
Level-dependent behaviour :

Bot level | GetGlyphSlot(i) ≠ 0 for… | Sockets actually processed
1-14 | none | none → command fails
15-29 | 0, 1 | 2 sockets
30-49 | 0, 1, 3 | 3 sockets
50-69 | 0, 1, 2, 3 | 4 sockets
70-79 | 0, 1, 2, 3, 4 | 5 sockets
80 | 0, 1, 2, 3, 4, 5 | 6 sockets

So the code never writes into a locked slot; it relies on AzerothCore’s existing GetGlyphSlot API, which automatically reflects the level thresholds.

If you prefer, 
We can keep the intent clearer by computing maxUnlocked first (same formula as in PlayerbotFactory::InitGlyphs) and limiting the outer loop:

```
uint8 maxUnlocked = 0;
uint32 level = bot->GetLevel();
if (level >= 15) maxUnlocked = 2;
if (level >= 30) maxUnlocked = 3;
if (level >= 50) maxUnlocked = 4;
if (level >= 70) maxUnlocked = 5;
if (level >= 80) maxUnlocked = 6;

for (auto const* g : glyphs)
{
    bool placed = false;
    for (uint8 i = 0; i < maxUnlocked; ++i)
    {
        if (used[i]) continue;
```
**But functionally the current code is already level-safe because of the gs == nullptr guard.**

Let me know if you need anything else!

PR #1470, opened 2025-07-26
Adds triggers for low/medium/high mana and health

EricksOliveira:
Previously, bots only used food/drink under very specific conditions, ignoring common scenarios like being at 50% mana. Now, bots can: Use food at low, medium, and high health, with different priorities. Use drinks at low, medium, and high mana, with different priorities. This change ensures more intelligent and reactive behavior outside of combat, covering a wider range of situations, especially when the bot is between 40–80% mana or health.

Closes https://github.com/liyunfan1223/mod-playerbots/issues/1457

### DIFF
--- a/src/strategy/generic/UseFoodStrategy.cpp
+++ b/src/strategy/generic/UseFoodStrategy.cpp
@@ -11,13 +11,25 @@
 void UseFoodStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
     Strategy::InitTriggers(triggers);
-    if (sPlayerbotAIConfig->freeFood)
-        triggers.push_back(new TriggerNode("medium health", NextAction::array(0, new NextAction("food", 3.0f), nullptr)));
-    else
-        triggers.push_back(new TriggerNode("low health", NextAction::array(0, new NextAction("food", 3.0f), nullptr)));
 
     if (sPlayerbotAIConfig->freeFood)
-        triggers.push_back(new TriggerNode("high mana", NextAction::array(0, new NextAction("drink", 3.0f), nullptr)));
+    {
+        triggers.push_back(new TriggerNode("high health", NextAction::array(0, new NextAction("food", 1.0f), nullptr)));
+        triggers.push_back(new TriggerNode("medium health", NextAction::array(0, new NextAction("food", 3.0f), nullptr)));
+        triggers.push_back(new TriggerNode("low health", NextAction::array(0, new NextAction("food", 5.0f), nullptr)));
+
+        triggers.push_back(new TriggerNode("high mana", NextAction::array(0, new NextAction("drink", 1.0f), nullptr)));
+        triggers.push_back(new TriggerNode("medium mana", NextAction::array(0, new NextAction("drink", 3.0f), nullptr)));
+        triggers.push_back(new TriggerNode("low mana", NextAction::array(0, new NextAction("drink", 5.0f), nullptr)));
+    }
     else
-        triggers.push_back(new TriggerNode("low mana", NextAction::array(0, new NextAction("drink", 3.0f), nullptr)));
+    {
+        triggers.push_back(new TriggerNode("low health", NextAction::array(0, new NextAction("food", 5.0f), nullptr)));
+        triggers.push_back(new TriggerNode("medium health", NextAction::array(0, new NextAction("food", 3.0f), nullptr)));
+        triggers.push_back(new TriggerNode("high health", NextAction::array(0, new NextAction("food", 1.0f), nullptr)));
+
+        triggers.push_back(new TriggerNode("low mana", NextAction::array(0, new NextAction("drink", 5.0f), nullptr)));
+        triggers.push_back(new TriggerNode("medium mana", NextAction::array(0, new NextAction("drink", 3.0f), nullptr)));
+        triggers.push_back(new TriggerNode("high mana", NextAction::array(0, new NextAction("drink", 1.0f), nullptr)));
+    }
 }


src/strategy/generic/UseFoodStrategy.cpp
```diff
@@ -11,13 +11,25 @@
...
-        triggers.push_back(new TriggerNode("low mana", NextAction::array(0, new NextAction("drink", 3.0f), nullptr)));
+    {
+        triggers.push_back(new TriggerNode("low health", NextAction::array(0, new NextAction("food", 5.0f), nullptr)));
+        triggers.push_back(new TriggerNode("medium health", NextAction::array(0, new NextAction("food", 3.0f), nullptr)));
+        triggers.push_back(new TriggerNode("high health", NextAction::array(0, new NextAction("food", 1.0f), nullptr)));
+
+        triggers.push_back(new TriggerNode("low mana", NextAction::array(0, new NextAction("drink", 5.0f), nullptr)));
+        triggers.push_back(new TriggerNode("medium mana", NextAction::array(0, new NextAction("drink", 3.0f), nullptr)));
+        triggers.push_back(new TriggerNode("high mana", NextAction::array(0, new NextAction("drink", 1.0f), nullptr)));
+    }
```
kadeshar, line 31: For me this PR is wrong.
1. Condition check "if (sPlayerbotAIConfig->freeFood)" is useless
2. Everything except "high health" and "high mana" is useless.
3. If current aproach is wrong for you use custom strategies. Now everytime bots will be eat/drink which even in current state is anoying

Maybe someone have other PoV then comment.

kadeshar:
Also this PR dont resolve https://github.com/liyunfan1223/mod-playerbots/issues/1457
I added explanation here https://github.com/liyunfan1223/mod-playerbots/issues/1457#issuecomment-3122133348

hermensbas:
Maybe its a idea to add /"drink force" instead? Which overwrites the default logic and forces them drink no matter what? Which is core the related issue if not mistaken, as in sometimes u want to force them for some reason.

src/strategy/generic/UseFoodStrategy.cpp
EricksOliveira: I can reduce for them to eat and drink below 30% and below 60%

kadeshar: Dont understand this statement. 
First question is why change current aproach? If they are bots without master that state is ok (they dont make raids/dungeons only killing mobs). If they are bots with master then master can order them to do that.

EricksOliveira: Restrict food/drink usage triggers to "low" and "medium" (health/mana < 60%);

Leave DrinkAction::isUseful() as is (threshold at 85%), ensuring that they only drink if they are truly below average;

This does not prevent the mestre from forcing manual actions, nor does it affect bots with freeFood enabled.

kadeshar: I continued conversation in issue to no double same topic

EricksOliveira: The main point here is that, even in its current state, bots end up sitting down to eat or drink **unnecessarily**, which makes the behavior artificial and even counterproductive, especially for masterless bots that constantly face mobs.

The proposal to limit triggers to "low" and "medium" (health/mana < 60%) aims to:

- Make the behavior **more efficient and natural**, similar to what a player would do;
- Prevent bots from constantly stopping at 80% or more health/mana, even out of combat;
- Keep the `DrinkAction::isUseful()` function as is (threshold < 85%), which ensures that the action will only be executed if it makes sense.

This change **does not prevent** the master from controlling the bots as they wish (via commands), and it also **does not affect bots with freeFood** enabled—they will still have access to resources normally.

kadeshar: Again, continue conversation in issue

kadeshar: Its not question. I answered in issue https://github.com/liyunfan1223/mod-playerbots/issues/1457 and there continue conversation

liyunfan1223:
Multiple triggers are useless, for example, "high mana" refers to a range between 0% to 65%, not a specific conditions. I haven't meet the issue you described.

Removing the high mana trigger does not make the bot behave more naturally, it causes the bot to drink too long at low mana instead of doing short drink at high mana (<65%).

hermensbas:
There a few issues described in the issue which are seperate issues. Which need to confirmed first.

 1.bots dont always seems to respond on drink command even they should? or ppl think they should?

2 The default threshold rule of 85% value only then the bot is applicable to eat or drink is not wanted or not always atleast.

3. The bots behaviour drinking/eating behaviour doesnt feel natural?

Not sure which we are trying to solve, or whether they are confirmed.

EricksOliveira:
> Multiple triggers are useless, for example, "high mana" refers to a range between 0% to 65%, not a specific conditions. I haven't meet the issue you described.
> 
> Removing the high mana trigger does not make the bot behave more naturally, it causes the bot to drink too long at low mana instead of doing short drink at high mana (<65%).

Thanks for the explanation. I think I can close this PR. Especially since I don't have the problem with bots not drinking or running low on mana. It's not a problem for me.

hermensbas:
> > Multiple triggers are useless, for example, "high mana" refers to a range between 0% to 65%, not a specific conditions. I haven't meet the issue you described.
> > Removing the high mana trigger does not make the bot behave more naturally, it causes the bot to drink too long at low mana instead of doing short drink at high mana (<65%).
> 
> Thanks for the explanation. I think I can close this PR. Especially since I don't have the problem with bots not drinking or running low on mana. It's not a problem for me.

well you tried so kuddos for that.

PR #1472, opened 2025-07-27
Change drink aura (free food) to speed up

liyunfan1223:
Changes drink aura from https://www.wowhead.com/wotlk/spell=24707/food to https://www.wowhead.com/wotlk/spell=25990/graccus-mince-meat-fruitcake (from 3%/s to 5%/s).


kadeshar:
Im curious, why freefood is not cheat but separate option in config?

liyunfan1223:
No special reason for that, I think it's okay to put it in cheat.

PR #1473, opened 2025-07-27
Add option to remove 'healer dps' strategy based on specified map.

Tierisch:
Based on recent discussions on whether to remove `healer dps` as a default strategy, here's my take on a compromise.

If enabled and left with the default list, it'll remove `healer dps` on maps that correspond to dungeons and raids, adding it back when bots are instead in the open world.

It adds a check to refrain from including 'healer dps' when default strategies are applied if on a restricted map, and calls a function in `IsBeingTeleportedFar` (where instance strategies get applied) to add/remove `healer dps` for healer bots depending on the map they're moving to.

It's working as intended in my local testing (both adding and removing the strategy) when initializing/resetting on given maps and when summoning bots to different maps, and I haven't been able to think of any problems it might cause.

By default, I have this option set as disabled in the conf, in case more people would rather leave it in the current state or set `healer dps` to be blanket removed from altbots instead of using this distinction. I made it a conf list so users could choose things like making it only get removed in raids or only high level instances.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -501,9 +501,10 @@ AiPlayerbot.AutoGearScoreLimit = 0
 # "mana" (bots have infinite mana)
 # "power" (bots have infinite energy, rage, and runic power)
 # "taxi" (bots may use all flight paths, though they will not actually learn them) 
+# "raid" (bots use cheats implemented into raid strategies) 
 # To use multiple cheats, separate them by commas below (e.g., to enable all, use "gold,health,mana,power,taxi")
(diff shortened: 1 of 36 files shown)


ThePenguinMan96:
Tested this locally on my own PC, can confirm that the PR is working as described! I changed the conf setting to enabled locally, logged in, went into an instance with a healer, and the healer dps strat was gone! Woot.

Dreathean:
This seems like a great compromise that directly targets the main issue people were having with it without impacting Rndbots and giving people plenty of flexibility, very nice!

PR #1475, opened 2025-07-27
Improve Arena Bot Behavior When Target is Behind Obstacles

EricksOliveira:
This update enhances bot behavior in arena scenarios by addressing the issue where bots remain idle if their target moves behind line-of-sight (LoS obstacles). The bot now attempts to reposition near the target to regain LoS instead of standing still, preventing situations where enemies can recover without pressure.

Could someone test it?

### DIFF
--- a/src/strategy/actions/BattleGroundTactics.cpp
+++ b/src/strategy/actions/BattleGroundTactics.cpp
@@ -26,7 +26,7 @@
 #include "Playerbots.h"
 #include "PositionValue.h"
 #include "PvpTriggers.h"
-#include "PathFinder.h"
+#include "PathGenerator.h"
 #include "ServerFacade.h"
 #include "Vehicle.h"
 
@@ -4167,7 +4167,7 @@ bool ArenaTactics::Execute(Event event)
     Unit* target = bot->GetVictim();
     if (target && (!bot->IsWithinLOSInMap(target) || fabs(bot->GetPositionZ() - target->GetPositionZ()) > 5.0f))
     {
-        PathFinder path(bot);
+        PathGenerator path(bot);
         path.CalculatePath(target->GetPositionX(), target->GetPositionY(), target->GetPositionZ(), false);
 
         if (path.IsValid() && path.GetPathType() != PATHFIND_NOPATH)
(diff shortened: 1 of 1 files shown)


yqsy2025:
The effect is not ideal when the enemy suddenly jumps off a bridge or when there is a vertical height difference.I conducted a preliminary test: when the enemy is behind an obstacle, the bot doesn't move and just stands still in place.

EricksOliveira:
> The effect is not ideal when the enemy suddenly jumps off a bridge or when there is a vertical height difference.I conducted a preliminary test: when the enemy is behind an obstacle, the bot doesn't move and just stands still in place.

Thanks for the feedback. Could you provide more details, photos, or videos? Since I don't have a machine to test with, I'm just making blind edits. My PC is too slow to compile and test.

yqsy2025:
It's an improvement compared to before, but when the enemy hides behind an obstacle while the bot is casting a spell, the bot still stands still and does not move toward the enemy.

EricksOliveira:
> It's an improvement compared to before, but when the enemy hides behind an obstacle while the bot is casting a spell, the bot still stands still and does not move toward the enemy.

Thanks again for the feedback. I hope you can test it again and have it fixed now, if possible.

yqsy2025:
This change works better, The bot now chases the enemy in a timely manner, but the bot still doesn't directly follow the enemy when they jump down, causing it to take a detour. As a result, the bot can still be kited to death.

Dreathean:
@yqsy2025 thank you for testing this! I think bots jumping down/falling off things is a different and very challenging issue because bots don't really jump or fall normally currently unless explicitly scripted in specific situations, so perhaps out of the scope for what this PR is trying to accomplish?

yqsy2025:
> @yqsy2025 thank you for testing this! I think bots jumping down/falling off things is a different and very challenging issue because bots don't really jump or fall normally currently unless explicitly scripted in specific situations, so perhaps out of the scope for what this PR is trying to accomplish?

@EricksOliveira  I want this follow-jump behavior to be used only in arenas or battlegrounds. Is that possible to achieve?
@Dreathean when bot->GetPositionZ() - target->GetPositionZ() > 5.0f then should jumping down/falling

kadeshar:
@EricksOliveira can we merge that? Or you have plans to make some more improvements?

EricksOliveira:
You can merge it. I'll open a new PR for combat improvements.

PR #1477, opened 2025-07-28
Channel Cancel checks for Channeled AOE DPS Spells

ThePenguinMan96:
Hello everyone,

I liked the channel cancel I did on mage recently, where they cancel blizzard if there is only 1 enemy left. I decided to do the same for the following classes:

Druid: Hurricane
Hunter: Volley
Priest: Mind Sear
Warlock: Rain of Fire

I moved the "ChannelCancel" action to it's own file, so other classes could benefit from it. I removed it from the mageactions.

Tested it on several fights in molten core, naxxramas, and utgarde pinnacle. Appears to be working as intended.

### DIFF
--- a/src/strategy/warlock/WarlockTriggers.cpp
+++ b/src/strategy/warlock/WarlockTriggers.cpp
@@ -6,6 +6,8 @@
 #include "WarlockTriggers.h"
 #include "GenericTriggers.h"
 #include "Playerbots.h"
+#include "PlayerbotAI.h"
+#include "Player.h"
 
 static const uint32 SOUL_SHARD_ITEM_ID = 6265;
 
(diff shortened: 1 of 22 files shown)


src/strategy/warlock/WarlockTriggers.cpp
```diff
@@ -228,3 +230,32 @@ bool WrongPetTrigger::IsActive()
...
+    5740,   // Rain of Fire Rank 1
+    6219,   // Rain of Fire Rank 2
+    11677,  // Rain of Fire Rank 3
+    11678,  // Rain of Fire Rank 4
+    27212,  // Rain of Fire Rank 5
+    47819,  // Rain of Fire Rank 6
+    47820   // Rain of Fire Rank 7
+};
+
+bool RainOfFireChannelCheckTrigger::IsActive()
```
kadeshar, line 244: This triggers IsActive is almost the same. You dont want write generic trigger on that which accepts list of spells ids and minimum enemy count as additional parameters?

ThePenguinMan96: I can if you want me to. I just did separate triggers for each class so someone can look at the strategy and understand what spell is affected. 

Originally I had the cancelchecktrigger apply to all channeled spells - but then it was stopping evocation, drain soul, etc. I switched it to just one spell for each class that had it, tested it, and it's working as intended.

ThePenguinMan96: And the number of enemies that it checks for is the "attacker count" or "min enemies". I could unify this trigger so it has a list of all of the spell IDs (blizzard, rain of fire, volley, hurricane, and mind sear) and checks an entry in the config (minimum attackers) that's adjustable based on the player preference

Issue #1350, opened 2025-05-29
roll <itemlink>


Mige5:
Hey, I think this would be relatively simple to implement.

So I would like it if im in group/raid/guild chat, I could say "roll <itemlink>" and then the bots would roll on the item if its an upgrade for them.

So I could easily distribute loot I get from 5-mans for example + would also allow me to manually decide if the bot gets the item even if it makes a good roll, I could first also check stats how theyr doing in the raid etc.

notOrrytrout:
Sounds cool. I don’t know how easy it would be to implement though.

Dreathean:
Added with https://github.com/liyunfan1223/mod-playerbots/pull/1482

PR #1482, opened 2025-07-30
New roll for item action

Celandriel:
If you type "roll [item]" in chat bots will respond by rolling. Useful for Masterlooting.


ThePenguinMan96:
This would solve issue #1350. Also, it would be nice to have bots make better decisions on what to roll on, and this supports that.

Dreathean:
Just to confirm based on what I'm seeing with the code, do bots use logic to only roll if it's an upgrade for them?

Celandriel:
Correct. Basically they follow the same logic that currently exists for deciding need or greed. That part is the same, the difference is where they would roll need on an item, now they send a roll command in chat.

Celandriel:
Added general ability to roll. 
Now bots will roll when prompted with the "roll" command.

PR #1483, opened 2025-07-31
Revise bot logic for initializing and using consumables

brighton-chi:
This PR overhauls the logic for how bots use oils and stones so that all classes and specs that would benefit from oils or stones will use them. This PR also introduces logic that provides for oils and stones to be generated upon initialization and level-up for randombots as well as upon maintenance for randombots and altbots. Food, potions, and reagents have also been added to the randombot level-up action, and accordingly, randombots will now also delete outleveled consumables. Additionally, corrections were made to the level brackets for generating reagents.

**Application of Oils and Stones:** 

Under the previous system, Druids, Hunters, Mages, and Priests had logic to apply oils to weapons. Whether wizard or mana oils were prioritized would be based on the type of weapon equipped--swords, staves, and daggers would prioritize wizard oils, while maces would prioritize mana oils. Other types of weapons would not have oils applied at all. Further, the order of priority for oil usage was mixed up--bots would prioritize the lowest rank oils first and then alternate between types of oils when determining which to apply. So for example, if a swordwielder had every single type of oil in its inventory, it would apply Minor Wizard Oil. If it had every type of oil except for Minor Wizard Oil, it would actually apply Minor Mana Oil. 

Meanwhile, only Warriors had logic to apply stones to weapons. Sharpening stones would be applied to swords, axes, and daggers, while weightstones would be applied to maces. Stones could never be applied to polearms, staves, and fist weapons.

This PR makes the following changes with respect to usage of oils and stones:
- Druids already applied oils but prioritized wizard vs. mana based on the type of weapon equipped. They will now prioritize wizard oils if they are Balance spec and will prioritize mana oils if they are Restoration spec. They will apply stones based on weapon type if they are Feral spec.
- Paladins will now apply oils if they are Protection spec (prioritizing wizard oils) or Holy spec (prioritizing mana oils). They will apply stones based on weapon type if they are Retribution spec.
- Hunters previously applied oils and prioritized wizard vs. mana based on the type of melee weapon equipped. They will now prioritize stones instead because their ranged attacks benefit from stones used on melee weapons. They will still apply oils (prioritizing mana) if they have no stones in their inventory.
- Mages and Priests already applied oils but prioritized wizard vs. mana based on the type of weapon equipped. Mages will now always prioritize wizard oils, and Priests will prioritize wizard oils if they are Shadow spec and mana oils if they are Discipline or Holy spec.
- Warlocks had partial logic implemented to apply oils, but doing so was not part of their class-specific source files so they did not apply them in practice. They should not be using oils since they have Spellstones and Firestones and therefore all code relating to Warlocks using oil has been removed. 
- Death Knights, Rogues, and Shamans also should not be using oils or stones since they have class-specific spells or items that are stronger, so no changes were made to their logic.
- The best wizard oil to prioritize according to the bots' logic is Brilliant Wizard Oil (36 SP and 14 crit rating) instead of Superior Wizard Oil (42 SP) because Brilliant Wizard Oil is better at level 60 and remains better in most cases beyond that. The exception is Protection Paladins, who will prioritize Superior Wizard Oil.

**Acquisition of Consumables:**

- The maintenance command will now add level/class/spec-appropriate oils/stones for bots up to level 75. I used 75 as the cutoff because WotLK made it so that oils/stones cannot be used on weapons with iLevels of 165 or higher, and level 76 is when green weapons hit iLevel 165 (166 actually). I'd be open to having bots receive oils/stones up to level 79, though, if that is preferred.
- Upon leveling, randombots will now receive food, potions, reagents, and, up to level 75, oils/stones. They will receive food only if FreeFood (i.e., infinite food) is disabled in the config.
- Level brackets for determining which type of reagent a bot should receive upon initialization or maintenance have been corrected. Previously, there were some errors at the margins. For example, Priests received only Holy Candles until level 60, but they actually need Sacred Candles as well starting at level 56 for Prayer of Shadow Protection (Prayer of Fortitude upgrades to Sacred Candles at level 60).
- Appropriate level requirements to initialize certain reagents were added, such as for Mage runes.
- I decreased the number of some reagents generated because I thought the existing amounts were excessive. Reagents are primarily relevant with respect to altbots for running instances, and for example, I cannot see the need for a Druid to have reagents to cast 40 Gift of the Wild and 40 Rebirth, given the fact that they can be replenished by maintaining, so I reduced them to 20 each. But I'll change these back if people feel strongly about wanting the existing quantities.
- Upon leveling, randombots will delete all food, potions, ammo, reagents, oils and stones, and poisons from their inventories. This is to make it so that their inventories do not get cluttered by lower-quality consumables when they level to a point where they are eligible to use a higher-quality consumable. I would definitely appreciate a review/test of this code, but I have designed it to avoid being overinclusive with deletions. The deletion of potions as a whole subclass of consumables will delete potions beyond just than healing/mana potions, but I consider this acceptable for randombots, and it does not delete flasks or elixirs. I note that deleting reagents as a whole subclass of miscellaneous items does delete Soulshards, but I tested it and believe that the logic Zhur implemented for Warlocks works fine in conjunction with the deletion because Warlocks will just generate new Soulshards after the level-up maintenance. Fun fact--there is only one actual reagent that is in the reagent class, which is the Ankh. The other five items in the class are Worthless Blinding Powder and Vanishing Powder (which is also worthless in WotLK) and three "NPC Equips."

**Refactoring:**

- The ImbueAction source file was modified to make ImbueWithPoisonAction consistent with the corresponding functions for imbuing oils and stones, but Rogues still apply instant poison to their mainhand and deadly poison to their offhand, consistent with the current logic. The only change is that now if a Rogue does not have deadly poison in its inventory, they will apply instant poison to both hands instead of applying nothing to their offhand (and vice versa). I plan to work on modifying Rogues' usages of poisons to include additional types of poisons and perhaps to make the choice of poison an assignable strategy, but that's beyond the scope of this PR.
- Oils, stones, potions, poisons, and (randomly) Healthstones are now defined by item ID instead of display ID. Changes were made to code that relied on the defined terms so that everything relevant should be adapted to use item ID.
- The previous function for initializing stones and oils was "AddConsumables." It has been changed to "InitConsumables" to be consistent with the functions used for ammo, reagents, food, and potions (e.g., InitAmmo).


Dreathean:
This all sounds really great, thank you for your effort with all of this!

ThePenguinMan96:
I did not know that hunter's ranged weapons benefited from the sharpening stone on their melee weapon! That is awesome. 

I enjoy the attention to detail, as far as what oil/stone should be used where. 

I also like the fine-tuning of the level-based reagents.

ThePenguinMan96:
Also, no worries about deleting soul shards - they automatically create one in their inventory when they hit 0.

PR #1485, opened 2025-08-01
Add config to disable hunters from generating specified pet families

brighton-chi:
This PR lets you add into the config any number of IDs from CreatureFamily.dbc to exclude any families of pets from the InitPet function. Essentially it should prevent randombots from generating pets from excluded families and also prevents excluded families from being generated via the maintenance command.

This was inspired by a personal desire to eliminate spiders, and it was very hard to resist having spiders excluded by default.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -699,6 +699,10 @@ AiPlayerbot.AutoUpgradeEquip = 1
 # Default: 0 (disabled)
 AiPlayerbot.HunterWolfPet = 0
 
+# Prohibit hunter bots from creating pets with any family ID listed below in ExcludedHunterPetFamilies
+# See the creature_family database table for all pet families by ID (note: ID for spiders is 3)
+AiPlayerbot.ExcludedHunterPetFamilies = ""
+
 #
 #
 #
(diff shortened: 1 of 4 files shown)


ThePenguinMan96:
Just to check, have you tested:

1. If the entry is misspelled, is there an error or does anything happen?

2. If all of the pet families are blocked, what happens?

3. Does this affect the "pet" chat command?

brighton-chi:
> Just to check, have you tested:
> 
> 1. If the entry is misspelled, is there an error or does anything happen?
> 2. If all of the pet families are blocked, what happens?
> 3. Does this affect the "pet" chat command?

1. What do you mean by misspelled? If somebody puts in a number that is not an existing ID for a creature family, there is no impact.

2. This was one of the first things I was curious about. I tested it with maintenance, and the command will not call a pet if you have all families disabled.

3. It does not. You can still use the chat commands to call a name or ID of a pet in the blocked family, and in fact you can still call the blocked family itself.

ThePenguinMan96:
> > Just to check, have you tested:
> > 
> > 1. If the entry is misspelled, is there an error or does anything happen?
> > 2. If all of the pet families are blocked, what happens?
> > 3. Does this affect the "pet" chat command?
> 
> 1. What do you mean by misspelled? If somebody puts in a number that is not an existing ID for a creature family, there is no impact.
> 
> 2. This was one of the first things I was curious about. I tested it with maintenance, and the command will not call a pet if you have all families disabled.
> 
> 3. It does not. You can still use the chat commands to call a name or ID of a pet in the blocked family, and in fact you can still call the blocked family itself. 

Perfect. I was making sure that the InitPet command would still work if there was an invalid id. Same with the pet command. This PR gets my support - I am all about more options for the users!

PR #1495, opened 2025-08-02
Don't apply XPRate multiplier when bot is in group with real player

hermensbas:
No XP rate multiplier for bots when in group with real player.  Depending on the outcome of the tests, considering adding additional XP rate multiplier config setting for bot in groups with real player(s). In theory an additional custom multiplier should not be needed though.

additionally optimized the code path; cheap condition/checks first, expensive last.

See request: https://github.com/liyunfan1223/mod-playerbots/issues/1490

Expecting that the issue creator to test the change, until that time keeping the PR on draft.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -221,7 +221,7 @@ public:
         if (Group* group = player->GetGroup())
         {
             Player* leader = group->GetLeader();
-            if (leader != player)
+            if (leader && leader != player)
             {
                 if (PlayerbotAI* leaderBotAI = GET_PLAYERBOT_AI(leader))
                 {
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -227,10 +227,10 @@ public:
             Player* leader = group->GetLeader();
             if (leader != player)
             {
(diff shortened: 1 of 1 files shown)


noisiver:
`considering adding additional XP rate multiplier config setting for bot in groups with real player(s)`

That shouldn't be necessary. The bots are affected by the experience multipliers in worldserver just like any player. If someone changes those rates they aren't required to change it for bots on top of that to get them the same rates as regular players.

That's why I think this is the best solution.

hermensbas:
> `considering adding additional XP rate multiplier config setting for bot in groups with real player(s)`
> 
> That shouldn't be necessary. The bots are affected by the experience multipliers in worldserver just like any player. If someone changes those rates they aren't required to change it for bots on top of that to get them the same rates as regular players.
> 
> That's why I think this is the best solution.

Agree, still waiting for this be tested with by @Mige5

hermensbas:
Seems to working on my end, but i need more verification. Only person i think off to test it, next to issue creator, is you @noisiver

hermensbas:
verified

Issue #1369, opened 2025-06-09
[Feature Request] Set bot level range that is allowed to login


NoxMax:
**Is your feature request related to a problem? Please describe.**
Set bot level range that is allowed to login, without having to edit min/max levels and then init bot.

**Describe the solution you'd like**
In the conf it would look like this:
```
# Minimum and maximum levels for randombots allowed to login
AiPlayerbot.RandomBotMinLoginLevel = 1
AiPlayerbot.RandomBotMaxLoginLevel = 80
```
or in this alternative format
```
# Range of levels for randombots allowed to login
AiPlayerbot.RandomBotLoginLevelRange = 1,80
```
Bonus is if there is also a check for **OnPlayerLevelChanged** so if a bot level goes out of the allowed login range, it gets logged out.

**Describe alternatives you've considered**
Editing 
```
AiPlayerbot.RandomBotMinLevel = 1
AiPlayerbot.RandomBotMaxLevel = 80
```
and then reinitializing the bots

**Additional context**
The main reason for this feature is if you want to select specific levels of which bots can login, without having to reinitialize bots, which might change levels for specific bots that you might not want altered. In this way however, no reinitializing is done, and only bots that are already at the desired range would be logged in.

hermensbas:
With all respect, i dont see the point of this feature. Other then adding more logic in a fragile init process. Just make sure you create the server  with the right config, if you wanna change the config then change it and re init the server/bots.

min/max should be equal bots level login/logout level, makes no sense to me to differiate between those. Maybe i am missing something?

NoxMax:
Well that's the whole point about init, to bypass it entirely in this situation. Not modify it.

Say you have 4000 bots in the database and they are more or less equally distributed b/w 1-80. You want to do something with just 50 bots at level 80. You have two options in that case:

1. Login all 4000 bots to the world just so those 50 bots at level 80 among them join in. But no init is needed.
2. Set Min/MaxRandomBots = 50 and RandomBotMin/MaxLevel = 80, and init. This would login the first 50 bot GUIDs and init them to level 80.

Both of these options come with disadvantages. The first one forces you to hog down the server with thousands of bot logins, just so you have access to some small pool of bots that you want. 
The second option, while it allows you to only login 50 bots, you will have to init them so that the are all level 80. The issue here is that those 50 bots might be part of a player arena team, they have a specific spec you don't want altered, or some other thing where you don't want them to get reinitialized.

But if have something like 
```
[...]

PR #1499, opened 2025-08-04
Feat: Filter bot logins by level range

NoxMax:
## Current State: 
Given that you want only bots from a certain level range to be logged into the world, your only option is to set `RandomBotMinLevel` and `RandomBotMaxLevel` to the desired levels and then reinitialize the bots.
 
 ## Changes Proposed:
 Instead of reinitializing bots, if your DB already has enough bots at the desired level range, you can now use the config `RandomBotMinLoginLevel` and `RandomBotMaxLoginLevel` to only login bots of certain level range. You can also enable `RandomBotLogoutOutsideLoginRange` which will log-out bots outside the login range.
 
 ## Reason For Change:
 The main reason for this change is to login bots at a certain range, without having to reinitialize bots, which might change levels for specific bots that you might not want altered. In this way however, no reinitializing is done, and only bots that are already at the desired range would be logged in.
 
 **Tests:** Tested that everything runs correctly when the number of eligible bots is < MinRandomBots, > MinRandomBots & < MaxRandomBots, and > MaxRandomBots. Tested that `RandomBotLogoutOutsideLoginRange` works when it's enabled and that it only affects RNDBots, not alts or AddClass. Recommend that others test as well before merging.
 
 Closes: #1369

### DIFF
--- a/src/RandomPlayerbotMgr.cpp
+++ b/src/RandomPlayerbotMgr.cpp
@@ -13,6 +13,7 @@
 #include <ctime>
 #include <iomanip>
 #include <random>
+#include <climits>
 
 #include "AccountMgr.h"
 #include "AiFactory.h"
@@ -426,7 +427,7 @@ void RandomPlayerbotMgr::UpdateAIInternal(uint32 elapsed, bool /*minimal*/)
         {
             LOG_ERROR("playerbots", "No eligible bots found with level filter {}-{}. Change the level range.",
                     sPlayerbotAIConfig->randomBotMinLoginLevel, sPlayerbotAIConfig->randomBotMaxLoginLevel);
-            SetEventValue(0, "bot_count", 0, 999999999);
+            SetEventValue(0, "bot_count", 0, INT_MAX);
             maxAllowedBotCount = 0;
             currentBots.clear();
             levelFilterAdjusted = true;


hermensbas:
In all honesty i am not huge fan of the amount of code changes in those specific classes for this feature, considering the init process in fragile as it as. Adding more logic will add to that. The feature itself sounds logical and nice to have, just not sure whether we should add more logic in the current phase we are in. I leave that @liyunfan1223

NoxMax:
The login filtration feature does not affect the initialization process at all. In fact it enables the user to login only select desired bots, and bypass the need to use init entirely in cases where the DB already has bots at the desired levels. That's the whole point of it.

NoxMax:
On a sidenote, the feature does override minRandomBots (only if the bots eligible for login are less than minRandomBots) and maxRandomBots (again only if the bots eligible for login are less than maxRandomBots, which is more likely for the max cap). So maybe I should leave some clarifying note in the conf about that. Maybe something like

```
# Minimum and maximum level range for randombots allowed to login
# If level filtration is used, minRandomBots and maxRandomBots might be automatically adjusted to lower values,
# depending on available eligible bots in the database
AiPlayerbot.RandomBotMinLoginLevel = 1
AiPlayerbot.RandomBotMaxLoginLevel = 80
```

hermensbas:
I will try and test it somewhere this week, i can imagine more ppl are waiting for this feature.

src/Playerbots.cpp
```diff
@@ -115,12 +116,39 @@ class PlayerbotsPlayerScript : public PlayerScript
...
     }
 
+    void OnPlayerLevelChanged(Player* player, uint8 oldLevel) override
+    {
+        // Check if feature is enabled and required objects are valid
+        if (!sPlayerbotAIConfig || !sPlayerbotAIConfig->randomBotLogoutOutsideLoginRange || !sRandomPlayerbotMgr)
+            return;
+
+        // Only apply to bots from rndBotTypeAccounts (type 1)
+        uint32 accountId = player->GetSession()->GetAccountId();
```
hermensbas, line 132: Could cause nullpointer exceptions

src/Playerbots.cpp
```diff
@@ -115,12 +116,39 @@ class PlayerbotsPlayerScript : public PlayerScript
...
+        // Check if the new level is outside the allowed login range
+        if (newLevel < sPlayerbotAIConfig->randomBotMinLoginLevel ||
+            newLevel > sPlayerbotAIConfig->randomBotMaxLoginLevel)
+        {
+            LOG_INFO("playerbots", "Bot {} changed levels from {} to {}, outside login range ({}-{}). Marking for logout",
+                    player->GetName(), oldLevel, newLevel,
+                    sPlayerbotAIConfig->randomBotMinLoginLevel, sPlayerbotAIConfig->randomBotMaxLoginLevel);
+
+            // Mark the bot for removal in the next update cycle
+            sRandomPlayerbotMgr->MarkBotForLogout(player->GetGUID().GetCounter());
```
hermensbas, line 147: Nice

hermensbas:
I will try make some time soon, still working on another issue atm.

src/Playerbots.cpp
NoxMax: I think the session is always gonna be valid at OnPlayerLevelChanged, particularly that the logout call is done with event marking as you noted below.

hermensbas:
I thought about it and i'm still not a fan of this. I would love hear from other ppl what they think about it.

Apart from the functionality can be handy, i don't feel like its really required. Also next the already existing options where can put bots to sleep in certain conditions. Feels more like mod addition functionality then core functionality.

Mostly because the process to logon/logout and the changes you need make in those specific classes, The classes involved are not stabilized as is so before adding more additional logic that will bring more risk and will complicate things atleast in the current state.

noisiver:
I agree. I don't personally see this as something that should be in the core module and more suited for a separate module.

hermensbas:
I know you really want this feature, and put a lot effect in it as research, testing, writing the code. But the current state it would be wise to decline these changes for now. So i will archive the PR and close it. That being that, if you really want this feature. You could consider creating mod or keeping the change on your own branch.  I hope you understand though.

NoxMax:
@hermensbas I understand where you're coming from. Excuse the long post but allow me to address few things.

> Also next the already existing options where can put bots to sleep in certain conditions.

True, but it tracks to the larger issue of disk writes we've been discussing. This feature would allow you to limit the number of bots to around the levels that your PC is. This way you wouldn't need to login bots you're not gonna see or interact with, in zones you're not gonna be around in.

Yes, you can already have a low number of bots and set RandomBotMin/MaxLevel around your level, but if you switch to a PC in a different level range, you will need to change the Min/Max values, and of course you will have to reinitialize the bots.

And like you, I'm not a fan of relying on frequent reinitialization. That's why I wanted a feature to not use init if I don't have to.

> Mostly because the process to logon/logout and the changes you need make in those specific classes, The classes involved are not stabilized as is so before adding more additional logic that will bring more risk and will complicate things atleast in the current state.

First the logout is disabled by default, and mostly I have let existing systems handle it, this feature only marks bots to be logged out. Second, if a user is not utilizing this feature by having the default values 1,80 then it is not doing 1 to 80 filtration on a technicality. Instead, it's logic is completely skipped. Meaning it only interacts with the rest code, only if a user wants to use it.

Can it be its own mod? I guess, but I don't think it's big or complicated enough for it to warrant being its own mod. And I hoped it would be of use to others, rather than just have it hidden in my own branch.

Now I have tested it from every angle I can think of, with all sorts of scenarios of eligible login bots that are less than minRandomBots, between min/maxRandomBots, greater than maxRandomBots, scenarios involving EnablePeriodicOnlineOffline, scenarios involving A/H ratios other than 50/50 and other parameters that I could think of that can interact with it. And it has been stable.

Basically when I submit a PR, I do try to break it first.

But I understand your concerns. For example I am aware that I am introducing new forms of logic like the hook OnPlayerLevelChanged. It is utilized by AC of course, but this will be the first time PB inherently is using it. So even if a feature is stable and working as intended, one has to also consider its potential effects on future development. This is why I try to keep clear comments in the code. I would not want someone in the future to have errors in their PR because they were confused by my pre-existing code.

So with all that in mind, and given the feature is stable, would you at least be willing to have a poll about it in Discord? Just showing people the intent behind this and see if it's of use to them. If most people think it is pointless and has potential to cause headaches... then hey, that's fine by me.

hermensbas:
hehe long post, i am read it when i can and come back to you. ❤️ Did a quickscan, but ofcourse a poll would be a nice way on developers channel. Although it doesnt change the fact the classes and processing your adding too are not stabile as is. As in when comes to threads problems or related to them. Adding more logic to that process feels wrong and possible gateway to more problems for a relative small feature.

hermensbas:
Did also consider player with the following settings?

```ini
AiPlayerbot.BotActiveFreezeDuringServerInit = 1
AiPlayerbot.BotActiveAloneForceWhenInRadius = 150
AiPlayerbot.BotActiveAloneForceWhenInZone = 1
AiPlayerbot.BotActiveAloneForceWhenInMap = 0
# BotActiveAloneForceWhenIsFriend = 1
AiPlayerbot.BotActiveAloneForceWhenIsFriend = 0
AiPlayerbot.BotActiveAloneForceWhenInGuild = 1

and 

# Randombots will behave more like real players (experimental)
# This option will override AiPlayerbot.AutoDoQuests, RandomBotTeleLowerLevel, and RandomBotTeleHigherLevel
# Default: 1 (enabled)
AiPlayerbot.EnableNewRpgStrategy = 1

# Control probability weights for RPG status of bots. Takes effect only when the status meets its premise.​
# Sum of weights need not be 100. Set to 0 to disable the status.
#
# WanderRandom  (Default: 15  Move randomly nearby to find and kill mobs)
# WanderNpc     (Default: 20  Randomly interact with nearby NPCs​)
# GoGrind       (Default: 15  Go to nearby level-appropriate locations to grind for killing mobs)
# GoCamp        (Default: 10  Return to a nearby camp depending on innkeeper/flightmaster)​
# DoQuest       (Default: 60  Select quest from the quest log and head to the location to attempt completion​)
# TravelFlight  (Default: 15  Go to the nearest flightmaster and fly to a level-appropriate area)
# Rest          (Default: 5   Take a break for a while and do nothing)
AiPlayerbot.RpgStatusProbWeight.WanderRandom = 15
AiPlayerbot.RpgStatusProbWeight.WanderNpc = 20
AiPlayerbot.RpgStatusProbWeight.GoGrind = 15
AiPlayerbot.RpgStatusProbWeight.GoCamp = 10
AiPlayerbot.RpgStatusProbWeight.DoQuest = 60
AiPlayerbot.RpgStatusProbWeight.TravelFlight = 15
AiPlayerbot.RpgStatusProbWeight.Rest = 5
```

hermensbas:
Lets re-open and pull discussion into the middle.

NoxMax:
It shouldn't have any interactions with smart scaling, but I will check it out to see if there is anything that stands. Sidenote: BotActiveAlone has been set to 100 default for a while now, unless a manual 10 works for you. BotActiveFreezeDuringServerInit is also not a parameter that exists anymore, but at least from its name I gather it had interactions with the login process when it was around. 

Similarly with the RPG ratios. Filtering here modifies the overall maxAllowedBotCount, but RPG strat doesn't care about what happens there. It just applies the weight probabilities on a per bot basis. Still, wouldn't hurt to test. I will throw at it some extreme ratios and see if anything pops.

NoxMax:
Update: Made better error handling when there are no bots at all at the requested level range

---

Scaling tests: Ran with a 70-80 filter and a manual 10% scaling. These were the stats

`Level filter active: 1287 eligible bots found in range 70-80. Effective maximum: 1287 bots`
Directly confirmed in the DB that these are the eligible bots, not counting altbots, not counting addclass

```
playerbot rndbot stats                                                                                                                                                                                     [2/1988]
AC> Random Bots Stats: 1287 online                                                                                                                                                                                 
Bots level:                                                                                                                                                                                                        
    70..79: 553 alliance, 614 horde
    80..80: 55 alliance, 65 horde
Bots race:
    Human: 126, avg lvl: 74.6
    Orc: 102, avg lvl: 74.6
    Dwarf: 98, avg lvl: 75.1
    Night Elf: 130, avg lvl: 74.8
    Undead: 121, avg lvl: 74.7
    Tauren: 128, avg lvl: 75
    Gnome: 103, avg lvl: 75
    Troll: 142, avg lvl: 75.4
    Blood Elf: 186, avg lvl: 74.7
    Draenei: 151, avg lvl: 75
Bots class:
    warrior: 94, avg lvl: 75
    paladin: 108, avg lvl: 74.3
    hunter: 99, avg lvl: 75.1
    rogue: 120, avg lvl: 74.6
    priest: 118, avg lvl: 74.2
    dk: 332, avg lvl: 75.2
    shaman: 105, avg lvl: 75.3
    mage: 114, avg lvl: 75
    warlock: 105, avg lvl: 74.8
    druid: 92, avg lvl: 74.8
Bots role:
    tank: 200, heal: 188, dps: 899
Bots status:
    Active: 60
    Moving: 7
    In flight: 0
    On mount: 24
    In combat: 23
    In BG: 0
    In Rest: 112
    Dead: 46
Bots rpg status:
    Idle: 864, Rest: 14, GoGrind: 80, GoCamp: 47, MoveRandom: 34, MoveNpc: 41, DoQuest: 164, TravelFlight: 43
Bots total quests:
    Accepted: 22, Rewarded: 2, Dropped: 0
Bots engine:
    Non-combat: 1200, Combat: 41, Dead: 46
```
60 is obviously much less than 10% of 1287, but that has been a pre-existing issue with manual values of BotActiveAlone. That's why it was defaulted to 100% and smart scale enabled.

---

RPG ratios: Again 1287 bots from level 70-80, but this time scaling is on auto. Gave it this extreme ratio
```
AiPlayerbot.RpgStatusProbWeight.WanderRandom = 1
AiPlayerbot.RpgStatusProbWeight.WanderNpc = 1
AiPlayerbot.RpgStatusProbWeight.GoGrind = 1
AiPlayerbot.RpgStatusProbWeight.GoCamp = 1
AiPlayerbot.RpgStatusProbWeight.DoQuest = 20
AiPlayerbot.RpgStatusProbWeight.TravelFlight = 1
AiPlayerbot.RpgStatusProbWeight.Rest = 1
```
Stats more or less aligned with the config:
```
Bots status:                                                                                                                                                                                                       
    Active: 920                                                                                                                                                                                                    
    Moving: 390                                                                                                                                                                                                    
    In flight: 1                                                                                                                                                                                                   
    On mount: 5                                                                                                                                                                                                    
    In combat: 128                                                                                                                                                                                                 
    In BG: 0                                                                                                                                                                                                       
    In Rest: 132                                                                                                                                                                                                   
    Dead: 21                                                                                                                                                                                                       
Bots rpg status:                                                                                                                                                                                                   
    Idle: 29, Rest: 17, GoGrind: 125, GoCamp: 106, MoveRandom: 26, MoveNpc: 119, DoQuest: 809, TravelFlight: 56   
```
However, and unrelated to level filtering and with it disabled, these weights caused a crash:
```
AiPlayerbot.RpgStatusProbWeight.WanderRandom = 0
AiPlayerbot.RpgStatusProbWeight.WanderNpc = 0
AiPlayerbot.RpgStatusProbWeight.GoGrind = 0
AiPlayerbot.RpgStatusProbWeight.GoCamp = 0
AiPlayerbot.RpgStatusProbWeight.DoQuest = 15
AiPlayerbot.RpgStatusProbWeight.TravelFlight = 0
AiPlayerbot.RpgStatusProbWeight.Rest = 0
```
I posted an issue for it #1584

hermensbas (approved):

src/RandomPlayerbotMgr.cpp
```diff
@@ -374,13 +374,76 @@ void RandomPlayerbotMgr::UpdateAIInternal(uint32 elapsed, bool /*minimal*/)
...
+                                sPlayerbotAIConfig->randomBotCountChangeMaxInterval));
+
+            currentBots.clear();
+            levelFilterAdjusted = true;
+        }
+        else
+        {
+            LOG_ERROR("playerbots", "No eligible bots found with level filter {}-{}. Change the level range.",
+                    sPlayerbotAIConfig->randomBotMinLoginLevel, sPlayerbotAIConfig->randomBotMaxLoginLevel);
+            SetEventValue(0, "bot_count", 0, 999999999);
```
kadeshar: 999 999 999 means 11574 days? I think you can provide 0 or int.max instead of magic number

NoxMax: INT_MAX is probably best. I will implement that, but I can't test it until late next week, though the change is minor so unlikely that there will be sideeffects. It just needs climits.

Issue #1489, opened 2025-08-01
Warlocks resummoning same pet after flight, soul shard issues


Vortikai:
**Bug Description**
I've noticed that the warlock altbot I use keeps summoning the felhunter it already has, especially after flying and landing.

I've also noticed that I can play on that character, get like 30 shards stored, and after a short while of it being an altbot, it just goes to 1. I haven't determined yet if it's immediate or just being used for the pets, soul stones, or what.

**Commit Hash**
[de9f8fb](https://github.com/liyunfan1223/mod-playerbots/commit/de9f8fbbeab01c2d0cfc8b55a1075106866e5d8b)

**How To Reproduce Bug**
1. Have an affliction warlock altbot with a felhunter already as a pet.
2. Fly somewhere (e.g. Argent Tournament to Dalaran).
3. Demount. The warlock summons a felhunter.

**Expected Behavior**
The warlock shouldn't summon pets again if it is a good option and already out. The reason is it slows it down, especially if not in a demonology spec.

**Screenshots**

**Modules**
none

**Playerbot Settings**
Default strategies, sometimes nc +ss healer.

**System**
OS: [e.g. Windows, Linux]
Hardware: [e.g. CPU if performance related]

**Additional Context**

ThePenguinMan96:
I will look into this over my weekend (Monday/Tuesday). I wonder if it lands and before the pet is brought out tries to summon a pet?

As far as soul shards go - warlock bots have built in triggers that delete soul shards if they have above 5 in their inventory, and creates one if they run out. It's essentially automating the soul shard creation/usage because demonology and destruction don't naturally use drain soul as their rotation. 

It makes it so when randombots are out grinding, they don't run out and fail to summon their pet. Also, I chose 5 as a maximum so their inventory doesn't get clogged with soulstones.

ThePenguinMan96:
Is it just affliction? Or is it all 3 specs?

Vortikai:
Yeah, that was what I was thinking. It could also be that it checks for a pet, not finding one since it is mounted and pets aren't out when mounted, then queues up summoning one after dismount.

For soul shards: That makes sense for randombots. I think I'd be fine with that for altbots as well. However, I don't have five built up. I did change the max quantity for soul shards to something like 20 or something, I don't remember, but it seems whenever I log in there's just one. I'll recheck later as well.

I haven't tried any other specs. I'll test demonology later tonight.

ThePenguinMan96:
[...]

PR #1501, opened 2025-08-04
Warlock Dismount Pet Fix

ThePenguinMan96:
Hello everyone,

This PR is to address #1489, where the warlock summons a pet when they dismount.

A tester found that the cause was the "wrong pet" triggering "summon (pet)". I looked into the "wrong pet" trigger, and noticed that there was not a clause if there was no active pet. It was inadvertently casting "summon (pet)" because for a brief second after dismounting, the warlock didn't technically have a pet.

I was able to recreate the issue based on tester feedback (dismounting with a warlock bot that has a pet). I tested this fix locally - it seems to work as intended. The warlock no longer attempts to summon a pet when dismounting. I tested it with nc +debug, and noticed that the "wrong pet" trigger was no longer firing. I also checked the logs - nothing.

Thank y'all for the testing and feedback!

### DIFF
--- a/src/strategy/warlock/WarlockTriggers.cpp
+++ b/src/strategy/warlock/WarlockTriggers.cpp
@@ -207,7 +207,11 @@ bool WrongPetTrigger::IsActive()
     if (enabledCount != 1)
         return false;
 
-    // Step 3: At this point, we know only one pet strategy is enabled.
+    // Step 3: If there is no pet, do not trigger.
+    if (!pet)
+        return false;
+
+    // Step 4: At this point, we know only one pet strategy is enabled.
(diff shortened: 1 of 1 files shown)


Vortikai:
Tests complete and all tests successful. The warlock altbots no longer attempts to summon a pet when a good option for a pet is already out when dismounting. Tested under various conditions and received expected behavior.

All other aspects of the play session where within normal parameters with no negative effects.

This PR gets my vote for merger.

Mr-Deadbeat:
some pets are not meant to be controlled take a glance at this
https://worldofwarcraft.fandom.com/et/wiki/Pet

PR #1512, opened 2025-08-08
Correct side effects of  merge f5ef5bd1c2039290be1b1108872d046e1e65ebe1

Wishmaster117:
Hi 👋,

This PR adjusts the earlier crash fix so we keep the stability (no more GetGuidValue AV) without the side-effects players reported (manual bots disappearing after logout or teleport, unable to summon/revoke).

Issue: https://github.com/liyunfan1223/mod-playerbots/issues/1509

Below is what changed since the previous patch and why.

### Root cause of the regressions
In the first fix i sometimes purged too aggressively:

PlayerbotsMgr::GetPlayerbotAI(Player* player) did a hard cleanup when ObjectAccessor::FindPlayer(...) failed (common during transitions like logout/teleport/flight).
Hard purge also erased the master ⇄ bots relation in _playerbotsMgrMap.
⇒ After a reconnect or a teleport, the module no longer knew which bots belonged to the player.

I also briefly gated OnBotLogin with IsInWorld(), which fires too early in this hook and prevented bots from being registered in playerBots. That gate has been removed.

**What this PR does (delta from the previous patch)**

**1. Introduce a soft vs hard purge for AIs**

- void RemovePlayerbotAI(ObjectGuid guid, bool removeMgrEntry = true);
   - Hard purge (default): remove AI entry and the manager relation — used on real logouts.
   - Soft purge: remove only the AI entry — used for transient states (out-of-world during teleport/flight/etc.).
   - GetPlayerbotAI(...) now calls soft purge only, so we never lose the master ⇄ bots linkage during transitions.

**2. Lookup by GUID and never dereference a null Player***

- GetPlayerbotAI(Player* player) first checks !player, then copies player->GetGUID() to a local guid, and all map lookups are done with that GUID.
  - This removes the original player->GetGUID() deref on a potential nullptr that caused Object::GetGuidValue crashes.

- Added GetPlayerbotAIByGuid(ObjectGuid guid) helper (used internally) to keep lookups uniform and safe.

**3. Keep the original gameplay semantics intact**
- No change to the “save only if: in group && not in BG && not in BG queue && master active” rules.
- I did not add any IsInWorld() checks to login paths (that was the regression).
- Teleport/flight are treated as transients: if a soft purge happens, only the AI cache is cleaned; the player-bot relationship is preserved and bots can be summoned again.

**4. Thread-safety stays**
- The maps are still protected by std::shared_mutex (shared_lock for reads, unique_lock for writes).
- This keeps the original race/crash fixed while avoiding data loss.

**5. Safer master check in AI**
- PlayerbotAI::HasRealPlayerMaster() now guards with ObjectAccessor::FindPlayer(master->GetGUID()) and nulls master if it’s gone, preventing the chain that reached GetGuidValue(this=nullptr).

**6. fmt/MSVC logging nit**
- When logging ObjectGuid, we cast to uint64 to avoid the “type is unformattable” error.

### Before / After

**Before (with the first crash fix):**

✅ Crash was gone.
❌ Manual bots could disappear after logout or teleport (relation erased by hard purge in transient moments).
❌ In some branches OnBotLogin IsInWorld() gating prevented bots from being registered.

**After (this PR):**
✅ Crash still gone (GetGuidValue no longer hit).
✅ Manual bots remain associated to their master after logout/teleport; you can summon/revoke them as usual.
✅ Original save rules (BG/queue/master) are unchanged.
✅ No IsInWorld() gating in login callbacks.
✅ Still thread-safe.

### Code touch points (high level)

- PlayerbotMgr.h/.cpp
  - Add GetPlayerbotAIByGuid(ObjectGuid) (GUID-based lookup).
  - Change RemovePlayerbotAI(guid, bool removeMgrEntry = true) and use soft purge (removeMgrEntry=false) from  GetPlayerbotAI(...).
  - Maintain shared_mutex protection around the maps.

- PlayerbotAI.cpp
  - Harden HasRealPlayerMaster() with FindPlayer() guard and early return.

Note: I considered a small helper GET_PLAYERBOT_AI_SAFE(p) (returns nullptr if p == nullptr) for call sites where the pointer can be transiently null, but it’s optional and not required for this fix.

### How I tested
- Repro’d the teleport case (Dalaran → Stormwind), then summon/revoke: works again.
- Repro’d logout/relogin with manual bots previously added via commands: bots stay associated and can be brought back.
- Soak test with 2+ map threads: no AVs, no stale master references.

### Risks / Compatibility
- API changes are internal to the module. The new RemovePlayerbotAI overload has a default parameter and preserves existing call sites (hard purge).
- Behavioural change only affects transient states: soft purge keeps relations, restoring expected UX.

Happy to squash or rename commits. Thanks for reviewing and for maintaining Playerbots!

### DIFF
--- a/src/PlayerbotMgr.h
+++ b/src/PlayerbotMgr.h
@@ -130,10 +130,16 @@ private:
 
 #define sPlayerbotsMgr PlayerbotsMgr::instance()
 
+// Temporary addition If it keeps crashing, we will use them.
+// Like
+// BEFORE : PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
+// AFTER (safe) : PlayerbotAI* botAI = GET_PLAYERBOT_AI_SAFE(bot);
+// BEFORE : if (PlayerbotAI* botAI = GET_PLAYERBOT_AI(player)) { ... }
+// AFTER (safe) : if (PlayerbotAI* botAI = GET_PLAYERBOT_AI_SAFE(player)) { ... }
 // --- SAFE helpers (append to PlayerbotMgr.h) ---
 inline PlayerbotAI* GET_PLAYERBOT_AI_SAFE(Player* p)
 {
-    // Evite tout déréférencement pendant des états transitoires (nullptr, téléport, vol, etc.)
+    // Avoid any dereference during transient states (nullptr, teleport, flight, etc.)
     return p ? sPlayerbotsMgr->GetPlayerbotAI(p) : nullptr;
 }
 


hermensbas:
No need to squash or rename the commits, amazing work though. Wish i had /debug skillset todo so hehe.  Although ppl might say test test, but best test after sanity is merging atleast in this case.

i like the inline comments, since the lack of testing or possible ability for testing these types of conditions.

Wishmaster117:
> No need to squash or rename the commits, amazing work though. Wish i head skillset todo so hehe.

I hope this time is the right one! :D

hermensbas:
"PlayerbotsMgr::GetPlayerbotAI(Player* player) did a hard cleanup when ObjectAccessor::FindPlayer(...) failed (common during transitions like logout/teleport/flight)”

Just for clarification, findPlaysr will fail when called during transitions?

Wishmaster117:
> "PlayerbotsMgr::GetPlayerbotAI(Player* player) did a hard cleanup when ObjectAccessor::FindPlayer(...) failed (common during transitions like logout/teleport/flight)”
> 
> Just for clarification, findPlaysr will fail when called during transitions?

Yes , during certain transitions, ObjectAccessor::FindPlayer(guid) can briefly return nullptr. That’s expected.

FindPlayer only returns a pointer when the Player is currently registered in the world (in ObjectAccessor’s containers). During transitions the core intentionally unregisters the player, so there is a short window where:

Logout / disconnect: the player is removed from the world before destruction finishes;

Map transfer / teleport (incl. far teleports to instances/BGs/other continents): the player is removed from the old map and only later added to the new one (loading screen in between). In that “limbo” phase, FindPlayer returns nullptr;

(Normal taxi flight usually keeps the player in world; if a flight segment uses a far-teleport under the hood, you may see the same short window.)

This is why the PR switches from a hard purge to a soft purge in GetPlayerbotAI(...) when FindPlayer fails: we clean only the in-memory AI cache but we do not delete the master⇄bot relationship. Once the player finishes the transition and re-enters the world, the relationship is still intact and bots can be summoned/revoked as expected.

hermensbas:
thanks for explaining!

hermensbas:
Added @liyunfan1223 as reviewer just to make him aware of the followup change.

Wishmaster117:
Yes, I’d prefer that he takes a look as well!
Since 1h i'm playing with the .tele, login, login ou and my bots still here and no crash

hermensbas:
> Yes, I’d prefer that he takes a look as well!

I am gonna merge though to fix the current issue's on hand, i rather do forward fix then taking the risk of reverting and the change to be forgotten at this point. I will continue monitoring the channels for issue's, we can still revert if needed. Its a bit of a risky approach, but imo worth it regarding to this longtime existing issue.  Maybe for some this has been minor issue when playing solo, but when playing with larger group of ppl/friends this has been such a pain.

Wishmaster117:
Now I’m going to sleep (long day...), but I’m keeping an eye on this as well — worst case, I’ve set up the safeguards in MultibotMgr.h.

Issue #1502, opened 2025-08-05
Warlock altbot soul shard management


Vortikai:
**Is your feature request related to a problem? Please describe.**
Currently altbots have their inventory touched along with randombots to limit a problem where their bags filled up with soul shards. It seems a bit cheat-y to me for soul shards to automatically be created and destroyed with actual player characters controlled by the playerbots system.

**Describe the solution you'd like**
Either disable the soul shard, soul stone, etc. systems for altbots only or make it an option. Open to suggestions of course. These are just what come to mind

**Describe alternatives you've considered**
None. This ticket serves as a forum for refining the system, if it's even needed. If we can't avoid it, then it is what it is and we can close this ticket.

**Additional context**

ThePenguinMan96:
I'm thinking a solution would be a config entry, something like this:

// Enable this to have all bots automatically 
// replenish soul shards without the use of 
// drain soul. Bots will carry between 1 and 5.
// Default - 1 (enabled)
Ai.AutomatedSoulShards - 1

That way, for those who don't want the soul shards to be automatic, they can disable it by making the value 0. Also, I might make a "drain soul execute" strategy that enables all specs to use drain soul as an execute ability until the bot has the max number of soul shards (32), or slightly before. And make IsUseful for drain soul to check if they have like 28 shards before it casts. I chose 28 because if it casts it at 31, and makes several soul shards, it will lead to the bot spamming the player with the error "I can't carry anymore of those".

ThePenguinMan96:
> I'm thinking a solution would be a config entry, something like this:
> 
> // Enable this to have all bots automatically 
> // replenish soul shards without the use of 
> // drain soul. Bots will carry between 1 and 5.
> // Default - 1 (enabled)
> Ai.AutomatedSoulShards - 1
> 
> That way, for those who don't want the soul shards to be automatic, they can disable it by making the value 0. Also, I might make a "drain soul execute" strategy that enables all specs to use drain soul as an execute ability until the bot has the max number of soul shards (32), or slightly before. And make IsUseful for drain soul to check if they have like 28 shards before it casts. I chose 28 because if it casts it at 31, and makes several soul shards, it will lead to the bot spamming the player with the error "I can't carry anymore of those".

ThePenguinMan96:
I'm thinking a solution would be a config entry, something like this:

# Enable this to have all bots automatically 
# replenish soul shards without the use of 
# drain soul. Bots will carry between 1 and 5.
[...]

PR #1514, opened 2025-08-09
Warlock Soul Shard Cap Increase / Firestone and Spellstone Modes

ThePenguinMan96:
Hello everyone,

This is a small change to warlocks that accomplishes 2 things:

1. Changes the firestone and spellstone weapon enchants so only one of them can be active - players reported to me that both strategies could be present before, resulting in a bug where the bot repeatedly applied the enchant.
2. Changes the soul shard deletion cap from 6 or more to 26 or more - players will now be able to stockpile soul shards up to 25 in a bot's inventory before the bot starts deleting them one at a time back down 25. I chose 25 because if it was higher, drain soul would get multiple shards above the 32 unique cap, and spam the player "I can't carry any more of those". It was super annoying, and with testing, I have not seen this error at 25. This aims to address issue #1502 .

@Vortikai

### DIFF
--- a/src/strategy/warlock/WarlockActions.cpp
+++ b/src/strategy/warlock/WarlockActions.cpp
@@ -21,8 +21,8 @@
 
 const int ITEM_SOUL_SHARD = 6265;
 
-// Checks if the bot has less than 20 soul shards, and if so, allows casting Drain Soul
-bool CastDrainSoulAction::isUseful() { return AI_VALUE2(uint32, "item count", "soul shard") < 20; }
+// Checks if the bot has less than 26 soul shards, and if so, allows casting Drain Soul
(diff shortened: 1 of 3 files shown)


Vortikai:
Will the warlock use the drain soul ability when it is useful to do so?

Also doesn't matter now, but the warlock was using it with the old PR and built up some. The problem is I kill things too fast in daily quests for it to register to use that.

ThePenguinMan96:
> Will the warlock use the drain soul ability when it is useful to do so?
> 
> Also doesn't matter now, but the warlock was using it with the old PR and built up some. The problem is I kill things too fast in daily quests for it to register to use that.

Currently warlocks use drain soul under the following conditions: 

Must be affliction 
Target must be at or below 15% (which I should change to 25%, thats when the 4x damage kicks in)
Must have less than 26 soul shards

PR #1518, opened 2025-08-10
Ignore GameObject IDs from vanilla dungeons

HennyWilly:
While leveling a new character, I noticed bots interacting with certain objects in vanilla dungeons that they shouldn't. 
This PR prevents that by adding their IDs to `AiPlayerbot.DisallowedGameObjects`.
The list may not be exhaustive.

**Added GameObject IDs:**
-  ~~**Large Solid Chest** ([153464](https://www.wowhead.com/wotlk/object=153464))~~
  ~~Located in Blackrock Spire and Stratholme.~~
- **Witch Doctor's Chest** ([141596](https://www.wowhead.com/wotlk/object=141596))
  Located in Zul'Farrak next to `Witch Doctor Zum'rah`. Approaching him will turn him hostile.
- **Relic Coffer** ([160836](https://www.wowhead.com/wotlk/object=160836)) and **Dark Coffer** ([160845](https://www.wowhead.com/wotlk/object=160845))
  Located in the vault room inside Blackrock Depths. Bots will loot them while the player is still unlocking the respective vault door.
- **Fengus's Chest** ([179516](https://www.wowhead.com/wotlk/object=179516))
  Located in Dire Maul North. Bots will loot the `Gordok Courtyard Key` inside.
- **Supply Crate** ([176224](https://www.wowhead.com/wotlk/object=176224),[181085](https://www.wowhead.com/wotlk/object=181085))
  Located in Stratholme. These are the "real" ones that can contain `Stratholme Holy Water`. Bots ignore the trapped crates ([176304](https://www.wowhead.com/wotlk/object=176304),[176307](https://www.wowhead.com/wotlk/object=176307),[176308](https://www.wowhead.com/wotlk/object=176308),[176309](https://www.wowhead.com/wotlk/object=176309))
- **Malor's Strongbox** ([176112](https://www.wowhead.com/wotlk/object=176112))
  Located in Stratholme. Bots will loot the `Medallion of Faith` inside.
- **Shallow Grave** ([128308](https://www.wowhead.com/wotlk/object=128308),[128403](https://www.wowhead.com/wotlk/object=128403))
  Located in Zul'Farrak. Spawns elite trolls. Opening multiple graves at the same time is dangerous for the group.
- **Grim Guzzler Boar** ([165739](https://www.wowhead.com/wotlk/object=165739)) and **Dark Iron Ale Mug** ([165738](https://www.wowhead.com/wotlk/object=165738))
  Located in the Grim Guzzler inside Blackrock Depths. Interacting with them a specific amount of times will turn the whole inn hostile.
- **Father Flame** ([175245](https://www.wowhead.com/wotlk/object=175245))
  Located in the "egg room" inside Upper Blackrock Spire. Interacting with it will spawn waves of enemies.
- **Unforged Runic Breastplate** ([175970](https://www.wowhead.com/wotlk/object=175970))
  Located in Upper Blackrock Spire, after the "egg room". This object is interactable even if the bot doesn't have [the corresponding quest](https://www.wowhead.com/wotlk/quest=5127). The object despawns afterwards...
- **Blacksmithing Plans** ([176325](https://www.wowhead.com/wotlk/object=176325),[176327](https://www.wowhead.com/wotlk/object=176327))
  Located in Stratholme. Interacting with these objects will each spawn an elite enemy.

I verified that bots no longer interact with the listed objects in my local tests. Further verification by others is welcome.

This is my first PR here, so feedback is appreciated 😄

### DIFF
--- a/src/PlayerbotAIConfig.cpp
+++ b/src/PlayerbotAIConfig.cpp
@@ -160,7 +160,7 @@ bool PlayerbotAIConfig::Initialize()
     LoadSet<std::set<uint32>>(
         sConfigMgr->GetOption<std::string>("AiPlayerbot.DisallowedGameObjects",
                                            "176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,2857,"
-                                           "179490,153464,141596,160836,160845,179516,176224,181085,176112,128308,128403,"
+                                           "179490,141596,160836,160845,179516,176224,181085,176112,128308,128403,"
                                            "165739,165738,175245,175970,176325,176327,123329"),
         disallowedGameObjects);
     botAutologin = sConfigMgr->GetOption<bool>("AiPlayerbot.BotAutologin", false);


HennyWilly:
Update: Removed the trapped supply crates (176304,176307,176308,176309). Bots ignore these.

hermensbas:
Could someone with more ingame knowlegde verify these? :)

brighton-chi:
Thanks for doing this. I want to raise a question for the community regarding the purpose of the default IDs. Are we wanting to stop bots from interacting with game objects in instances entirely, unless commanded otherwise? Some things like preventing bots from spawning ZF zombies or looting profession quest objects clearly should be on the list. But how expansive should we be? Do we really want to exclude standard loot chests, as opposed to having them operate in accordance with the looting method?

In any case, we should have the default in the dist match the default in the source. It looks like the source currently has the Defias Gunpowder and Blood of Heroes as the only default exclusions so it needs to be updated regardless.

HennyWilly:
@brighton-chi Thanks for the feedback. I updated the default value in the source.

hermensbas:
I like the current black list approach, i would prefer keeping the list as small as possible. Only those who that break quests or always create a situation you gonna wipe 95% of the time.

When it comes to more dangerious chest spawning elites, thats kinda interesting. Preferable this list would be two list.

config:
InteractionItemListNever
InteractionItemListOnChance
InteractionItemOnChanceValue
 

Having bots opening chest which spawns elites there and there at random would be fun and one would expect in real life as well. Maybe something for followup feature?

brighton-chi:
The current list is pretty arbitrary. Here is what it contains:

AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,17155
QuestItems: 176213 = Blood of Heroes, 17155 = Defias Gunpowder, 2656 = Waterlogged Envelope
Chests: Large Solid Chest = 74448, Box of Assorted Parts = 19020, Food Crate = 3719, Water Barrel = 3658, Barrel of Milk = 3705, Barrel of sweet Nectar = 3706, Tattered Chest = 105579, Large bettered Chest = 75293, Solid Chest = 2857, Battered Foodlocker = 179490

The list should definitely contain 17155 since that's an item needed to proceed through the instance (and I think was the inspiration for this feature in the first place). And 2656 starts a quest--I'm not sure if it despawns upon looting but if so, it definitely makes sense to exclude. But the rest of the items are just general interactable game objects. And they're not comprehensive--for example, 179490 is on there, and that's for a lockpicking footlocker in the Badlands. But there are lots of different lockpicking footlockers in various zones--why exclude only the Badlands ones? I don't know if the bots can even open lockpicking footlockers.

If we want to make it more limited, we should keep 17155 (though I note it's on the list twice so one should be deleted), 2656, and arguably 176213, which spawns elites and is a quest item, albeit one with multiple spawns in EPL so it's not quite as gamebreaking. Some of the additions in this PR are definitely worthy of being added because it is gamebreaking for bots to loot--for example, 175970 is for a blacksmithing quest and for some reason it has always been lootable by players not on the quest or even players without blacksmithing, and it can be looted only once per instance. IMO things like the ZF shallow graves should also be excluded because if we don't we are basically having the bots intentionally play badly (the point during that fight is not to accidentally open the graves). I'm not opposed to having a separate config option that lets the bots be Leeroy Jenkins if somebody wants, but in the meantime, I think we should exclude anything that would be stupid behavior for a player to loot randomly.

In my opinion, the best default list from what we currently have is to keep only 176213, 17155, and 2656 from the current list (meaning delete everything else that's on the list already) and then add everything in this PR except for 153464, which is just a standard loot chest in an instance AFAIK. There surely are other things that should also be added since this PR is only for vanilla; I'm just looking at everything that is already included and everything being proposed.

Tierisch:
A number of those low level objects containing food and drink are ones that bots will often go for and set off terminal warnings about possible hacking when multiple try it at the same time (since it's despawned by the time the second gets it).

I don't have a personal preference about whether they're in the blacklist by default, but I'd add them to my own to avoid the spam until they can be made to not try opening things someone else is already using.

HennyWilly:
I’m not very familiar with the codebase yet, but the crate-related warnings seem like they might stem from a deeper underlying issue that’s outside the scope of this PR. It might be better handled in a separate issue for targeted discussion, where it can also be decided whether the crates should be removed entirely.
For this PR, I’d prefer to focus on objects that break the instance flow, break quests or spawn enemies, and leave crates (including 153464) for separate consideration.

hermensbas:
> A number of those low level objects containing food and drink are ones that bots will often go for and set off terminal warnings about possible hacking when multiple try it at the same time (since it's despawned by the time the second gets it).
> 
> I don't have a personal preference about whether they're in the blacklist by default, but I'd add them to my own to avoid the spam until they can be made to not try opening things someone else is already using.

This might indicate a bug, adding small randomizer and check before actual should fix that problem.

hermensbas:
Ive added: https://github.com/liyunfan1223/mod-playerbots/pull/1559

hermensbas:
> The current list is pretty arbitrary. Here is what it contains:
> 
> AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,17155 QuestItems: 176213 = Blood of Heroes, 17155 = Defias Gunpowder, 2656 = Waterlogged Envelope Chests: Large Solid Chest = 74448, Box of Assorted Parts = 19020, Food Crate = 3719, Water Barrel = 3658, Barrel of Milk = 3705, Barrel of sweet Nectar = 3706, Tattered Chest = 105579, Large bettered Chest = 75293, Solid Chest = 2857, Battered Foodlocker = 179490
> 
> The list should definitely contain 17155 since that's an item needed to proceed through the instance (and I think was the inspiration for this feature in the first place). And 2656 starts a quest--I'm not sure if it despawns upon looting but if so, it definitely makes sense to exclude. But the rest of the items are just general interactable game objects. And they're not comprehensive--for example, 179490 is on there, and that's for a lockpicking footlocker in the Badlands. But there are lots of different lockpicking footlockers in various zones--why exclude only the Badlands ones? I don't know if the bots can even open lockpicking footlockers.
> 
> If we want to make it more limited, we should keep 17155 (though I note it's on the list twice so one should be deleted), 2656, and arguably 176213, which spawns elites and is a quest item, albeit one with multiple spawns in EPL so it's not quite as gamebreaking. Some of the additions in this PR are definitely worthy of being added because it is gamebreaking for bots to loot--for example, 175970 is for a blacksmithing quest and for some reason it has always been lootable by players not on the quest or even players without blacksmithing, and it can be looted only once per instance. IMO things like the ZF shallow graves should also be excluded because if we don't we are basically having the bots intentionally play badly (the point during that fight is not to accidentally open the graves). I'm not opposed to having a separate config option that lets the bots be Leeroy Jenkins if somebody wants, but in the meantime, I think we should exclude anything that would be stupid behavior for a player to loot randomly.
> 
> In my opinion, the best default list from what we currently have is to keep only 176213, 17155, and 2656 from the current list (meaning delete everything else that's on the list already) and then add everything in this PR except for 153464, which is just a standard loot chest in an instance AFAIK. There surely are other things that should also be added since this PR is only for vanilla; I'm just looking at everything that is already included and everything being proposed.

Shall we merge this initial setup and take it from there?

PR #1521, opened 2025-08-10
Playerbots/LFG: fix false “not eligible” & dungeon 0/type 0, add clear diagnostics

Wishmaster117:
### Summary
This PR makes the Playerbots module play nicely with the WotLK LFG flow after logout/teleport and during early queue initialization . It prevents the two common user-visible issues:

- “One or more members are not eligible…” even though the group looks fine.
- issue: https://github.com/liyunfan1223/mod-playerbots/issues/1509
- Wrong dungeon type 0 for dungeon 0 right after clicking Random Dungeon.

It also adds lightweight logs to see exactly what the LFG check is evaluating.

### What changed

1. Reset transient LFG state when a bot logs in

- In PlayerbotHolder::OnBotLogin:

```
// after ResetStrategies(...) and before DB load
botAI->Reset(true); // clears "lfg proposal"/role-check and other transient flags
sPlayerbotDbStore->Load(botAI);
```

- Rationale: logging out in the middle of a proposal/role-check could leave stale flags that block the next queue.

2. Keep 5-man groups as “party” (don’t convert to raid at 5)

- In the group-attach path:

```
if (mgroup->GetMembersCount() + 1 > 5) {
  mgroup->ConvertToRaid();
}
mgroup->AddMember(bot);
```

- Rationale: LFG requires a party (not a raid). Converting to raid at exactly 5 caused early rejections and bypassed our LFG hook entirely.

3. Robust LFG membership check with clear instrumentation

- In OnPlayerbotCheckLFGQueue(lfg::Lfg5Guids const& guidsList):

- Ignore placeholders: empty GUIDs and non-player GUIDs (except group GUIDs).
- Count real players, bots, and offline members; log a short per-slot line and a summary.
- Hybrid policy:

- If we only see placeholders (very early frame), do not veto — let the core finish initializing (prevents dungeon 0/type 0).
- Otherwise, allow only if no offline members and (at least one real player OR a group GUID) — preserves original behavior.

- If we do see real members and at least one offline, we perform a soft, inlined preflight reset (just like users were doing manually by toggling bots off/on) and return false once; the next click succeeds:

```
// inside OnPlayerbotCheckLFGQueue, only when offline>0 and not placeholders-only
for (GroupReference* ref = g->GetFirstMember(); ref; ref = ref->next()) {
  if (Player* member = ref->GetSource())
    if (PlayerbotAI* ai = GET_PLAYERBOT_AI(member))
      ai->Reset(true); // clears proposal/role-check etc.
}
return false; // small, deliberate retry after reset
```

4. Logging nits (MSVC/fmt)

- When logging ObjectGuid, we cast to uint64 to avoid the fmt “unformattable type” error:

`static_cast<uint64>(guid.GetRawValue())`

### Why this fixes the issues

- “Not eligible” after relog: stale per-bot LFG flags (proposal/role-check) are cleared at bot login; the queue no longer inherits dirty state.
- dungeon 0/type 0: early LFG checks sometimes pass placeholders; we now don’t veto in that specific frame, allowing the core to finish populating the request before validation.
- Raid vs Party: keeping a 5-man as a party ensures the LFG path is taken and our check actually runs.

### User impact

- Players can queue Random Dungeon with their bot party immediately after login/teleport, without having to toggle bots offline/online.

### Testing done

- 5-man party (leader + 4 bots), queue Random LK Dungeon → proposal appears normally.
- Logout/login of the leader, immediate queue → OK (no “not eligible”).
- Teleport Dalaran → Stormwind, immediate queue → OK.
- Forcing an “offline” member during the check triggers one soft reset and a clean retry.

### Risks

- Minimal. The gameplay rules aren’t changed; we only:
- reset transient state on bot login,
- avoid converting to raid at 5,
- ignore placeholder GUIDs in the hook,
- and perform a one-shot soft reset if offline is detected.

Thanks for reviewing!

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -529,6 +529,9 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
     {
         botAI->ResetStrategies(!sRandomPlayerbotMgr->IsRandomBot(bot));
     }
+
+    botAI->Reset(true);  // Reset transient states (incl. LFG "proposal") to avoid the "one or more players are not eligible" error after reconnect.
+
     sPlayerbotDbStore->Load(botAI);
 
     if (master && !master->HasUnitState(UNIT_STATE_IN_FLIGHT))
@@ -548,16 +551,21 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
     if (master && master->GetGroup() && !group)
     {
         Group* mgroup = master->GetGroup();
-        if (mgroup->GetMembersCount() >= 5)
+        // if (mgroup->GetMembersCount() >= 5)
+        if (mgroup->GetMembersCount() + 1 > 5)  // only convert in raid if the add of THIS bot make group > 5
         {
             if (!mgroup->isRaidGroup() && !mgroup->isLFGGroup() && !mgroup->isBGGroup() && !mgroup->isBFGroup())
             {
                 mgroup->ConvertToRaid();
             }
-            if (mgroup->isRaidGroup())
-            {
-                mgroup->AddMember(bot);
-            }
+            //if (mgroup->isRaidGroup())
+            //{
+                //mgroup->AddMember(bot);
+            //}
+            mgroup->AddMember(bot);
+
+            LOG_DEBUG("playerbots", "[GROUP] after add: members={}, isRaid={}, isLFG={}",
+                      (int)mgroup->GetMembersCount(), mgroup->isRaidGroup() ? 1 : 0, mgroup->isLFGGroup() ? 1 : 0);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


Wishmaster117:
hermensbas added the **Waiting to be Tested** label 2 minutes ago
You don’t take my word for it? :D :D :D

Regrad:
I just want to express my gratitude to @Wishmaster117. We have already lost hope that all this will ever be fixed. I hope that you will continue to fix those difficult and very important problems! Thank you!

Wishmaster117:
It is a pleasure for me to help the community to the extent of my possibilities and skills. You should know that the last time I touched the cpp was during my engineering studies at the university, so for a very long time... what I could do I would do with pleasure in my free time.

Issue #1528, opened 2025-08-11
Crash in RandomPlayerbotMgr:1018


VincentVanclef:
This one is weird, as i would have assumed the player->InBattleground check would make sure it doesnt crash but it did.

Perhaps changing this

```
                if (player->InBattleground())
                {
                    std::vector<uint32>* instanceIds = nullptr;
                    uint32 instanceId = player->GetBattleground()->GetInstanceID();

                    instanceIds = &BattlegroundData[queueTypeId][bracketId].bgInstances;
                    if (instanceIds &&
                        std::find(instanceIds->begin(), instanceIds->end(), instanceId) == instanceIds->end())
                        instanceIds->push_back(instanceId);

                    BattlegroundData[queueTypeId][bracketId].bgInstanceCount = instanceIds->size();
                }
```

to

```
                if (Battleground const* bg = player->GetBattleground())
                {
                    std::vector<uint32>* instanceIds = nullptr;
                    uint32 instanceId = bg->GetInstanceID();

                    instanceIds = &BattlegroundData[queueTypeId][bracketId].bgInstances;
                    if (instanceIds &&
                        std::find(instanceIds->begin(), instanceIds->end(), instanceId) == instanceIds->end())
                        instanceIds->push_back(instanceId);

                    BattlegroundData[queueTypeId][bracketId].bgInstanceCount = instanceIds->size();
                }
```

Crashlog:
```
Thread 1 "worldserver" received signal SIGSEGV, Segmentation fault.
RandomPlayerbotMgr::CheckBgQueue (this=<optimized out>) at /home/ubuntu/rtgcore/modules/mod-playerbots/src/RandomPlayerbotMgr.cpp:1018
1018	                    uint32 instanceId = player->GetBattleground()->GetInstanceID();
#0  RandomPlayerbotMgr::CheckBgQueue (this=<optimized out>) at /home/ubuntu/rtgcore/modules/mod-playerbots/src/RandomPlayerbotMgr.cpp:1018
#1  0x0000555557cec360 in RandomPlayerbotMgr::instance()::instance ()
#2  0x0100000400000000 in ?? ()
#3  0x0000555557cec358 in RandomPlayerbotMgr::instance()::instance ()
#4  0x0000000000000000 in ?? ()
#0  RandomPlayerbotMgr::CheckBgQueue (this=<optimized out>) at /home/ubuntu/rtgcore/modules/mod-playerbots/src/RandomPlayerbotMgr.cpp:1018
[...]

PR #1530, opened 2025-08-11
[Fix issue #1528] Close small window where the “in a BG/arena” state can change between the check (InBattleground() / InArena()) and grabbing the pointer (GetBattleground()), which leads to a null dereference.

Wishmaster117:
This PR will fix issue https://github.com/liyunfan1223/mod-playerbots/issues/1528.

There’s a small window where the “in a BG/arena” state can change between the check (InBattleground() / InArena()) and grabbing the pointer (GetBattleground()), which leads to a null dereference.
The robust fix is to fetch the bg pointer once, check it for null, and then use it.

### Why this fixes the crash

- InBattleground() / InArena() return a state that can change immediately after the check (e.g. switching to STATUS_WAIT_LEAVE, teleport, etc.).
- GetBattleground() can therefore return nullptr even if the previous check was true → immediate crash on dereference.
- By first grabbing the bg pointer and checking that it is not null, we eliminate this TOCTOU (time-of-check / time-of-use).

### DIFF
--- a/src/BroadcastHelper.cpp
+++ b/src/BroadcastHelper.cpp
@@ -947,8 +947,21 @@ bool BroadcastHelper::BroadcastSuggestThunderfury(PlayerbotAI* ai, Player* bot)
     {
         std::map<std::string, std::string> placeholders;
         ItemTemplate const* thunderfuryProto = sObjectMgr->GetItemTemplate(19019);
-        placeholders["%thunderfury_link"] = GET_PLAYERBOT_AI(bot)->GetChatHelper()->FormatItem(thunderfuryProto);
-
+        // placeholders["%thunderfury_link"] = GET_PLAYERBOT_AI(bot)->GetChatHelper()->FormatItem(thunderfuryProto); // Old code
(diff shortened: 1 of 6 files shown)


hermensbas:
I will test/review it tonight, 

Was the the commit 'Harden playerbot logout & packet dispatch; add null-safety in chat ho' also intended in this PR? Nevermind the commit question i see what you did there, you made the branch from your other fix to prevent merging issue. 

note: can be merged after https://github.com/liyunfan1223/mod-playerbots/pull/1529 is merged

Wishmaster117:
> I will test/review it tonight,
> 
> Was the the commit 'Harden playerbot logout & packet dispatch; add null-safety in chat ho' also intended in this PR? Nevermind the commit question i see what you did there, you made the branch from your other fix to prevent merging issue.
> 
> note: can be merged after #1529 is merged

I’m an old guy — I used to code on typewriters, and I’m still not 100% familiar with this darn GitHub. :D
I made this branch 'Harden playerbot logout & packet dispatch; add null-safety in chat ho', after that i made an other one from it and anotherone from the other one, So you have to merge :
1529 and after 1530 and after 1531

PR #1532, opened 2025-08-11
[Fix issue #1527] : startup crash in tank target selection — add TOCTOU & null-safety guards

### LINKED ISSUE #1527: crash at the startup
**Bug Description**
Crash at the loading of the server

**Commit Hash**
commit as of today : 11/08/2025

**Screenshots**
(https://drive.google.com/file/d/1yxFiuOzF2zlBpI9eZKdvTmHSVJ1OJXdl/view?usp=sharing)

**Modules**
AHBOT

**Playerbot Settings**
400

**System**
OS: Windows 11

Wishmaster117:
### Summary
This PR fixes a crash that could occur at server startup when the tank target value is computed during the first AI ticks. The root cause was a TOCTOU window and missing null/“in world” checks while iterating attackers and reading threat/victim data.

**Changes are limited to:**

- strategy/values/TargetValue.cpp
- strategy/values/TankTargetValue.cpp

### Why
Early in the boot sequence, some Unit/threat structures can be not yet fully “in world” or in the middle of removal. Without guards, we could dereference:

- an attacker that just left the world,
- a currentVictim that is null/stale,
- a previously cached result that became invalid.

This led to sporadic segfaults/AVs reported in #1527.

### What changed
**TargetValue.cpp**

- Harden FindTarget(FindTargetStrategy*):

- Guard the AI context.
- For each attacker GUID, skip units that are nullptr, not alive, not in world, or during remove from world.
- Only pass safe Unit* to strategy->CheckAttacker(...).

Example:
```
if (!unit || !unit->IsAlive() || !unit->IsInWorld() || unit->IsDuringRemoveFromWorld())
    continue;
strategy->CheckAttacker(unit, &unit->GetThreatMgr());
```

**TankTargetValue.cpp**

###  FindTargetForTankStrategy::CheckAttacker(...) (kept in sync even if not the default):

- Guard creature, threatMgr, and bot.
- Filter units not in world / during removal.
- If the current result became invalid, reset it before comparison.
- Safely read currentVictim (null-check before use).

### -FindTankTargetSmartStrategy::CheckAttacker(...) (used by Calculate()):

- Guard attacker (null, dead, not in world, during removal).
- Keep raid icon `#4` logic, but only after attacker validation.
- If cached result is invalid, clear it.
- Then run the existing IsBetter(...) comparison.

###  FindTankTargetSmartStrategy::IsBetter(...):

- Guard new_unit/old_unit (alive, in world, not during removal).
- Guard bot before distance/threat queries.
- Keep existing interval/threat/distance logic.

### FindTankTargetSmartStrategy::GetIntervalLevel(...):

- Guard unit and bot (alive, in world) before melee/aggro checks.

`No intentional behavioral change once the world is stabilized; we only avoid dereferencing transient/invalid pointers during boot and fast transitions.`

### Impact

- During initialization or heavy transitions, the strategies may now return no target instead of crashing. Callers are expected to handle a nullptr target (common pattern).
- Negligible performance cost (simple boolean checks).

### Testing

- Repeated startups with AHBOT and ~400 bots: no crash at load.
- Stress with rapid teleports/login/logout while re-evaluating tank targets: no invalid dereferences.
- Multi-tank scenarios (raid icon focus, current target preference) behave as before once units are in world.

### Follow-ups

- Apply the same guard pattern to any other strategies/values that read threat/victim data or rely on units that can be mid-removal.

Fixes #1527.

### DIFF
--- a/src/strategy/values/TankTargetValue.cpp
+++ b/src/strategy/values/TankTargetValue.cpp
@@ -14,7 +14,7 @@ class FindTargetForTankStrategy : public FindNonCcTargetStrategy
 public:
     FindTargetForTankStrategy(PlayerbotAI* botAI) : FindNonCcTargetStrategy(botAI), minThreat(0) {}
 
-    void CheckAttacker(Unit* creature, ThreatMgr* threatMgr) override
+    /*void CheckAttacker(Unit* creature, ThreatMgr* threatMgr) override
     {
         if (!creature || !creature->IsAlive())
         {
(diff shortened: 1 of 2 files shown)


hermensbas:
I am first gonna test the first two commit, before merging this one

PR #1538, opened 2025-08-11
[Fix] teleport to invalid map or invalid coordinates (x , y , z  200000, o ) given when teleporting player (g UI d full type player low , name , map , x , y , z , o ) 

Wishmaster117:
### Playerbots: guard against invalid-Z teleports

Avoid spam in console when bots decide to do a silly teleport:

Like:
```
TeleportTo: invalid map (0) or invalid coordinates (X: -940.0883, Y: -2843.7368, Z: -200000, O: 7.331777) given when teleporting player (GUID Full: 0x00000000000002ed Type: Player Low: 749, name: Ozuy, map: 0, X: -940.7874, Y: -2844.9495, Z: 61.477768, O: 7.331777).
TeleportTo: invalid map (0) or invalid coordinates (X: -11621.91, Y: 203.48108, Z: -200000, O: 3.2240791) given when teleporting player (GUID Full: 0x0000000000000289 Type: Player Low: 649, name: Justinor, map: 0, X: -11620.515, Y: 203.5972, Z: 23.282, O: 3.2240791).
TeleportTo: invalid map (1) or invalid coordinates (X: -7499.1714, Y: -3402.504, Z: -200000, O: 3.1415927) given when teleporting player (GUID Full: 0x000000000000022f Type: Player Low: 559, name: Geronavan, map: 1, X: -7497.7715, Y: -3402.5046, Z: 12.673878, O: 3.1415927).
TeleportTo: invalid map (571) or invalid coordinates (X: 8303.741, Y: 2736.2776, Z: -200000, O: 5.7079563) given when teleporting player (GUID Full: 0x0000000000000243 Type: Player Low: 579, name: Gantun, map: 571, X: 8302.567, Y: 2737.0388, Z: 653.82086, O: 5.7079563).
```

### Summary:
Stops bots from spamming TeleportTo with invalid height (Z = -200000). Adds module-level safe wrappers and updates call sites so bad teleports are skipped instead of flooding logs or looping.

### What changed

- Added TeleportToSafe(...) in Playerbots.h:

- Validates Z; if invalid, tries one GetHeight(...) on the same map; otherwise aborts.
- Overloads for (mapId, x, y, z, o), WorldPosition (by value), and Position.

- Replaced bot ->TeleportTo(...) calls with TeleportToSafe(...) in affected actions (e.g., ReviveFromCorpseAction, ICC gunship helper, Ulduar Kologarn block).

### Why

- -200000 is the “invalid height” sentinel; passing it to Player::TeleportTo caused console spam and retry loops.

### Scope

- Module-only; no core edits. Affects bots’ teleports only.

### Testing

- Reproduced previous spam case; with this patch, bots skip invalid teleports, logs are clean, normal teleports still work.

### DIFF
--- a/src/BotMovementUtils.h
+++ b/src/BotMovementUtils.h
@@ -4,27 +4,31 @@
  */
  
 #pragma once
+#include "Unit.h"
 #include "Player.h"
 #include "MotionMaster.h"
-#include "Unit.h"
 
-inline bool CanStartMoveSpline(Unit* u)
-{
+inline bool CanStartMoveSpline(Player* bot) {
+    if (!bot) return false;
+    if (!bot->IsAlive()) return false;
+    if (bot->IsBeingTeleported() || bot->IsInFlight()) return false;
+    if (bot->HasUnitState(UNIT_STATE_LOST_CONTROL) || bot->HasRootAura() ||
+        bot->HasStunAura() || bot->IsCharmed() || bot->isFrozen() || bot->IsPolymorphed())
+        return false;
+    if (bot->GetMotionMaster()->GetMotionSlotType(MOTION_SLOT_CONTROLLED) != NULL_MOTION_TYPE)
+        return false;
+    if (bot->GetSpeed(MOVE_RUN) <= 0.01f) return false;
+    return true;
+}
+
+inline bool CanStartMoveSpline(Unit* u) {
     if (!u) return false;
     if (!u->IsAlive()) return false;
-
-    // states that block movement
     if (u->HasUnitState(UNIT_STATE_ROOT | UNIT_STATE_STUNNED | UNIT_STATE_CONFUSED | UNIT_STATE_FLEEING))
         return false;
-
-    // no spline if a "CONTROLLED" movement is in progress (knockback/fear, etc.)
     if (u->GetMotionMaster()->GetMotionSlotType(MOTION_SLOT_CONTROLLED) != NULL_MOTION_TYPE)
         return false;
-
-    // MoveSpline constraint: speed > 0.01f
-    if (u->GetSpeed(MOVE_RUN) <= 0.01f)
-        return false;
-
+    if (u->GetSpeed(MOVE_RUN) <= 0.01f) return false;
     return true;
 }
 
--- a/src/BotMovementUtils.h
+++ b/src/BotMovementUtils.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license, you may redistribute it
+ * and/or modify it under version 2 of the License, or (at your option), any later version.
+ */
+ 
+#pragma once
+#include "Player.h"
+#include "MotionMaster.h"
+#include "Unit.h"
+
+inline bool CanStartMoveSpline(Unit* u)
+{
+    if (!u) return false;
+    if (!u->IsAlive()) return false;
+
+    // states that block movement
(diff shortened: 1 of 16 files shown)


noisiver:
Why create a new function instead of changing existing teleport functions to address the issue? It feels like it's just filling the codebase up for the sake of filling it up.

hermensbas:
> Why create a new function instead of changing existing teleport functions to address the issue? It feels like it's just filling the codebase up for the sake of filling it up.

 Player->TeleportTo is a core implementation

Wishmaster117:
**Why a new function instead of changing teleport itself?**

- **Scope & risk:** The core TeleportTo is used by players, GMs, scripts, transports, BGs, etc. Changing it risks regressions everywhere. The wrapper confines the behavior to bots only.
- **Upstream-friendly:** We keep the engine API untouched, so merging/updating from upstream stays clean. Our change lives entirely in the module.
- **Single responsibility:** Invalid-Z handling is a bot policy, not an engine rule. It belongs in the playerbots module.
- **Reversible & auditable:** One small wrapper + targeted call-sites = easy to review, easy to revert if needed. No semantic drift in the core.
- **Safety without side effects:** We stop bot spam and loops without altering legitimate teleports for real players.
- **Path to core later:** If there’s consensus, we can upstream a core guard later. The wrapper is a low-risk proof that the fix works.

 In short: minimal blast radius, zero core churn, and we fix the bot problem without creating new ones.
 
Also, about the module code itself, parts of it are dated. It works, but it deserves a light refresh. Pragmatic plan (incremental, zero core changes):

- Const-correctness & API: add const where it belongs (or take by value like we did for WorldPosition); replace C-style casts with static_cast.
- Macros → constexpr/inline: swap utility #defines (angles, helpers) for constexpr and small inline functions.
- Logging hygiene: avoid noisy LOG_ERROR; use DEBUG with a simple rate-limit (counter/timestamp) to prevent spam loops.
- Backoff/throttle: keep a short delay (≈ 500–1500 ms) after worldports to break rapid retries.
- Dedicated helpers: centralize “bot policy” (e.g., height validation) in a tiny, tested utility instead of scattering logic.
- clang-tidy/format: enable modernize-, readability-, performance- checks and enforce consistent formatting.
- Strong enums: replace magic flags with enum class where possible.

Net result: a cleaner, easier-to-maintain playerbots module—no core churn, fewer regressions.

DustinHendrickson:
This is a lot of LLM changes to be merged in so quickly.

Wishmaster117:
> This is a lot of LLM changes to be merged in so quickly.

You’re probably right, and I really appreciate the feedback.
That said, we’ve been seeing more and more crashes lately, and I’m trying to tackle them — any help from you would be greatly appreciated.
At the moment, it seems that only @hermensbas is actively reviewing, so until we have more people involved, I’ll slow things down a bit to avoid having too many LLM changes merged too quickly.

I’ll still keep digging into the crash logs and sharing potential fixes, so we can be ready to move faster once there are more hands on deck. 🚀

hermensbas:
Its abit risky, but knowingly that Wishmaster knows what he is doing. I just barely know enough about c++ to review what the changes doing and why. Having multiple ppl involved on the testing part. I rather keep the momentum. 

These problems been there for ages with no ppl able to solve them in a correct way. Its abit tricky, hence i am not merging other PR other then strongly related to these type of fixes. So in worst we could revert, although i rather see us learning from the fixes and take it from there.

I have my eyes on issue listst,l and i am testing constantly on general stability in various situations.

In all honosty i am just happy Wishmaster has the time and expertise and will to fix these type of issues.

Issue #1545, opened 2025-08-12
Crash report on server restart


Regrad:
**Bug Description**
I do not know if this needs to be fixed, but I assume that it could potentially cause a crash not only during reboot.

**Commit Hash**
3fff58df1a2058894e9b758be07869aec87c2c70

**How To Reproduce Bug**
Restart the server (server restart 1)

**Expected Behavior**
The server should reboot without crashes.

**Screenshots**
-

**Modules**
Transmog

**Playerbot Settings**
2000 bot

**System**
OS: Windows 10
Hardware: Xeon 22 cores

**Additional Context**
@Wishmaster117

[e801e623210c_worldserver.exe_\[12-8_14-2-8\].txt](https://github.com/user-attachments/files/21734916/e801e623210c_worldserver.exe_.12-8_14-2-8.txt)

Wishmaster117:
The crash is an ACCESS_VIOLATION in Object::HasFlag (Object.cpp:885), called during playerbot logout, via PlayerbotHolder::LogoutPlayerBot (PlayerbotMgr.cpp:525) then LogoutAllBots (PlayerbotMgr.cpp:379). This is exactly the sequence of a restart where the module attempts to disconnect the bots. e801e623210c_worldserve…

In the local variables, the bot in question (Player* bot named Heraen) has its update-value pointers corrupted:
m_int32Values = 0x30489
m_uint32Values = 0x30489

These addresses are clearly invalid (it’s the low GUID, not a real address), indicating a use-after-free / double logout: the Player object has already been partially destroyed (or removed from the world) when it’s still being used to read a flag, hence the crash in HasFlag. The stack and line numbers indicate this happens in the OnPlayerbotLogout → LogoutAllBots → LogoutPlayerBot phase during the restar…

Solution: During your automated restarts, force a bot logout ahead of time (via command/module) and add a short delay before the worldserver shutdown/restart to let the disconnections finish.

Wishmaster117:
I'me coding a fix but there  is a lot of LLM changes.....

hermensbas:
> I'me coding a fix but there is a lot of LLM changes.....

says who?

kadeshar:
Could check that crash still exists when you update azerothcore?
You can use for example this fork https://github.com/kadeshar/azerothcore-wotlk/tree/Playerbot-updated

kadeshar:
Now update is part of main fork. Server still crashes after update?

PR #1547, opened 2025-08-12
Fix] crash report on server restart issue #1545

Wishmaster117:
**Simplify PlayerbotHolder::LogoutPlayerBot to make restarts safe and deterministic Issue: https://github.com/liyunfan1223/mod-playerbots/issues/1545**

### Summary
This PR simplifies PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid) and hardens LogoutAllBots() to eliminate a restart-time crash (use-after-free) while keeping gameplay behavior and persistence exactly the same. The function is now shorter, easier to reason about, and more robust under concurrent shutdown conditions.

### Background and root cause
During server restarts the module sometimes read runtime flags on a Player (HasFlag, HasUnitState, etc.) while that object was already being torn down elsewhere. That led to use-after-free and ACCESS_VIOLATION crashes. In practice, the function also ended up forcing instant logout anyway, which made the prior “compute logout conditions” code both redundant and risky.

### What changed (high level)

- Always use instant logout for bots. We remove conditional reads of Player state (no more HasFlag/HasUnitState during teardown).
- Preserve all saves and side effects. We still save bot AI data (when applicable) and call SaveToDB before destroying the session.
- Deterministic teardown order. We:

1. notify the master,
2. clear AI context pointers defensively,
3. remove the bot from internal maps,
4. call LogoutPlayer(true),
5. delete the bot’s WorldSession.

- Companion hardening in LogoutAllBots(). We iterate over a snapshot of GUIDs (not Player*) and re-lookup each bot, with a reentrancy guard, to avoid stale pointers and double-logout.

### Why this is better

- Crash-proof under restart pressure. No more reading fields from an object that might already be half-destroyed.
- No gameplay changes. Bots already ended up in instant logout. We simply make that explicit from the start.
- Persistence remains intact. AI data and characters are still saved before logout, just as before.
- Less code, fewer branches. Easier to audit, easier to maintain, and less surface area for race conditions.
- Deterministic order and ownership. We stop touching bot after removing it from maps and handing control to LogoutPlayer(true).

### Alternatives considered

- Keeping conditional logout and trying to “guard” every read of Player state.
=> Rejected: fragile and still prone to races during restart; also redundant since bots log out instantly anyway.

- Adding more locks around state checks.
=> Rejected: higher contention without eliminating the core hazard (reads during teardown).

### Risks & mitigations

- Risk: Behavior change if someone expected non-instant logout for bots.
 Mitigation: Previous code effectively forced instant logout; this codifies the real behavior and removes unsafe reads.

- Risk: Reentrancy/duplicate teardown.
Mitigation: LogoutAllBots() now uses GUID snapshots + re-lookup + a reentrancy guard; LogoutPlayerBot early-returns if a session is already logging out.

### Testing notes

- Restarted the server with many active bots: no crashes, no double-logout, characters and AI state persisted as expected.
- Verified master notification still occurs, and that no access to bot or botAI happens after RemoveFromPlayerbotsMap.

### Implementation sketch (before vs. after)
**Before:** compute logout using multiple HasFlag/HasUnitState reads → sometimes read freed memory → crash → then force instant logout anyway.
**After:** skip risky reads, force instant logout immediately, perform save + deterministic teardown. Same outcome, no crash.

This change makes the logout path **predictable and safe** under all shutdown scenarios, without altering gameplay or persistence semantics.

PS: Sorry for the  lot of LLM changes.
PS2: I commented the originals functions
....

PS6: Not till 2027

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -720,14 +720,16 @@ void PlayerbotAI::HandleTeleportAck()
         {
             bot->GetSession()->HandleMoveWorldportAck();
         }
-        // SetNextCheckDelay(urand(2000, 5000));
-		SetNextCheckDelay(urand(500, 1500)); // short delay to break bursts without hindering gameplay
+
         if (sPlayerbotAIConfig->applyInstanceStrategies)
             ApplyInstanceStrategies(bot->GetMapId(), true);
         EvaluateHealerDpsStrategy();
         Reset(true);
+		
+		SetNextCheckDelay(urand(300, 700)); // Safety : short delay after a worldport to prevent rapid re-teleports.
+		return;
     }
-
+    // If teleportNear : we stay cool and keep the standard GCD.
     SetNextCheckDelay(sPlayerbotAIConfig->globalCoolDown);
 }
 
--- a/src/PlayerbotAIBase.h
+++ b/src/PlayerbotAIBase.h
@@ -13,7 +13,6 @@ class PlayerbotAIBase
 {
 public:
     PlayerbotAIBase(bool isBotAI);
-
     bool CanUpdateAI();
     void SetNextCheckDelay(uint32 const delay);
     void IncreaseNextCheckDelay(uint32 delay);
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -336,34 +336,39 @@ void PlayerbotHolder::HandleBotPackets(WorldSession* session) // [Crash Fix] Sec
     }
 }
 
-void PlayerbotHolder::LogoutAllBots()
+// void PlayerbotHolder::LogoutAllBots() // Original function
+// {
+//     /*
+//     while (true)
+//     {
+//         PlayerBotMap::const_iterator itr = GetPlayerBotsBegin();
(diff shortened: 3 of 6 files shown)


Celandriel:
Commenting out the old code functions is Bad practice. It makes code review more difficult, and git maintains the history anyway.

Wishmaster117:
Don't merge yet, nedd to text some function.
If my teste succeed i clean the code

Wishmaster117:
> Commenting out the old code functions is Bad practice. It makes code review more difficult, and git maintains the history anyway.

Please, it’s exhausting when you come to give moral lessons instead of helping to make this mod work properly.
I’m doing my best to stop these random crashes from happening, so if I sometimes comment out code, it’s not the end of the world, is it?
Even with commented code, if you understand what is written and what it’s for, you’ll have no trouble seeing what has been done.

Celandriel:
Its about maintaining code quality. The majority of the PRs you have submitted are full of defunct code for no reason.

hermensbas:
> Its about maintaining code quality. The majority of the PRs you have submitted are full of defunct code for no reason.

what are you talking about?

Celandriel:
Entire sections commented out and then reimplemented later on in the code base.

PR #1559, opened 2025-08-13
Update playerbots.conf.dist

Raz0r1337:
Added Baelogs Chest (123329) to blocked GOs so that the chest cannot be stolen by bots and Duplicate entry removed

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1975,7 +1975,7 @@ AiPlayerbot.AllowedLogFiles = ""
 ####################################################################################################
 # A list of gameObject GUID's that are not allowed for bots to interact with.
 #
-AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,17155,2857,179490,123329
+AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,2857,179490,123329
 #
 # List of GUID's:
 # QuestItems:
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1975,11 +1975,11 @@ AiPlayerbot.AllowedLogFiles = ""
 ####################################################################################################
 # A list of gameObject GUID's that are not allowed for bots to interact with.
 #
-AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,17155,2857,179490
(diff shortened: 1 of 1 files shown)


brighton-chi:
> Looks good to me
> 
> 
> 
> Also agree on not updating the defaults in: https://github.com/liyunfan1223/mod-playerbots/blob/master/src/PlayerbotAIConfig.cpp

Why have the two not match?

Raz0r1337:
> Looks good to me
> 
> Also agree on not updating the defaults in: https://github.com/liyunfan1223/mod-playerbots/blob/master/src/PlayerbotAIConfig.cpp

What would be so wrong with changing the defaults in Playerbots.cpp? Personally, I didn't even know I could have done that, but if it makes sense, I think it makes more sense when the ID list is complete, or at least nearly complete.

What do the defaults in playerbot.cpp even exist for? The only thing I can imagine is that they serve as a fallback if the config file is missing or corrupted, so the server can simply load the defaults instead of crashing. Is that correct, and are there other reasons?

hermensbas:
> > Looks good to me
> > Also agree on not updating the defaults in: https://github.com/liyunfan1223/mod-playerbots/blob/master/src/PlayerbotAIConfig.cpp
> 
> Why have the two not match?

The question is why to have them matched. The whole idea of having config file is having the option to modify without changing the code. 

So the default provide two functionalities atm if not mistaken. Thats having a fallback that a application/game critical settings are always set and secondly being able to add custom config file with only those values that differ from the defaults.

This field not being part of critical path being able to run the game its in my opinion it isnt that relevent. But that being that it does not hurt either to modify the defaults in the code. 

Although now thinking of it maybe my second argument is good enough reason to add them. That being said they already differ atm: https://github.com/liyunfan1223/mod-playerbots/blob/8d51092d4234e7cb606c54f557930312c974d90c/src/PlayerbotAIConfig.cpp#L160

hermensbas:
Also consider: https://github.com/liyunfan1223/mod-playerbots/pull/1518

brighton-chi:
> Also consider: https://github.com/liyunfan1223/mod-playerbots/pull/1518

Yeah these two will need to be consolidated for sure.

I thought the default in the code won't matter unless the config is screwed up, such as if DisallowedGameObjects is deleted from the config. I don't think it should impact anything if the config is set properly. So if somebody makes DisallowedGameObjects blank in the config, even if there are defaults in the code, they should not take effect. That was my understanding anyway. So it seems to me the dist should always match the default in the code.

hermensbas:
I will merge this changes into 1518, so we can combine and close this PR

hermensbas:
Added into: https://github.com/liyunfan1223/mod-playerbots/pull/1518

PR #1560, opened 2025-08-13
WIP:  Agnostic Naxx40+Ony40 strat integration (mod-individual-progression) without adding dependency to external code/module

valsan-azerty-boi:
No worry if in the end you don't want to merge it (because vanilla strat for other modules are not really in the actual scope of mod-playerbots).

I'm taking your advice on how to improve this.
The code is a bit borderline because it use a LOT of reinterpret_casts.

Overall, it allows bots to play the vanilla fights (for mod-individual-progression) of Heigan / Grobbulus / Gluth / Razuvious / Loatheb / Sapphiron / KT and also to properly focus the adds on Onyxia.
Right now it's not well suited for Four Horsemen and it remains difficult on Thaddius, but overall the bots try to play the strat.

Works better with 10 players than with 40 (for example, with an autobalance module).

### DIFF
--- a/src/strategy/raids/naxxramas/RaidNaxxActions.cpp
+++ b/src/strategy/raids/naxxramas/RaidNaxxActions.cpp
@@ -1,5 +1,6 @@
 
 #include "RaidNaxxActions.h"
+#include "RaidNaxxAi40.h"
 
 #include "LastMovementValue.h"
 #include "ObjectGuid.h"
@@ -13,7 +14,9 @@
 bool GrobbulusGoBehindAction::Execute(Event event)
 {
     Unit* boss = AI_VALUE(Unit*, "boss target");
-    if (!boss)
+    if (!boss 
+        || (boss->GetEntry() != 15931 // Default Azerothcore Grobbulus
+            && boss->GetEntry() != 351003)) // mod-individual-progression Grobbulus
     {
         return false;
     }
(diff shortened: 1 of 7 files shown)


jordan3176:
How does the Thaddius fight work? 
I'm happy doing group splits using RTSC for the platforms, but my main show stopper is the polarity shift mechanic. 
For vanilla naxx, its by default not working, they just stand still, unlike the level 80 version.  

If your code allows bots to move on polarity shift, this is a game changer for us! 
I'm happy to test it out and provide some feedback.


In naxx 40 with 40 bots, every fight is doable apart from Thaddius. 
4 Horsemen, you just have to have enough dps to kill them in the middle before they kill you.

valsan-azerty-boi:
> How does the Thaddius fight work? I'm happy doing group splits using RTSC for the platforms, but my main show stopper is the polarity shift mechanic. For vanilla naxx, its by default not working, they just stand still, unlike the level 80 version.

If you use RTSC on Thaddius phase 1 it's ok

> If your code allows bots to move on polarity shift, this is a game changer for us! I'm happy to test it out and provide some feedback.

They move on polarity shift with this PR
on the other hand I noticed that Thaddius' damage on the raid in case of a bad raid placement are not correct, wotlk numbers instead of vanilla numbers (but that's a problem of the mod-individual-progression module, not mod-playerbots)

brighton-chi:
I'm not seeing an LoS strategy for Razuvious. Have you tried this since IP was updated to restore the mana burn from Naxx40? I think Grim used a different spell instead of modifying Disrupting Shout. I'm not sure if it's working or not as I haven't tested.

valsan-azerty-boi:
> LoS strategy for Razuvious

I didn't add any specific new things, only that bots can refer to vanilla boss timers/event_map like with the wotlk version of the raid

jordan3176:
I just did the Construct quarter for the first phase of testing, since this is the most challenging quarter in the raid.  

Grobbulus was perfect, its now the same as wotlk and the tanks kite correctly.  

Gluth, is actually harder now and impossible, since they are trying to do the ZERG tactic from wotlk.  
- No matter what you try, if you position tanks to stay at the door away from the zombies, the entire raid still keeps moving back to the grates. I've tried the stay commands, and the guard command, the raid tactics override everything. So this fight is impossible.  

Thaddius, yes the bots move on Polarity, but its all wrong. They are not going to the correct places, and the raid wipes every time.  
- Update, I used autobalance and had 10 raid members kill Thaddius without issues. The issue is with more raid members.  


An important note, I am running my original tests with a full raid of 40.

Fero-Fero:
The idea of this PR is freaking epic, i hope some genius minds can come together to pull this off!

valsan-azerty-boi:
https://github.com/ZhengPeiRu21/mod-individual-progression/pull/737

PR #1561, opened 2025-08-13
NamedObjectContext improvements

Vigerus:
remove unneccessary pass by copy
replace typedef with using std::fuction to allow lambda ObjectCreators simplification in strategies

### DIFF
--- a/src/strategy/NamedObjectContext.h
+++ b/src/strategy/NamedObjectContext.h
@@ -11,6 +11,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <functional>
 
 #include "Common.h"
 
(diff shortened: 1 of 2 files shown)


hermensbas:
The std::function alias i find  much more readable then then a raw function pointer as function/method reference. And being able to use lamba doesnt hurt either.

src/strategy/NamedObjectContext.h
```diff
@@ -286,11 +283,9 @@ class NamedObjectFactoryList
         if (creators.find(name) == creators.end())
             return nullptr;
 
-        ObjectCreator creator = creators[name];
-        if (!creator)
-            return nullptr;
+        const ObjectCreator& creator = creators[name];
```
hermensbas, line 286: should be we add?

```
if (!creator) 
      return nullptr;
```

src/strategy/NamedObjectContext.h
```diff
@@ -286,11 +283,9 @@ class NamedObjectFactoryList
         if (creators.find(name) == creators.end())
```
hermensbas, line 283: Consider

```c++
    auto it = creators.find(name);
    if (it == creators.end() || !it->second)
        return nullptr;

    // no insert or second lookup
    const ObjectCreator& creator = it->second;

    // earlier return before cast.
    T* object = creator(botAI);
    if (!object)
        return nullptr;

    if (auto* q = dynamic_cast<Qualified*>(object); q && sep != std::string::npos)
    {
        q->Qualify(qualifier);
    }    

    return object;
```

kadeshar (commented):
My feeling is positive about this PR but i dont have full knowledge to confirm that.

Vigerus:
> The std::function alias i find much more readable then then a raw function pointer as function/method reference. And being able to use lamba doesnt hurt either.

After this PR being accepted and submitted we can reduce all strategies into single liners, instead of having all those private: static Strategy* methods all over the place.

PR #1564, opened 2025-08-15
Prefer group/raid buffs when possible (Paladin/Druid/Mage), with reagent checks and no rebuff if equivalent aura present

Wishmaster117:
### Small change to the non-combat buffing logic for bots:

- When in a party/raid of 3+ players, Paladin/Druid/Mage bots now prefer the group/raid variant of their buffs (Greater/Gift/Brilliance) if the spell is known and the required components are in the bags.
- If not possible (solo/duo, spell not learned, or component missing), bots fall back to the single-target version.
- Target selection now treats single and group variants as equivalent auras, so we don’t rebuff someone who already has the “other” version.

No changes to rotations/strategies beyond buffing behaviour. Priests were already covered by Prayer of ... logic, so untouched.

Based on my knowledge of wotlk 3.3.5, Paladins, Druids and Mages use the group buff out of combat to save GCDs and keep things blizzlike. The current code only ever asks for the single-target spell by name, so bots never attempt the raid version even when they have the component. Also, target lookup didn’t consider “equivalent” auras, causing some pointless rebuffs if someone else applied the group version.
This PR fixes both pain points while keeping the implementation tight and centralised.

Mecanics:
**Target selection (BuffOnPartyAction::GetTargetValue):**
Build an aura qualifier that lists both variants (exemple:"mark of the wild,gift of the wild"), so “party member without aura” ignores units already covered by either one.

**Cast decision (BuffOnPartyAction::Execute):**

- If group->GetMembersCount() >= 3, try to map the single-target spell name to its group variant:

Paladin: Kings/Might/Wisdom/Sanctuary => Greater ...
Druid: Mark of the Wild =>Gift of the Wild
Mage: Arcane Intellect =>Arcane Brilliance

- Resolve the highest learned rank via existing "spell id" value.

- Check component from SpellInfo::Reagent[i]/ReagentCount[i]. If any are missing, stick to the single-target spell.

- Cast proceeds as usual.

**Hard coded choice (may be changed, i don't bor better roleplay?):**

- Threshold = 3 members before using group buffs. This avoids burning componants in a 2 man party where the benefit is marginal.

- No new strategies/triggers, just a centralised decision in the generic action.

**Results:**

- Paladin: uses Greater Blessings (consuming Symbol of Kings) in groups of 3+, otherwise single blessings.
- Druid: uses Gift of the Wild (consuming Wild Spineleaf at max rank) in groups of 3+.
- Mage: uses Arcane Brilliance (consuming Arcane Powder) in groups of 3+.
- Priest: unchanged (Prayer of Fortitude/Spirit/Shadow Protection already supported).

**Bonus:**
Added some funny RP sentences

**How to test this:**

- Solo: bot sticks to single-target buffs only.
- Duo (2 players): still single-target (by design).
- 3+ players, with componant in bags: bot uses Greater/Gift/Brilliance and does not rebuff players who already have the “other” variant.
- 3+ players, no component: clean fallback to single-target, no spam.
- Mixed parties where someone else applied the group aura: target selection correctly skips already-buffed players.

### DIFF
--- a/src/strategy/actions/GenericSpellActions.cpp
+++ b/src/strategy/actions/GenericSpellActions.cpp
@@ -5,6 +5,8 @@
 
 #include "GenericSpellActions.h"
 
+#include <ctime>
+
 #include "Event.h"
 #include "ItemTemplate.h"
 #include "ObjectDefines.h"
@@ -13,6 +15,9 @@
 #include "Playerbots.h"
 #include "ServerFacade.h"
 #include "WorldPacket.h"
+#include "Group.h"
+#include "Chat.h"
+#include "Language.h"
 
 CastSpellAction::CastSpellAction(PlayerbotAI* botAI, std::string const spell)
     : Action(botAI, spell), range(botAI->GetRange("spell")), spell(spell)
@@ -216,10 +221,120 @@ Value<Unit*>* CurePartyMemberAction::GetTargetValue()
     return context->GetValue<Unit*>("party member to dispel", dispelType);
 }
 
+// Make Bots Paladin, druid, mage use the greater buff rank spell
+// TODO Priest doen't verify il he have components
+// local util send "simple,group if buff has greater variant, here we choose te target of the buff
+static std::string MakeAuraQualifierForBuff(std::string const& name)
+{
+    if (name == "blessing of kings")        return "blessing of kings,greater blessing of kings";
+    if (name == "blessing of might")        return "blessing of might,greater blessing of might";
+    if (name == "blessing of wisdom")       return "blessing of wisdom,greater blessing of wisdom";
+    if (name == "blessing of sanctuary")    return "blessing of sanctuary,greater blessing of sanctuary";
+    if (name == "mark of the wild")         return "mark of the wild,gift of the wild";
+    if (name == "arcane intellect")         return "arcane intellect,arcane brilliance";
+    return name;
+}
+
+// Seache who need buffs
 Value<Unit*>* BuffOnPartyAction::GetTargetValue()
 {
-    return context->GetValue<Unit*>("party member without aura", spell);
+    return context->GetValue<Unit*>("party member without aura", MakeAuraQualifierForBuff(spell));
 }
+
+//  Here we choose the spell to cas
+static std::string GroupVariantFor(std::string const& name)
+{
+    if (name == "blessing of kings")        return "greater blessing of kings";
+    if (name == "blessing of might")        return "greater blessing of might";
+    if (name == "blessing of wisdom")       return "greater blessing of wisdom";
+    if (name == "blessing of sanctuary")    return "greater blessing of sanctuary";
+    if (name == "mark of the wild")         return "gift of the wild";
+    if (name == "arcane intellect")         return "arcane brilliance";
+    return std::string();
+}
+
+static bool HasRequiredReagents(Player* bot, uint32 spellId)
+{
+    if (!spellId)
+        return false;
+
+    if (SpellInfo const* info = sSpellMgr->GetSpellInfo(spellId))
+    {
+        for (int i = 0; i < 8; ++i)
+        {
+            if (info->Reagent[i] > 0)
+            {
+                uint32 const itemId = info->Reagent[i];
+                int32  const need   = info->ReagentCount[i];
+                if ((int32)bot->GetItemCount(itemId, false) < need)
+                    return false;
+            }
+        }
+        return true; // no composant required or the bot have composants
+    }
+    return false;
+}
+
+// Decide what buff launch
+bool BuffOnPartyAction::Execute(Event event)
+{
+    std::string castName = spell; // default: single-target
+
+    // helper to send RP line to the right channel
+    auto SendGroupRP = [&](std::string const& msg)
+    {
+        if (Group* g = bot->GetGroup())
+        {
+            WorldPacket data;
+            ChatMsg type = g->isRaidGroup() ? CHAT_MSG_RAID : CHAT_MSG_PARTY; // <-- isRaidGroup()
+            ChatHandler::BuildChatPacket(data, type, LANG_UNIVERSAL, bot, /*receiver=*/nullptr, msg.c_str());
+            g->BroadcastPacket(&data, true, -1, bot->GetGUID());
+        }
+        else
+        {
+            bot->Say(msg, LANG_UNIVERSAL);
+        }
+    };
+
+    // Use group variant only in parties of 3+ to save components in duo
+    if (bot->GetGroup() && bot->GetGroup()->GetMembersCount() > 2)
+    {
+        if (std::string const groupName = GroupVariantFor(spell); !groupName.empty())
+        {
+            uint32 const groupId = AI_VALUE2(uint32, "spell id", groupName);
+			
+			bool const allowGroup = AI_VALUE2(bool, "spell cast useful", groupName);
+
+		    if (groupId && allowGroup && HasRequiredReagents(bot, groupId))
+            {
+                castName = groupName; // learned + have component we use group variant
+            }
+			else if (groupId && allowGroup)
+            {
+                static time_t s_lastWarn = 0;
+                time_t now = std::time(nullptr);
+                if (!s_lastWarn || now - s_lastWarn >= 30) // 30s cooldown to avoid spam
+                {
+                    std::string rp;
+                    if (groupName.find("greater blessing") != std::string::npos)
+                        rp = "By the Light... I forgot my Symbols of Kings. We’ll make do with the simple blessings!";
+                    else if (groupName == "gift of the wild")
+                        rp = "Nature is generous, my bags are not... out of herbs for Gift of the Wild. Take Mark of the Wild for now!";
+                    else if (groupName == "arcane brilliance")
+                        rp = "Out of Arcane Powder... Brilliance will have to wait. Casting simple Intellect!";
+                    else
+                        rp = "Oops, I’m out of components for the group version. We’ll go with the single one!";
+
+                    SendGroupRP(rp);
+                    s_lastWarn = now;
+                }
+            }
+        }
+    }
+
+    return botAI->CastSpell(castName, GetTarget());
+}
+// End greater buff fix
 
 CastShootAction::CastShootAction(PlayerbotAI* botAI) : CastSpellAction(botAI, "shoot")
 {
--- a/src/strategy/actions/GenericSpellActions.h
+++ b/src/strategy/actions/GenericSpellActions.h
@@ -215,17 +215,18 @@ protected:
     uint32 dispelType;
 };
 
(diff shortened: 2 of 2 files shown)


Dreathean:
This is great! I know a lot of people have been asking about this, and will close https://github.com/liyunfan1223/mod-playerbots/issues/693. Out of curiosity does this also fix the issue of multiple paladins not applying different buffs in a party/raid? Someone correct me if I'm wrong but I believe that is also an issue.

brighton-chi:
Thanks for tackling what has been a pain point for quite some time for many of us. As I understand, Paladins currently have some specialized logic to determine which blessing to apply (i.e., setting a particular blessing strategy does not mean that the Paladin will cast only that blessing--it will depend on the class/spec of the recipient still). In light of that, how does a Paladin determine which greater blessing to apply, since it will apply on a class-wide basis?

Given the unique mechanics of greater blessings, I think it's worth a discussion about the right logic for using them (considering that only one per Paladin can be applied per character, they are cast on a class-wide basis, not party or raid-wide, and they significantly increase the duration of the buff from 10 minutes to 30 minutes), 

Edit: Separately, for Priests, is it right that currently they are supposed to use the group versions if they are in a group of at least 3?

Wishmaster117:
> Thanks for tackling what has been a pain point for quite some time for many of us. As I understand, Paladins currently have some specialized logic to determine which blessing to apply (i.e., setting a particular blessing strategy does not mean that the Paladin will cast only that blessing--it will depend on the class/spec of the recipient still). In light of that, how does a Paladin determine which greater blessing to apply, since it will apply on a class-wide basis?
> 
> Given the unique mechanics of greater blessings, I think it's worth a discussion about the right logic for using them (considering that only one per Paladin can be applied per character, they are cast on a class-wide basis, not party or raid-wide, and they significantly increase the duration of the buff from 10 minutes to 30 minutes),
> 
> Edit: Separately, for Priests, is it right that currently they are supposed to use the group versions if they are in a group of at least 3?

Hello,

I'll try to answer you, i didn’t change the “which blessing fits this target?” logic. The Paladin still decides the type (Might , Wisdom ,  Sanctuary  , Kings) based on the target’s class/spec exactly like before. The only thing the patch does is, right before casting, upgrade the single target blessing to the greater version if it makes sense (party/raid size threshold and reagents available).
Because greater blessings are class-wide, casting say Greater Blessing of Wisdom on any member of that class will cover the whole class that is still missing it. The target selection already looks for “party member without the aura (single or greater)” so the Paladin naturally ends up casting the right greater on a class that still needs it. No change to the class/spec logic, we just reduce GCDs and respect reagents.

I also added a small safety threshold (defaults to 3+ players) so you don’t burn reagents in a duo. If the reagent is missing or the threshold isn’t met, it falls back to the normal 10 mn blessing. 
Auras are checked equivalently (single vs greater) to avoid rebuff spam.

On the unique mechanics (one blessing per Paladin per character, class-wide, 30 min):
Agreed, that’s why the upgrade is conservative: only when it’s actually beneficial (group big enough, reagent present). Otherwise we keep the current single-target behavior. The patch doesn’t change stacking rules or overwrite logic, just which version is cast.

Priests

Priests were already handled by the module. They use Prayer of … when it’s efficient (there’s existing trigger logic based on how many party members are missing the aura), and single-target when not. My change does not alter Priest behavior, and it’s not tied to the 3+ threshold we use for Paladin/Druid/Mage.

Things we may could do:

Make the threshold configurable (hardcoded things are not the best)
Optional heuristic: prefer greater only if at least two members of that class still need it. Use greater only if 2 players of the same class need the buff.
Out-of-combat preference to save reagents. Because now its allowed.

About Paladin:
The paladin doesn’t blindly apply the requested blessing.
He chooses which blessing suits the target based on their class/spec (and on what the paladin knows):
Key functions in src/strategy/paladin/PaladinActions.cpp:
GetActualBlessingOfMight(Unit* target)
GetActualBlessingOfWisdom(Unit* target)
GetActualBlessingOfSanctuary(Unit* target, Player* bot)
These helpers return the name of the “single-target” blessing to cast according to the target’s class/spec:
Might for physical classes (warrior/rogue/dk/hunter, etc.)
Wisdom for casters (mage/priest/warlock, etc.)
With special cases (e.g., enhancement shaman / feral druid → Might).
Sanctuary if the target is a tank and the paladin knows the spell; otherwise, Kings by default.

Exemple:

```
inline std::string const GetActualBlessingOfMight(Unit* target)
{
    if (!target->ToPlayer())
    {
        return "blessing of might";
    }

    int tab = AiFactory::GetPlayerSpecTab(target->ToPlayer());
    switch (target->getClass())
    {
        case CLASS_MAGE:
        case CLASS_PRIEST:
        case CLASS_WARLOCK:
            return "blessing of wisdom";
            break;
        case CLASS_SHAMAN:
            if (tab == SHAMAN_TAB_ELEMENTAL || tab == SHAMAN_TAB_RESTORATION)
            {
                return "blessing of wisdom";
            }
            break;
        case CLASS_DRUID:
            if (tab == DRUID_TAB_RESTORATION || tab == DRUID_TAB_BALANCE)
            {
                return "blessing of wisdom";
            }
            break;
        case CLASS_PALADIN:
            if (tab == PALADIN_TAB_HOLY)
            {
                return "blessing of wisdom";
            }
            break;
    }

    return "blessing of might";
}
```
Cheers,

Wishmaster117:
I’ll add this before writing a whole novel:

My code is just a start; it’s far from complete.
The function (GetActualBlessingOfMight) is only used to choose the type of blessing (Might vs. Wisdom) based on the target’s class/spec.
My “upgrade” to Greater only triggers for actions that actually go through BuffOnPartyAction::Execute.

Specifically:
Blessing of Kings (there’s no GetActualBlessingOfKings) uses the generic path, so yes,  with group ≥ 3 + reagents, it gets upgraded to Greater Blessing of Kings.

Blessing of Might / Wisdom / Sanctuary, on the paladin side, have dedicated actions that override Execute and call GetActualBlessingOf…(...) directly, then CastSpell(...).
As a result, they don’t go through BuffOnPartyAction::Execute, so no — they’re not yet upgraded to Greater with your current modification.

So… I still have a bit of work to do.

src/strategy/actions/GenericSpellActions.cpp
```diff
@@ -216,10 +221,120 @@ Value<Unit*>* CurePartyMemberAction::GetTargetValue()
...
+    return false;
+}
+
+// Decide what buff launch
+bool BuffOnPartyAction::Execute(Event event)
+{
+    std::string castName = spell; // default: single-target
+
+    // helper to send RP line to the right channel
+    auto SendGroupRP = [&](std::string const& msg)
```
hermensbas: I think playerbotAI have functions for these already. e.g. bool PlayerbotAI::SayToRaid(const std::string& msg) etc.

Wishmaster117:
You guys are a pain! :D
Now you’ve made me even more curious, I’m going to spend hours on this...

src/strategy/actions/GenericSpellActions.cpp
Wishmaster117: > I think playerbotAI have functions for these already. e.g. bool PlayerbotAI::SayToRaid(const std::string& msg)

Dont' wast time for now i have to use another way because this damn Paladin

Dreathean:
Paladin buffs are a pain!

Regarding the:
"About Paladin:
The paladin doesn’t blindly apply the requested blessing.
He chooses which blessing suits the target based on their class/spec (and on what the paladin knows):"

Just to clarify, because it certainly is confusing, how would the upgrade to greater blessings work in these situations:

1) Raid has just 1 paladin; there are for example 2 druids, one feral and one resto. Would it just see whichever comes first (ie the feral) and apply greater blessing of might for both since it's class-wide? Or would it stick with regular blessing for might on the feral and wisdom on the resto?


2) Raid has 3 paladins - would they have 1 apply greater blessing of kings to all, another apply greater blessing of might to all, and a third apply greater blessing of wisdom to all?

I think this is why they have stayed as non-greater blessings for the longest time, because there's so many different configurations to consider - indeed a pain! Lol

Wishmaster117:
Ok got it, i don't spent my time i can use all this code juste another way.

Dreathean:
I'm only asking because I personally have no idea haha, just things to consider for how it works in practice. At a minimum this is great improvement for mages and druids! I'm sorry I can't be more helpful.

hermensbas:
Not only are paladins a pain in the ass ingame, but also in code it seems.

brighton-chi:
**Paladins**

The simplest way I can think of is to keep the existing logic but add a scenario in which, say, if there is a raid group consisting of at least X players/bots, Y of which are Paladins (say, X = 20 and Y = 3), you override existing logic such that paladins will cast only greater blessings and they will cast only their designated blessing in their strats. Thus, in that situation, if a paladin was set to bstats, it would cast Greater Blessing of Kings on all 10 classes; if it was set to bdps, it would cast Greater Blessing of Might on all classes except those that would not ever physically attack, and it would just not buff those classes at all (Priests, Mages, Warlocks); if it was set to bmana, it would cast Greater Blessing of Wisdom on all classes except those that don't use mana, and it would just not buff those classes at all. This isn't a perfect solution, but it seems more straightforward to code, and I think in practice it would get you to the desired result for 25- and 40-man raids at least.

An aside: Should a tank really want Sanctuary over Kings? Kings is not the better first buff for every class/spec?

**Druids, Mages, and Priests**

It seems to me that the logic should be the same for Druids, Mages, and Priests. Personally, since I just use maintenance to give reagents to bots, I would rather all of them just use the group buffs whenever there are 3 or more bots in a party/raid. This means the timers are always synced on the entire raid so there isn't one-off buffing going on. However, there certainly is inefficiency there for somebody who wants to micromanage reagents. Does anybody prefer the existing Priest logic that checks for how many don't already have the buff?

kadeshar:
That PR should be marked as draft?

src/strategy/actions/GenericSpellActions.cpp
noisiver: RP shouldn't be enforced since you can disable chat in the config. Maybe a check should be added to make sure chat is actually enabled?

Wishmaster117:
> That PR should be marked as draft?

Yes please

Wishmaster117:
Ok patch-1 tell me what yoy guys think and what to need be changed?
I updated the desc of the PR.
Now go to sleep it's 2AM
Good Night

Wishmaster117:
> That PR should be marked as draft?

Ready to review and Merge

ThePenguinMan96:
> > That PR should be marked as draft?
> 
> Ready to review and Merge

I like the detail that went into the explanation. If I have time later today I will test it

Vortikai:
Great work 👍

Wishmaster117:
I’m going to close all my PRs about paladin buffs and post a single one with everything. I changed my approach midway through the code and made it blizzlike.

PR #1565, opened 2025-08-15
Yogg-Saron strategy

kadeshar:
Simplified Yogg-Saron strategy. Added cheats but all of them can be disabled via config. 
I wrote in code comments which cheat can be deleted when (for example phase 3 adds cheat can be disabled when better handling with lunatic gaze will be implemented)

Test: https://youtu.be/ibZIGBRqRdE


src/PlayerbotAI.cpp
```diff
@@ -1840,6 +1840,29 @@ bool PlayerbotAI::HasAggro(Unit* unit)
...
+    Group* group = player->GetGroup();
+    if (!group)
+    {
+        return -1;
+    }
+    int counter = 0;
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
+        if (player == member)
```
hermensbas, line 1870: Would be wise to add, just incase the state although seems slim chance.

```c++
if (!member)
            continue;
```

src/strategy/triggers/RangeTriggers.cpp
```diff
@@ -214,3 +214,86 @@ bool FarFromMasterTrigger::IsActive()
...
+
+    return !debuffedPlayers.empty();
+
+    if (debuffedPlayers.empty())
+    {
+        return false;
+    }
+
+    for (Unit* debuffedPlayer : debuffedPlayers)
+    {
```
hermensbas, line 249: what happens if a player has logged out between the vector buildup and getting the distance from possible player that logged out?

src/strategy/triggers/RangeTriggers.cpp
```diff
@@ -214,3 +214,86 @@ bool FarFromMasterTrigger::IsActive()
...
+        float dist = debuffedPlayer->GetExactDist2d(bot->GetPositionX(), bot->GetPositionY());
```
hermensbas, line 289: Same applies, the state of player can be changed between buildup vector and using the player in that vector

hermensbas (commented):
Left a few comments, you decide whether they should be addressed or not.

hermensbas (approved):

hermensbas:
Got my approval from a code perspective, nice talk btw :)

biaanyuanren11:
Perfect!

src/strategy/actions/BossAuraActions.cpp
```diff
@@ -22,6 +22,7 @@ bool BossFireResistanceAction::Execute(Event event)
 {
     PaladinFireResistanceStrategy paladinFireResistanceStrategy(botAI);
     botAI->ChangeStrategy(ADD_STRATEGY_CHAR + paladinFireResistanceStrategy.getName(), BotState::BOT_STATE_COMBAT);
+    botAI->DoSpecificAction("fire resistance aura", Event(), true);
```
privatecore, line 25: @kadeshar hey, can you please explain why you use the default constructor `Event()` here instead of the `event` from the parameters?

kadeshar: Tbh i dont know :D This usage is copied from other action :)

privatecore: oh :) I thought it was some magic behind or some sort of bug you are tryin to avoid

PR #1572, opened 2025-08-17
Feat(paladin) configurable sanctuary vs kings policy on tanks and safe targeting

Wishmaster117:
Ok Guys tis @/!%** Paladin  is driving me crazy!

### Summary

This PR introduces a configurable policy to prefer Blessing of Sanctuary over Blessing of Kings on tanks, and hardens the casting logic so Kings won’t overwrite Sanctuary (and vice-versa). 
In WotLK, Sanctuary provides 3% damage reduction and returns resources on avoidance, making it a strong tank blessing; Kings remains a universal fallback. 

### Why

- Many groups want Sanctuary on tanks (mitigation + resource return), Kings for everyone else. Playerbots previously defaulted to Kings in some paths or kept Sanctuary from replacing Kings immediately.
- I also ensure the “one Paladin blessing per target” rule is respected by preventing accidental overwrites. 

### What changed

1. New policy switch

- AiPlayerbot.Paladin.SanctuaryPriority (default 1)

- 1 => Prefer Sanctuary on tanks (if learned), Kings elsewhere.
- 0 => Prefer Kings; the Sanctuary action won’t auto-apply.

2. Sanctuary action is “strict”

- The Sanctuary action never falls back to Kings. If the target isn’t a tank (or the spell isn’t known), it does nothing and lets the Kings action handle the rest.

3. Allow replacement when prioritized

- When SanctuaryPriority=1, the Sanctuary action targets party/raid members missing Sanctuary only, even if they already have Kings, so the tank gets Sanctuary immediately (Kings => Sanctuary upgrade).

4. Kings won’t overwrite Sanctuary

- The Kings action’s target filter now excludes Sanctuary (mono + greater), preventing Kings from replacing an already-applied Sanctuary.

5. Main-Tank focus for fast application

- If SanctuaryPriority=1 and the Main Tank lacks Sanctuary, the Sanctuary action forces its target to the MT before casting. This avoids selector ordering issues (master/healer first) and guarantees prompt application to the tank.

### Configuration
```
# Paladin — Sanctuary policy on tanks
# 1 = Sanctuary on tanks if learned (fallback Kings)
# 0 = Kings has priority; Sanctuary is not used automatically
# Default: 1
AiPlayerbot.Paladin.SanctuaryPriority = 1
```
Works alongside Strict Role Mode (OFF/ON/AUTO) and Greater/Single upgrade logic; no change required to existing settings.

### Implementation notes:

**Selector changes : **

**Sanctuary GetTargetValue():**

- Priority ON => match  without Sanctuary  (allows replacing Kings).
- Priority OFF => legacy behavior (requires neither Kings nor Sanctuary).

**Kings GetTargetValue() now excludes both Kings and Sanctuary.**

**Decision function**

- GetActualBlessingOfSanctuary(...) returns Sanctuary only for tanks (or the explicit Main Tank), otherwise empty (no cast). No implicit Kings fallback here.

**Execute path**

**Sanctuary Execute(...) optionally retargets the Main Tank if he lacks Sanctuary, then applies normal Greater vs Single upgrade with reagent handling.**

### Acceptance tests

**Default (priority=1)**

**Group with Prot Paladin (bstats), Warrior Prot, others:**

- Everyone receives Kings (Greater when appropriate).
- Warrior Prot’s Kings is immediately replaced by Sanctuary.
- Subsequent Kings passes do not overwrite Sanctuary.

**Priority=0**

- Same setup: tanks receive Kings; Sanctuary is not applied automatically.

**No Sanctuary learned**

- Priority=1 or 0: Kings behavior unchanged; no Sanctuary attempts.

### Backward compatibility

- Default is priority=1, which aligns with common WotLK tank buff practice (Sanctuary on tanks; Kings elsewhere). No behavior change for non-tank targets. The change is localized and low-risk.

**Why Sanctuary?**

- Greater Blessing of Sanctuary on WotLK: 3% damage taken reduction; +10% Strength/Stamina; returns mana/rage/runic power on block/parry/dodge. 
- Greater Blessing of Kings (WotLK): +10% stats; only one Paladin blessing per target at a time. 

Thanks for reviewing and tests.

### DIFF
--- a/src/strategy/paladin/PaladinActions.cpp
+++ b/src/strategy/paladin/PaladinActions.cpp
@@ -48,7 +48,7 @@ namespace {
 
     struct StrictAnnounceState
     {
-        bool lastStrict  = false;
+        bool lastStrict = false;
         time_t lastAnnounce = 0;
     };
 
@@ -321,7 +321,20 @@ bool CastBlessingOfMightAction::Execute(Event event)
     if (!target)
         return false;
 
-    std::string castName = GetActualBlessingOfMight(target);
+    // Strict mode: Might only for relevant targets (no fallback to Wisdom)
+    if (StrictRoleModeEnabledFor(bot))
+    {
+        std::string desired = GetActualBlessingOfMight(target);
+        if (desired != "blessing of might")
+            return false; // Ignore irrelevant target (non-physique)
+    
+        std::string castName = "blessing of might";
+        auto RP = ai::chat::MakeGroupAnnouncer(bot);
+        castName = ai::buff::UpgradeToGroupIfAppropriate(bot, botAI, castName, /*announceOnMissing=*/true, RP);
+        return botAI->CastSpell(castName, target);
+    }
+    
+    std::string castName = GetActualBlessingOfMight(target);	
 
     auto RP = ai::chat::MakeGroupAnnouncer(bot);
 
@@ -372,8 +385,20 @@ bool CastBlessingOfWisdomAction::Execute(Event event)
     if (!target)
         return false;
 
-    std::string castName = GetActualBlessingOfWisdom(target);
-
+    // Strict mode: Wisdom only for mana users (no fallback to Might)
+    if (StrictRoleModeEnabledFor(bot))
+    {
+        std::string desired = GetActualBlessingOfWisdom(target);
+        if (desired != "blessing of wisdom")
+            return false; // Ignore irrelevant target (non-mana)
+    
+        std::string castName = "blessing of wisdom";
+        auto RP = ai::chat::MakeGroupAnnouncer(bot);
+        castName = ai::buff::UpgradeToGroupIfAppropriate(bot, botAI, castName, /*announceOnMissing=*/true, RP);
+        return botAI->CastSpell(castName, target);
+    }
+    
+    std::string castName = GetActualBlessingOfWisdom(target);	
 
     auto RP = ai::chat::MakeGroupAnnouncer(bot);
 


brighton-chi:
Is there a standalone greater blessing of sanctuary strategy? Like bhealth means cast greater blessing of sanctuary on everybody if you meet the group size threshold? So then with four paladins, you could cover all blessings?

Wishmaster117:
> Is there a standalone greater blessing of sanctuary strategy? Like bhealth means cast greater blessing of sanctuary on everybody if you meet the group size threshold? So then with four paladins, you could cover all blessings?


not yet.
There isn’t a standalone bhealth = cast (Greater) Blessing of Sanctuary on everyone strategy in the module today.

What exists now after my PRs
StrictRoleMode (OFF/ON/AUTO) + AutoStrict thresholds.
SanctuaryPriority: when enabled, a Prot Paladin in bstats will prefer Sanctuary on tanks (and won’t fall back to Kings), with guards so Kings won’t overwrite Sanctuary and vice-versa.

For raid coverage, the intended pattern is:
Kings (raid-wide baseline),
Might for physical DPS,
Wisdom for mana users,
Sanctuary only on tanks (Prot paladin handles it).

We intentionally don’t have a Sanctuary for everyone mode because:
Players can have only one Paladin blessing at a time   putting Sanctuary on non-tanks would displace Kings/Might/Wisdom, which are typically stronger for them. 

Greater blessings are class-wide casts (they hit the whole class of the target), which would unintentionally put Sanctuary on DPS Warriors/Paladins/Etc. as well. 

With four paladins: you can already cover everything that makes sense  Kings (all), Might (physicals), Wisdom (mana users), and Sanctuary on the tanks  but not Sanctuary on everyone (by design).

If there’s strong interest, we could consider adding an opt-in bhealth role later (behind a config flag), but the default/sane WotLK setup is to keep Sanctuary on tanks only and Kings/Might/Wisdom elsewhere.

brighton-chi:
Hmm then is there a way to have kings and sanctuary auto cast on tanks when you are past the threshold where greater blessings are triggered? Seems even with four paladins then it wouldn't be possible with this setup unless I'm understanding wrong. Or if you have two with bstats, the tanks will get both?

Wishmaster117:
Hi i put all last PR on Draft because i'me taking that by the wrong way, the code made still now it's ok but i think that some things have to be done another way, so i post the final Pr where i'he finished all

Wishmaster117:
I’m going to close all my PRs about paladin buffs and post a single one with everything. I changed my approach midway through the code and made it blizzlike.

PR #1573, opened 2025-08-17
Fix(License): Non-technical. Update license note from GPL v2 to AGPL v3

NoxMax:
Updated the license note for all files to match AC's and PB's AGPL v3.

No technical code edits made. Only CTRL+H this
```
 * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU GPL v2 license, you may redistribute it
 * and/or modify it under version 2 of the License, or (at your option), any later version.
```
to this, across 837 files
```
 * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU AGPL v3 license, you may redistribute it
 * and/or modify it under version 3 of the License, or (at your option), any later version.
```

This commit does not change Playerbots' license. The license remains AGPL v3. This commit only makes sure to correctly state that AC was released under AGPL v3.

Some files also had trailing whitespaces removed by VSC.


noisiver:
I wonder if <https://github.com/azerothcore/azerothcore-wotlk/discussions/19869> will ever happen, in which case there would be issues converting this back to v2.

NoxMax:
I guess that depends on what you mean "back" is. What I found out is AC has always used AGPL v3. Meaning that mentions of GPLv2 (whether here or in AC) are not because of the original license of AC, but based on the legacy code of Mangos and TC, which do use GPLv2.

But also I found the sentiment is that it was a mistake to use AGPL v3 for AC, because it creates issues with porting AC changes to Mangos and TC, thus there is the effort to shift to GPLv2 to make AC backwards compatible. To be honest, I'm not sure that's gonna happen after all these years.

kadeshar:
@NoxMax could you resolve conflicts? When you do that i will merge that PR.

hermensbas:
recreated: https://github.com/liyunfan1223/mod-playerbots/pull/1674

NoxMax:
> @NoxMax could you resolve conflicts? When you do that i will merge that PR.

My apologies. I have been away from GH for a month for personal reasons and I just logged in. I see that Bash recreated this PR less than a day ago owing to my abcence, so we can continue over there. Given that Yunfan has approved this PR, #1674 would only need one more maintainer approval and I think we can merge it.

hermensbas:
I did, might have redo it, since i first would like to see unused warning merged. If needed i will create a new one.

PR #1574, opened 2025-08-17
Revert commits from PR#s 1507 and 1512

### LINKED ISSUE #1549: The bots' strategies started to reset after teleportation.
**Bug Description**
It seems that after the latest changes, the bots' teleportation strategies have started to reset. 

**Commit Hash**
3fff58df1a2058894e9b758be07869aec87c2c70

**How To Reproduce Bug**
Create bots, set up strategies for them, enter the dungeon.

**Expected Behavior**
Bots should keep their strategies.

**Screenshots**
No

**Modules**
Transmog

**Playerbot Settings**
No

**System**
OS: Windows 10
Hardware: Xeon 22 cores

**Additional Context**
@Wishmaster117 Maybe a regression after one of your stability fixes.

Tierisch:
This should resolve #1549.

These two were left out of the merged PRs that were recently reverted, and I've locally confirmed them to be the cause for bots resetting strategies when they shouldn't, *not* resetting them when they should in some cases, delays before being able to summon already grouped bots after changing maps, and other odd behavior around bots joining and leaving groups.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -3970,36 +3970,14 @@ bool IsAlliance(uint8 race)
 
 bool PlayerbotAI::HasRealPlayerMaster()
 {
-//     if (master)
-//     {
-//         PlayerbotAI* masterBotAI = GET_PLAYERBOT_AI(master);
-//         return !masterBotAI || masterBotAI->IsRealPlayer();
-//     }
-// 
-//     return false;
-
-     // Removes a long-standing crash (0xC0000005 ACCESS_VIOLATION)
-    /* 1) The "master" pointer can be null if the bot was created
-          without a master player or if the master was just removed. */
-    if (!master)
-        return false;
-
- /* 2) Is the master player still present in the world?
-          If FindPlayer fails, we invalidate "master" and stop here. */
-    if (!ObjectAccessor::FindPlayer(master->GetGUID()))
+    if (master)
     {
-        master = nullptr;           // avoids repeating the check on the next tick
-        return false;
+        PlayerbotAI* masterBotAI = GET_PLAYERBOT_AI(master);
+        return !masterBotAI || masterBotAI->IsRealPlayer();
     }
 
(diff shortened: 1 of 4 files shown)


hermensbas:
I know @Wishmaster117 still figuring those patches, in regard some stuff we talked about in admin channel. So i am kinda waiting on this new PR of him which includes those that been reverted. The general idea that we want to have more time to go over the changes and test it more proper.

noisiver:
The current behavior is unacceptable though. If this PR addresses those issues at least we can get the module back to working order.

It may take some time before a new PR by Wishmaster, that could fix these issues, would be ready to be merged and I don't think we can keep it this way until then.

Wishmaster117:
> The current behavior is unacceptable though. If this PR addresses those issues at least we can get the module back to working order.
> 
> It may take some time before a new PR by Wishmaster, that could fix these issues, would be ready to be merged and I don't think we can keep it this way until then.

I also think that those PR should be reverted, they have no sens if the other one are not merged. We were too eager and wanted to move too quickly.
I'm going to take my time and, with the help of those who post the crash logs on my Git, analyze them calmly and create a coherent global patch rather than putting a "fix" on every bleeding cut.
This is not the right approach.

Wishmaster117:
> > The current behavior is unacceptable though. If this PR addresses those issues at least we can get the module back to working order.
> > It may take some time before a new PR by Wishmaster, that could fix these issues, would be ready to be merged and I don't think we can keep it this way until then.
> 
> I also think that those PR should be reverted, they have no sens if the other one are not merged. We were too eager and wanted to move too quickly. I'm going to take my time and, with the help of those who post the crash logs on my Git, analyze them calmly and create a coherent global patch rather than putting a "fix" on every bleeding cut. This is not the right approach.

I mean on my GIT or Here!!!

hermensbas:
https://github.com/liyunfan1223/mod-playerbots/pull/1583

Two leftover indeed.

PR #1578, opened 2025-08-18
Greater buffs and RP anti-spam and reliable Sanctuary on tanks and ( optional 2 paladin logic)

Wishmaster117:
### Summary

This PR improves Paladin buffing in Playerbots  without changing which buff is chosen per target. It adds:

- Greater vs Single: automatic upgrade to Greater ranks when the group meets a configurable threshold and reagents are available (centralized helper).
- RP lines on missing reagents with a configurable anti-spam cooldown.
- Sanctuary prioritized on tanks: the bstats paladin always applies Blessing of Sanctuary to tanks (MT/OT, including itself if it’s a tank), even if Kings is already present (WotLK behavior: Sanctuary replaces Kings on a tank).
- Might/Wisdom never overwrite Sanctuary (selector guarded).

 **Optional “two-paladin” logic (exactly 1 bstats + 1 bmana), behind a config switch:**

ON :

- bstats → Sanctuary on all tanks; does not cast Kings on tanks.
- bmana → Kings (single) on tanks that have neither Kings nor Sanctuary, then normal Wisdom/Might on non-tanks.

OFF (default) 

- Original behavior (no special handling when there are exactly two paladins).

No configuration is required to benefit from Greater/Single, RP, or the tank-Sanctuary reliability. The two-paladin logic is opt-out via config.

### What’s included

**Shared utilities (reused, no API change)**

- GenericBuffUtils.h/.cpp
- UpgradeToGroupIfAppropriate(...) – chooses single/greater based on:
- group size ≥ MinBotsForGreaterBuff,
- greater spell known and enabled,
- reagents available.
- If reagents are missing: emits an RP warning with per-bot per-buff cooldown.

- Paladin  reliable Sanctuary & no unwanted overwrites
**Sanctuary (OnParty)**

- Trigger: fires whenever Sanctuary is missing (no longer blocked by Kings presence).
- Targeting: retargets tanks (MT preferred), and self-buffs if the bstats paladin is itself a tank.
- Kings no longer prevents Sanctuary from being applied to tanks (Sanctuary will replace Kings on tanks, as in WotLK).

**Kings (OnParty): never overwrites Sanctuary.**

**Might/Wisdom (OnParty): never attempt to buff units that already have Sanctuary (so Sanctuary on tanks is not overwritten).**

**Optional “exactly two paladins: bstats + bmana”**

- New config flag: AiPlayerbot.Paladin.TwoPallyKingsSanctuary
- 1 (default) – special handling enabled:
- bstats casts Sanctuary on tanks and skips Kings on tanks.
- bmana casts Kings (single) on tanks only if they have neither Kings nor Sanctuary, and skips Might/Wisdom on tanks.
- 0 – special handling disabled (original distribution logic).

**With 3+ paladins, nothing is changed by this PR: natural distribution already covers tanks properly.**

### Configuration (snippets)
```
# Minimum group size for using Greater ranks (default 3)
AiPlayerbot.MinBotsForGreaterBuff = 3

# Cooldown for reagent-missing RP lines (seconds), per bot & per buff (default 30)

# Optional: enable/disable special handling when there are exactly two paladins (bstats + bmana)
# 1 = ON (default), 0 = OFF
AiPlayerbot.Paladin.TwoPallyKingsSanctuary = 1
```

Behavior

**Normal mode (default):**
Uses Greater when the group meets the threshold and reagents are present (Paladin/Druid/Mage families).

**Exactly two paladins (bstats + bmana) with the flag ON:**

- bstats: Sanctuary on tanks; does not cast Kings on tanks.
- bmana: Kings (single) on tanks missing both Kings and Sanctuary; then normal Wisdom/Might on non-tanks.

Three or more paladins: unchanged.

### Test plan

**Greater threshold**

- Group of 3 single.
- Group ≥ threshold with reagents → Greater.
- Change AiPlayerbot.MinBotsForGreaterBuff and verify.

**Reagents handling**

- Remove Symbols of Kings → RP line once per cooldown; single rank cast.
- Restore reagents → Greater resumes.

**Sanctuary on tanks after resurrection**
Two-paladin flag
AiPlayerbot.Paladin.TwoPallyKingsSanctuary = 1 → special handling active (Sanctuary on tanks + Kings single on tanks).
= 0 → original behavior (no special case).
- Kill and resurrect MT/OT:
- bmana applies Kings (single) to tanks lacking buffs,
- bstats applies Sanctuary afterwards which replaces Kings on tanks,
- Might/Wisdom never overwrite Sanctuary.

### Backwards compatibility

- Safe by default: Greater upgrade respects existing thresholds/reagents; Sanctuary remains tank-only; Sanctuary is never overwritten by Kings/Might/Wisdom.
- The “two-paladin” behavior can be disabled via config (= 0) to restore the original behavior entirely.

Thanks for reviewing


noisiver:
How does this work if there are one of each specialization paladin? A protection, a holy and a retribution. The protection paladin should apply sanctuary on himself but could still apply kings (or something else) on the other two; or would that be too difficult to handle?

I am looking forward to this one, thank you for submitting it and I hope the tests go alright. I plan to test it as soon as I can but wanted to know before how it works with multiple paladins.

Wishmaster117:
> How does this work if there are one of each specialization paladin? A protection, a holy and a retribution. The protection paladin should apply sanctuary on himself but could still apply kings (or something else) on the other two; or would that be too difficult to handle?
> 
> I am looking forward to this one, thank you for submitting it and I hope the tests go alright. I plan to test it as soon as I can but wanted to know before how it works with multiple paladins.

Yes the prot Paladin can make Sanctuary in himself and the other tanks and it put King in the rest of the group.
With two paladins, the prot one put Sanctuary in tanks and himself and the bmana Paladin put Might or Mana and king in Tanks of you active the optional settings 2 paladins, otherwise it put Might in Tanks

noisiver:
You shouldn't drop and recreate ai_playerbot_texts when you change it. There are users who change this table on their end and wouldn't want it reset. You should use update queries.

Wishmaster117:
> You shouldn't drop and recreate ai_playerbot_texts when you change it. There are users who change this table on their end and wouldn't want it reset. You should use update queries.

Omg i use my smartphone and made a mess in the PR 
No table change required let me one hour to fix this sorry

Wishmaster117:
Wiil replaced by a new one

noisiver:
> > You shouldn't drop and recreate ai_playerbot_texts when you change it. There are users who change this table on their end and wouldn't want it reset. You should use update queries.
> 
> Omg i use my smartphone and made a mess in the PR No table change required let me one hour to fix this sorry

It's ok, I just figured I'd tell you since I believe some people will be unhappy with replacing the table. It doesn't matter to me personally.

PR #1582, opened 2025-08-19
feat(paladin): Greater buffs + RP anti-spam + reliable Sanctuary on tanks (+ optional 2-paladin logic)

Wishmaster117:
### Summary

This PR improves Paladin buffing in Playerbots  without changing which buff is chosen per target. It adds:

- Greater vs Single: automatic upgrade to Greater ranks when the group meets a configurable threshold and reagents are available (centralized helper).
- RP lines on missing reagents with a configurable anti-spam cooldown.
- Sanctuary prioritized on tanks: the bstats paladin always applies Blessing of Sanctuary to tanks (MT/OT, including itself if it’s a tank), even if Kings is already present (WotLK behavior: Sanctuary replaces Kings on a tank).
- Might/Wisdom never overwrite Sanctuary (selector guarded).

 **Optional “two-paladin” logic (exactly 1 bstats + 1 bmana), behind a config switch:**

ON :

- bstats → Sanctuary on all tanks; does not cast Kings on tanks.
- bmana → Kings (single) on tanks that have neither Kings nor Sanctuary, then normal Wisdom/Might on non-tanks.

OFF (default) 

- Original behavior (no special handling when there are exactly two paladins).

No configuration is required to benefit from Greater/Single, RP, or the tank-Sanctuary reliability. The two-paladin logic is opt-out via config.

### What’s included

**Shared utilities (reused, no API change)**

- GenericBuffUtils.h/.cpp
- UpgradeToGroupIfAppropriate(...) – chooses single/greater based on:
- group size ≥ MinBotsForGreaterBuff,
- greater spell known and enabled,
- reagents available.
- If reagents are missing: emits an RP warning with per-bot per-buff cooldown.

- Paladin  reliable Sanctuary & no unwanted overwrites
**Sanctuary (OnParty)**

- Trigger: fires whenever Sanctuary is missing (no longer blocked by Kings presence).
- Targeting: retargets tanks (MT preferred), and self-buffs if the bstats paladin is itself a tank.
- Kings no longer prevents Sanctuary from being applied to tanks (Sanctuary will replace Kings on tanks, as in WotLK).

**Kings (OnParty): never overwrites Sanctuary.**

**Might/Wisdom (OnParty): never attempt to buff units that already have Sanctuary (so Sanctuary on tanks is not overwritten).**

**Optional “exactly two paladins: bstats + bmana”**

- New config flag: AiPlayerbot.Paladin.TwoPallyKingsSanctuary
- 1 (default) – special handling enabled:
- bstats casts Sanctuary on tanks and skips Kings on tanks.
- bmana casts Kings (single) on tanks only if they have neither Kings nor Sanctuary, and skips Might/Wisdom on tanks.
- 0 – special handling disabled (original distribution logic).

**With 3+ paladins, nothing is changed by this PR: natural distribution already covers tanks properly.**

### Configuration (snippets)
```
# Minimum group size for using Greater ranks (default 3)
AiPlayerbot.MinBotsForGreaterBuff = 3

# Cooldown for reagent-missing RP lines (seconds), per bot & per buff (default 30)

# Optional: enable/disable special handling when there are exactly two paladins (bstats + bmana)
# 1 = ON (default), 0 = OFF
AiPlayerbot.Paladin.TwoPallyKingsSanctuary = 1
```

Behavior

**Normal mode (default):**
Uses Greater when the group meets the threshold and reagents are present (Paladin/Druid/Mage families).

**Exactly two paladins (bstats + bmana) with the flag ON:**

- bstats: Sanctuary on tanks; does not cast Kings on tanks.
- bmana: Kings (single) on tanks missing both Kings and Sanctuary; then normal Wisdom/Might on non-tanks.

Three or more paladins: unchanged.

### Test plan

**Greater threshold**

- Group of 3 single.
- Group ≥ threshold with reagents → Greater.
- Change AiPlayerbot.MinBotsForGreaterBuff and verify.

**Reagents handling**

- Remove Symbols of Kings → RP line once per cooldown; single rank cast.
- Restore reagents → Greater resumes.

**Sanctuary on tanks after resurrection**
Two-paladin flag
AiPlayerbot.Paladin.TwoPallyKingsSanctuary = 1 → special handling active (Sanctuary on tanks + Kings single on tanks).
= 0 → original behavior (no special case).
- Kill and resurrect MT/OT:
- bmana applies Kings (single) to tanks lacking buffs,
- bstats applies Sanctuary afterwards which replaces Kings on tanks,
- Might/Wisdom never overwrite Sanctuary.

### Backwards compatibility

- Safe by default: Greater upgrade respects existing thresholds/reagents; Sanctuary remains tank-only; Sanctuary is never overwritten by Kings/Might/Wisdom.
- The “two-paladin” behavior can be disabled via config (= 0) to restore the original behavior entirely.

Thanks for reviewing

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -492,7 +492,7 @@ AiPlayerbot.AutoGearQualityLimit = 3
 AiPlayerbot.AutoGearScoreLimit = 0
 
 # Enable/Disable cheats for bots
-# "food" (bots use cheat when eat or drink)
+# "food" (bots eat or drink without using food or drinks from their inventory)
 # "gold" (bots have infinite gold)
 # "health" (bots have infinite health)
 # "mana" (bots have infinite mana)
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -3244,7 +3244,7 @@ std::vector<uint32> PlayerbotFactory::GetCurrentGemsCount()
 
 void PlayerbotFactory::InitFood()
 {
-    if (!botAI->HasCheat(BotCheatMask::food))
+    if (botAI && botAI->HasCheat(BotCheatMask::food))
     {
         return;
     }


Wishmaster117:
Something wrong with this PR?

brighton-chi:
What do you mean by no behavior change with 3 paladins? Meaning bstats doesn't do sanctuary? Wouldn't that mean that tanks wouldn't get sanctuary at all (3 paladins would presumably be kings, might, and wisdom)?

What does bhealth do currently?

Dreathean:
> Something wrong with this PR?

People probably just been busy recently to take a close look at it

kadeshar:
I think its too complicate for me :D
Im not so fluent in c++
Will check only minor things :P

conf/playerbots.conf.dist
```diff
@@ -461,6 +462,36 @@ AiPlayerbot.FleeingEnabled = 1
...
+# When enabled, in a setup with EXACTLY 2 paladins where one uses the
+# non-combat strategy 'bstats' and the other 'bmana':
+#  - bstats casts SANCTUARY on all tanks (overwrites Kings if present) and
+#    does not cast Kings on tanks;
+#  - bmana casts KINGS (single) on tanks that have neither Kings nor Sanctuary,
+#    and does not attempt Might/Wisdom on tanks.
+# Set to OFF (=0) to restore the original behavior.
+#
+# 1 = ON  |  0 = OFF  (default)
+AiPlayerbot.Paladin.TwoPallyKingsSanctuary = 0
```
kadeshar: Suggesting more official naming

Also this config should be added to PlayerbotAIConfig

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+        if (name == "blessing of might")        return "greater blessing of might";
+        if (name == "blessing of wisdom")       return "greater blessing of wisdom";
+        if (name == "blessing of sanctuary")    return "greater blessing of sanctuary";
+        // Druid
+        if (name == "mark of the wild")         return "gift of the wild";
+        // Mage
+        if (name == "arcane intellect")         return "arcane brilliance";
+
+        return std::string();
+    }
```
kadeshar: Personally i have problem with Prayer of Fortitude which is not on this list.
There is some reason on that?

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    bool HasRequiredReagents(Player* bot, uint32 spellId)
+    {
+        if (!spellId)
+            return false;
+
+        if (SpellInfo const* info = sSpellMgr->GetSpellInfo(spellId))
+        {
+		    for (int i = 0; i < 8; ++i)
```
kadeshar: 8 = magic number
Consider using MAX_SPELL_REAGENTS

Wishmaster117:
> What do you mean by no behavior change with 3 paladins? Meaning bstats doesn't do sanctuary? Wouldn't that mean that tanks wouldn't get sanctuary at all (3 paladins would presumably be kings, might, and wisdom)?
> 
> What does bhealth do currently?

With 3 Paladins no problem, the problem is only with 2 paladins because bstats put sanctuary and King so one paladin = 1 buff so it overwrite the sanctuary so i make the special strat you can activate, so when you have 2 paladins.
Bstats put greater Sanctuary on tanks and Greater King in all exept tanks and bmana put Great Might and Great Windsom in alle except tank, and he put minor king in tankd, so all raid have 2 buffs.

Wishmaster117:
Better it's to try it live before merge

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+        std::function<void(std::string const&)> announce
+    )
+    {
+        std::string castName = baseName;
+		
+		Group* g = bot->GetGroup();
+        if (!g || g->GetMembersCount() < static_cast<uint32>(MinBotsForGreaterBuff()))
+            return castName; // Group too small: stay in solo mode
+
+        if (std::string const groupName = GroupVariantFor(baseName); !groupName.empty())
```
kadeshar: Consider changing name on more user-friendly like:

UpgradeToGroupIfAppropriate -> UpgradeToGreaterIfAppropriate
GroupVariantFor -> GetGreaterSpellName
groupName -> greaterSpellName
groupId -> greaterSpellId

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -13,6 +13,83 @@
...
+#include "Log.h"
+	
+using ai::buff::MakeAuraQualifierForBuff;
+using ai::buff::UpgradeToGroupIfAppropriate;
+
+// Paladin — special logic "2 paladins (bstats + bmana)"
+namespace {
+    inline bool TwoPallyLogicEnabled()
+    {
+        return sConfigMgr->GetOption<bool>("AiPlayerbot.Paladin.TwoPallyKingsSanctuary", false);
```
kadeshar: sPlayerbotAIConfig should be used i think

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -13,6 +13,83 @@
...
+
+    inline bool IsTwoPaladinsBstatsAndSelfBmana(Player* me)
+    {
+        Group* g = me->GetGroup();
+        if (!g) return false;
+        uint32 palCount = 0;
+        bool otherHasBstats = false, selfHasBmana = false;
+        for (auto const& slot : g->GetMemberSlots())
+        {
+            Player* p = ObjectAccessor::FindPlayer(slot.guid);
```
kadeshar: Im not sure its proper way. This method is marked as unsafe
obraz

In playerbots module is often using such aproach
obraz

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +263,198 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+                 bot->GetName(), self->HasTankSpec(), selfHasSanct, needSelf);
+
+        if (needSelf)
+            target = self;
+    }
+
+    // Paladin — special logic "2 paladins (bstats + bmana)"
+    if (TwoPallyLogicEnabled() && IsTwoPaladinsSelfBstatsOtherBmana(bot))
+    {
+        if (auto* ai = GET_PLAYERBOT_AI(bot))
```
kadeshar: Use botAI instead of that

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +263,198 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+        {
+            // Main Tank prioritized if missing Sanctuary (Kings present does not block Sanctuary)
+            if (Unit* mt = ai->GetAiObjectContext()->GetValue<Unit*>("main tank")->Get())
```
kadeshar: use context instead of that

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +263,198 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+            {
+                Player* mtp = mt->ToPlayer();
+                bool mtHasSanct = botAI->HasAura("blessing of sanctuary", mt) ||
+                                  botAI->HasAura("greater blessing of sanctuary", mt);
```
kadeshar: This code is repeated 5 times in this method. 
Consider making method for that

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +263,198 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+    // Paladin — special logic "2 paladins (bstats + bmana)"
+    if (TwoPallyLogicEnabled() && IsTwoPaladinsSelfBstatsOtherBmana(bot))
+    {
+        if (Player* tp = target->ToPlayer())
+        {
+            if (tp->HasTankSpec())
+            {
+                return false;
+            }
+        }
```
kadeshar: In playerbots most often we using snippet

if (botAI->IsTank(bot))

which is shorter

src/strategy/actions/GenericBuffUtils.cpp
Wishmaster117: Historically, in WotLK/DBC, a spell entry (SpellInfo) has 8 reagent slots: Reagent[8] and ReagentCount[8]. That's why I literally used 8. But you're right, it's a "magic number."
You can see I still have some gaps in the module's code base...

src/strategy/paladin/PaladinActions.cpp
Wishmaster117: Refactored

kadeshar:
Additionally what this line doing?

Player* pp = p->ToPlayer();

Its very weird.

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -67,7 +67,9 @@ namespace {
         bool selfHasBstats = false, otherHasBmana = false;
         for (auto const& slot : g->GetMemberSlots())
         {
```
kadeshar: That should be deleted

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -268,14 +270,15 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
     {
         if (Group* g = bot->GetGroup())
         {
-            for (auto const& slot : g->GetMemberSlots())
+        if (Group* g = bot->GetGroup())
+        {
```
kadeshar: Thats already exists

Wishmaster117:
> Additionally what this line doing?
> 
> Player* pp = p->ToPlayer();
> 
> Its very weird.

I'm working since 6 am now it's 11 pm .
I'me weird. 
Tomorrow I will be in work travel so I see that tomorrow night maybe. 
Better i stop before mess up all this thing. :D

Too late have to fix the compil error  but not now

kadeshar:
Wise :D Ok i dont review that more until you say so.

Wishmaster117:
Hello everyone, to finalize this change and keep it blizzlike, I need your help.
Despite searching the internet, I’m finding a lot of contradictory information, so I’m reaching out to the paladin specialists. Here are my questions:
1 paladin in the group — how should they buff the various classes and specs?
2 paladins in the group, bmana + bstats or bmana + bdps, or bstats + bdps or bmana + bdps.... — same question.
3 paladins in the group, bmana + bstats + bdps  — same question.
4 paladins in the group, bmana + bstats + bdps  and another one — same question.

Do paladins’ Greater buffs stack?
For example, does Sanctuary stack with KINGS, whether single-target or Greater? Does Mana stack with KINGS? Etc…

If someone could share all these paladin buff strategies for dungeons and raids, I could finally code them properly, because right now I’m going around in circles a bit.
Thanks, everyone.

Dreathean:
Hey Wishmaster! Yes they do all stack when you have multiple paladins, as in 1 player can simultaneously have all four greater/normal blessings of sanctuary, kings, might, and wisdom IF you have 4 paladins. 

It does get very complicated in practice on how to apply them optimally in party/raid settings when you have limited paladins. It is easiest in fact if you have 4 paladins because then each paladin can just apply one of each greater blessing and cover everything. 

Because a greater blessing automatically applies to every player of a class, greater blessings don't work too well in a raid setting if you only have 1 or 2 paladins, since if you only have 1 paladin with a raid that has multiple druids of different specs for example, it would be very suboptimal to apply greater blessing might to the feral druid when it then automatically gives that same blessing to the resto druid as well.

So I think unfortunately for the average raid, if you only have 1 paladin, you may just need to stick with regular blessings in order to get the optimal outcome.

I don't know enough to say what is the best way to optimize for 2 or 3 paladins, maybe someone with more expertise can chime in on that.

PR #1585, opened 2025-08-22
Food refactoring

kadeshar:
- Changed standalone config on cheat
- Changed drink condition

### DIFF
--- a/src/strategy/actions/NonCombatActions.cpp
+++ b/src/strategy/actions/NonCombatActions.cpp
@@ -17,7 +17,7 @@ bool DrinkAction::Execute(Event event)
     if (!hasMana)
         return false;
 
-    if (sPlayerbotAIConfig->freeFood)
+    if (botAI->HasCheat(BotCheatMask::food))
     {
         // if (bot->IsNonMeleeSpellCast(true))
         //     return false;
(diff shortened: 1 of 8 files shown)


src/strategy/actions/NonCombatActions.cpp
```diff
@@ -54,19 +54,19 @@ bool DrinkAction::Execute(Event event)
     return UseItemAction::Execute(event);
 }
 
-bool DrinkAction::isUseful() { return UseItemAction::isUseful() && AI_VALUE2(uint8, "mana", "self target") < 85; }
+bool DrinkAction::isUseful() { return UseItemAction::isUseful() && AI_VALUE2(uint8, "mana", "self target") < 100; }
```
NoxMax, line 57: Bots drink and eat too frequently already to be honest, it's a bit uncanny how after a small action they sit down and pullout a cheese wheel. I think the current value of 85 is okay. I would even drop it to 80.

Tierisch: > Bots drink and eat too frequently already to be honest, it's a bit uncanny how after a small action they sit down and pullout a cheese wheel. I think the current value of 85 is okay. I would even drop it to 80.

This is essentially just changing the threshold for them to do it when told to. The one for them doing it on their own is staying the same and has a conf option for users to change it to anything below this.

It didn't really make sense that you couldn't tell them to `drink` to full mana if they were already above 85%.

kadeshar: Yeap Trigger:isActive and Action:isUseful are 2 different methods.
That PR resolve this https://github.com/liyunfan1223/mod-playerbots/issues/1457 and here is more explanation

NoxMax: Ahhh I see I see. Okay then <100 makes perfect sense here.

PR #1588, opened 2025-08-24
Update playerbots.conf.dist

Raz0r1337:
Added Sturdy Chest ID to Ignored GUIDs

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1985,13 +1985,13 @@ AiPlayerbot.AllowedLogFiles = ""
 ####################################################################################################
 # A list of gameObject GUID's that are not allowed for bots to interact with.
 #
-AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,2857,179490,141596,160836,160845,179516,176224,181085,176112,128308,128403,165739,165738,175245,175970,176325,176327,123329
+AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,2857,179490,141596,160836,160845,179516,176224,181085,176112,128308,128403,165739,165738,175245,175970,176325,176327,123329,221671
 #
 # List of GUID's:
 # QuestItems:
 # 176213 = Blood of Heroes, 17155 = Defias Gunpowder, 2656 = Waterlogged Envelope, 123329 = Baelogs Chest
 # Chests:
-# Large Solid Chest = 74448, Box of Assorted Parts = 19020, Food Crate = 3719, Water Barrel = 3658, Barrel of Milk = 3705, Barrel of sweet Nectar = 3706, Tattered Chest = 105579, Large bettered Chest = 75293, Solid Chest = 2857, Battered Foodlocker = 179490, Witch Doctor's Chest = 141596, Relic Coffer = 160836, Dark Coffer = 160845, Fengus's Chest = 179516, Supply Crate = 176224/181085, Malor's Strongbox = 176112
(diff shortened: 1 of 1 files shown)


conf/playerbots.conf.dist
```diff
@@ -1985,13 +1985,13 @@ AiPlayerbot.AllowedLogFiles = ""
...
 #
 # List of GUID's:
 # QuestItems:
 # 176213 = Blood of Heroes, 17155 = Defias Gunpowder, 2656 = Waterlogged Envelope, 123329 = Baelogs Chest
 # Chests:
-# Large Solid Chest = 74448, Box of Assorted Parts = 19020, Food Crate = 3719, Water Barrel = 3658, Barrel of Milk = 3705, Barrel of sweet Nectar = 3706, Tattered Chest = 105579, Large bettered Chest = 75293, Solid Chest = 2857, Battered Foodlocker = 179490, Witch Doctor's Chest = 141596, Relic Coffer = 160836, Dark Coffer = 160845, Fengus's Chest = 179516, Supply Crate = 176224/181085, Malor's Strongbox = 176112
+# Large Solid Chest = 74448, Box of Assorted Parts = 19020, Food Crate = 3719, Water Barrel = 3658, Barrel of Milk = 3705, Barrel of sweet Nectar = 3706, Tattered Chest = 105579, Large bettered Chest = 75293, Solid Chest = 2857, Battered Foodlocker = 179490, Witch Doctor's Chest = 141596, Relic Coffer = 160836, Dark Coffer = 160845, Fengus's Chest = 179516, Supply Crate = 176224/181085, Malor's Strongbox = 176112, Sturdy Chest = 221671
```
kadeshar, line 1994: I dont see reason why disallow bots looting this chest. This chest dont contains any item which is require to quest/usage.
Could explain more?

EDIT: that is even wotlk item?
obraz

noisiver:
221671 isn't even a valid object ID.
{DDCECAA9-C72F-42C0-AC54-CB8E1E93ACE9}

conf/playerbots.conf.dist
Raz0r1337: I honestly thought that bots shouldn't just loot all the possible chests on their own. So I'm constantly looking for chests that bots aren't ignoring and adding them to the list. Personally, it annoys me when bots constantly loot all the chests, but if that's not the point, then feel free to close this PR. I just wanted to give something back and thought you might have overlooked this one, since all the other chests I put on the list stayed there, and this one doesn't really contain anything different than most of the others.

noisiver: It's a retail chest only existing in Timeless Isle. It's not a WotLK Classic or 3.3.5 object. I don't know where you got the ID for it, but I would also guess you took the GUID rather than the ID. (maybe?)

Personally I don't mind the bots looting the chests since that means I don't have to bother. I can just remove entries from the list on my end though, it's not to say this shouldn't be in here.

kadeshar: I think that is very personal thought.
Putting whole https://www.wowhead.com/wotlk/objects/chests and/or https://www.wowhead.com/wotlk/objects/containers to config is not solution.
You can always remove loot strategy from bots.
Another solution is to make config which disallow bots to loot any chest or/and containers.

Raz0r1337:
https://www.wowhead.com/object=221671/sturdy-chest

It's located in Stranglethorn Vale

kadeshar:
Links you provided say something different
obraz

noisiver:
> https://www.wowhead.com/object=221671/sturdy-chest
> 
> It's located in Stranglethorn Vale

No, it's not. https://wowgaming.altervista.org/aowow/?object=221671

Raz0r1337:
I think it's a mistake on my side.
It seems that there was a solid chest that are not ignored by the bots. It was shown to me above my mouse as a sturdy-chest. 🤷
On my GM account it says it is a solid chest.

I will research about it and if I find anything I will continue. 😉

Raz0r1337:
It is an translation issue!

In German solid chest and sturdy chest are both robuste truhe. 🤦

brighton-chi:
IMO the approach for the default list should be to include only quest items and objects that a competent player would not loot without thought due to the consequences (e.g., ZF zombie graves).

Any other objects, like generic chests, should be for players to add if they want them to be excluded.

PR #1589, opened 2025-08-25
Add SMV area IDs to PvP Prohibited Areas

brighton-chi:
Sanctum of the Stars and the Altar of Sha’tar are cross-faction quest/flight hubs in Shadowmoon Valley where PvP should be disabled.


brighton-chi:
You can confirm the area IDs here: https://www.azerothcore.org/wiki/areatable

It is very confusing, but I believe area ID in that table is actually the zone ID, and field nb is the area ID. That being said, due to a quirk in the code and the way IDs work for the Core (we had a discussion on Discord about this a few weeks ago), it does not matter whether an ID is included in zone or area IDs in the config. Either way works.

I tested Sanctum of the Stars, and it works. I did not test Altar of Sha'tar due to my lack of an Aldor character, but I see no reason why it should not work the exact same way.

PR #1591, opened 2025-08-25
Has spell check

Badgermilk0:
Meant to reduce amount of duplicate spell entry errors when bots changes levels: 

SQL(p): INSERT INTO character_spell (guid, spell, specMask) VALUES (xxxxx, xxxxx, xxx)
 [ERROR]: [1062] Duplicate entry 'xxxxx-xxxxx' for key 'character_spell.PRIMARY'
 
 Does not solve all errors - still get a lot specifically for:
 54197 - cold weather flying

### DIFF
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -2697,12 +2697,18 @@ void PlayerbotFactory::InitSpecialSpells()
          i != sPlayerbotAIConfig->randomBotSpellIds.end(); ++i)
     {
         uint32 spellId = *i;
-        bot->learnSpell(spellId);
+        if (!bot->HasSpell(spellId))
+        {
+            bot->learnSpell(spellId);
+        }
     }
     // to leave DK starting area
     if (bot->getClass() == CLASS_DEATH_KNIGHT)
     {
-        bot->learnSpell(50977, false);
+        if (!bot->HasSpell(50977))
+        {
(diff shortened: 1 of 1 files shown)


src/factory/PlayerbotFactory.cpp
```diff
@@ -2574,8 +2574,8 @@ void PlayerbotFactory::InitAvailableSpells()
 
             if (tSpell->IsCastable())
                 bot->CastSpell(bot, tSpell->spell, true);
-            else
-                bot->learnSpell(tSpell->learnedSpell[0], false);
+            else if (!bot->HasSpell(tSpell->learnedSpell[0]))
```
kadeshar, line 2577: learnSpell method have his own check that player have this spell (even more restictive). For me that checks do nothing but maybe im wrong

HasSpell
return (itr != m_spells.end() && itr->second->State != PLAYERSPELL_REMOVED && itr->second->IsInSpec(m_activeSpec));

learnSpell->HasActiveSpell
return (itr != m_spells.end() && itr->second->State != PLAYERSPELL_REMOVED && itr->second->Active && itr->second->IsInSpec(m_activeSpec));

Badgermilk0: Then my testing might have been circumstantial, when applying the hasSpell the console output of duplicate spells severely got reduced.
I will iterate longer to see if the issue is elsewhere.

PR #1593, opened 2025-08-26
Fixed bug with RTSC despawning objects

kadeshar:
obraz

Resolves: https://github.com/liyunfan1223/mod-playerbots/issues/1444


brighton-chi:
FYI, this issue was due to the aedm spell, which is an existing WoW spell that was adapted, having the effect of despawning objects.

Credit to @Tierisch for identifying. This is huge.

kadeshar:
Thanks, didnt know he has github account :D

noisiver:
The file should probably be called world_spell_dbc.sql to match the naming style of the other files in there. It's also missing a newline at the end.

kadeshar:
Sql structure dont have any normalization. Every folder have different naming convention/structure.
For me that dont fit anywhere. Thats should be custom script to world but structure dont assume that. Other scripts in this folder create table. Not insert row.

noisiver:
They do have a specific structure of names and it doesn't really matter if it creates table or inserts data but sure. It's more a standard to name it based on the table it changes.

kadeshar:
Dont agree

Character database:
Dont have structure for updates
It have only creation scripts named from table name

World database:
Dont have structure for updates
It have only creation scripts named from database type and table name

Playerbots database
Have structure for updates
It have scripts for row modification in updates folder

I can unify that ofc because currently is mess.

kadeshar:
@noisiver could you look now, that is correct?

hermensbas (approved):

PR #1596, opened 2025-08-27
Possible hacking attempt bugfix

kadeshar:
- Fixed bug where bot looting gameobject which is on respawn time


hermensbas:
nice work

PR #1601, opened 2025-08-28
Add chat filters for aura, aggro, and spec and make all filters case insensitive

brighton-chi:
This PR adds three new chat filters (in addition to some other tweaks noted below). If merged, this PR will require an update to the Wiki.

**AggroByChatFilter**
The new command prefix "`@`aggroby" will allow filtering by bots that are being targeted by a specified enemy. The enemy can be entered by their entryid or by their name (if the enemy has a multiple-word name, it will need to be enclosed in quotation marks). It is optional whether to put a space between the filter and the entryid/creature name. I am hoping that this can provide for more ability to control bots in fights with multiple mobs, adds, aggro dumps, and fixates (and when combined with RTSC, serve as a poor man’s kiting strategy for bots).

**AuraChatFilter**
The new command prefixes "`@`aura" and "`@`noaura" will allow filtering by whether a bot has or does not have, respectively, a particular aura (specified by its id). It is optional whether or not to put a space between the filter name and the id. I think this can be useful for boss fights to react to certain boss abilities. For example, I think this filter may actually make Thaddius in Naxx40 killable without coded strategies--I have not tested that fight, but I expect you could set two separate RTSC spots, and then when he casts Polarity Shift, you could execute a macro that has all bots with the Positive Charge aura go to one RTSC spot and all bots with the Negative Charge aura go to the other RTSC spot.

**SpecChatFilter**
You can now filter by spec (with Feral Druids and Blood DKs further filterable by tank vs. dps). The command prefixes for specs are not very intuitive (e.g., "hpal" for holy paladin) because I wanted to keep them very short. I expect any use of the filter to be through macros and not on the fly and therefore I think it is better to have the prefixes be short in light of the 255-character macro limit. Perhaps the wiki could point users to ChatFilter.cpp for the correct command prefixes.

**Case Insensitivity**
All filters (new and existing) are now case insensitive, which should be much more user friendly.

**ClassFilter for Death Knights**
The command prefix to filter for Death Knights was previously “`@`death_knight”, which is unwieldy to type on the fly and takes up too many characters for macros. I have shortened it to “`@`dk”.

**Logic For Running ChatFilters**
I modified the process to apply filters to make it a single loop that runs through each filter once from top to bottom. For some reason, the prior logic would run each filter multiple times--specifically, it would run each filter the number of times that is equal to the total number of filters, meaning each filter 6 times before this PR and 9 times after. I see no reason why this is necessary, and it was driving me nuts trying to log for testing. If there is a good reason for the existing logic, I can revert.


dk:
Dear Brighton

In case you don't know, setting @ before a user id spams that user id with the full message.
That was the case when you decided to shorten @ death knights to @ dk. 
I assume you did that in ignorance, not because you wanted to reach me.
But even if I wrong, kindly don't do that again, it's a bit annoying

Thank you

brighton-chi:
Weird that tagging is the default. Seems like the best way to avoid it is to make the `@` show up as code, so we'll go with that.

src/ChatFilter.cpp
```diff
@@ -295,6 +308,290 @@ class SubGroupChatFilter : public ChatFilter
...
+
+        size_t spacePos = auraIdOrName.find(' ');
+        std::string idStr = (spacePos != std::string::npos) ? auraIdOrName.substr(0, spacePos) : auraIdOrName;
+        if (spacePos != std::string::npos)
+        {
+            rest = auraIdOrName.substr(spacePos + 1);
+        }
+        else
+        {
+            rest = "";
```
kadeshar: rest in declariation have assigned this value

src/ChatFilter.cpp
```diff
@@ -295,6 +308,290 @@ class SubGroupChatFilter : public ChatFilter
...
+            }
+        }
+
+        // Use nearest hostile npcs value and filter by entryID or name within 100 yards
+        const float radius = 100.0f;
+        GuidVector npcs = botAI->GetAiObjectContext()->GetValue<GuidVector>("nearest hostile npcs")->Get();
+        bool match = false;
+        for (const auto& guid : npcs)
+        {
+            Unit* unit = botAI->GetUnit(guid);
```
kadeshar: Use GetCreature instead of GetUnit making check in 573 and cast from 577 unnecesary

kadeshar (approved):

brighton-chi:
I will update for your comments soon. Got focused on Karazhan strats in the meantime.

brighton-chi:
@kadeshar 

This should be good now. I made the changes based on your comments. I also noticed neutral creatures were not being included by the aggroby filter, even after they aggroed and were targeting the bot, because "nearest hostile npcs" excludes them (changed to "nearest npcs"). Finally, aura and aggroby filters now trim spaces so it's optional whether or not to put a space between the filter and the ID/creature name.

kadeshar:
I approved PR. Its waiting on other reviewer.

brighton-chi:
> I approved PR. Its waiting on other reviewer.

I know. I mentioned it to you only because you approved it before I even incorporated your comments =P

hermensbas:
Should we update the wiki first?

kadeshar:
> Should we update the wiki first?

I think wiki dont describe well chat filters.
Maybe new section is required on that.

brighton-chi:
Yeah, I think a new heading for Chat Filters is good. I think we can do something like add the following under the existing header of Party/Raid General Commands: "Commands can be issued to the entire party or raid by messaging them in party or raid chat (note that when in a raid, party chat will issue commands to the entire raid). Commands can also be send to individual bots by whisper. Finally, commands can be issued to a subset of bots in the party or raid by using chat filters (see below)."

Then we can add a new section like the attached doc (can't send directly in the message due to there being a table, but LMK if there's a better way to provide the language).

[Chat filters.docx](https://github.com/user-attachments/files/22391503/Chat.filters.docx)

PR #1603, opened 2025-08-29
Paladin buff logic: Sanctuary+Kings synergy, role-aware targeting, safer Greater buffs

Wishmaster117:
### What this changes

This PR revises how Paladin bots choose and apply blessings so the behavior matches WotLK 3.3.5a expectations and typical raid assignments, while fixing a few stability issues.

### Core behavior

- Sanctuary + Kings can now coexist on the same target (as in WotLK).
- No overwrites on tanks: Kings will never overwrite Sanctuary on tanks.
- Greater vs Single: still uses AiPlayerbot.MinBotsForGreaterBuff + reagents, but:

-  Sanctuary is always single on tanks (any class).
-  bmana Kings is always single (tanks only).
-  bstats Kings is Greater only on non-tankable classes (Rogue, Hunter, Mage, Warlock, Priest, Shaman). It won’t GBOK on   Paladin/Warrior/Druid/DK to avoid splashing tanks.

### Per-strategy assignments (non-combat)

- bstats (tank/def):

- Cast Sanctuary (single) on all tanks, including Paladin, Warrior, Druid (Bear), Death Knight.
- Tank detection is now role-aware (spec or active tank strategies like tank, tank face, bear).
- Then cast Kings (Greater) on everyone else (classes that cannot be tanks).

- bmana (holy/support):

- Cast Greater Wisdom/Might on non-tanks (auto-select by role).
- Cast Kings (single) only on tanks.
- Never casts Wisdom/Might on tanks.

- bdps (ret): unchanged (Might/Wisdom on appropriate roles), and will not fight with Sanctuary.

### Why

- Aligns bot behavior with WotLK mechanics: Kings stacks with Might/Wisdom when coming from different Paladins; Sanctuary stacks with Kings (unique DR/mana return + 10% stats).
- Produces sensible raid-like assignments automatically with two Paladins (e.g., tank = Sanctuary + Kings, melee = Might, casters = Wisdom).
- Prevents GBOK from unintentionally landing on tanks via class-wide casts.

### Gameplay impact

- Tanks (of any class) reliably end up with Sanctuary (single) + Kings (single).
- Non-tanks receive Greater versions when the group meets the threshold and reagents are available.
- Less buff churn; no accidental overwrites of Sanctuary on tanks.

### Config notes

- AiPlayerbot.MinBotsForGreaterBuff still controls when Greater buffs are considered.
- Requires Symbols of Kings for Greater casts (unchanged).

### Testing checklist

1. Group with 2 Paladins: bstats + bmana and a mix of melee/casters + a non-paladin tank (e.g., Bear).

- Bear/Prot/DK tanks → Sanctuary (single) + Kings (single).
- Melee → Greater Might, Casters → Greater Wisdom.
- No GBOK on tankable classes.

2. Remove Symbols of Kings → Paladins fall back to single buffs with an RP warning (existing behavior).

3. Toggle tank strategies on a Druid (add/remove bear/tank face) → Sanctuary targeting follows role.


src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+namespace {
+    inline int32 RPWarningCooldown()
+    {
+        static int32 v = sConfigMgr->GetOption<int32>("AiPlayerbot.RPWarningCooldown", 30);
+        return v;
+    }
+
+    inline int32 MinBotsForGreaterBuff()
+    {
+        static int32 v = sConfigMgr->GetOption<int32>("AiPlayerbot.MinBotsForGreaterBuff", 3);
```
kadeshar: All configs should be implement into PlayerbotAIConfig and sPlayerbotAIConfig used in code

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+        // Paladin
+        if (name == "blessing of kings")        return "greater blessing of kings";
+        if (name == "blessing of might")        return "greater blessing of might";
+        if (name == "blessing of wisdom")       return "greater blessing of wisdom";
+        if (name == "blessing of sanctuary")    return "greater blessing of sanctuary";
+        // Druid
+        if (name == "mark of the wild")         return "gift of the wild";
+        // Mage
+        if (name == "arcane intellect")         return "arcane brilliance";
+
```
kadeshar, line 56: Still missing Prayer of Fortitude. Is intentionally?

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    bool HasRequiredReagents(Player* bot, uint32 spellId)
+    {
+        if (!spellId)
+            return false;
+
+        if (SpellInfo const* info = sSpellMgr->GetSpellInfo(spellId))
+        {
+		    for (int i = 0; i < 8; ++i)
```
kadeshar: 8 = magic number
Consider using MAX_SPELL_REAGENTS

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,147 @@
...
+                        return false;
+                }
+            }
+            // No reagent required
+            return true;
+        }
+        return false;
+    }
+
+    std::string UpgradeToGroupIfAppropriate(
```
kadeshar: Consider changing name on more user-friendly like:

UpgradeToGroupIfAppropriate -> UpgradeToGreaterIfAppropriate
GroupVariantFor -> GetGreaterSpellName
groupName -> greaterSpellName
groupId -> greaterSpellId

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -13,6 +13,38 @@
...
+            otherAI->HasStrategy("tank face", BOT_STATE_COMBAT)     ||
+            otherAI->HasStrategy("bear", BOT_STATE_NON_COMBAT) ||
+            otherAI->HasStrategy("bear", BOT_STATE_COMBAT))
+            return true;
+    }
+    return false;
+}
+
+
+static inline bool GroupHasTankOfClass(Group* g, uint8 klass)
```
kadeshar: We using english in code. Should be classId or something

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
-    return botAI->CastSpell(GetActualBlessingOfWisdom(target), target);
+    if (Group* g = bot->GetGroup())
+        if (Player* tp = target->ToPlayer())
+            if (!g->IsMember(tp->GetGUID()))
+                return false;
+
+    if (botAI->HasStrategy("bmana", BOT_STATE_NON_COMBAT))
+        if (Player* tp = target->ToPlayer())
+            // if (tp->HasTankSpec())
+			if (IsTankRole(tp))
```
kadeshar: I dont even now that in c++ multiline code below if without {} work.
Even that shouldnt be used because readability

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+            {
+                LOG_DEBUG("playerbots", "[PallyBuff][Wisdom/bmana] Skip tank {} (Kings only)", target->GetName());
```
kadeshar: [Palladin buff]

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+    );
 }
 
 bool CastBlessingOfSanctuaryOnPartyAction::Execute(Event event)
+{
+    if (!bot->HasSpell(SPELL_BLESSING_OF_SANCTUARY)) return false;
+
+    Unit* target = GetTarget();
+    if (!target)
+    target = bot;
```
kadeshar: inline if without tab. I dont know even why assigning bot to target.

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+
+    if (Group* g = bot->GetGroup())
+        if (Player* tp = target->ToPlayer())
+            if (!g->IsMember(tp->GetGUID()))
+            {
+                LOG_DEBUG("playerbots", "[PallyBuff][Sanct] Initial target not in group, ignoring");
+                target = bot;
+            }
```
kadeshar: Another multiline code for inline if

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+
+    if (Player* self = bot->ToPlayer())
+    {
+        bool selfHasSanct = botAI->HasAura("blessing of sanctuary", self) ||
+                            botAI->HasAura("greater blessing of sanctuary", self);
+        // bool needSelf = self->HasTankSpec() && !selfHasSanct;
+		bool needSelf = IsTankRole(self) && !selfHasSanct;
+
+        LOG_DEBUG("playerbots", "[PallyBuff][Sanct] {} isTank={} selfHasSanct={} needSelf={}",
```
kadeshar: [Palladin buff]

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+                 //bot->GetName(), self->HasTankSpec(), selfHasSanct, needSelf);
+				 bot->GetName(), IsTankRole(self), selfHasSanct, needSelf);
+
+        if (needSelf)
+            target = self;
+    }
+
+    {
```
kadeshar, line 313: Why this {}

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+			targetOk = IsTankRole(tp) && !hasSanct;
+        }
+
+        if (!targetOk)
+        {
+            if (Group* g = bot->GetGroup())
+            {
+                for (auto const& slot : g->GetMemberSlots())
+                {
+                    if (Player* p = ObjectAccessor::FindPlayer(slot.guid))
```
kadeshar: obraz

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+                }
+            }
+        }
+    }
+
+    {
+        bool hasKings = botAI->HasAura("blessing of kings", target) ||
+                        botAI->HasAura("greater blessing of kings", target);
+        bool hasSanct = botAI->HasAura("blessing of sanctuary", target) ||
+                        botAI->HasAura("greater blessing of sanctuary", target);
```
kadeshar: 3 times same code in one method

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+            auto RP = ai::chat::MakeGroupAnnouncer(bot);
+            castName = ai::buff::UpgradeToGroupIfAppropriate(bot, botAI, castName, /*announceOnMissing=*/true, RP);
+        }
+    }
+    else
+    {
+        castName = "blessing of sanctuary";
+    }
+    bool ok = botAI->CastSpell(castName, target);
+    LOG_DEBUG("playerbots", "[PallyBuff][Sanct] Cast {} on {} result={}", castName, target->GetName(), ok);
```
kadeshar: Remove all PallyBuff

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +213,244 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+    std::string castName = "blessing of kings";
+
+    bool allowGreater = true;
+
+    if (botAI->HasStrategy("bmana", BOT_STATE_NON_COMBAT))
+        allowGreater = false;
+
+    if (allowGreater && botAI->HasStrategy("bstats", BOT_STATE_NON_COMBAT))
+    {
+        if (Player* tp = target->ToPlayer())
```
kadeshar: 4 times is one method

Wishmaster117: Ok removed All

src/strategy/actions/GenericBuffUtils.cpp
kadeshar: That is not resolved. I still this that using word "Group" in this place is wrong

Wishmaster117: I do see the point that “Greater” is a clearer game-term.
In this code path, though, the naming with “group” is intentional and maps to the behavior of these spells in 3.3.5a: the “greater” variants are group-wide buffs (e.g., Gift of the Wild, Arcane Brilliance, Greater Blessings), not single-target. That’s why the helpers currently use GroupVariantFor, UpgradeToGroupIfAppropriate, and local variables like groupName/groupId — they describe upgrading from a single-target buff to its group counterpart, across classes (Paladin/Druid/Mage), not only Paladin “Greater Blessing” terminology.

Dreathean: I can definitely see the argument in favor of maintaining “groupvariant” here instead of “greatervariant” given it is then uniform across the different classes. Personally as long as the functionality is an improvement and this part isn’t visible to the end user in the game, I don’t think this should hold up the PR by itself?

Wishmaster117: The mod is working properly in my test server, as all things it may need some features added later i don't know, but it's better than before.

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+    {
+        std::string castName = baseName;
+		
+		Group* g = bot->GetGroup();
+        if (!g || g->GetMembersCount() < static_cast<uint32>(sPlayerbotAIConfig->minBotsForGreaterBuff))
+            return castName; // Group too small: stay in solo mode
+
+        if (std::string const groupName = GroupVariantFor(baseName); !groupName.empty())
+        {
+         uint32 const groupId = botAI->GetAiObjectContext()
```
kadeshar: Its not groupId then this name provides confusion. Should be groupVariantSpellId or something like this.

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +228,255 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+    if (Group* g = bot->GetGroup())
+        if (targetPlayer && !g->IsMember(targetPlayer->GetGUID()))
+            return false;
+
+    if (botAI->HasStrategy("bmana", BOT_STATE_NON_COMBAT))
+        if (targetPlayer && IsTankRole(targetPlayer))
+        {
+            LOG_DEBUG("playerbots", "[Wisdom/bmana] Skip tank {} (Kings only)", target->GetName());
+            return false;
+        }
```
kadeshar: Multiline code below if without {}

src/strategy/paladin/PaladinActions.cpp
kadeshar: No response, and not fixed.

src/strategy/paladin/PaladinActions.cpp
kadeshar: No response, and not fixed.

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +228,255 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+            else
+                return false;
+        }
+        else
+            return false;
+    }
+
+    if (targetPlayer)
+    {
+		if (IsTankRole(targetPlayer))
```
kadeshar: if (targetPlayer && !IsTankRole(targetPlayer)) reduce one if nest

src/strategy/paladin/PaladinActions.cpp
```diff
@@ -155,21 +228,255 @@ bool CastBlessingOfWisdomOnPartyAction::Execute(Event event)
...
+    LOG_DEBUG("playerbots", "[Sanct] Cast {} on {} result={}", castName, target->GetName(), ok);
+    return ok;
+}
+
+Value<Unit*>* CastBlessingOfKingsOnPartyAction::GetTargetValue()
+{
+    /*return context->GetValue<Unit*>(
+        "party member without aura",
+        "blessing of kings,greater blessing of kings"
+    );*/ //old version before solo paladin patch
```
kadeshar: Commented code

src/strategy/paladin/PaladinTriggers.h
```diff
@@ -234,13 +234,22 @@ class BlessingOfMightOnPartyTrigger : public BuffOnPartyTrigger
...
 
-class BlessingOfSanctuaryOnPartyTrigger : public BuffOnPartyTrigger
+/*class BlessingOfSanctuaryOnPartyTrigger : public BuffOnPartyTrigger
 {
 public:
     BlessingOfSanctuaryOnPartyTrigger(PlayerbotAI* botAI)
         : BuffOnPartyTrigger(botAI, "blessing of sanctuary,blessing of kings", 2 * 2000)
     {
     }
+};*/
```
kadeshar: Commented code

src/strategy/paladin/PaladinBuffStrategies.cpp
```diff
@@ -9,11 +9,15 @@
 
 void PaladinBuffManaStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
-    triggers.push_back(
+    /*triggers.push_back(
         new TriggerNode("blessing of wisdom on party",
-                        NextAction::array(0, new NextAction("blessing of wisdom on party", 11.0f), nullptr)));
-    // triggers.push_back(new TriggerNode("blessing", NextAction::array(0, new NextAction("blessing of wisdom",
-    // ACTION_HIGH + 8), nullptr)));
+                        NextAction::array(0, new NextAction("blessing of wisdom on party", 11.0f), nullptr)));*/
```
kadeshar: Commented code

src/strategy/actions/GenericBuffUtils.cpp
noisiver: Missing spaces in the new line after the curly bracket as well

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+            static std::map<std::pair<uint32, std::string>, time_t> s_lastWarn; // par bot & par buff
+            time_t now = std::time(nullptr);
+            uint32 botLow = static_cast<uint32>(bot->GetGUID().GetCounter());
+            time_t& last = s_lastWarn[ std::make_pair(botLow, groupName) ];
+			if (!last || now - last >= sPlayerbotAIConfig->rpWarningCooldown) // Configurable anti-spam
+              {
+                  std::string rp;
+                  if (groupName.find("greater blessing") != std::string::npos)
+                      rp = "By the Light... I forgot my Symbols of Kings. We’ll make do with the simple blessings!";
+                  else if (groupName == "gift of the wild")
```
VincentVanclef: Question, why are we using magic strings and not enums in this project? I see it all over the place, even in the core logic of playerbots e.g. passing in "actions" as "strings" why is it not enums? Its a million times faster and less error prone. Just really curious why strings are prefered

kadeshar: There is a lot of code to refactor and no peoples which want to do that :D

VincentVanclef: That is understandable, it could perhaps be done using Codex AI to easily refactor it

kadeshar: Without good design which codex for sure will not provide is impossible.

brighton-chi:
Do Sanctuary and Kings stack? They are not supposed to (nor is Sanctuary supposed to stack with Grace). Blizzlike behavior would be that Sanctuary is useless for Warriors and Bears, and it's a trade-off for Paladins, though I think it is generally considered the better buff than Kings for them (you lose 10% agility and 10% intellect in exchange for the mana restoration).

Wishmaster117:
> Do Sanctuary and Kings stack? They are not supposed to (nor is Sanctuary supposed to stack with Grace). Blizzlike behavior would be that Sanctuary is useless for Warriors and Bears, and it's a trade-off for Paladins, though I think it is generally considered the better buff than Kings for them (you lose 10% agility and 10% intellect in exchange for the mana restoration).

Blizzlike behavior:
One blessing per paladin per target (the same paladin cannot place two different blessings on you).

Multiple blessings in total are possible if they come from different paladins (e.g., Kings from Paladin A + Sanctuary from Paladin B). Identical effects (such as Stamina) did not stack, but the blessings themselves did coexist.

Source : https://www.bluetracker.gg/wow/topic/us-en/17899680722-paladin-buffs-and-a-nerf/

brighton-chi:
I understand they can coexist. What I mean is that if their stat increases do not stack, since the damage reduction also does not stack with Grace and it is very likely a raid will have a priest with that talent (or there could be a check for a priest with that talent), then Sanctuary is effectively worthless for non-Paladin tanks because Kings is strictly better.

If it's complicated to address, I don't think it matters much for non-Paladin tanks to lose out on 10% agility. I know this is a hell of an endeavor. I just wanted to raise the point.

hermensbas (approved):

Wishmaster117:
> I understand they can coexist. What I mean is that if their stat increases do not stack, since the damage reduction also does not stack with Grace and it is very likely a raid will have a priest with that talent (or there could be a check for a priest with that talent), then Sanctuary is effectively worthless for non-Paladin tanks because Kings is strictly better.
> 
> If it's complicated to address, I don't think it matters much for non-Paladin tanks to lose out on 10% agility. I know this is a hell of an endeavor. I just wanted to raise the point.

It's possible to don't add Sanctuary to a non paladin tank, just few conditions to add, but for now need this PR be merged and after i make adjustments

hermensbas:
> t's complicated to address, I don't think it matters much for non-Paladin tanks to lose out on 10% agility. I know this is a hell of an endeavor. I just wanted to raise the point.

I think we need add additional strategy pattern for finding the applicable buffs, but more in general purpose usage apart from this merge possibly?

Wishmaster117:
> > t's complicated to address, I don't think it matters much for non-Paladin tanks to lose out on 10% agility. I know this is a hell of an endeavor. I just wanted to raise the point.
> 
> I think we need add additional strategy pattern for finding the applicable buffs, but more in general purpose usage apart from this merge possibly?

With this PR we already have a solid foundation, it’s much better than the default system. Now we can make adjustments so that paladins buff as accurately as possible. Try it like this first and report back the adjustments that need to be made.

It0w:
I got the following error at server start after compiling the server.

```
Updating Playerbots database...
>> Applying update "2025_09_17_00_paladin_buff_reagent_texts.sql" 'B780AE6'...
ERROR 1364 (HY000) at line 17 in file: '/mnt/raid0/wow_server/azerothcore-playerbots/modules/mod-playerbots/data/sql/playerbots/updates/2025_09_17_00_paladin_buff_reagent_texts.sql': Field 'text_loc3' doesn't have a default value

Applying of file '/mnt/raid0/wow_server/azerothcore-playerbots/modules/mod-playerbots/data/sql/playerbots/updates/2025_09_17_00_paladin_buff_reagent_texts.sql' to database 'acore_playerbots' failed! If you are a user, please pull the latest revision from the repository. Also make sure you have not applied any of the databases with your sql client. You cannot use auto-update system and import sql files from AzerothCore repository with your sql client. If you are a developer, please fix your sql query.

Could not update the Playerbots database, see log for details.
Closing down DatabasePool 'acore_playerbots'. Waiting for 0 queries to finish...
Asynchronous connections on DatabasePool 'acore_playerbots' terminated. Proceeding with synchronous connections.
All connections on DatabasePool 'acore_playerbots' closed.
```

noisiver:
Looking at the code, it doesn't look like the announcements can be disabled. In that case it's not good. The ability to turn that stuff off is kind of required as not everyone wants it.

kadeshar:
I would say that we need some flags or something unified for that. Adding new config for each notification sounds bad.

noisiver:
There are options in the config for various types of chat, it wouldn't be bad to use one of those. I don't know which one though.

I just know that a lot of people don't need the bots telling them that they're missing reagents, especially not when they're nowhere near a vendor to buy them and they don't want to use the maintenance command. Maybe I'm wrong but looking at the history of playerbots it should be correct.

Wishmaster117:
> There are options in the config for various types of chat, it wouldn't be bad to use one of those. I don't know which one though.
> 
> I just know that a lot of people don't need the bots telling them that they're missing reagents, especially not when they're nowhere near a vendor to buy them and they don't want to use the maintenance command. Maybe I'm wrong but looking at the history of playerbots it should be correct.

You have a anti spam config :

Just increase the value.

In my idea bots have to react like a real player, and sometimes a real player say things like that. 

I posted a fix to the problem of It0w I think it's because the text fields have no value in the insert

noisiver:
Not everyone wants the bots to do that, simply telling them to increase the value (they still say it) isn't the best solution. At least I can just scrap that code on my end since I know how to but others may not.

Wishmaster117:
> Not everyone wants the bots to do that, simply telling them to increase the value (they still say it) isn't the best solution. At least I can just scrap that code on my end since I know how to but others may not.

You're free to make a PR to remove it, i don't mind.

It0w:
@Wishmaster117 
> I posted a fix to the problem of It0w I think it's because the text fields have no value in the insert

Thanks! I have tested it, and now the server starts successfully.

PR #1611, opened 2025-09-02
Fix thresholds for LimitEnchantExpansion

brighton-chi:
The thresholds for LimitEnchantExpansion are wrong. I noticed that bots were applying 23 agility to gloves at level 70, and that is a WotLK enchant. The source determines which enchants to limit by spell ID thresholds and calls enchants with spell ID < 25072 as being Vanilla enchants and enchants with spell ID <= 48577 as being WotLK enchants. This results in too wide of a window for TBC—there are Vanilla enchants excluded from level 60 and below bots, and there are WotLK enchants available to level 61 through 70 bots.

The lowest spell ID for a TBC enchant is 27899 (Enchant Bracer – Brawn). No Vanilla enchants have a greater spell ID. See the following filtered search results:

- No Vanilla enchants with >= 27900 spell ID: https://www.wowhead.com/wotlk/spells/professions/enchanting?filter=16:14;1:2;0:27900
- No TBC enchants with < 27899 spell ID: https://www.wowhead.com/wotlk/spells/professions/enchanting?filter=16:14;2:5;0:27899

The lowest spell ID for a WotLK enchant is 44483 (Enchant Cloak – Superior Frost Resistance). There are three TBC enchants with a greater spell ID that need to be accounted for individually. See the following filtered search results:

- No WotLK enchants with < 44483 spell ID: https://www.wowhead.com/wotlk/spells/professions/enchanting?filter=16:14;3:5;0:44483
- Three TBC enchants with >= 44483 spell ID: https://www.wowhead.com/wotlk/spells/professions/enchanting?filter=16:14;2:2;0:44483 

I don't know if any bots would actually apply any of the three TBC enchants with spell ID >= 44483, but to be technically correct, I have excluded them specifically from the level 70 limit.

- Enchant Weapon – Deathfrost (46578)
- Enchant Chest – Defense (46594)
- Enchant Cloak – Steelweave (47051)

I also cleaned up what I thought was a very confusing comment in the config (the existing language makes it sound like setting MinEnchantingBotLevel to be greater than the maximum allowable bot level would somehow disable any minimum level requirement for bots to enchant their equipment, when really it prevents bots from enchanting their equipment at all).

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -659,7 +659,7 @@ AiPlayerbot.RandomGearScoreLimit = 0
 # Default: 1 (enabled)
 AiPlayerbot.IncrementalGearInit = 1
 
-# Set minimum level of bots that will enchant their equipment (Maxlevel + 1 to disable)
+# Set minimum level of bots that will enchant their equipment (if greater than RandomBotMaxlevel, bots will not enchant equipment)
 # Default: 60
 AiPlayerbot.MinEnchantingBotLevel = 60
 
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -126,6 +126,9 @@ void PlayerbotFactory::Init()
         if (id == 15463 || id == 15490) // Legendary Arcane Amalgamation
             continue;
 
+        if (id == 29467 || id == 29475 || id == 29480 || id == 29483) // Naxx40 Sapphiron Shoulder Enchants
+            continue;
+
         SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(id);
         if (!spellInfo)
             continue;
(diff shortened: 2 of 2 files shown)


brighton-chi:
Changed to draft because I realized I haven't checked spells from items (e.g., ZG enchants).

brighton-chi:
Okay, I confirmed that the new thresholds are correct, except for a very small number of item enchants that are not at all useful and would never be applied to bots, so I think it is unnecessary clutter in the code to exclude those from the threshold calculations.

The one exception is Naxx40 shoulder enchants. They are (before and after this PR) outside the vanilla threshold and therefore before this PR are not available at level 60 but are at level 61 and onward. I have made the decision to exclude them entirely from bot enchants because they are technically not available in 3.3.5. The alternative is to add specific exceptions to make them available at level 60--it is easy to do, but I'm not sure that is the right thing since Naxx40 is not in the Core.

Note that enchanting logic is still not perfect. Bots will apply the best available ZG enchant for their spec, but this often means they are applying another class's ZG enchant (which should not be available to them). But that is outside the scope of this PR.

PR #1615, opened 2025-09-03
Add weighted bot to banker teleport logic and config

zeb139:
The other day I was trying to fiddle with the AiPlayerbot.ProbTeleToBankers setting to make cities feel more populated/immersive. I found that by increasing this value and the total # of bots I could get pretty decent results, but it bothered me that the distribution between cities was equal ('cause who actually hangs out in Exodar?).

I added a configurable list of weights for each city so now admins can sway which cities receive more bots (or disable them from the table). This should allow (I hope) a similar level of immersion while reducing the resource requirements (# of bots).

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1014,6 +1014,20 @@ AiPlayerbot.RandomBotMaps = 0,1,530,571
 # Default: 0.25
 AiPlayerbot.ProbTeleToBankers = 0.25
 
+# Control probability weights for bots teleporting to Capital city bankers
+# Sum of weights need not be 100. Set to 0 to disable teleporting to the city.
+AiPlayerbot.EnableWeightTeleToCityBankers = 0
+AiPlayerbot.TeleToStormwindWeight = 2
+AiPlayerbot.TeleToIronforgeWeight = 1
+AiPlayerbot.TeleToDarnassusWeight = 1
+AiPlayerbot.TeleToExodarWeight = 1
+AiPlayerbot.TeleToOrgrimmarWeight = 2
+AiPlayerbot.TeleToUndercityWeight = 1
+AiPlayerbot.TeleToThunderBluffWeight = 1
+AiPlayerbot.TeleToSilvermoonCityWeight = 1
+AiPlayerbot.TeleToShattrathCityWeight = 1
+AiPlayerbot.TeleToDalaranWeight = 1
+
 # How far randombots are teleported after death
 AiPlayerbot.RandomBotTeleportDistance = 100
 
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1016,7 +1016,7 @@ AiPlayerbot.ProbTeleToBankers = 0.25
 
 # Control probability weights for bots teleporting to Capital city bankers
 # Sum of weights need not be 100. Set to 0 to disable teleporting to the city.
-AiPlayerbot.EnableWeightTeleToCityBankers = 0
+AiPlayerbot.EnableWeightTeleToCityBankers = 1
 AiPlayerbot.TeleToStormwindWeight = 2
 AiPlayerbot.TeleToIronforgeWeight = 1
 AiPlayerbot.TeleToDarnassusWeight = 1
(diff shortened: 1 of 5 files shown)


src/RandomPlayerbotMgr.cpp
```diff
@@ -1660,11 +1660,6 @@ void RandomPlayerbotMgr::RandomTeleport(Player* bot, std::vector<WorldLocation>&
         return;
     }
 
-    if (tlocs.empty())
-    {
-        LOG_DEBUG("playerbots", "Cannot teleport bot {} - no locations available", bot->GetName().c_str());
-        return;
-    }
```
zeb139, line 1667: This block was a duplicate of the lines just above it

conf/playerbots.conf.dist
```diff
@@ -1014,6 +1014,20 @@ AiPlayerbot.RandomBotMaps = 0,1,530,571
 # Default: 0.25
 AiPlayerbot.ProbTeleToBankers = 0.25
 
+# Control probability weights for bots teleporting to Capital city bankers
+# Sum of weights need not be 100. Set to 0 to disable teleporting to the city.
```
icemansparks, line 1018: The default weights should follow the guideline: "The sum needs to be 100"
The weights below currently do not meet this requirement and could lead to confusion if it is activated and not adjusted before use.

zeb139: The weights intentionally do not add to 100 ("Sum of weights need **not** be 100"). 
Here's another example of config totals not adding to 100: https://github.com/liyunfan1223/mod-playerbots/issues/1535.  

The logic behind the weighting is basically the same as pulling colored balls from a bag. The more balls of a given color, the more likely you are to withdraw it. Maybe I can express this better in the config guideline?

Or do you mean there are project guidelines outlined somewhere in the repo? I haven't been able to find anything, but would be happy to take a look.

icemansparks: It's late - looks like I just misread the commented line about "not" needing 100 (probably because it's what is used in other places) :-) 
Thanks for the details on the logic 👍

src/RandomPlayerbotMgr.cpp
```diff
@@ -2138,11 +2136,137 @@ void RandomPlayerbotMgr::RandomTeleportForLevel(Player* bot)
...
+            {IRONFORGE,       {2460, 2461, 5099}},
+            {DARNASSUS,       {4155, 4208, 4209}},
+            {EXODAR,          {17773, 18350, 16710}},
+            {ORGRIMMAR,       {3320, 3309, 3318}},
+            {UNDERCITY,       {4549, 2459, 2458, 4550}},
+            {THUNDER_BLUFF,   {2996, 8356, 8357}},
+            {SILVERMOON_CITY, {17631, 17632, 17633, 16615, 16616, 16617}},
+            {SHATTRATH_CITY,  {19246, 19338, 19034, 19318}}, 
+            {DALARAN,         {30604, 30605, 30607, 28675, 28676, 28677, 29530}}
+        };
```
kadeshar: I think objects like const/enums should be on file beginning not hidden inside method body. Consider refactoring that.

And im curious. Why you choose bankers? Not better flight masters?

zeb139: Good call! I've moved the lookup tables and enums to the top of the file, and initialized bankerEntryToLocation during RandomPlayerbotMgr::PrepareTeleportCache() so that it's only done once. (I assume PrepareTeleportCache() is only run 1 time during startup anyway). 
Let me know if I missed anything.

I didn't really choose bankers myself since some of the logic was already implemented to teleport bots to bankers (see AiPlayerbot.ProbTeleToBankers). But I think there a couple of advantages Bankers have over Flight Masters in this regard:
1. Bankers in capital cities are all found near the "center" where most real players would gather, so the bots are more likely to be seen. Flight Masters are usually a little further out of the way.  
2. After spawning at Bankers (near the city center), existing bot behavior already has them seek random nearby NPCs to interact with, so they tend to spread relatively evenly throughout the city
3. In the new RPG Strategy behavior, bots will already travel to Flight Masters occasionally

hermensbas (approved):

PR #1623, opened 2025-09-10
Implement Karazhan strategy

brighton-chi:
Thank you to @noisiver for starting this project and implementing the strategies for Moroes, Maiden of Virtue, the Opera Events, Curator, and Terestian Illhoof and parts of the strategies for Shade of Aran. I have spent far more time on this than was warranted, but it was worth the frustration for my favorite instance.

This PR includes strategies for 11/14 Karazhan bosses, as a result of which Karazhan is fully clearable by a single player with level-appropriate gear. I have done so with a party in a mix of Outland blues and T3 gear, with damage nerfed to 55%, healing nerfed to 60%, and health nerfed to 95%. I also reverted the patch 3.0.2 nerfs that reduced all TBC raid mobs' HP by 30%. So suffice to say that if you play with no modifications, Karazhan should be a breeze.

Below is a description of what this PR does (and doesn't do), boss-by-boss, along with some guidance regarding how to do the fights using bots (with and without strategies). I suggest bringing 2 tanks and 2 or 3 healers (I ran with 2, but if you play with the nerfs/buffs that I did, you may experience very high pressure on healer mana in the later fights).

**Servant's Quarters**
I have nothing to say. You weren't really expecting a strategy here, were you?

**Attumen the Huntsman**
This is a very simple tank-and-spank fight, but this PR includes a strategy for all bots (excluding the tank with aggro) to stack up 5 yards behind the boss in Phase 2 after he mounts Midnight. This will eliminate the threat of his most dangerous attack, Berserker Charge, which has a minimum range of 8 yards.

**Moroes**
You probably can just stack up Moroes and all adds and cleave them all down, but this PR includes a strategy that automatically marks adds by the skull icon in recommended kill order (shadow priest with mana burn first, holy priest second, etc.)

**Maiden of Virtue**
Ranged bots fan out in a fixed layout around Maiden to mitigate the damage of Holy Wrath. When Maiden casts Repentence, stunning the raid, the tank will bring Maiden over to a healer so that Holy Ground breaks the healer's stun.

**The Big Bad Wolf**
If you even get to his first Little Red Riding Hood cast, the bot with the debuff will run around the stage.

**Romulo and Julianne**
In Phase 3 when both bosses are active, because they must die within 10 seconds of each other, the strategy switches the skull icon back-and-forth between the bosses to try to keep them within 10% HP of each other.

**Wizard of Oz**
The strategy marks the bosses with a skull icon in their recommended kill order: Dorothee, Tito (though you will also certainly kill Dorothee before he spawns), Roar, Strawman, Tinhead, and the Crone. Additionally, Mages will spam Scorch on Strawman to daze him. I originally implemented a strategy for Warlocks to fear Roar, but I removed it because honestly it works better just to put a moon icon on Roar if you want to do that.

**The Curator**
The tank will drag the boss to a fixed spot down the hallway. Ranged bots will spread out to avoid chain damage from Arcing Sear, and bots will mark Astral Flares with the skull icon to prioritize them down. I recommend saving cooldowns for Evocate.

**Terestian Illhoof**
The bots will mark targets with the skull icon in the following order: Demonic Chains, Kil'rek (since his death places a damage debuff on Illhoof), and Illhoof himself.

**Shade of Aran**
 This one took a lot of time. It is probably a doable fight without a coded strategy, but it would be challenging. Here's what the strategy does:

- Flame Wreath: Bots will stop moving and will yell the Flame Wreath chant when Aran starts casting Flame Wreath. They will remain in place until the aura fades. Aran is buggy, and sometimes Flame Wreath will persist long beyond the 20-second duration and will not go away until his next special spellcast, or it will cast between special spellcasts without warning. The bots should hold position regardless.

- Arcane Explosion: Bots will run out immediately and will disable movement when they are out of the explosion's range. Immediately after the explosion, they will run/charge back into the fight.

- Conjured Elementals: When they spawn, they will be marked one-by-one by the skull icon, except that the marking will skip any elemental that is banished (so feel free to have your warlock(s) banish if you want). If Flame Wreath is up, the bots will not chase the Conjured Elementals, even the one with the skull icon.

- Positioning: Ranged bots spread out 5 yards apart, no farther than 12 yards from the boss.

**Netherspite**
Most of my work was on this boss. It's easily the most mechanically complicated fight in Karazhan and is absolutely not soloable without a strategy. If you play without nerfs, you might be able to down Netherspite with 3 players, but that's the best-case scenario. Note that AC's version of Netherspite has an annoying bug, which is that the beams end for purposes of game mechanics at his hitbox, even though they visually continue to his center (and they should be blockable within his hitbox). This means that blockers need to stay at the edge of his hitbox (18 yards) or the beam will be lost.

Here is what I did for the strategy, which I think works pretty well (I've watched 9 bots kill him while I walked around just creeping on what they were doing):

- Red (Tank) Beam: One tank will be assigned to block the beam for each Portal Phase. The assigned tank will announce their status and proceed to dance in and out of the beam (5 seconds in, 5 seconds out). This for me was a good compromise between maintaining the tank's survivability without buffing Netherspite too much. Tanks will stand in Void Zones--their positioning on the beam is too important, and this version of Netherspite too finicky, to afford avoiding them. You will have to heal through it (not too hard).

- Blue (DPS) Beam: DPS other than Rogues and Warriors are eligible to be assigned to block this beam. Each assigned blocker will announce their status and then will move to and stand in the beam. If a Void Zone drops under the bot while the bot is still the assigned blocker, the bot will move along the beam to get out of the Void Zone so that they do not drop the beam. When a blocker reaches 25 stacks of the debuff, they will announce that they are leaving the beam (in practice, it will take several more stacks before they actually move). After stepping out of the beam, their avoidance behavior will revert to that of a normal non-blocking bot. They will not be called upon to block again before Nether Exhaustion expires. The next eligible bot will simultaneously become the assigned blocker and will take the previous bot's spot to block the beam. If nobody dies, you need only two blue beam blockers per Portal Phase.

- Green (Healer) Beam: This works the same way as the Blue Beam, except that eligible blockers are healers, Rogues, and DPS Warriors. There is nothing wrong with having the second tank take this beam, but I chose not to make them eligible because I think it is a waste of the beam's buff to have your offtank sit in it (Rogues and DPS Warriors can make use of it because their energy/rage costs will be reduced). Healers that are assigned to block will swap in the same way as Blue Beam blockers. If a Rogue or DPS Warrior is the assigned blocker, however, they will stand in the beam for the entire Portal Phase since they do not suffer any adverse effects from the beam.

- Non-Blockers: They will stay at least 5 yards away from each beam until called to be an assigned blocker. They will also avoid Void Zones.

- Banish Phase: The only strategy I implemented was for bots to avoid residual Void Zones from the Portal Phase. It's up to you whether you want to run away or burn Netherspite. I recommend the latter, and actually he can be cheesed during the Banish Phase because he fixates on a single player instead of turning and randomly blasting the raid like he is supposed to.

**Chess Event**
I have no idea how to implement a strategy for this one, technically or theoretically. Luckily, you don't need one as the event is fairly easy to solo. I'm going to use chess notation here (from the white side of the board). For your first three moves, move pawns to b3, d3, and g3. For your next two moves, move your rooks diagonally (yeah, wtf) to b2 and g2. If at any time your king or queen is on fire, move them out. Take over your queen and proceed to spam the 4k nuke that can hit the whole board (do not bother with the AoE--both factions' versions are bugged--the Horde version does no damage, and the Alliance version just casts right around the queen instead of having a targeting circle like it's supposed to). Online guides will tell you to just nuke the opposing king, but I don't think that's always the right answer--because Medivh will cheat and fully heal his pieces (including the king), I found it more effective to first set up more sustained DPS on the opposing king. So I have a lot more success killing pieces in front of the opposing king to basically lure him into a trap where I can flank him with my king and a rook, with me taking over the queen and nuking at that point. There is still some luck involved (largely depending on if the opposing side's king stays back or not and also what cheats Medivh uses and when), but I win probably 1/3 of the time.

**Prince Malchezaar**
This fight is definitely doable without a strategy, but I think it would be very difficult if you have stats adjusted to make Karazhan similar difficulty to what it was in TBC. The strategy does the following:

- Bots other than the main tank will maintain a distance of 22 yards from all Infernals. The Hellfire radius is supposed to be 20 yards, but for whatever reason if I set the distance to 20 or even 21 yards, the bots will get killed. There is a little difficulty here in that if bots get stuck directly behind an Infernal, they have trouble getting around it. I implemented arc movement, and it does work, but it's not at all smooth and takes the bot a while to circle around the Infernal.
- Bots that get Enfeebled will run to 35 yards from the boss, taking the shortest possible route that does not cross within any Infernal's Hellfire radius. This is to survive Shadow Nova, which is supposed to be a spell with a radius of 30 yards, but if I set the radius to 30 or even up to 32 yards, the bots will get killed. I think they are not running far enough as opposed to the AoE radius being wrong, but I could not figure out the issue. All I know is setting it to 35 works fine (I didn't bother testing between 32 and 35 and decided just to give some more distance since there is plenty of time to get there). The bots will spasm in place when they reach the maximum distance, and I cannot figure out what is causing it (I've tried to solve it in basically the same way I did for Aran and Netherspite, and it's not successful). I did however disable them using Charge and similar abilities so they should survive OK.
- The active tank will maintain a distance of at least 30 yards from all Infernals. This is to give DPS a little more room to manuever. There is no Enfeeble logic for them because the tank cannot be Enfeebled.

I added LoS checks to everything so the bots will not run out of the map to escape hazards (they were doing so without the LoS checks). This can cause them to get slowed down by walls so I suggest tanking Malchezaar away from the walls (which is contrary to the retail strategy, I know).

**Nightbane**
This fight sucks with bots. And I don't mean because it's overly hard (although it is the hardest fight in Karazhan, in both TBC and AC). The bots just have a serious case of the call of the void on the balcony and absolutely lose their shit when Nightbane is flying around. They fall through the ground, fall off the castle wall, go running off into the terrain, cause Nightbane and skeletons to go off the map, it's awful. Any strat I tried only made the behavior worse. There are no strats for Nightbane as a result. You'll just need to power through, and it's a gear check for sure.

You should clear out all the lower-floor ghosts or they are very likely to be pulled through the ground during the fight. Turn off avoid AoE for your tank (or they'll spin Nightbane around trying to get out of Charred Earth and kill everybody). Your healers will have to heal through it, and you probably want 3 healers for this fight. Try to get melee positioned at the sides. When Nightbane goes into his Flight Phase, put the moon icon on him so the bots don't try to chase him off the map, follow him with the bots so he doesn't use Fireball Barrage, and then have the bots stay in place near where Nightbane stops. You will take more damage from Rain of Bones as a result, but trust me, it's much better than bots running off the map and causing skeletons to spawn under you or pulling mobs from elsewhere in the instance. Make sure to get back in position and take the moon icon off before Nightbane lands.


Dreathean:
I definitely second the idea of having detailed notes on a boss by boss basis for each raid like what you have put here. Because it is certainly easy for details on past strategies and their individual quirks to get lost to the nether otherwise - I'll get another section added onto the Wiki for this purpose.

Thank you and Revision for all your work on Karazhan!

brighton-chi:
I significantly redid Aran, Netherspite, and Malchezaar. The first two are working much, much better now. Malchezaar still has some jank that I cannot figure out how to solve, but the strategy is pretty effective, and I think I've made as much progress on it as I can.

noisiver:
I forgot to say this earlier but thank you for picking this up and for redoing Aran. That one was a mess so it really needed a rewrite.

I forgot to mention that Big Bad Wolf *can* be hit and miss. The logic is that they follow the 4 waypoints on the stage to circle it and stay away from the wolf, but the issue is if the wolf is positioned badly he will catch up with them at the first waypoint. Ideally they should calculate the best waypoint to go to first based on where he is. What I mean is if the wolf is positioned at or near waypoint 2 (bottom-right) or 3 (top-right) and the bot runs to 1 (bottom-left) he will almost certainly intercept the bot. It works but could be improved, I'm sorry to say I just never found the time to do so.

brighton-chi:
> I forgot to say this earlier but thank you for picking this up and for redoing Aran. That one was a mess so it really needed a rewrite.
> 
> 
> 
> I forgot to mention that Big Bad Wolf *can* be hit and miss. The logic is that they follow the 4 waypoints on the stage to circle it and stay away from the wolf, but the issue is if the wolf is positioned badly he will catch up with them at the first waypoint. Ideally they should calculate the best waypoint to go to first based on where he is. What I mean is if the wolf is positioned at or near waypoint 2 (bottom-left) or 3 (top-right) and the bot runs to 1 (bottom-left) he will almost certainly intercept the bot. It works but could be improved, I'm sorry to say I just never found the time to do so.

No problem. I was happy to have the opportunity to work on my favorite instance.

I'll see if I can figure out what's going on with BBW. That's actually the only fight I did not test because I only got him once and killed him right as he used the fixate, and I didn't bother to specifically select and test the encounter.

Edit: I can't really get BBW to work well. I've added an action for the tank to drag BBW to a corner, which solves the issue of bots cutting across BBW when running. They still almost always get caught at some point, however. I've tried a few different things and cannot get the strategy to work better. I don't think it's worthwhile to spend a bunch more time on this since that fight is very easy even without a strategy so I'm going to open the PR for review.

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
```diff
@@ -0,0 +1,1179 @@
...
+    int maxIndex = 7;
+    int index = 0;
+
+    const GuidVector members = AI_VALUE(GuidVector, "group members");
+
+    for (const auto& memberGuid : members)
+    {
+        Unit* member = botAI->GetUnit(memberGuid);
+
+    if (!member || !botAI->IsRanged(member->ToPlayer()))
```
kadeshar: Tab

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
```diff
@@ -0,0 +1,1179 @@
...
+        index++;
+    }
+
+    float distance = bot->GetExactDist2d(KARAZHAN_MAIDEN_OF_VIRTUE_RANGED_POSITION[index]);
+    const float maxDistance = 2.0f;
+
+    if (distance > maxDistance)
+        return MoveTo(bot->GetMapId(), KARAZHAN_MAIDEN_OF_VIRTUE_RANGED_POSITION[index].GetPositionX(),
+                      KARAZHAN_MAIDEN_OF_VIRTUE_RANGED_POSITION[index].GetPositionY(), bot->GetPositionZ(), false,
+                      false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 163: Multiline statement in if without {}

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
```diff
@@ -0,0 +1,1179 @@
...
+    std::vector<Player*> blueBlockers = karazhanHelper.GetBlueBlockers();
+    static std::map<ObjectGuid, bool> wasBlockingBlueBeam;
+    ObjectGuid botGuid = bot->GetGUID();
+    Player* assignedBlueBlocker = blueBlockers.empty() ? nullptr : blueBlockers.front();
+    bool isBlockingNow = (bot == assignedBlueBlocker);
+    bool wasBlocking = wasBlockingBlueBeam[botGuid];
+
+    if (wasBlocking && !isBlockingNow)
+    {
+        bot->Yell("I'm leaving the blue beam--next blocker up!", LANG_UNIVERSAL);
```
kadeshar: No transaltion and possibility to turn off

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
```diff
@@ -0,0 +1,1179 @@
...
+        wasBlockingBlueBeam[botGuid] = false;
+    }
+    else if (isBlockingNow)
+    {
+        wasBlockingBlueBeam[botGuid] = true;
+    }
+    if (isBlockingNow)
+    {
+        if (!wasBlocking)
+            bot->Yell("I'm moving to block the blue beam!", LANG_UNIVERSAL);
```
kadeshar: No transaltion and possibility to turn off

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
```diff
@@ -0,0 +1,1179 @@
...
+
+    if (wasBlocking && !isBlockingNow)
+    {
+        bot->Yell("I'm leaving the green beam--next blocker up!", LANG_UNIVERSAL);
+        wasBlockingGreenBeam[botGuid] = false;
+    }
+    else if (isBlockingNow)
+    {
+        if (!wasBlocking)
+            bot->Yell("I'm moving to block the green beam!", LANG_UNIVERSAL);
```
kadeshar: No transaltion and possibility to turn off

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
```diff
@@ -0,0 +1,1179 @@
...
+                bestCandidate = Position(cx, cy, cz);
+                bestDist = moveDist;
+                found = true;
+            }
+        }
+    }
+    if (found && karazhanHelper.IsSafePosition(bestCandidate.GetPositionX(), 
+        bestCandidate.GetPositionY(), bestCandidate.GetPositionZ(), 
+        voidZones, 4.0f))
+
```
kadeshar, line 935: if without {} with strange empty line

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
```diff
@@ -0,0 +1,362 @@
...
+        if (skullGuid != target->GetGUID())
+            group->SetTargetIcon(skullIconId, bot->GetGUID(), target->GetGUID());
+    }
+}
+
+Unit* RaidKarazhanHelpers::GetFirstAliveUnit(const std::vector<Unit*>& units)
+{
+    for (Unit* unit : units)
+        if (unit && unit->IsAlive())
+            return unit;
```
kadeshar, line 60: Multiline statement for for loop without {}

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
```diff
@@ -0,0 +1,362 @@
...
+bool RaidKarazhanHelpers::IsFlameWreathActive()
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "shade of aran");
+    if (!boss)
+        return false;
+
+    Spell* currentSpell = boss->GetCurrentSpell(CURRENT_GENERIC_SPELL);
+    if (currentSpell && currentSpell->m_spellInfo && currentSpell->m_spellInfo->Id == SPELL_FLAME_WREATH)
+    {
+        bot->Yell("I will not move when Flame Wreath is cast or the raid blows up.", LANG_UNIVERSAL);
```
kadeshar: No transaltion and possibility to turn off

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    if (Group* group = bot->GetGroup())
+    {
+        for (GroupReference* itr = group->GetFirstMember(); itr != nullptr; itr = itr->next())
+        {
+            Player* member = itr->GetSource();
+            if (!member || !member->IsAlive())
+                continue;
+            PlayerbotAI* memberAI = sPlayerbotsMgr->GetPlayerbotAI(member);
+            if (!memberAI || !memberAI->IsTank(member))
```
kadeshar: getting memberAI only to check isTank is not required

botAI->IsTank(member)

src/strategy/raids/karazhan/RaidKarazhanMultipliers.cpp
```diff
@@ -0,0 +1,199 @@
...
+    return 1.0f;
+}
+
+float KarazhanPrinceMalchezaarMultiplier::GetValue(Action* action)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "prince malchezaar");
+    if (!boss)
+        return 1.0f;
+
+    if (boss && bot->HasAura(SPELL_ENFEEBLE))
```
kadeshar: if (boss condition was checked in previous if

src/strategy/raids/karazhan/RaidKarazhanMultipliers.cpp
```diff
@@ -0,0 +1,199 @@
...
+    {
+        if (IsChargeAction(action))
```
kadeshar: That can be merged in previous if

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
brighton-chi: What languages do we need to translate into? If it's complicated, I can take this phrase out (and the other beam yells), but it's pretty valuable information IMO to know which bots are going to what beam (if at all) because as the player you may need to adjust and fill in temporarily or for the phase because there is little margin for error.

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
brighton-chi: This one is a meme from the English version of the game from back in the day. I'll remove it.

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
Dreathean: I would agree that this is very important in-fight information, not like debug information that people might want toggled off. 

I don’t know how this is setup as far as accessibility to supporting other languages, but I feel adding other languages could be added in a later PR?

kadeshar: I mean only INSERT record in table with texts and use it in code. You dont have too fill other languages.

brighton-chi: What table is used for this purpose? Just the ai_playerbot_texts table? I didn't see yells being used for any other bosses like this so I'm not sure what you have in mind (and I'm not sure if I enter the text into that table if there's some other code that is going to start calling the yell in other contexts).

brighton-chi:
I addressed all comments other than the open points re: translation.

I made some general tweaks based on what I've learned in subsequent PRs (sorry!), which include some changes to the multiplier that improve melee bot behavior during Malchezaar.

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
kadeshar: All text used by bots should be translatable and there is issue on that

brighton-chi: That's fine, I'm just not sure what is needed to be done. What table needs to be referenced for the text?

kadeshar: Here example https://github.com/liyunfan1223/mod-playerbots/pull/1644/files

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
brighton-chi: Actually, I think GetPlayerbotAI(member) is needed here--won't botAI->IsTank(member) return real players as well? This needs to be limited to bots (it's for determining who the current beam blocker is so other bots will engage in non-blocker avoidance logic--if I don't limit this to bots, the helper could return that a real player is the beam blocker and cause the bot blocking the beam to engage in non-blocker avoidance).

But let me know if I'm not understanding this correctly.

noisiver: Strategies don't apply to non-bots. In order for the strategies to be active the bot AI has to be active and have the karazhan strategy. If you use `.playerbot bot self` you can have it but then you're also a bot so it controls your character. You would have to use something other than botAI to find regular players.

brighton-chi: I mean that GetFirstMember doesn't inherently identify only bots. This is just a helper to identify a particular bot so that other bots know how to react. I think it can return a player if not specifically excluded.

src/strategy/raids/karazhan/RaidKarazhanActions.cpp
brighton-chi: Should be resolved now--let me know if what I did in the latest commit works (seems fine based on my testing).

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
kadeshar: For check that member is bot you can use flag member->GetSession()->IsBot()

src/strategy/raids/karazhan/RaidKarazhanMultipliers.cpp
```diff
@@ -19,13 +19,17 @@ float KarazhanBigBadWolfMultiplier::GetValue(Action* action)
...
+    }
 
     if (bot->HasAura(SPELL_LITTLE_RED_RIDING_HOOD))
     {
         if ((dynamic_cast<MovementAction*>(action) && !dynamic_cast<KarazhanBigBadWolfRunAwayAction*>(action)) || 
            (dynamic_cast<AttackAction*>(action)))
-            return 0.0f;
+           {
+               return 0.0f;
+           }
```
kadeshar: Remove one tab

kadeshar:
I made PR (https://github.com/liyunfan1223/mod-playerbots/pull/1678) which add method allowing getting translation or default text to avoid multiple code repetition

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
brighton-chi: I think it should be !GET_PLAYERBOT_AI(member) to exclude players. That's the macro for sPlayerbotsMgr->GetPlayerbotAI(member), and actually it's what you used in Ulduar strats too.

brighton-chi:
Ugh, back to draft for a bit due to some weird stuff happening on Netherspite.

Edit: Ok, made some changes to improve some issues at the margins.

For Netherspite, beam positioning should be more precise, and the bots are more responsive now ing eneral. The order of beam blockers will no longer override the current blocker (previously, if a higher priority blocker became eligible to block again, they would immediately take over from a lower-priority blocker who was in the beam at the time--that's not desirable). Also, Rogues will no longer use Killing Spree when blocking beams.

I finally managed to fix the janky movement with the Enfleebled bot avoidance vs. Malchezaar. Additionally, I replaced all LOS checks with Collision checks, which just work a lot better for bot movement. I separately stripped out all the arc movement logic because it wasn't working properly and is more complicated than necessary to try to work out.

Sorry to keep revising this. I'm done on modifying further before merge, other than to address specific comments.

hermensbas (approved):

hermensbas:
Only reviewed the code not the functional gameplay

brighton-chi:
> Only reviewed the code not the functional gameplay

Thank you. As far as I know, nobody has tried these in game other than me.

PR #1628, opened 2025-09-13
Refactor GossipHelloAction with overload

Tierisch:
This just allows `GossipHelloAction::Execute` to also be used silently with a specified guid and menu option from code instead of needing packets or reporting anything to a master. It is in addition to the existing method, so the current one still works as usual.

I'm using it to have bots interact with a nearby NPC at specific waypoints based on table values in the dungeon pathing I'm working on, but it could be used in other cases as well, so I thought it worth submitting on its own.

### DIFF
--- a/src/strategy/actions/GossipHelloAction.cpp
+++ b/src/strategy/actions/GossipHelloAction.cpp
@@ -26,55 +26,12 @@ bool GossipHelloAction::Execute(Event event)
         p >> guid;
     }
 
-    if (!guid)
-        return false;
-
-    Creature* pCreature = bot->GetNPCIfCanInteractWith(guid, UNIT_NPC_FLAG_NONE);
-    if (!pCreature)
-    {
-        LOG_DEBUG("playerbots",
-                  "[PlayerbotMgr]: HandleMasterIncomingPacket - Received  CMSG_GOSSIP_HELLO {} not found or you can't "
-                  "interact with him.",
-                  guid.ToString().c_str());
-        return false;
-    }
-
-    GossipMenuItemsMapBounds pMenuItemBounds =
-        sObjectMgr->GetGossipMenuItemsMapBounds(pCreature->GetCreatureTemplate()->GossipMenuId);
-    if (pMenuItemBounds.first == pMenuItemBounds.second)
-        return false;
-
     std::string const text = event.getParam();
     int32 menuToSelect = -1;
-    if (text.empty())
-    {
-        WorldPacket p1;
-        p1 << guid;
-        bot->GetSession()->HandleGossipHelloOpcode(p1);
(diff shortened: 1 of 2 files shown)


src/strategy/actions/GossipHelloAction.cpp
```diff
@@ -141,3 +141,36 @@ bool GossipHelloAction::ProcessGossip(int32 menuToSelect)
...
+    {
+        if (!bot->PlayerTalkClass)
+            return false;
+        if (!ProcessGossip(menuToSelect))
+            return false;
+    }
+
+    bot->TalkedToCreature(pCreature->GetEntry(), pCreature->GetGUID());
+    return true;
+}
```
kadeshar, line 152: - this method overload dont work like old one which is confusing (other return false/true places)
- if should work same then dont duplicate code but use like
obraz

only add parameter silent to avoid tells and dont skip log_debug

- if shouldnt work same then is wrong place for this method

Tierisch: I'm not sure I understand all of what you mean. I do understand having left out the LOG_DEBUG that I should fix and that the existing method could be improved with a silent parameter, but not the rest.

Unless I missed something, the overload *does* return true/false for the same conditions as the event method. Being silent in general is because a bot using gossip with this method isn't being directed by a master and may not have one at all.

Unless I'm misunderstanding how events work, your suggestion would still require making one with a fake packet to be parsed instead of giving the variables directly whenever this is called. Avoiding the need for that is the point of the overload, since that seems like a roundabout way to reach the same end.

Tierisch: Oh, wait. Do you mean refactoring the event method so it just handles the packets part and passes the rest to the overload, then have that be called either from within the event method or when the variables are given directly?

If so, that does make sense and I just misunderstood your screenshot at first.

kadeshar: > Oh, wait. Do you mean refactoring the event method so it just handles the packets part and passes the rest to the overload, then have that be called either from within the event method or when the variables are given directly?
> 
> If so, that does make sense and I just misunderstood your screenshot at first.

Yes, because:
- if your new method is not used by action then shouldnt be it part but some helper or something
- to avoid same code duplication which in future only makes changes hardest

Tierisch:
Alright, how's this?

It wasn't until doing this that I realized my previous overload wouldn't have been entirely silent anyway since there's another `Tell` through `ProcessGossip` (I hadn't tested the previous one when actually using a menu option, so I missed that it would've still reported some parts). The new `silent` bool now gets passed to that as well.

I've tested this both with true and false for silent (and when choosing menu options), and it acts the same as it did previously unless that's called with true (which it won't be when using the event method).

PR #1635, opened 2025-09-14
Update playerbots.conf.dist

Raz0r1337:
Added Half-Burried Bottle to the list of disallowed GO's

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1997,11 +1997,11 @@ AiPlayerbot.AllowedLogFiles = ""
 ####################################################################################################
 # A list of gameObject GUID's that are not allowed for bots to interact with.
 #
-AiPlayerbot.DisallowedGameObjects = 176213,17155,2656,74448,19020,3719,3658,3705,3706,105579,75293,2857,179490,141596,160836,160845,179516,176224,181085,176112,128308,128403,165739,165738,175245,175970,176325,176327,123329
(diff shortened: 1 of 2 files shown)


brighton-chi:
I would say the source also needs to be updated, but I think default parameters are going to be irrelevant once the pending PR to the core is merged. Maybe it's worth continuing to update them anyway for completeness and in case the approach to fallbacks changes in the future?

Raz0r1337:
While questing, I noticed that my bots were looting the item that started the quest before I had a chance. So I entered the ID in my system, and it's working. I'll find out where I need to change it in the source code and add it.

What PR? Is another system currently being developed that prevents this kind of thing, or why do you think the list shouldn't be further developed?

Tierisch:
> What PR? Is another system currently being developed that prevents this kind of thing, or why do you think the list shouldn't be further developed?

There's an upstream core PR that will make `.conf.dist` files not be read by the server for default values anymore if/when it gets merged. They'll still be there for a user to see what the defaults would be, but only what's in a `.conf` will matter. I'm assuming that's the one they meant.

brighton-chi:
> > What PR? Is another system currently being developed that prevents this kind of thing, or why do you think the list shouldn't be further developed?
> 
> 
> 
> There's an upstream core PR that will make `.conf.dist` files not be read by the server for default values anymore if/when it gets merged. They'll still be there for a user to see what the defaults would be, but only what's in a `.conf` will matter. I'm assuming that's the one they meant.

I think it will make the default parameters in the source also no longer be read. So it was a general question as to whether we care what the default parameters in the source are anymore. Since in the past, whenever we changed a default value in the .dist, we should have done it in the source also.

Raz0r1337:
Ok, now I understand exactly what you meant. Thanks for the clarification.

I think as long as the default values are still used or present in the source, they should be maintained. Who knows, maybe this will be removed in the future and all default values in the source code will be deleted.

PR #1639, opened 2025-09-16
Fix typo in RandomPlayerbotMgr.cpp

brighton-chi:
These entries print at the beginning of every debug log. I don't know if we need this info to be logged at all by default, but I'm just fixing the text.


brighton-chi:
Also modified the comment to the health cheat-it's actually a very relevant distinction because the cheat will not protect bots from getting one-shot (it merely restores them to full every tick).

PR #1643, opened 2025-09-19
Replaced tabs with spaces

noisiver:
The standard in AzerothCore and Playerbots is 4 spaces, not tabs.


kadeshar:
Never heard about such standard either here or in cmangos playerbots.
Also i though default from most IDE is tab.

EDIT:
Nvm, you have right

hermensbas (approved):

Raz0r1337:
It's really true, and I've been told the same thing myself. Unfortunately, i think it's not written down anywhere that I know of. It's more of an unspoken rule. ;-)

noisiver:
I also went through and removed unnecessary spaces in all files. Like blank lines with spaces for no reason or spaces at the end of a line and so on.

PR #1646, opened 2025-09-19
Big update for the forbidden GO ID's

Raz0r1337:
I'm doing this because the list that is currently implemented is, at best, incomplete. There are often multiple IDs for the same object name, and all of them need to be included to effectively block access to all those objects. So I extracted all lootable game objects (type 3) from the world database and used ChatGPT to create a list from them.
The IDs are grouped in the commented section for easier readability.

If possible, please double-check that the IDs are correct.
I'm already testing it on my own server and it seems to work perfectly. I've been playing for three hours now, and the bots haven't stolen anything from me at all. But please check it yourself as well.

If anything needs to be removed, please let me know. I'll take a look at it.

I would also like to suggest that this list be integrated into the Playerbots database; that way, the configuration file wouldn't become so bloated.


kadeshar:
This settings is too block interaction/looting important objects which can block player like important items in dungeon or something.
I dont see reason why we should block looting from all chest for all bots.
Second such block should be done in code, for example by checking that gameobject have type GAMEOBJECT_TYPE_CHEST or something

Raz0r1337:
Okay, I understand that. However, I still find it much easier to remove an ID from a complete list than to spend a long time searching, testing, and then inserting it especially when you're playing normally and find out that your bot has stolen an object you are searching for. Very often you don't even notice it and spend a long time looking for the object. This way, you can simply search for the object name in the commented-out area and immediately have the corresponding ID, which you can quickly delete. After a while, you'll be left with a complete and functioning list of all the GOs that would otherwise be stolen. I fear otherwise, this list will always have serious holes and will never be truly complete.

hermensbas:
i dont agree with this list, as mentioned this list only there for exclusions that otherwise would break the gameplay, questline, trigger unwanted behaviour in dungeons etc. Exceptions sort of say.

Its blacklist not a whitelist 😁

hermensbas:
Sorry buddy

Raz0r1337:
Hehe, I work in IT myself and in my experience and that of all my colleagues, whitelists are never properly maintained. As the current situation clearly shows. But whatever. If the collective rejects this PR, then that's just how it is. But then I would ask that all IDs be removed from the current list, with the exception of the dungeon GOs. That's not a whole story. A sturdy chest that's blocked in one zone is open to everyone in the next. That way, everything is definitely in there and only needs to be edited if bots are to take something. In this case, I think it would be good if the whitelist was treated like a blacklist. 🙃

But like I said, that's not my decision. I'll just wait and see what happens next with this PR.

PR #1647, opened 2025-09-20
Implement Gruul's Lair strategy

brighton-chi:
This PR implements strategies for the two boss encounters in Gruul’s Lair: High King Maulgar and Gruul the Dragonkiller. I put together these strategies playing with 50% damage and 60% healing (via Individual Progression) and also after reverting the patch 3.0.2 30% HP nerfs to TBC raid mobs. Like Karazhan, I'm marking this as a draft for testing purposes, even though there is nothing further I am intending to do at this time.

### High King Maulgar:

**Overview**: This fight is, to put it lightly, a pain in the ass. It’s an easy fight that is won or lost on the pull, and it’s pretty messy getting bots to set it up properly, even with a strategy, considering that they cannot be manually positioned on an individual basis like real players (at least not by coded strategies). I think this fight is probably doable without strategies if you have really good DPS, as you can probably burn down a couple of the adds quickly to get away with a messy pull. I do not think it is doable without strategies if you adjust the numbers to create a more authentic 2007 experience. You will have bots that get one-shot, so without a strategy it’s a pure race.

**Strategies**:

_Standard Tanks_: This fight requires three traditional tanks (i.e., warrior, bear, or paladin). They are selected by the standard GetFirstMember logic (the final tank that joined your raid will be the first-selected tank and will tank Maulgar, the second-to-last tank that joined your raid will tank Olm, and the third-to-last tank that joined your raid will tank Blindeye). The Maulgar tank will drag Maulgar to the back wall near the entrance of the room. The Olm and Blindeye tanks will drag their respective targets to the back wall but closer to the exit of the room (out of range of Maulgar). Moving Olm is kind of finicky as he tends to want to just keep casting and refuse to chase the tank. I wasn’t able to come up with a smoother way of doing it, but as long as he’s out of range of Krosh and Maulgar, it’s not the end of the world.

_Moonkin Tank_: Kiggler will be tanked by the Moonkin Druid in your raid with the highest max HP (if you’re crazy enough to have more than one Moonkin Druid). There is no designated spot for Kiggler—ideally the Druid just tanks him where he stands, but he doesn’t always end up in the same spot. If you have no Moonkin Druid, you will need to figure out how to tank Kiggler on your own (basically any two ranged DPS can do it, or potentially a bear if you’re feeling lucky).

_Mage Tank_: This is the most important role in the fight and will be your Mage with the highest max HP (9K+ at a minimum not to get one shot, but ideally more). There is no specified location in the strategy; ideally Krosh is tanked where he stands, and in practice that’s generally how it works. The Mage tank will Spellsteal Krosh’s Spell Shield, which will reduce the damage of Krosh’s Greater Fireballs by 75%. There will be a gap between the Spell Shield’s expiration on the Mage tank and Krosh generating a new Spell Shield to be stolen; the Mage will apply Fire Ward on themselves at that time to try to survive the hit without the Spell Shield.

_DPS Priorities_: There is significant variance in kill order strategies—the only universally agreed-upon strategy is that all DPS should kill Blindeye first and Maulgar will die last. In this strategy, melee DPS will follow the kill order of Blindeye, Olm, Kiggler, and Maulgar. Ranged DPS will follow the kill order of Blindeye, Olm, Krosh, Kiggler (if melee has not killed him already), and Maulgar. As adds die, the respective tanks will join the melee/ranged DPS logic (as applicable). I understand that Kiggler casts Arcane Explosion, but it’s only every 30 seconds, and it does piss for damage. I find it more effective for melee to kill him than for melee to move directly to Maulgar after Olm.

_General Avoidance_: Bots will attempt to stay out of Krosh’s Blast Wave radius, Kiggler’s Arcane Explosion radius (when not attacking), and Maulgar’s Whirlwind (only while being channeled). Healers will attempt to stay near the center of the raid.

_Wild Fel Stalkers_: The intended strategy was for Warlocks to enslave the Wild Fel Stalkers and have them tank Olm (who casts a stacking debuff that, back in 2007 anyway, made it difficult for a traditional tank to remain on Olm). Unfortunately, after banging my head on my keyboard for hours, I realized that bots are not able to control enslaved demons. That is a gap in the general playerbots logic. So instead, Warlocks will banish Fel Stalkers as they are summoned.

_Misdirection_: The last Hunter that joined your party will cast Misdirection on your Moonkin tank and direct threat from Kiggler. The second-to-last Hunter that joined your party will cast Misdirection on your Mage tank and direct threat from Krosh. I discovered that Misdirection does not place an aura on the character on which Misdirection is cast so that it is not possible to cast Misdirection before the pull and have the bots attack the correct target. This was the best compromise I could come to.

### Gruul the Dragonkiller: 

**Overview**: This guy is kind of a giant piñata, and I’m certain no strategy is needed if you’re playing the base 3.3.5 version. But since the whole point of the fight is that it gets more difficult over time due to him gaining 15% damage with Growth every 30 seconds, any nerfs will compound the difficulty on this fight, so I think some strategies do make a big difference if you want to have a more authentic experience. But even with the nerfs I played with (which I believe result in feasible numbers), the bots could take down Gruul in under 10 Growth stacks. Back in the day, it was not uncommon for guilds to end up at 15+ stacks (players were just not as knowledgeable and therefore not as effective back then).

**Strategies**: 

_Tanking Location_: The tank with aggro will move Gruul to the center of the room. This is particularly important after Shatters, when Gruul will chase the tank to their post-Shatter location. I prefer using the tank with aggro in the strategies, even though Playerbots does recognize the main tank set in the UI, because some players might not be aware of the ability to actually set a main tank and therefore the strategy would fail for them.

_Spreading Ranged:_ At the beginning of the fight, ranged bots will fan out around Gruul. This is just so that they start spread through the room (spreading immediately takes them forever, and they end up not utilizing the whole space. This makes it much less likely that bots will end up stacked when Shatter hits.

_Shatter_: Bots will attempt to spread out after a Ground Slam. There is only so much that can be done though based on where the bots are thrown (the spreading of ranged in advance is really the key part of the fight). You will get some weird errors in WorldServer complaining about velocity <0.01f during this action. I do not know what causes this, but it doesn't appear to have any adverse effect. I see that one of the recent PRs that was reverted attempted to address this issue.

Good luck with getting a Dragonspine Trophy!

### DIFF
--- a/src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
+++ b/src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
@@ -1,0 +1,1066 @@
+#include "RaidGruulsLairActions.h"
+#include "RaidGruulsLairHelpers.h"
+#include "PlayerbotAI.h"
+#include "SpellAuras.h"
+#include "Unit.h"
+
+bool HighKingMaulgarRemoveTankAssistAction::Execute(Event event)
+{
+    if (!botAI->IsTank(bot))
+    {
+        return false;
+    }
+
+    Unit* maulgar  = AI_VALUE2(Unit*, "find target", "high king maulgar");
+    Unit* kiggler  = AI_VALUE2(Unit*, "find target", "kiggler the crazed");
+    Unit* krosh    = AI_VALUE2(Unit*, "find target", "krosh firehand");
+    Unit* olm      = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
+
+    bool maulgarEncounterActive = (maulgar  && maulgar->IsInCombat()) ||
+                                  (kiggler  && kiggler->IsInCombat()) ||
+                                  (krosh    && krosh->IsInCombat()) ||
+                                  (olm      && olm->IsInCombat()) ||
+                                  (blindeye && blindeye->IsInCombat());
+
+    if (maulgarEncounterActive && botAI->HasStrategy("tank assist", BOT_STATE_COMBAT))
+    {
+        botAI->ChangeStrategy("-tank assist", BOT_STATE_COMBAT);
+        {
+        return true;
+        }
+    }
+    return false;
+}
+
+bool HighKingMaulgarMaulgarTankAction::Execute(Event event)
+{
+    if (!IsMaulgarTank(botAI, bot))
+    {
+        return false;
+    }
+
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* maulgar = AI_VALUE2(Unit*, "find target", "high king maulgar");
+    if (!maulgar || !maulgar->IsAlive())
+    {
+        return false;
+    }
+
+    ObjectGuid currentIconGuid = group->GetTargetIcon(squareIcon);
+    if (currentIconGuid.IsEmpty() || currentIconGuid != maulgar->GetGUID())
+    {
+        group->SetTargetIcon(squareIcon, bot->GetGUID(), maulgar->GetGUID());
+    }
+
+    if (botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get() != "square" && 
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get() != maulgar)
+    {
+        botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("square");
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(maulgar);
+    }
+
+    if (bot->GetVictim() != maulgar)
+    {
+        Attack(maulgar);
+
+        if (!bot->IsWithinMeleeRange(maulgar))
+        {
+            return MoveTo(maulgar->GetMapId(), maulgar->GetPositionX(), maulgar->GetPositionY(), maulgar->GetPositionZ());
+        }
+    }
+
+    if (maulgar->GetVictim() == bot)
+    {
+        const TankSpot& spot = GruulsLairTankSpots::Maulgar;
+        const float maxDistance = 3.0f;
+        float distanceToMaulgar = maulgar->GetExactDist2d(spot.x, spot.y);
+        
+        if (distanceToMaulgar > maxDistance)
+        {
+            float dX = spot.x - maulgar->GetPositionX();
+            float dY = spot.y - maulgar->GetPositionY();
+            float moveX = spot.x + (dX / distanceToMaulgar) * maxDistance;
+            float moveY = spot.y + (dY / distanceToMaulgar) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, MovementPriority::MOVEMENT_COMBAT);
+        }
+        
+        float orientation = atan2(maulgar->GetPositionY() - bot->GetPositionY(), maulgar->GetPositionX() - bot->GetPositionX());
+        bot->SetFacingTo(orientation);
+        return false;
+    }
+
+    else if (!bot->IsWithinMeleeRange(maulgar))
+    {
+        return MoveTo(maulgar->GetMapId(), maulgar->GetPositionX(), maulgar->GetPositionY(), maulgar->GetPositionZ());
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarOlmTankAction::Execute(Event event)
+{
+    if (!IsOlmTank(botAI, bot))
+    {
+        return false;
+    }
+
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* olm = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    if (!olm || !olm->IsAlive())
+    {
+        return false;
+    }
+
+    ObjectGuid currentIconGuid = group->GetTargetIcon(circleIcon);
+    if (currentIconGuid.IsEmpty() || currentIconGuid != olm->GetGUID())
+    {
+        group->SetTargetIcon(circleIcon, bot->GetGUID(), olm->GetGUID());
+    }
+
+    if (botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get() != "circle" && 
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get() != olm)
+    {
+        botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("circle");
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(olm);
+    }
+
+    if (bot->GetVictim() != olm || olm->IsNonMeleeSpellCast(false))
+    {
+        Attack(olm);
+        
+        if (!bot->IsWithinMeleeRange(olm))
+        {
+            return MoveTo(olm->GetMapId(), olm->GetPositionX(), olm->GetPositionY(), olm->GetPositionZ());
+        }
+    }
+
+    if (olm->GetVictim() == bot && !olm->IsNonMeleeSpellCast(false))
+    {
+        const TankSpot& spot = GruulsLairTankSpots::Olm;
+        const float maxDistance = 3.0f;
+        float distanceToOlm = olm->GetExactDist2d(spot.x, spot.y);
+
+        if (distanceToOlm > maxDistance)
+        {
+            float dX = spot.x - olm->GetPositionX();
+            float dY = spot.y - olm->GetPositionY();
+            float moveX = spot.x + (dX / distanceToOlm) * maxDistance;
+            float moveY = spot.y + (dY / distanceToOlm) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, MovementPriority::MOVEMENT_COMBAT);
+        }
+
+        float orientation = atan2(olm->GetPositionY() - bot->GetPositionY(), olm->GetPositionX() - bot->GetPositionX());
+        bot->SetFacingTo(orientation);
+        return false;
+    }
+
+    else if (!bot->IsWithinMeleeRange(olm))
+    {
+        return MoveTo(olm->GetMapId(), olm->GetPositionX(), olm->GetPositionY(), olm->GetPositionZ());
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarBlindeyeTankAction::Execute(Event event)
+{
+    if (!IsBlindeyeTank(botAI, bot))
+    {
+        return false;
+    }
+
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
+    if (!blindeye || !blindeye->IsAlive())
+    {
+        return false;
+    }
+
+    ObjectGuid currentIconGuid = group->GetTargetIcon(starIcon);
+    if (currentIconGuid.IsEmpty() || currentIconGuid != blindeye->GetGUID())
+    {
+        group->SetTargetIcon(starIcon, bot->GetGUID(), blindeye->GetGUID());
+    }
+
+    if (botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get() != "star" && 
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get() != blindeye)
+    {
+        botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("star");
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(blindeye);
+    }
+
+    if (bot->GetVictim() != blindeye)
+    {
+        Attack(blindeye);
+        
+        if (!bot->IsWithinMeleeRange(blindeye))
+        {
+            return MoveTo(blindeye->GetMapId(), blindeye->GetPositionX(), blindeye->GetPositionY(), blindeye->GetPositionZ());
+        }
+    }
+
+    if (blindeye->GetVictim() == bot && !blindeye->IsNonMeleeSpellCast(false))
+    {
+        const TankSpot& spot = GruulsLairTankSpots::Blindeye;
+        const float maxDistance = 3.0f;
+        float distanceToBlindeye = blindeye->GetExactDist2d(spot.x, spot.y);
+
+        if (distanceToBlindeye > maxDistance)
+        {
+            float dX = spot.x - blindeye->GetPositionX();
+            float dY = spot.y - blindeye->GetPositionY();
+            float moveX = spot.x + (dX / distanceToBlindeye) * maxDistance;
+            float moveY = spot.y + (dY / distanceToBlindeye) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, MovementPriority::MOVEMENT_COMBAT);
+        }
+
+        float orientation = atan2(blindeye->GetPositionY() - bot->GetPositionY(), blindeye->GetPositionX() - bot->GetPositionX());
+        bot->SetFacingTo(orientation);
+        return false;
+    }
+
+    else if (!bot->IsWithinMeleeRange(blindeye))
+    {
+        return MoveTo(blindeye->GetMapId(), blindeye->GetPositionX(), blindeye->GetPositionY(), blindeye->GetPositionZ());
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarMageTankAction::Execute(Event event)
+{
+    if (!IsMageTank(botAI, bot))
+    {
+        return false;
+    }
+
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* krosh = AI_VALUE2(Unit*, "find target", "krosh firehand");
+    if (!krosh || !krosh->IsAlive())
+    {
+        return false;
+    }
+
+    ObjectGuid currentIconGuid = group->GetTargetIcon(triangleIcon);
+    if (currentIconGuid.IsEmpty() || currentIconGuid != krosh->GetGUID())
+    {
+        group->SetTargetIcon(triangleIcon, bot->GetGUID(), krosh->GetGUID());
+    }
+
+    if (botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get() != "triangle" && 
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get() != krosh)
+    {
+        botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("triangle");
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(krosh);
+    }
+
+    if (krosh->HasAura(SPELL_AURA_SPELL_SHIELD) && botAI->CanCastSpell(SPELL_SPELLSTEAL, krosh, true))
+    {
+        botAI->CastSpell(SPELL_SPELLSTEAL, krosh);
+    }
+
+    if (!bot->HasAura(SPELL_AURA_SPELL_SHIELD))
+    {
+        botAI->CastSpell("fire ward", bot);
+    }
+
+    if (bot->GetVictim() != krosh)
+    {
+        Attack(krosh);
+    }
+
+    float const OPTIMAL_RANGED_DISTANCE = 27.0f;
+    float const MIN_RANGE = OPTIMAL_RANGED_DISTANCE - 2.0f;
+    float const MAX_RANGE = OPTIMAL_RANGED_DISTANCE + 2.0f;
+
+    if (bot->IsWithinRange(krosh, MIN_RANGE) || !bot->IsWithinRange(krosh, MAX_RANGE) || 
+        !IsPositionSafe(botAI, bot, bot->GetPosition()))
+    {
+        Position safePos = FindSafePosition(botAI, bot, krosh, OPTIMAL_RANGED_DISTANCE);
+        return MoveTo(krosh->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarMoonkinTankAction::Execute(Event event)
+{
+    if (!IsMoonkinTank(botAI, bot))
+    {
+        return false;
+    }
+
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* kiggler = AI_VALUE2(Unit*, "find target", "kiggler the crazed");
+    if (!kiggler || !kiggler->IsAlive())
+    {
+        return false;
+    }
+
+    ObjectGuid currentIconGuid = group->GetTargetIcon(diamondIcon);
+    if (currentIconGuid.IsEmpty() || currentIconGuid != kiggler->GetGUID())
+    {
+        group->SetTargetIcon(diamondIcon, bot->GetGUID(), kiggler->GetGUID());
+    }
+
+    if (botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get() != "diamond" && 
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get() != kiggler)
+    {
+        botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("diamond");
+        botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(kiggler);
+    }
+
+    if (bot->GetVictim() != kiggler)
+    {
+        Attack(kiggler);
+    }
+
+    float const OPTIMAL_RANGED_DISTANCE = 29.0f;
+    float const MIN_RANGE = OPTIMAL_RANGED_DISTANCE - 1.0f;
+    float const MAX_RANGE = OPTIMAL_RANGED_DISTANCE + 1.0f;
+
+    if (bot->IsWithinRange(kiggler, MIN_RANGE) || !bot->IsWithinRange(kiggler, MAX_RANGE) || 
+        !IsPositionSafe(botAI, bot, bot->GetPosition()))
+    {
+        Position safePos = FindSafePosition(botAI, bot, kiggler, OPTIMAL_RANGED_DISTANCE);
+        return MoveTo(kiggler->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarMeleeDPSAction::Execute(Event event)
+{
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* maulgar = AI_VALUE2(Unit*, "find target", "high king maulgar");
+    Unit* kiggler = AI_VALUE2(Unit*, "find target", "kiggler the crazed");
+    Unit* olm = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
+
+    if (!botAI->IsMelee(bot) || 
+        (IsMaulgarTank(botAI, bot) && maulgar && maulgar->IsAlive()) || 
+        (IsOlmTank(botAI, bot) && olm && olm->IsAlive()) || 
+        (IsBlindeyeTank(botAI, bot) && blindeye && blindeye->IsAlive()))
+        {
+            return false;
+        }
+
+    // Target priority 1: Blindeye
+    if (blindeye && blindeye->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+
+        if (rtiValue != "star" || rtiTarget != blindeye)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("star");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(blindeye);
+        }
+
+        if (bot->GetVictim() != blindeye)
+        {
+            Attack(blindeye);
+        }
+
+        if (!bot->IsWithinMeleeRange(blindeye) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, blindeye, 5.0f);
+            return MoveTo(blindeye->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+
+        return false;
+    }
+
+    // Target priority 2: Olm
+    if (olm && olm->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "circle" || rtiTarget != olm)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("circle");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(olm);
+        }
+
+        if (bot->GetVictim() != olm)
+        {
+            Attack(olm);
+        }
+        
+        if (!bot->IsWithinMeleeRange(olm) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, olm, 5.0f);
+            return MoveTo(olm->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+
+        return false;
+    }
+
+    // Target priority 3: Kiggler
+    if (kiggler && kiggler->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+
+        if (rtiValue != "diamond" || rtiTarget != kiggler)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("diamond");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(kiggler);
+        }
+
+        if (bot->GetVictim() != kiggler)
+        {
+            Attack(kiggler);
+        }
+
+        if (!bot->IsWithinMeleeRange(kiggler) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, kiggler, 5.0f);
+            return MoveTo(kiggler->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+
+        return false;
+    }
+
+    // Target priority 4: Maulgar
+    if (maulgar && maulgar->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "square" || rtiTarget != maulgar)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("square");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(maulgar);
+        }
+
+        if (bot->GetVictim() != maulgar)
+        {
+            Attack(maulgar);
+        }
+        
+        if (!bot->IsWithinMeleeRange(maulgar) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, maulgar, 5.0f);
+            return MoveTo(maulgar->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarRangedDPSAction::Execute(Event event)
+{
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* maulgar = AI_VALUE2(Unit*, "find target", "high king maulgar");
+    Unit* kiggler = AI_VALUE2(Unit*, "find target", "kiggler the crazed");
+    Unit* krosh = AI_VALUE2(Unit*, "find target", "krosh firehand");
+    Unit* olm = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
+
+    if (!botAI->IsRanged(bot) || 
+        (IsMageTank(botAI, bot) && krosh && krosh->IsAlive()) || 
+        (IsMoonkinTank(botAI, bot) && kiggler && kiggler->IsAlive()) || 
+        botAI->IsHeal(bot))
+        {
+            return false;
+        }
+
+    float const OPTIMAL_RANGED_DISTANCE = 25.0f;
+    float const MIN_RANGE = OPTIMAL_RANGED_DISTANCE - 4.0f;
+    float const MAX_RANGE = OPTIMAL_RANGED_DISTANCE + 4.0f;
+
+    // Target priority 1: Blindeye
+    if (blindeye && blindeye->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "star" || rtiTarget != blindeye)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("star");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(blindeye);
+        }
+
+        if (bot->GetVictim() != blindeye)
+        {
+            Attack(blindeye);
+        }
+
+        if (bot->IsWithinRange(blindeye, MIN_RANGE) || 
+            !bot->IsWithinRange(blindeye, MAX_RANGE) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, blindeye, OPTIMAL_RANGED_DISTANCE);
+            return MoveTo(blindeye->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+
+        return false;
+    }
+
+    // Target priority 2: Olm
+    if (olm && olm->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "circle" || rtiTarget != olm)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("circle");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(olm);
+        }
+
+        if (bot->GetVictim() != olm)
+        {
+            Attack(olm);
+        }
+        
+        if (bot->IsWithinRange(olm, MIN_RANGE) || 
+            !bot->IsWithinRange(olm, MAX_RANGE) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, olm, OPTIMAL_RANGED_DISTANCE);
+            return MoveTo(olm->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+        
+        return false;
+    }
+
+    // Target priority 3: Krosh
+    if (krosh && krosh->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "triangle" || rtiTarget != krosh)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("triangle");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(krosh);
+        }
+
+        if (bot->GetVictim() != krosh)
+        {
+            Attack(krosh);
+        }
+        
+        if (bot->IsWithinRange(krosh, MIN_RANGE) || 
+            !bot->IsWithinRange(krosh, MAX_RANGE) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, krosh, OPTIMAL_RANGED_DISTANCE);
+            return MoveTo(krosh->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+        
+        return false;
+    }
+
+    // Target priority 4: Kiggler
+    if (kiggler && kiggler->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "diamond" || rtiTarget != kiggler)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("diamond");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(kiggler);
+        }
+
+        if (bot->GetVictim() != kiggler)
+        {
+            Attack(kiggler);
+        }
+        
+        if (bot->IsWithinRange(kiggler, MIN_RANGE) || 
+            !bot->IsWithinRange(kiggler, MAX_RANGE) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, kiggler, OPTIMAL_RANGED_DISTANCE);
+            return MoveTo(kiggler->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+        
+        return false;
+    }
+
+    // Target priority 5: Maulgar
+    if (maulgar && maulgar->IsAlive())
+    {
+        Unit* rtiTarget = botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Get();
+        std::string rtiValue = botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Get();
+        
+        if (rtiValue != "square" || rtiTarget != maulgar)
+        {
+            botAI->GetAiObjectContext()->GetValue<std::string>("rti")->Set("square");
+            botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(maulgar);
+        }
+
+        if (bot->GetVictim() != maulgar)
+        {
+            Attack(maulgar);
+        }
+        
+        if (bot->IsWithinRange(maulgar, MIN_RANGE) || 
+            !bot->IsWithinRange(maulgar, MAX_RANGE) || 
+            !IsPositionSafe(botAI, bot, bot->GetPosition()))
+        {
+            Position safePos = FindSafePosition(botAI, bot, maulgar, OPTIMAL_RANGED_DISTANCE);
+            return MoveTo(maulgar->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
+        }
+    }
+
+    return false;
+}
+
+bool HighKingMaulgarHealerAvoidanceAction::Execute(Event event)
+{
+    if (!botAI->IsHeal(bot))
+    {
+        return false;
+    }
+
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+
+    Unit* kiggler = AI_VALUE2(Unit*, "find target", "kiggler the crazed");
+    
+    bool needToMove = !IsPositionSafe(botAI, bot, bot->GetPosition());
+    if (kiggler && kiggler->IsAlive())
+    {
+        float distanceToKiggler = bot->GetDistance(kiggler);
+        if (distanceToKiggler < 30.0f)
+        {
+            needToMove = true;
+        }
+    }
+    
+    if (needToMove)
+    {
+        float centerX = 0, centerY = 0, centerZ = 0;
+        uint32 count = 0;
+        
+        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+        {
+            Player* member = ref->GetSource();
+            if (!member || !member->IsAlive() || member == bot) 
+            {
+                continue;
+            }
+            centerX += member->GetPositionX();
+            centerY += member->GetPositionY();
+            centerZ += member->GetPositionZ();
+            count++;
+        }
+        
+        if (count == 0)
+        {
+            return false;
+        }
+
+        centerX /= count;
+        centerY /= count;
+        centerZ /= count;
+        Player* centerPlayer = nullptr;
+        float minDistToCenter = 9999.0f;
+        
+        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+        {
+            Player* member = ref->GetSource();
+            if (!member || !member->IsAlive() || member == bot)
+            {
+                continue;
+            }
+
+            float dist = sqrt(pow(member->GetPositionX() - centerX, 2) +
+                             pow(member->GetPositionY() - centerY, 2) +
+                             pow(member->GetPositionZ() - centerZ, 2));
+                             
+            if (dist < minDistToCenter)
+            {
+                minDistToCenter = dist;
+                centerPlayer = member;
+            }
+        }
+        
+        if (centerPlayer)
+        {
+            Position safePos = FindSafePosition(botAI, bot, centerPlayer, 30.0f);
+            
+            if (kiggler && kiggler->IsAlive())
+            {
+                float safeDistToKiggler = sqrt(pow(safePos.GetPositionX() - kiggler->GetPositionX(), 2) +
+                                             pow(safePos.GetPositionY() - kiggler->GetPositionY(), 2));
+                
+                if (safeDistToKiggler < 31.0f)
+                {
+                    float dx = safePos.GetPositionX() - kiggler->GetPositionX();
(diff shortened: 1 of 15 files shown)


kadeshar:
Then you waiting until someone test that and then you mark that as normal PR which will check reviewed? Same is with Karazhan? Based on my experience dont expect that someone will test that in this state. Onyxia raid exists in current repo long time and dont work. Broken strategy or this which you dont want to use you can still easily remove from bots. And still something is better than nothing.

brighton-chi:
> Then you waiting until someone test that and then you mark that as normal PR which will check reviewed? Same is with Karazhan? Based on my experience dont expect that someone will test that in this state. Onyxia raid exists in current repo long time and dont work. Broken strategy or this which you dont want to use you can still easily remove from bots. And still something is better than nothing.

Ok, if you think that makes sense I can just mark this as ready for review.

Karazhan I need to recheck the BBW fight so that one I would say is not ready yet.

kadeshar:
Better test version after review especial that is new strategy and even with bugs merging to main repo should broke anything

brighton-chi:
Converted back to draft because I realized I can probably get rid of tank assist issues with Maulgar via multipliers instead of a direct removal of the strat, so I'll need to implement that and try it out.

brighton-chi:
Tank assist multiplier works well so now I don't need to mess with individual bot strategies. 

I’ve also modified the DPS kill order so melee now goes to Kiggler after Olm (ranged still goes to Krosh). This is because I found sending melee directly to Maulgar after Olm was more dangerous due to Kiggler potentially blasting bots into Maulgar’s whirlwind with arcane explosion, and also taking Maulgar into his second phase while Kiggler was not dead yet was unhelpful. So melee will probably kill Kiggler by themselves now (ranged will go to Kiggler after Krosh if he is still alive, but in my experience, melee will kill Kiggler faster). Some people have melee avoid Kiggler because of his arcane explosion, but he casts it only every 30 seconds, and it does piss damage, so I think this is the better strategy.

One thing I forgot to mention, which I implemented early on, is that I've set bots' sight distance to 150 yards during the Maulgar fight (it's configurable, and the default is 75 yards). I was finding 75 not to be enough, particularly if bots got blown away by Kiggler's arcane explosion, as in some cases a bot would get out of sight range of its RTI target, which would result in the bot becoming idle. The sight distance should return to whatever you have in your config after the fight.

One poor aspect of the fight is that Olm will not pull into the intended tanking spot near Blindeye against the wall and instead stands in the middle of the room, which is a dangerous position near Kiggler and Krosh. I could not fix this because the issue is caused by a longstanding Core issue with casters not moving to players when not casting (and it looks like the issue has even been reported with respect to Olm specifically). I haven’t checked the fight since the very recent PR that was intended to address idle casters but will try to do that soon.

noisiver:
I'm temporarily removing the waiting to be tested tag. The build is failing for reasons other than the workflow being bad.

brighton-chi:
The build should be fixed now. I also made a fix that should make Warlocks a lot better on HKM and redid the tank movement for Gruul a bit.

Unfortunately, Olm was not fixed by the Core caster movement commit. So it goes.

noisiver:
```
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp:11:31: error: ‘GruulsLairTankSpots::Maulgar’ was declared ‘extern’ and later ‘static’ [-fpermissive]
   11 |         static const TankSpot Maulgar  = { 90.686f, 167.047f, -13.234f, 3.009f };
      |                               ^~~~~~~
In file included from /home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp:1:
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.h:54:27: note: previous declaration of ‘GruulsLairTankSpots::Maulgar’
   54 |     extern const TankSpot Maulgar;
      |                           ^~~~~~~
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp:12:31: error: ‘GruulsLairTankSpots::Olm’ was declared ‘extern’ and later ‘static’ [-fpermissive]
   12 |         static const TankSpot Olm      = { 99.392f, 192.834f, -10.883f, 6.265f };
      |                               ^~~
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.h:55:27: note: previous declaration of ‘GruulsLairTankSpots::Olm’
   55 |     extern const TankSpot Olm;
      |                           ^~~
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp:13:31: error: ‘GruulsLairTankSpots::Blindeye’ was declared ‘extern’ and later ‘static’ [-fpermissive]
   13 |         static const TankSpot Blindeye = { 100.728f, 206.389f, -10.514f, 6.218f };
      |                               ^~~~~~~~
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.h:56:27: note: previous declaration of ‘GruulsLairTankSpots::Blindeye’
   56 |     extern const TankSpot Blindeye;
      |                           ^~~~~~~~
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp:14:31: error: ‘GruulsLairTankSpots::Gruul’ was declared ‘extern’ and later ‘static’ [-fpermissive]
   14 |         static const TankSpot Gruul    = { 241.238f, 365.025f, -4.220f, 0.0f };
      |                               ^~~~~
/home/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/strategy/raids/gruulslair/RaidGruulsLairHelpers.h:57:27: note: previous declaration of ‘GruulsLairTankSpots::Gruul’
   57 |     extern const TankSpot Gruul;
      |                           ^~~~~
 ```

**Comment by brighton-chi** [comment]
Derp. Well, should be good now.

File: src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp [hunk 1]
```diff
@@ -0,0 +1,1068 @@
...
+            float moveX = spot.x + (dX / distanceToMaulgar) * maxDistance;
+            float moveY = spot.y + (dY / distanceToMaulgar) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, MovementPriority::MOVEMENT_COMBAT);
+        }
+        
+        float orientation = atan2(maulgar->GetPositionY() - bot->GetPositionY(), maulgar->GetPositionX() - bot->GetPositionX());
+        bot->SetFacingTo(orientation);
+        return false;
+    }
+
```
kadeshar: enter to delete

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -0,0 +1,1068 @@
...
+            float moveX = spot.x + (dX / distanceToOlm) * maxDistance;
+            float moveY = spot.y + (dY / distanceToOlm) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, MovementPriority::MOVEMENT_COMBAT);
+        }
+
+        float orientation = atan2(olm->GetPositionY() - bot->GetPositionY(), olm->GetPositionX() - bot->GetPositionX());
+        bot->SetFacingTo(orientation);
+        return false;
+    }
+
```
kadeshar: enter to delete

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -0,0 +1,1068 @@
...
+        if (count == 0)
+        {
+            return false;
+        }
+
+        centerX /= count;
+        centerY /= count;
+        centerZ /= count;
+        Player* centerPlayer = nullptr;
+        float minDistToCenter = 9999.0f;
```
kadeshar: Consider using std::numeric_limits<float>::max()

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
```diff
@@ -0,0 +1,275 @@
...
+        return bestPos;
+    }
+
+    if (IsPositionSafe(botAI, bot, bestPos))
+    {
+        return bestPos;
+    }
+
+    const int NUM_POSITIONS = 32;
+    float bestScore = 99999.0f;
```
kadeshar: consider using std::numeric_limits<float>::max()

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: Sorry, I don't understand what you mean by this.

kadeshar: between if block and else block is empty line. Delete that line

brighton-chi:
I noticed a silly mistake for Gruul that I fixed but also had some problems in general with HKM (in particular, some types of avoidance behaviors were not working well at all). I decided to rewrite HKM, and I think it's in much better shape now. Among other things:

- With some tweaks to the misdirection action and tank movement, Olm now pulls out of the pack and goes to a spot by Blindeye away from the certain of the room, as intended. It takes a bit for him to get over there due to the Core issues with caster movement, so there isn't really opportunity to cleave him and Blindeye, but at least it gets him out of the danger area.

- The Krosh mage tank now has a dedicated spot. I was previously concerned about how relying on a tank spot would work out with a ranged bot, but it's actually working really well. The mage and Krosh just blast each other off by themselves until it's time for Krosh to die, as is supposed to be the case.

The fight is still pretty contingent on how the pull goes, but it's generally much smoother and less chaotic now. Apologies to the reviewers--like Karazhan, I'm not touching this one again until it gets merged, unless it's to respond to feedback.

Zerathane:
My comments having tested the High King Maulgar encounter from this PR.
Tested on a fresh build, Azerothcore playerbots fork with no additional modules. (No IP Mod)
Raid composition - 25 players: 3 tanks, 5 healers, 6 melee dps, 10 ranged dps.
Raid gear - average item level 106. autogear quality limit in config set to 3 (rare).

<ins>Boss pull</ins>
* On initial boss pull using "@ tank do attack my target" with HKM selected, all 3 tanks successfully and consistently attack their designated target and start to move them to their locations. The 3rd tank consistently takes aggro of Maulgar and moves him into the south east corner of the room. The second tank consistently aggros Olm and moves to the south west corner of the room. Due to his spell casting Olm, can take longer to move into position, but regularly does so without much issue. The 1st tank consistently aggros Blindeye and moves him to the south west corner of the room. 

* For Kriggler, the raid's moonkin focuses on him from the start and is generally able to hold and keep threat on him throughout the fight until a free tank taunts him away. He is most often tanked in his original location, perhaps slightly towards the centre of the room

* Krosh - The highest HP mage is always successfully selected to start tanking Krosh. I have also tested myself as a mage player giving myself the highest HP and this doesn't break the logic, and the playerbot mage spell steals as it should. The designated mage successfully uses its molten armour spell to help with tanking. Behavour i've noted - It can take a small while for the mage to build up it's aggro, especially if a tank has used a cleaving ability on pull for example avenger's shield.

* On pull both hunters cast misdirection. I can't confirm this is applied to the correct targets as the MD buff goes onto themselves which is what it shown in Skada's damage report.

<ins>The fight </ins>

* All playerbots adhere to the target kill orders assigned to them. Blindeye is always the first boss to die followed by Olm. After this the melee correctly target Kiggler and ranged target Krosh. Maulgar is always the last boss to die.

* Raid members are occasionally hit by blast wave. This seems to happen the most if Kiggler is too close to Krosh while the melee are killing Kiggler. This alone has not caused a wipe, and does not consistently happen, however if blast wave and arcane explosion happen close to each other it can cause how hp members to occasionally die.

* Wild Fel stalker is always banished out of the fight and disappears during the last part of the encounter when only Maulgar is alive.

* All raid members successfully avoid Maulgar's whirlwind ability. Only the main tank stays on the boss and takes the hits.

hkmposition

I've included this screenshot of a kill as it gives a very good indication of how the bosses are spread out. The vast majority of my kills have looked this this. Occasionally, Krosh and Kriggler being this close or closer can lead to some melee being hit by blast wave. 

hkmdps

This is an indication of my average raid dps during a kill. Just for consideration I'm not using IP module or any nerfs to player damage or healing, however the gear level I have chosen for the kill I'd regard as being low for this content. Blue gear lower, lower ilv than heroic dungeon drops. 

In total I managed 25 attempts with 23 kills, 2 wipes. 1 wipe seemed down to very slow movement of all bosses to their respective positions,  the majority of the raid being hit with an arcane explosion and blast wave together and losing a few low hp healers. I have not been able to replicate this. Not all kills went super smoothly, I think down to my gear / hp level Krosh can take out players if the mage tank isn't quick enough to gain aggro. Promisngly however, I have noticed if the main mage tank dies, the next highest hp mage successfully takes over and the fight has been saved.

Overall I consider the strategies implemented very successful. No input was needed by me as a player during any of the kills.

I've possibly missed some details but let me know if you have questions - I'll move on to Gruul and give that feedback in a separate comment.

brighton-chi:
Awesome, thank you. This is consistent with my observations. I think the current weak point of the fight is Kiggler tanking. I don't have a great way to solve it because his Arcane Explosion has the same radius as a Boomkin's spells, if the Boomkin has not taken the talent to increase spell range. Because of this, I cannot implement a forced tanking spot for Kiggler, and as a result, even if the Boomkin has the increased range talent, there is a risk of the Boomkin getting blasted around the room. It is not great, but I have not had issues with it causing wipes (just chaos and some deaths sometimes), so in lieu of a better solution, I've left that as is.

I had not considered the matter of tanks cleaving on pull (and in particular, Avenger's Shield with a Paladin tank). Has that resulted in Krosh launching Fireballs before the Mage tank gets aggro? If a couple hit the tanks, that is probably OK, but if he is one-shotting healers, then it's a problem. I understand Avenger's Shield is the standard Paladin pull skill though and they spam it on cooldown, so a solution is not immediately evident to me. If there is minimal impact on the result, I will probably leave that as is.

Zerathane:
> I had not considered the matter of tanks cleaving on pull (and in particular, Avenger's Shield with a Paladin tank). Has that resulted in Krosh launching Fireballs before the Mage tank gets aggro? If a couple hit the tanks, that is probably OK, but if he is one-shotting healers, then it's a problem. I understand Avenger's Shield is the standard Paladin pull skill though and they spam it on cooldown, so a solution is not immediately evident to me. If there is minimal impact on the result, I will probably leave that as is.

Yeah I can't think of an immediate solution this however it hasn't exactly been causing issues either. Similar can happen with Kriggler - The moonkin can need a few moments to gather his aggro.

I haven't had issues with this causing wipes either - perhaps the odd death on a few attempts but we've been able to recover each time.

Zerathane:
My comments regarding the Gruul the Dragonkiller ecounter of this PR.
As with HKM, Tested on a fresh build, Azerothcore playerbots fork with no additional modules. (No IP Mod)
Raid composition - 25 players: 3 tanks, 5 healers, 6 melee dps, 10 ranged dps.
Raid gear - average item level 106. autogear quality limit in config set to 3 (rare).

<ins> The pull </ins>

* The initial pull goes very well for this encounter. The designated main tank aggros the boss and stays in position in the middle of the room. I'll refer to this area as the white circle moving forward as it's literally white in the center.

* The ranged dps also spread around the boss form a circle around him, keep their distance from one another and using the entire room.

<ins> The fight </ins>

* For the fight, the tanks do a decent job at keeping the boss within the white circle. When they're thrown out by ground slam they will move back into the circle so the boss can be positioned more centrally. However, I've noticed on occasion, if the main tank is ground slammed to the north of the room and Gruul follows him. When the tank moves back to the middle, Gruul is quite off centred. The room isn't a perfect circle and there's physically less room between the white circle and the northern wall. RNG plays a factor here.

* Ground slam + shatter. The mechanic seems to be challenging for the bots at times and I can average between 0 to 4 deaths per shatter. Sometimes, a bot can be running away, but parallel to another bot who's running in the same direction. Furthermore if a group of bots are slammed in the same direction, they sometimes don't react quick enough to find a direction to start running thus causing deaths. Naturally the longer the fight goes on and more bots die, the better they succeed at doing the shatter mechanic. Whilst artificially keeping Gruul alive - I've been able to get the raid team to hit the soft enrage of 11 stacks until tanks are getting one shot. This happens at about 7 minutes. 

* A consideration, a main tank standing in a cave in will not move out of this aoe. I'm guessing because it's his designated tanking position and it doesn't want to move out of there. Ranged dps / healers can exhibit strange movement behaviour too where they're stuck in a loop of, moving out of cave in aoe. then moving back into it, then out again and so on until the aoe ends. It doesn't result in the bot's death but they are not dpsing at this time when it's happening.

gruulpositioning

The image above tries to show Gruul in that northern position as I mentioned. The tank was thrown to the northern wall, and his turning and moving back to centre doesn't fully bring Gruul with him, causing bots to be conjested around him. 

Overall though the boss is definitely killable with the strategies implemented, I totalled 16 attempts with 10 kills. (those 6 wipes being testing runs where I made Gruul invincible just to test the strategy rather than kill the boss). However it felt a bit RNG at times as to how many bots we'd lose during a shatter phase. Similar to HKM fight my raid team has been purposely under geared for this but I know the damage values will still be inflated compared to TBC era or those running IP Module. I believe the fight will be completable using the IP module, but with my gear level < ilv 106 7 minutes was the maximum time we could survive the encounter before growth stacks became too high, so IP module users may find this fight more challenging until their gear improves.

Zerathane:
Edit I forgot to add to my previous comment, The HKM and Gruul encounter do exhibit console errors but I believe the author is aware of it and it's not harmful?

gruulconsole

brighton-chi:
Thank you, re: the cave-ins, to be clear, melee (other than the tank) exhibit no issues with avoiding them? I have not accounted for the cave-in in the strategy so avoidance is reliant on the standard bot avoid aoe strategy, but it is certainly possible that ranged trying to spread vs. also avoiding cave-in is causing some conflict. That is something I can probably improve.

I can try something else with the tank positioning, which is to have the tank move Gruul to the center point (instead of the bot moving itself to the center point). I'm not sure if that will introduce other issues, but it is worth a try. Given that the pain point is shatters and Gruul being centered is key to shatter survival, I can see if this positioning can also account for cave ins, though if avoidance messes with the positioning itself, I am inclined to let the tank continue to stand in cave-ins since it seems that damage is not a difference maker for the encounter (it is a very simple matter for a human tank to step out of the cave in and keep Gruul positioned, but that is the kind of thing that can get bots all out of whack).

brighton-chi:
> Edit I forgot to add to my previous comment, The HKM and Gruul encounter do exhibit console errors but I believe the author is aware of it and it's not harmful?
> 
> gruulconsole

It's not clear to me what causes these errors, but they occur when Gruul uses Shatter. They don't result from the strategy as I tried with the strategy disabled and still get the errors.

kadeshar:
Someone which know how reproduce error can make breakpoint here and check stacktrace?
obraz

Zerathane:
> Thank you, re: the cave-ins, to be clear, melee (other than the tank) exhibit no issues with avoiding them? I have not accounted for the cave-in in the strategy so avoidance is reliant on the standard bot avoid aoe strategy, but it is certainly possible that ranged trying to spread vs. also avoiding cave-in is causing some conflict. That is something I can probably improve.
> 
Checked again this morning and melee do move out of the cave in aoe. Only the main tank seems to remain. But for the MT the damage seems negligible to me.

brighton-chi:
I played around with this and gave it some more thought and decided I am going to leave things as they are.

1. For the cave-ins, ideally how the strategy should work is that when spreading, bots pick locations that aren't in cave-ins (i.e., combine the avoidance and spreading into a single method instead of relying on the general avoid aoe strategy). However, I am not sure how to do this. The issue is that there are three types of AoEs in WoW (as far as mechanics of how they are generated). The first type is an AoE generated by a trigger NPC, and the second type is an AoE generated by a game object. I can account for these types of AoEs because bots can find the NPC or game object (even if invisible) and maintain a specified radius from it when moving. The third type, however, which includes cave-in, is an aura linked to a dynamic object. I do not know how to make bots detect such AoEs before actually walking into them. I could probably make them not walk back into an AoE after walking into it, but that would be reliant on a timer, and the bot wouldn't know how long the AoE has been there, so they would have to avoid the spot for the full 15-second duration. It would be restrictive and would be a lot to code in, and I'm not sure it is something that has been done before in boss strategies so I wouldn't have existing code to leverage.

2. With respect to the tank positioning, I'm not sure what the issue is with Gruul or Gruul's room, but applying the same logic that I use for tanking other bosses results in the tank running side to side repeatedly (hence why I have a multiplier for Gruul's tank to lock him in place). Whatever I try in terms of modifying the movement creates that side-to-side result again. And I'm not really sure how to make a method to pull Gruul into the middle of the room better since a bot, unlike a person, cannot be told where to put a boss and be able to figure out how to move itself to get the boss into place.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -1,24 +1,17 @@
...
-    Group* group = bot->GetGroup();
-    if (!group)
-    {
-        return false;
-    }
-
     Unit* maulgar = AI_VALUE2(Unit*, "find target", "high king maulgar");
-    if (!maulgar || !maulgar->IsAlive())
+    Group* group = bot->GetGroup();
+    if (!IsMaulgarTank(botAI, bot) || !group || !maulgar || !maulgar->IsAlive())
```
kadeshar: Things which are not stricte related with action execution should be checked in trigger, isPossible or isUseful. For example:
- that bot is in group
- that bot is maulgar tank

I think that bot is within melee range also.

Additionally condition checking that bot is in group also exists in IsMaulgarTank and is doubled.

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
kadeshar: Consider using existing methods to calculate which tank tanks which boss like:
MainTank - tank Maulgar (method botAI->IsMainTank(bot))
First assist tank - tank Olm (method botAI->IsAssistTankOfIndex(bot, 0)
Second assist tank - tank Blindeye (method botAI->IsAssistTankOfIndex(bot, 1)

That also give master more natural control which tank will tank which boss using raid leader options.

brighton-chi: So I thought about this. Yes, I can have it so that one of the tanks is set by main tank and another is set by main assist and then the player can set on the fly which tank has which role. The concern I have is then if the player doesn't assign those roles, the strategy doesn't work at all. Of course we could put it in the wiki, but I suspect it is unlikely people will read the wiki.

The way it is now, the fight works as is, and technically, players can still pick the tanks by the order in which tanks are invited to the group (which can be noted in the wiki for anybody who cares enough to investigate).

Or probably the code could have it so that if the player sets a main tank and/or assist, that will be used, but if main tank/assist is null, the tanks are picked by GetFirstMember. Maybe that's the best way?

kadeshar: It work. If player dont assign this roles, main tank and assist tank will be choose by their group order.
obraz

brighton-chi: Oh really? That's great. I'll definitely change it then. Probably I'll update the Karazhan code to allow picking too (would be new PR now of course hah).

Dreathean: One factor to consider also, if you don’t use that method then would that mean you couldn’t assign main tank to a real player?

kadeshar: Why? If player mark himself as main tank then he should tank maulgar. And that i real server assumption because maulgar is hardest to tank if i remember correctly

brighton-chi: Yes, that's another weakness of all the automatic role assigning. I need to look at the code when I'm at a computer, but it's theoretically possible that it could allow the player to tank by putting the main tank flag on themselves. Basically, if the main tank flag is set on a player OR a bot, then that player or bot will tank Maulgar, but if no main tank flag is set, GetFirstMember will be used to fetch a tank to so it.

kadeshar: Even if this first member is player. I created workaround. There is method IsBotMainTank(). You can check that also

brighton-chi: I think Dreathean wants the player to be able to tank if they want. Probably if somebody is aware enough to set the main tank flag, they are intending to tank all bosses so maybe it's best to allow the main tank to be a player or bot (so if you set a player to the main tank, no bot will tank Maulgar).

Dreathean: Yes, what Crow said - in other words I'm agreeing with adding what you suggested Grom, because I think without that a player can't identify themselves as the main tank from the bot's perspective?

brighton-chi:
Updated to address comments, clean up some stuff (e.g., remove excess includes), and also to modernize the code/make edits for best practice as I learn a little bit more about C++. Now, you can choose the tank for Gruul or Maulgar by setting the main tank (should work for a player too). You can also choose the tank for Olm by setting a tank as an assistant. If you don't set anything, it's fine because the logic will default to the standard GetFirstMember so all the tanks will still be assigned (based on the standard reverse group join order).

Gromleq, regarding what we talked about for group checks, I think actually we should have them for every method that references group because I think there is one case in which there could be a crash due to no group, which is if the player disbands the party inside the instance and then engages the boss (maybe on accident) before the player and bots are kicked from the instance.

brighton-chi:
> Someone which know how reproduce error can make breakpoint here and check stacktrace?
> 
> obraz
> 
> 
> 
> 

Purely speculating here, but it might be because of the Gronn Lord's Grasp aura from Ground Slam (slowing the player before they freeze and Gruul uses Shatter). Since the aura slowly reduces your speed until you stop, maybe there's a moment right before you totally freeze that the velocity is low enough to cause the error.

I know it's not related to the strategy and is an error arising in some fashion from the Shatter ability mechanics. It happens every time.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -52,31 +41,37 @@ bool HighKingMaulgarMaulgarTankAction::Execute(Event event)
             float dY = spot.y - maulgar->GetPositionY();
             float moveX = spot.x + (dX / distanceToMaulgar) * maxDistance;
             float moveY = spot.y + (dY / distanceToMaulgar) * maxDistance;
-            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, 
-                          MovementPriority::MOVEMENT_COMBAT);
+            {
+                return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, 
+                              MovementPriority::MOVEMENT_COMBAT, true, false);
+            }
```
kadeshar: What tool do you use which generate this {} without if, for or other?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -114,26 +104,33 @@ bool HighKingMaulgarOlmTankAction::Execute(Event event)
             float dY = spot.y - olm->GetPositionY();
             float moveX = spot.x + (dX / distanceToOlm) * maxDistance;
             float moveY = spot.y + (dY / distanceToOlm) * maxDistance;
-            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, 
-                          MovementPriority::MOVEMENT_COMBAT);
+            {
+                return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, 
+                              MovementPriority::MOVEMENT_COMBAT, true, false);
+            }
```
kadeshar: Another...

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -171,8 +162,10 @@ bool HighKingMaulgarBlindeyeTankAction::Execute(Event event)
             float dY = spot.y - blindeye->GetPositionY();
             float moveX = spot.x + (dX / distanceToBlindeye) * maxDistance;
             float moveY = spot.y + (dY / distanceToBlindeye) * maxDistance;
-            return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, 
-                          MovementPriority::MOVEMENT_COMBAT);
+            {
+                return MoveTo(bot->GetMapId(), moveX, moveY, spot.z, false, false, false, false, 
+                              MovementPriority::MOVEMENT_COMBAT, true, false);
+            }
```
kadeshar: Another

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -1,20 +1,15 @@
...
-#include "SpellAuras.h"
+#include "Playerbots.h"
 #include "Unit.h"
 
+using namespace GruulsLairHelpers;
+
 bool HighKingMaulgarMaulgarTankAction::Execute(Event event)
 {
     Unit* maulgar = AI_VALUE2(Unit*, "find target", "high king maulgar");
     Group* group = bot->GetGroup();
```
kadeshar: That is redundant now

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -114,26 +104,33 @@ bool HighKingMaulgarOlmTankAction::Execute(Event event)
...
+    Unit* olm = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    Group* group = bot->GetGroup();
+
+    return botAI->IsAssistTankOfIndex(bot, 0) && olm && olm->IsAlive() && group;
+}
+
 bool HighKingMaulgarBlindeyeTankAction::Execute(Event event)
 {
     Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
     Group* group = bot->GetGroup();
```
kadeshar: That is redundant now

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -188,14 +181,18 @@ bool HighKingMaulgarBlindeyeTankAction::Execute(Event event)
...
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
+    Group* group = bot->GetGroup();
+
+    return botAI->IsAssistTankOfIndex(bot, 1) && blindeye && blindeye->IsAlive() && group;
+}
+
 bool HighKingMaulgarKroshMageTankAction::Execute(Event event)
 {
     Unit* krosh = AI_VALUE2(Unit*, "find target", "krosh firehand");
     Group* group = bot->GetGroup();
```
kadeshar: That is redundant now

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -210,12 +207,12 @@ bool HighKingMaulgarKroshMageTankAction::Execute(Event event)
         botAI->GetAiObjectContext()->GetValue<Unit*>("rti target")->Set(krosh);
     }
 
-    if (krosh->HasAura(SPELL_AURA_SPELL_SHIELD) && botAI->CanCastSpell("spellsteal", krosh))
+    if (krosh->HasAura(static_cast<uint32>(GruulsLairSpells::SPELL_SHIELD)) && botAI->CanCastSpell("spellsteal", krosh))
```
kadeshar: I never used such cast. Really that is not automatic?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -188,14 +181,18 @@ bool HighKingMaulgarBlindeyeTankAction::Execute(Event event)
     return false;
 }
 
+bool HighKingMaulgarBlindeyeTankAction::isUseful()
+{
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
+    Group* group = bot->GetGroup();
+
+    return botAI->IsAssistTankOfIndex(bot, 1) && blindeye && blindeye->IsAlive() && group;
```
kadeshar: Group check is inside IsAssistTankOfIndex

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -114,26 +104,33 @@ bool HighKingMaulgarOlmTankAction::Execute(Event event)
...
 
     return false;
 }
 
+bool HighKingMaulgarOlmTankAction::isUseful()
+{
+    Unit* olm = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    Group* group = bot->GetGroup();
+
+    return botAI->IsAssistTankOfIndex(bot, 0) && olm && olm->IsAlive() && group;
```
kadeshar: Group check is inside IsAssistTankOfIndex

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -727,24 +728,18 @@ bool HighKingMaulgarHunterMisdirectionAction::Execute(Event event)
         {
             continue;
         }
-        if (IsOlmTank(botAI, member)) olmTank = member;
-        if (IsBlindeyeTank(botAI, member)) blindeyeTank = member;
+        if (botAI->IsAssistTankOfIndex(bot, 0)) olmTank = member;
+        if (botAI->IsAssistTankOfIndex(bot, 1)) blindeyeTank = member;
```
kadeshar: else if

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -799,41 +773,118 @@ bool HighKingMaulgarHunterMisdirectionAction::Execute(Event event)
     return false;
```
kadeshar, line 525: This method never return true which sound wrong

src/strategy/raids/gruulslair/RaidGruulsLairStrategy.cpp
kadeshar: I not sure that single trigger for boss is good idea.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -799,41 +773,118 @@ bool HighKingMaulgarHunterMisdirectionAction::Execute(Event event)
...
-                      MovementPriority::MOVEMENT_COMBAT);
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
+        if (!member || !member->IsAlive())
+        {
+            continue;
+        }
+        if (botAI->IsAssistTankOfIndex(bot, 0)) olmTank = member;
+        if (botAI->IsAssistTankOfIndex(bot, 1)) blindeyeTank = member;
```
kadeshar: else if

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: Oops. That's not generated, that's me--I was going through adding braces and shouldn't have put them around this.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: What is redundant about it? Don't I need to define maulgar and group for the function?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: I have to do this cast if I am using enum class instead of enum.

src/strategy/raids/gruulslair/RaidGruulsLairStrategy.cpp
brighton-chi: Why not?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: Will check this later

src/strategy/raids/gruulslair/RaidGruulsLairStrategy.cpp
kadeshar: I almost sure that at least few of this action have same isUseful condition. 
Additionally could you check how `nc +debug` looks like in such approach. That is spammed by actions not executed?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
kadeshar: I dont see variable group usage in method HighKingMaulgarMaulgarTankAction::Execute

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
kadeshar: Hmmm i not sure that using enum class intead of enum in such case when there are none overload's using this enum class exists is good

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: What about group->GetTargetIcon and group->SetTargetIcon?

kadeshar: Ehhhhh ctrl + f is too hard for me :)

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: So the reason I did class was because the spell names are sometimes pretty generic and potentially could be used in another boss strat but represent a different spell ID. For example, "Banish" is specifically rank 2 here because it is a level 70 strategy--potentially a different rank could be used for a non-TBC strategy. Maybe the headers would never be included together so there's no real risk.

I don't have a strong opinion. If you think it's better to just use enum then that's fine with me.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: Just looked at it. It does potentially return true with the below.

    switch (hunterIndex)
    {
        case 0:
        return olm && olm->IsAlive() && olm->GetHealth() > 0.98f * olm->GetMaxHealth() && 
               olmTank && olmTank->IsAlive() && botAI->CanCastSpell("misdirection", olmTank);

        case 1:
        return blindeye && blindeye->IsAlive() && blindeye->GetHealth() > 0.90f * blindeye->GetMaxHealth() && 
               blindeyeTank && blindeyeTank->IsAlive() && botAI->CanCastSpell("misdirection", blindeyeTank);

        default:
            break;
    }

src/strategy/raids/gruulslair/RaidGruulsLairStrategy.cpp
brighton-chi: I've done debugging while testing and yeah you do get a lot of failed prereqs and useless.

You think there should be a different trigger for each method? What would you put in the trigger vs IsUseful in that case?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
kadeshar: I marked HighKingMaulgarHunterMisdirectionAction::Execute not HighKingMaulgarHunterMisdirectionAction::isUseful()

src/strategy/raids/gruulslair/RaidGruulsLairStrategy.cpp
kadeshar: As i mentioned:
- not duplicate conditions in trigger:IsActive and action:IsUseful
- to make debug proccess clean

All raid/dungeon strategies have that approach.

brighton-chi: Yeah I understand that, but what is not clear to me is what to put in IsUseful vs the trigger. Or is the point to put everything in the trigger and not have IsUseful?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: Oh I see what you mean. Fixed.

src/strategy/raids/gruulslair/RaidGruulsLairStrategy.cpp
kadeshar: I wrote why. What you dont understand?

brighton-chi: I'm not questioning why. I'll send you a message on Discord.

brighton-chi:
All IsUseful checks have been converted to triggers for each individual action method. Tested and still works.

I do want to start splitting up action methods into separate functions, but that's a future thing.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -0,0 +1,889 @@
...
+        Attack(maulgar);
+    }
+
+    if (maulgar->GetVictim() == bot)
+    {
+        const Location& tankPosition = GruulsLairLocations::MaulgarTankPosition;
+        const float maxDistance = 3.0f;
+
+        float dX = tankPosition.x - bot->GetPositionX();
+        float dY = tankPosition.y - bot->GetPositionY();
```
kadeshar: This value is never used because override in line 47

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -0,0 +1,889 @@
...
+    {
+        Attack(maulgar);
+    }
+
+    if (maulgar->GetVictim() == bot)
+    {
+        const Location& tankPosition = GruulsLairLocations::MaulgarTankPosition;
+        const float maxDistance = 3.0f;
+
+        float dX = tankPosition.x - bot->GetPositionX();
```
kadeshar: This value is never used because override in line 46

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
+        if (!member || !member->IsAlive())
+        {
+            continue;
+        }
+        else if (botAI->IsAssistTankOfIndex(bot, 0)) olmTank = member;
+        else if (botAI->IsAssistTankOfIndex(bot, 1)) blindeyeTank = member;
```
kadeshar: Shouldnt be botAI->IsAssistTankOfIndex(member, 1)?

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -0,0 +1,889 @@
...
+    Player* blindeyeTank = nullptr;
+
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
+        if (!member || !member->IsAlive())
+        {
+            continue;
+        }
+        else if (botAI->IsAssistTankOfIndex(bot, 0)) olmTank = member;
```
kadeshar: Shouldnt be botAI->IsAssistTankOfIndex(member, 0)?

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+bool FindSafePosition(PlayerbotAI* botAI, Player* bot, Position& outPos)
+{
+    const float SEARCH_RADIUS = 30.0f;
+    const int NUM_POSITIONS = 32;
+    outPos = { bot->GetPositionX(), bot->GetPositionY(), bot->GetPositionZ() };
+
+    if (IsPositionSafe(botAI, bot, outPos))
+    {
+        return false;
```
kadeshar: Not true? Position is safe

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: Jesus I completely screwed this one up last night.

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
brighton-chi: Should be false because if the position is safe then there is no need to find a safe position

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: lol yes. Thanks.

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
kadeshar: Then the method name is wrong. Something like TryGetNewSafePosition fit better
False always should return outPos empty (false and value are confusing)
True always should return outPos filled

brighton-chi:
Alright, no more changes except for responding to comments/testing.

I can't believe I was forgetting to set the priority on so many MoveTos. That defaults to MOVEMENT_NORMAL, which is super slow to respond in combat.

hermensbas:
Gonna review soon

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -56,36 +42,23 @@ bool HighKingMaulgarMaulgarTankAction::Execute(Event event)
         bot->SetFacingTo(orientation);
     }
     else if (!bot->IsWithinMeleeRange(maulgar))
-    {
-        return MoveTo(maulgar->GetMapId(), maulgar->GetPositionX(), 
-                      maulgar->GetPositionY(), maulgar->GetPositionZ());
-    }
+        return MoveTo(maulgar->GetMapId(), maulgar->GetPositionX(), maulgar->GetPositionY(), 
+                      maulgar->GetPositionZ(), false, false, false, false, 
+                      MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 48: I dont know why you deleted {} return was and it's not in one line.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -97,61 +70,48 @@ bool HighKingMaulgarOlmTankAction::Execute(Event event)
...
         }
     }
     else if (!bot->IsWithinMeleeRange(olm))
-    {
-        return MoveTo(olm->GetMapId(), olm->GetPositionX(), 
-                      olm->GetPositionY(), olm->GetPositionZ());
-    }
+        return MoveTo(olm->GetMapId(), olm->GetPositionX(), olm->GetPositionY(), 
+                      olm->GetPositionZ(), false, false, false, false, 
+                      MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 88: I dont know why you deleted {} return was and it's not in one line.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -161,46 +121,29 @@ bool HighKingMaulgarBlindeyeTankAction::Execute(Event event)
         bot->SetFacingTo(orientation);
     }
     else if (!bot->IsWithinMeleeRange(blindeye))
-    {
-        return MoveTo(blindeye->GetMapId(), blindeye->GetPositionX(), 
-                      blindeye->GetPositionY(), blindeye->GetPositionZ());
-    }
+        return MoveTo(blindeye->GetMapId(), blindeye->GetPositionX(), blindeye->GetPositionY(), 
+                      blindeye->GetPositionZ(), false, false, false, false, 
+                      MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 131: I dont know why you deleted {} return was and it's not in one line.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -225,44 +166,30 @@ bool HighKingMaulgarKroshMageTankAction::Execute(Event event)
         else
         {
             Position safePos;
-            if (FindSafePosition(botAI, bot, safePos))
-            {
-                return MoveTo(krosh->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
-            }
+            if (TryGetNewSafePosition(botAI, bot, safePos))
+                return MoveTo(krosh->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ,
+                              false, false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 183: if without {} with non oneline code

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -213,10 +156,8 @@ bool HighKingMaulgarKroshMageTankAction::Execute(Event event)
         if (distanceToKrosh > minDistance && distanceToKrosh < maxDistance)
         {
             if (!bot->IsWithinDist2d(tankPosition.x, tankPosition.y, tankPositionLeeway))
-            {
-                return MoveTo(bot->GetMapId(), tankPosition.x, tankPosition.y, tankPosition.z, false, false, false, false, 
-                              MovementPriority::MOVEMENT_COMBAT, true, false);
-            }
+                return MoveTo(bot->GetMapId(), tankPosition.x, tankPosition.y, tankPosition.z, false, 
+                              false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 170: if without {} with non oneline code

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -225,44 +166,30 @@ bool HighKingMaulgarKroshMageTankAction::Execute(Event event)
...
+        return Attack(kiggler);
 
     Position safePos;
-    if (FindSafePosition(botAI, bot, safePos))
-    {
-        return MoveTo(kiggler->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
-    }
+    if (TryGetNewSafePosition(botAI, bot, safePos))
+        return MoveTo(kiggler->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ,
+                      false, false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 209: if without {} with non oneline code

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -551,22 +407,21 @@ bool HighKingMaulgarHealerAvoidanceAction::Execute(Event event)
...
     }
 
     Position safePos;
-    if (FindSafePosition(botAI, bot, safePos))
-    {
-        return MoveTo(bot->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ);
-    }
+    if (TryGetNewSafePosition(botAI, bot, safePos))
+        return MoveTo(bot->GetMapId(), safePos.m_positionX, safePos.m_positionY, safePos.m_positionZ,
+                      false, false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 366: if without {} with non oneline code

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -749,25 +600,23 @@ bool GruulTheDragonkillerPositionBossAction::Execute(Event event)
...
                                   gruul->GetPositionX() - bot->GetPositionX());
         bot->SetFacingTo(orientation);
     }
     else if (!bot->IsWithinMeleeRange(gruul))
-    {
-        return MoveTo(gruul->GetMapId(), gruul->GetPositionX(), 
-                      gruul->GetPositionY(), gruul->GetPositionZ());
-    }
+        return MoveTo(gruul->GetMapId(), gruul->GetPositionX(), gruul->GetPositionY(), gruul->GetPositionZ(), 
+                      false, false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
```
kadeshar, line 564: if without {} with non oneline code

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
```diff
@@ -675,46 +533,40 @@ bool HighKingMaulgarHunterMisdirectionAction::Execute(Event event)
...
+                    pet->GetCharmInfo()->SetIsFollowing(false);
+                    pet->GetCharmInfo()->SetIsCommandFollow(false);
+                    pet->GetCharmInfo()->SetIsReturning(false);
                 }
-                botAI->CastSpell("steady shot", olm);
+                pet->ToCreature()->AI()->AttackStart(blindeye);
             }
-            return true;
+            return botAI->CastSpell("steady shot", olm);
+        }
```
kadeshar, line 512: I think after all case should be break to not go to next case check

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
kadeshar: Doubled space before '='

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.h
kadeshar: Spells have `= spellid` in different place. Dont adjust this position or adjust correctly

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
brighton-chi: I was thinking if there is just one statement, then braces should be removed. If it's one statement but split into multiple lines because of length, the style is to still use braces?

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.h
brighton-chi: Thanks, will fix. I think it looks better to have all the spells aligned, but if you prefer just one space after each spell I can do that instead.

src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp
hermensbas: One statement it fine, but generally is one statement oneliners

hermensbas:
This might help @brighton-chi https://www.azerothcore.org/wiki/cpp-code-standards 

Meanwhile we, or @Dreathean, will deal with the wiki once merged.

brighton-chi:
> This might help @brighton-chi https://www.azerothcore.org/wiki/cpp-code-standards
> 
> Meanwhile we, or @Dreathean, will deal with the wiki once merged.

I can coordinate with Dreathean after this is merged. Some of the stuff in the original post is not accurate since the strategy has evolved over time.

The latest push looks like a lot of changes but it's just because I renamed almost all of the actions and triggers--you can see that otherwise it is just addressing comments with like two other tiny tweaks. I realized it was kind of silly to have the actions and triggers have the same names so I wanted to try to make them different in a way that hopefully makes more sense.

hermensbas (dismissed):

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
kadeshar: Still not fixed

brighton-chi: The double space is so it lines up with all the locations below it. I moved the comments to above the locations--does that make it more clear?

brighton-chi:
@kadeshar 

Sending message here because for some reason it won't let me reply to your comment. Regarding the constant location alignment, the double space is so the maulgar location lines up with all the locations below it. I moved the comments to above the locations--does that make it more clear?

kadeshar:
> @kadeshar
> 
> Sending message here because for some reason it won't let me reply to your comment. Regarding the constant location alignment, the double space is so the maulgar location lines up with all the locations below it. I moved the comments to above the locations--does that make it more clear?

Ok, im blind :D

kadeshar (dismissed):

brighton-chi:
I made the same edits here as for Magtheridon so that the action for assigning the kill order can return true.  I also combined the melee and ranged DPS priority actions because there is no need for them to be separate (easy enough to just put one extra condition in to account for the fact that only ranged should attack Krosh). II tested and everything works fine.

brighton-chi:
Looks like a lot happened but it's due to needing an extra level of indents inside the helper namespaces. I made this commit just to move the enums in the helper header inside the helper namespace because I am running into issues with conflicting definitions in other strategies I am working on.

src/strategy/raids/gruulslair/RaidGruulsLairHelpers.cpp
```diff
@@ -6,230 +6,229 @@
...
+        Unit* kiggler = botAI->GetAiObjectContext()->GetValue<Unit*>("find target", "kiggler the crazed")->Get();
+        Unit* krosh = botAI->GetAiObjectContext()->GetValue<Unit*>("find target", "krosh firehand")->Get();
+        Unit* olm = botAI->GetAiObjectContext()->GetValue<Unit*>("find target", "olm the summoner")->Get();
+        Unit* blindeye = botAI->GetAiObjectContext()->GetValue<Unit*>("find target", "blindeye the seer")->Get();
+
+        return (maulgar && maulgar->IsAlive()) ||
+               (kiggler && kiggler->IsAlive()) ||
+               (krosh && krosh->IsAlive()) ||
+               (olm && olm->IsAlive()) ||
+               (blindeye && blindeye->IsAlive());
```
kadeshar: Hmmm for example when maulgar is alive why getting all other boss. I think now this method is inefficient. Every get boss iterrate all hostiles units.
Consider changing that on single iteration similar to this
obraz

src/strategy/raids/gruulslair/RaidGruulsLairTriggers.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    switch (hunterIndex)
+    {
+    case 0:
+        return olm && olm->IsAlive() && olm->GetHealth() > 0.98f * olm->GetMaxHealth() && 
+               olmTank && olmTank->IsAlive() && botAI->CanCastSpell("misdirection", olmTank);
+
+    case 1:
+        return blindeye && blindeye->IsAlive() && blindeye->GetHealth() > 0.90f * blindeye->GetMaxHealth() && 
+               blindeyeTank && blindeyeTank->IsAlive() && botAI->CanCastSpell("misdirection", blindeyeTank);
```
kadeshar, line 128: Consider changing on blindeye->GetHealthPct() > 90.0f which is more readable

src/strategy/raids/gruulslair/RaidGruulsLairTriggers.cpp
```diff
@@ -0,0 +1,160 @@
...
+            continue;
+        else if (botAI->IsAssistTankOfIndex(member, 0)) olmTank = member;
+        else if (botAI->IsAssistTankOfIndex(member, 1)) blindeyeTank = member;
+    }
+
+    switch (hunterIndex)
+    {
+    case 0:
+        return olm && olm->IsAlive() && olm->GetHealth() > 0.98f * olm->GetMaxHealth() && 
+               olmTank && olmTank->IsAlive() && botAI->CanCastSpell("misdirection", olmTank);
```
kadeshar, line 124: Consider changing on olm->GetHealthPct() > 98.0f which is more readable

src/strategy/raids/gruulslair/RaidGruulsLairMultipliers.cpp
```diff
@@ -0,0 +1,116 @@
...
+           dynamic_cast<CastFeralChargeCatAction*>(action);
+}
+
+float HighKingMaulgarDisableTankAssistMultiplier::GetValue(Action* action)
+{
+    Unit* maulgar = AI_VALUE2(Unit*, "find target", "high king maulgar");
+    Unit* kiggler = AI_VALUE2(Unit*, "find target", "kiggler the crazed");
+    Unit* krosh = AI_VALUE2(Unit*, "find target", "krosh firehand");
+    Unit* olm = AI_VALUE2(Unit*, "find target", "olm the summoner");
+    Unit* blindeye = AI_VALUE2(Unit*, "find target", "blindeye the seer");
```
kadeshar: This variables are now unused

brighton-chi:
Updated for comments and to remove whitespaces but no need for review at the moment, pending resolution on playerbots approach to braces around single-statements that are broken across multiple lines.

PR #1649, opened 2025-09-21
Possible loop fix for Call of the Elements

kadeshar:


kadeshar:
Possible fix: https://github.com/liyunfan1223/mod-playerbots/issues/1648

Issue #1648, opened 2025-09-20
Randombot shamans do not heal in dungeon groups, seems new "resto" strategy is broken


odt1:
**Bug Description**
It seems that [after new shaman rework](https://github.com/liyunfan1223/mod-playerbots/pull/1566) low level randombot shamans (~34) stopped healing in dungeongroups _in combat_, even with "resto" and "healer dps" strategies (specifically in Razorfen Kraul). They don't do **anything** in combat at all, and start healing only after the combat ends. 

The default CO list they get summoned (with "/1 lfg" command) is this - 

> "Strategies: aoe, avoid aoe, boost, cast time, chat, cure, default, dps assist, duel, flametongue, formation, healer dps, mana spring, potions, racials, resto, save mana, stoneskin, wrath of air"

Disabling "resto" makes then do some dps in combat, but they are mostly useless sitting at 5%, despite being highest level in group. 

**Commit Hash**
e175eb117812e2a93411c5fede1c927f28b99f0c

kadeshar:
You tried command `reset botAI`?

odt1:
multiple times, with different bots and groups, and can confirm it still happening on latest git build. the only thing they do - is place totems and then nothing, even if getting killed.

kadeshar:
Before combat start whisper to shaman `nc +debug`. That provides his deciscions. Provide that from combat.

Bricedmur255:
I've been encountering a similar issue. I have a shaman that was enhancement that would behave normally. I respecc'd them to elemental and all they would do is place totems. I've been trying everything I could think of and checking recent issues on discord and have managed to get them to deal damage but every other action is placing totems.

<img width="754" height="414" alt="Image" src="https://github.com/user-attachments/assets/507ce807-3b9e-47a6-b893-28e495656e75" />

codengine:
I can confirm that Resto Shamans only heal after the fight and focus on totems only, even at level 62+.

They are useless right now.

Tierisch:
> I can confirm that Resto Shamans only heal after the fight and focus on totems only, even at level 62+.
> 
> They are useless right now.

They are fine for me after building with the updates. This one was level 15 that I grabbed to run RFC.

<img width="499" height="50" alt="Image" src="https://github.com/user-attachments/assets/f7c04f50-fa01-42a0-a6c6-bc8a1bd2355b" />

odt1:
This is typical fight log

```
[P] [Odt]: @tank attack 
[...]

PR #1651, opened 2025-09-22
Fix SetTotemAction - add `isUseful` and get array size instead of pointer size

Tierisch:
*Edit - My following text doesn't take into account the additional commits made to this after I marked it as draft, and I don't claim to have tested it after those were added.*

Related to #1648 

With this, I have been unable to find any more shamans stuck looping trying to set totems for `Call of the Elements`.

This isn't perfect, as a shaman dropping defaults with `Call of the Elements` that are different from what its strategies want will still replace them individually afterwards, but it at least fixes the looping and has shamans dropping as many as they can at a time.

I think the `isUseful` could be where more checks are added to dynamically change what's set on the totem bar based on the current strategies the shaman has instead of just ensuring it has totems assigned at all, but I didn't want to get into that when stopping the looping itself was more important right now.

### DIFF
--- a/src/strategy/shaman/ShamanActions.cpp
+++ b/src/strategy/shaman/ShamanActions.cpp
@@ -99,7 +99,7 @@ bool SetTotemAction::Execute(Event event)
 
     uint32 totemSpell = 0;
 
-    // Iterate backwards due signed/unsigned int
+    // Iterate backwards to prioritize the highest-rank totem spell the bot knows
     for (size_t i = spellIdsCount; i-- > 0;)
     {
         const uint32 spellId = totemSpellIds[i];


src/strategy/shaman/ShamanActions.h
```diff
@@ -531,132 +531,134 @@ class CastNatureResistanceTotemAction : public CastTotemAction
...
+    size_t totemSpellIdsCount;
     int actionButtonId;
 };
 
 class SetStrengthOfEarthTotemAction : public SetTotemAction
 {
 public:
     SetStrengthOfEarthTotemAction(PlayerbotAI* ai)
-        : SetTotemAction(ai, "strength of earth totem", STRENGTH_OF_EARTH_TOTEM, TOTEM_BAR_SLOT_EARTH) {}
+        : SetTotemAction(ai, "strength of earth totem", STRENGTH_OF_EARTH_TOTEM, sizeof(STRENGTH_OF_EARTH_TOTEM)/sizeof(uint32), TOTEM_BAR_SLOT_EARTH) {}
```
kadeshar: I intentionally wrote method to not have this as parameter because its repeatable and calculable

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -109,3 +107,27 @@ bool SetTotemAction::Execute(Event event)
     bot->addActionButton(actionButtonId, totemSpell, ACTION_BUTTON_SPELL);
     return true;
 }
+
+bool SetTotemAction::isUseful()
+{
+    Player* bot = botAI->GetBot();
```
kadeshar: Override for already existing bot.

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -109,3 +107,27 @@ bool SetTotemAction::Execute(Event event)
...
+        {
+            highestKnown = totemSpellIds[i];
+            break;
+        }
+    }
+    if (!highestKnown)
+        return false; // Bot doesn't know any valid rank
+
+    // Only consider the bar set if the highest rank is assigned
+    return button->GetAction() != highestKnown;
```
kadeshar: Idea is ok. But that should be in trigger not in action. Now trigger fires and action fails. Its not good scenario.

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -109,3 +107,27 @@ bool SetTotemAction::Execute(Event event)
...
+    if (!button || button->GetType() != ACTION_BUTTON_SPELL || button->GetAction() == 0)
+        return true; // No totem assigned
+
+    // Find the highest rank the bot knows
+    uint32 highestKnown = 0;
+    for (int i = (int)totemSpellIdsCount - 1; i >= 0; --i)
+    {
+        if (bot->HasSpell(totemSpellIds[i]))
+        {
+            highestKnown = totemSpellIds[i];
```
kadeshar: Consider deleting variable highestKnown and returning value asap like:
return button->GetAction() != totemSpellIds[i];

src/strategy/shaman/ShamanActions.h
Tierisch: I think this part was the actual fix. I was trying the `isUseful` without much success for this specific issue until I noticed a build warning about getting the size of the pointer rather than the array.

It was after changing this that they stopped looping, and I left the `isUseful` in because I thought it should be there anyway.

src/strategy/shaman/ShamanActions.cpp
Tierisch: I'm not sure if you mean this needs an override or is acting as an unnecessary override. It does have an override in the header for it.

src/strategy/shaman/ShamanActions.h
Tierisch: > I think this part was the actual fix. I was trying the `isUseful` without much success for this specific issue until I noticed a build warning about getting the size of the pointer rather than the array.
> 
> It was after changing this that they stopped looping, and I left the `isUseful` in because I thought it should be there anyway.

Well, maybe not. I tried just removing the `isUseful` block again and some were back to looping. I know they were still doing it for me before I added the size changes earlier, though.

Tierisch:
Well, I partly addressed the review comments with the second commit.

I'm either getting Shamans looping again or missing things and getting crashes when I also try to redo the size changes or move the `isUseful` logic to the trigger, and I'm out of time for the weekend to work those out.

I'm setting this as draft and will return to it next weekend if someone else doesn't have a better fix in the meantime.

odt1:
This pr fixed #1648 at first glance.

kadeshar:
I though about such changes
https://github.com/kadeshar/mod-playerbots/commit/eff7a3dcfacec66b35b9f5103a775bdfebc9a3ea

After that shaman's healing in combat
obraz

kadeshar:
@hermensbas i modified PR applying things from my review. After check merge.

hermensbas:
I will check either later tonight or tomorrow.

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -94,9 +94,8 @@ bool CastSpiritWalkAction::Execute(Event event)
 bool SetTotemAction::Execute(Event event)
```
hermensbas, line 94:
```c++
bool SetTotemAction::Execute(Event event)
{
    const size_t spellIdsCount = sizeof(totemSpellIds) / sizeof(uint32);
    if (spellIdsCount == 0)
        return false; // early return

    uint32 totemSpell = 0;

    // Iterate backwards due signed/unsigned int
    for (size_t i = spellIdsCount; i-- > 0; )
    {
        const uint32 spellId = totemSpellIds[i];
        if (bot->HasSpell(spellId))
        {
            totemSpell = spellId;
            break; 
        }
    }

    if (totemSpell == 0)
        return false;

    bot->addActionButton(actionButtonId, totemSpell, ACTION_BUTTON_SPELL);
    return true;
}
```

src/strategy/shaman/ShamanTriggers.cpp
```diff
@@ -432,12 +432,15 @@ bool SetTotemTrigger::IsActive()
         return true;
 
     size_t totemSpellIdsCount = sizeof(totemSpellIds) / sizeof(uint32);
-    for (size_t i = 0; i < totemSpellIdsCount; ++i)
+    for (int i = (int)totemSpellIdsCount - 1; i >= 0; --i)
```
hermensbas: kinda same optimizations:

```c++
bool SetTotemTrigger::IsActive()
{
    if (!bot->HasSpell(SPELL_CALL_OF_THE_ELEMENTS))
        return false;

    if (!bot->HasSpell(requiredSpellId))
        return false;

    ActionButton const* button = bot->GetActionButton(actionButtonId);
    if (!button || button->GetType() != ACTION_BUTTON_SPELL || button->GetAction() == 0)
        return true;

    const size_t totemSpellIdsCount = sizeof(totemSpellIds) / sizeof(uint32);
    if (totemSpellIdsCount == 0)
        return false;

    for (size_t i = totemSpellIdsCount; i-- > 0; )
    {
        const uint32 spellId = totemSpellIds[i];
        if (bot->HasSpell(spellId))
        {
            return button->GetAction() != spellId;
        }
    }

    return false;
}
```

src/strategy/shaman/ShamanActions.cpp
hermensbas: not mandatory, but just optimizations. Although the early returns and removing explicit cast does not hurt either.

src/strategy/shaman/ShamanTriggers.cpp
hermensbas: not mandatory, but just optimizations

kadeshar:
@hermensbas i commited this optimalizations. I dont plan changing anything due copilot suggestions

noisiver:
The comment from CoPilot about the comment being unclear makes perfect sense though. The comments should be clear as to what the code does. `Iterate backwards due signed/unsigned int` is unclear, and `Iterate backwards to prioritize the highest-rank totem spell the bot knows` makes perfect sense.

hermensbas:
> The comment from CoPilot about the comment being unclear makes perfect sense though. The comments should be clear as to what the code does. `Iterate backwards due signed/unsigned int` is unclear, and `Iterate backwards to prioritize the highest-rank totem spell the bot knows` makes perfect sense.

True, that was the initial comment actually lel

PR #1659, opened 2025-09-24
Fix: Shaman bots stuck spamming “Call of the Elements” / “set … totem” (totem rank detection & trigger loop)

Wishmaster117:
### Summary

This PR Fix Issue : https://github.com/liyunfan1223/mod-playerbots/issues/1648

Shaman bots could get stuck in a loop continuously attempting “set mana spring/stoneskin/flametongue” and “Call of the Elements”, effectively doing no healing or DPS. This happened especially on lower levels (e.g., level 34) where not all totem ranks are known. This PR fixes the rank detection logic and the trigger conditions, and adds lightweight debug messages to help diagnose totem configuration.

### Root cause

**1. Array length bug:**

Code used sizeof(totemSpellIds) / sizeof(uint32) where totemSpellIds is a pointer, not a fixed array in that scope. This returned an incorrect count, so the selection logic and the “already set?” checks were unreliable.

**2. Rank selection order:**

The loops iterated ranks in the wrong direction relative to how arrays are ordered (arrays are sorted highest → lowest rank). The bot often failed to find a known rank or picked the wrong one.

**3. Missing counts for some totems:**

Constructors for Set/Trigger classes didn’t pass the *_COUNT , so the base class couldn’t know the real array size.

**4. Noisy re-trigger:**

Since the action bar didn’t match a valid known rank, the “set … totem” triggers kept firing, which in turn kept “Call of the Elements” active, starving heal/DPS actions.

### What’s changed

- Pass the real array size (*_COUNT) into all SetTotemAction and SetTotemTrigger constructors (including Totem of Wrath).

- Select the highest known rank correctly by iterating arrays from index 0 (arrays are sorted highest→lowest).

- Fix trigger “already set?” checks to use the provided count instead of sizeof(pointer).

- Add minimal debug logs:

  - When no known rank exists for a totem family (useful for low-level bots).

  - When a totem spell is assigned to the totem bar (prints action slot and spell id).

  - When Call of the Elements is successfully cast (single line).

**Files touched (high level)**

- src/strategy/shaman/ShamanActions.h/.cpp

- src/strategy/shaman/ShamanTriggers.h/.cpp

**Behavior after the fix**

- Shaman bot assigns each totem slot once with the highest rank it actually knows.

- “Call of the Elements” is not spammed; it fires when needed, then the bot proceeds to heal/DPS normally.

- If a bot doesn’t know any rank for a given totem, a single [totem] message indicates it—no loop.

### Repro & validation

**Before (broken):**

- Level ~30–40 Shaman bot with incomplete totem ranks.
- Observe repeated logs such as:

  - set mana spring totem, set stoneskin totem, set flametongue totem
  - call of the elements
  - Very few healing/DPS actions.

**After (fixed):**

- First combat: one-time [totem] Affecte la barre (...) -> spell id X lines per family.
- Occasional [totem] Call of the Elements lancé once per actual cast.
- Regular heals/DPS resume; no “set … totem”/CoE spam.

**Backward compatibility & risk**

- Changes are internal to Shaman totem setup logic; no behavior changes for other classes.

**Notes**

This PR does not add new features; it corrects rank resolution, trigger conditions, and adds minimal diagnostics to aid future debugging.

Tested ingame with a level 34 heal shaman

**I've intentionally left the old code commented out; we'll remove it once everything is confirmed to be OK.
Same goes for debug messages.**


Wishmaster117:
> This PR i doubled with #1651 Waiting for author to check/merge that [kadeshar@eff7a3d](https://github.com/kadeshar/mod-playerbots/commit/eff7a3dcfacec66b35b9f5103a775bdfebc9a3ea)
> 
> Fast PR checking:
> 
> * i deleted passing counts to method on purpose
> * messages dont have translations and config to turn that off 
> * commented code

Thanks for reviewing! 
I'me a bit lost with your two mods.

A quick note on why this PR isn’t so duplicate as this and what’s still missing to make the overall fix rock-solid.
What this PR ensures

Correct rank selection (High → Low): Totem arrays are ordered highest→lowest. I iterate forward and pick the first known ID, guaranteeing the highest known rank gets bound.

Fix applied to both Actions and Triggers: The loop came from both sides. I make triggers compare the action button against the entire family of valid IDs using the real size, so once a valid rank is bound, triggers stay off and the bot returns to heal/DPS.

No sizeof(pointer) pitfalls: I don’t rely on sizeof(totemSpellIds) where it’s a pointer.


**What’s still missing in the other work and would make it “complete”**

Avoid passing counts but keep correct sizes:
If counts shouldn’t be passed (as mentioned), i think you may use a templated constructor taking const uint32 (&ids)[N] in both SetTotemAction and SetTotemTrigger. This deduces N at compile time (no extra parameter, no sizeof(pointer) bug, no repeated sizeof(array)/sizeof(uint32) in 14 places).

Iterate in the right direction:
Because arrays are High→Low, iterate 0 → N-1 and stop at the first HasSpell(...) to guarantee the highest known rank, not a lower one.

Propagate the fix to the TRIGGERS:
In SetTotemTrigger::IsActive, compare the current action button against all IDs in the family using the real size N. This is key to stop the “set … totem / Call of the Elements” spam loop.


The other PRs address parts of the problem but either:
still compute sizes from a pointer or rely on repeated sizeof(array) at call sites, iterate backwards, which can pick lower ranks,
or don’t update the triggers, leaving room for the loop to reappear.

If we fold in the 3 points above (template size deduction, forward iteration, trigger coverage), we’ll have a clean, maintainable, and complete fix.


> * messages dont have translations and config to turn that off  => As a tell in the PR this messages will be deleted before Merge
> * commented code => As i said in the PR the commented code will be deleted befores merge

Celandriel:
> commented code => As i said in the PR the commented code will be deleted befores merge
If this is not ready to merge "as is" from your point of view because there is still additional testing to do, please mark as a draft.

Wishmaster117:
> > commented code => As i said in the PR the commented code will be deleted befores merge
> > If this is not ready to merge "as is" from your point of view because there is still additional testing to do, please mark as a draft.



> > commented code => As i said in the PR the commented code will be deleted befores merge
> > If this is not ready to merge "as is" from your point of view because there is still additional testing to do, please mark as a draft.

Ok here's a final working fix, i tried with all 3 shaman class specs.

Changes:
Switch player chat to server logs: Replaced all TellMasterNoFacing debug messages with LOG_DEBUG in the Shaman totem setup and Call of the Elements flow.

No more “count” parameters (per review): Removed passing *_COUNT to constructors/methods. Instead, added templated constructors that take a reference to the ID array (const uint32 (&ids)[N]) and deduce N at compile time for both SetTotemAction and SetTotemTrigger. All derived totem actions/triggers now call these ctors without a count argument.

No more commented code

Why?

Avoids chat spam; diagnostics now live in server logs if DEBUG selected.

Aligns with reviewer feedback (“i deleted passing counts to method on purpose”) while eliminating the sizeof(pointer) pitfall and removing repeated sizeof(array)/sizeof(uint32) boilerplate across classes.

What’s logged

Successful Call of the Elements cast.

No known rank for a totem family (includes array size and bot level).

Slot binding details (slot index, spell id).

Notes

No behavioral changes to combat/rotation—only logging/output and array-size handling were adjusted.

This also guarantees we always pick the highest known rank (arrays are High→Low; we iterate accordingly).

odt1:
this fix seems to work the best

Wishmaster117:
Yes im' testing in since two days with different shaman class and it works well, no more spams of call of element and totems correctly summoned

> this fix seems to work the best

Yes im' testing in since two days with different shaman class and it works well, no more spams of call of element and totems correctly summoned

hermensbas:
Solved the merged conflict kept the original, since there indications this fix works. I havent review the code yet though, since @kadeshar was working on a fix.

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -87,32 +89,42 @@ bool CastSpiritWalkAction::Execute(Event event)
     return false;
 }
 
+bool CastCallOfTheElementsAction::Execute(Event event)
+{
+    const bool ok = CastSpellAction::Execute(event);
+    if (ok)
+    {
+        LOG_DEBUG("playerbots", "[Totem] Call of the Elements cast");
```
kadeshar: For that is special strategy "debug spell" and such override for single spell in bad

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -87,32 +89,42 @@ bool CastSpiritWalkAction::Execute(Event event)
...
+        LOG_DEBUG("playerbots", "[Totem] No known rank for {} (count={}, level={})",
+                 getName(), totemSpellIdsCount, static_cast<uint32>(bot->GetLevel()));
         return false;
-
+    }
     bot->addActionButton(actionButtonId, totemSpell, ACTION_BUTTON_SPELL);
+    // Debug: report which spell id was assigned
+    {
+    LOG_DEBUG("playerbots", "[Totem] Bind slot {} -> spellId={}", actionButtonId, totemSpell);
+    }
```
kadeshar, line 119: {} without purpose

src/strategy/shaman/ShamanTriggers.cpp
```diff
@@ -424,28 +424,19 @@ bool SetTotemTrigger::IsActive()
...
-    }
-
-    for (int i = (int)totemSpellIdsCount - 1; i >= 0; --i)
+    for (size_t i = 0; i < totemSpellIdsCount; ++i)
     {
-        const uint32 spellId = totemSpellIds[i];
-        if (bot->HasSpell(spellId))
+        if (button->GetAction() == totemSpellIds[i])
         {
-            return button->GetAction() != spellId;
```
kadeshar, line 446: HasSpell + button->GetAction() != spellId should be checked in trigger.
In current shape shaman never upgrade his spell when he got higher rank.
Action should be executed only when needed and not fail because conditions are in wrong place

src/strategy/shaman/ShamanActions.cpp
hermensbas: @Wishmaster117 consider using "nc +debug whispers to bots if you wanna analyse there actions" that might help giving more insight when working on it.

Wishmaster117:
Hi, i post a fix tomorrow

src/strategy/shaman/ShamanActions.cpp
Wishmaster117: Solved, i put a nc  +debug for the 4 casted totem, i tried with different level shaman and they always cast th highest rank known.

src/strategy/shaman/ShamanTriggers.cpp
Wishmaster117: Solveed

src/strategy/shaman/ShamanActions.cpp
Wishmaster117: Debug removed because useless now

Wishmaster117:
Ok Ready to final review, i added a nc +debug in CoE tu display what totem is pulled and the spell ID so you can see that the shaman pull his highest totem rank

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -10,7 +10,47 @@
...
+
+        auto spellName = [loc](uint32 id) -> std::string {
+            if (!id) return "none";
+            if (const SpellInfo* si = sSpellMgr->GetSpellInfo(id))
+            {
+                const char* n = si->SpellName[loc];
+                return (n && *n) ? std::string(n) : std::string("spell ") + std::to_string(id);
+            }
+            return std::string("spell ") + std::to_string(id);
+        };
```
kadeshar: Module support only english dbc

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -10,7 +10,47 @@
...
+        for (int i = 0; i < 4; ++i)
+        {
+            const ActionButton* button = bot->GetActionButton(slots[i]);
+            uint32 id = (button && button->GetType() == ACTION_BUTTON_SPELL) ? button->GetAction() : 0;
+            out << slotLbl[i] << "=" << spellName(id) << " (id " << id << ")";
+            if (i != 3) out << ", ";
+        }
+        botAI->TellMasterNoFacing(out.str());
+    }
+    return CastSpellAction::Execute(event);
```
kadeshar: Why you so consistent want to make yourself logging system. Such debugging already exists as strategy "debug spell". Also we support debug only in english.

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -89,42 +129,37 @@ bool CastSpiritWalkAction::Execute(Event event)
...
+{
+    // Pick the highest-rank spell the bot knows (arrays are sorted from highest to lowest rank).
     uint32 totemSpell = 0;
     for (size_t i = 0; i < totemSpellIdsCount; ++i)
     {
         if (bot->HasSpell(totemSpellIds[i]))
         {
             totemSpell = totemSpellIds[i];
-            break;
+            break; // first known => highest known rank
```
kadeshar: Comment before loop is enough.

src/strategy/shaman/ShamanActions.cpp
```diff
@@ -89,42 +129,37 @@ bool CastSpiritWalkAction::Execute(Event event)
...
-    bot->addActionButton(actionButtonId, totemSpell, ACTION_BUTTON_SPELL);
-    // Debug: report which spell id was assigned
+
+    // Micro-safety: if the slot already holds this spell, do nothing.
+    if (const ActionButton* button = bot->GetActionButton(actionButtonId))
     {
-    LOG_DEBUG("playerbots", "[Totem] Bind slot {} -> spellId={}", actionButtonId, totemSpell);
+        if (button->GetType() == ACTION_BUTTON_SPELL && button->GetAction() == totemSpell)
+        {
+            return true; // already set correctly
```
kadeshar: That not mean action execute sucessful. That means that road trigger->isPossible->isUseful was bad. Im not sure that such check should be consider in execute. 
For that are isPossible and isUseful
Even yes then in such situation action should return false.

src/strategy/shaman/ShamanTriggers.cpp
```diff
@@ -422,21 +422,35 @@ bool NoAirTotemTrigger::IsActive()
...
-
-    for (size_t i = 0; i < totemSpellIdsCount; ++i)
-    {
-        if (button->GetAction() == totemSpellIds[i])
-        {
-            return false;
-        }
-    }
-    return true;
+   Player* bot = botAI->GetBot();
```
kadeshar: ....
Trigger have bot variable

src/strategy/shaman/ShamanActions.cpp
Wishmaster117: Ok removed locale check

src/strategy/shaman/ShamanActions.cpp
Wishmaster117: Comments removed

src/strategy/shaman/ShamanActions.cpp
Wishmaster117: You're right we return false

src/strategy/shaman/ShamanTriggers.cpp
Wishmaster117: True, Get remoded

src/strategy/shaman/ShamanActions.cpp
Wishmaster117: Custom code removed

Wishmaster117:
@kadeshar , by the way, spell debug doesn't display totems

kadeshar:
> @kadeshar , by the way, spell debug doesn't display totems

I know, display all CastSpellAction which dont have Execute overrides.
And CastCallOfTheElementsAction is such action.

Wishmaster117:
> > @kadeshar , by the way, spell debug doesn't display totems
> 
> I know, display all CastSpellAction which dont have Execute overrides. And CastCallOfTheElementsAction is such action.

Yes but i wanted to see what totems and ranks the shaman put to see if it's acurate is for that i added a custom debug code buts it's ok shaman always use the greatest known rank of the totem

kadeshar:
Im also sometimes making custom debug while development but i delete that before making PR.
That only make code harder to read.

Wishmaster117:
> Im also sometimes making custom debug while development but i delete that before making PR. That only make code harder to read.

Copy that i will remove all before posting

hermensbas:
Boyscouted the ccp files:

- added braces when in doubt
- removed all invalid Player* bot = botAI->GetBot();
- added missing 'variable check' (map) before usage

hermensbas:
ps: we should add spellIds header and namespace per class. But for another time. As not related to this implementation but in general it would help.

Wishmaster117:
> ps: we should add spellIds header and namespace per class. But for another time.

Oupsiee red flag😁

kadeshar:
Please next time make refactoring outside hotfix. We should merge that as fast as possible.

hermensbas:
> Please next time make refactoring outside hotfix. We should merge that as fast as possible.

Doesnt make sense removing the botAI->GetBot from every line and then skip 3 others in the same files. Leaves even more of a mess.

PR #1661, opened 2025-09-25
Fix: Prevent non-mana users

gacuna89:
This change modifies the `HasManaValue::Calculate()` function in `src/strategy/values/StatsValues.cpp`.

Previously, bots from classes that do not use mana (e.g., Warriors, Rogues) might still attempt to restore mana by drinking water or using other mana-restoring actions, which is illogical and wasteful.

This fix introduces a check to ensure that `HasManaValue` only returns `true` if:
1. The bot's power type is explicitly `POWER_MANA`.
2. The bot has a maximum mana pool greater than zero.

This prevents non-mana-using classes from trying to drink water or use other mana-restoring actions unnecessarily, improving bot AI logic and resource management.


kadeshar:
Addition look ok but action dont use that value in isUseful (only in trigger)

EDIT:
HasManaValue is checked inside Execute not isUseful which is no bad. Maybe i will fix that later.

kadeshar (approved):

PR #1662, opened 2025-09-25
Fix: Prevent bots from eating and drinking while mounted

gacuna89:
This change modifies the `DrinkAction::Execute()` and `EatAction::Execute()` functions in `src/strategy/actions/NonCombatActions.cpp`.

Previously, playerbots could attempt to eat food or drink water while mounted, which is not possible in the game and creates unrealistic behavior.

This fix adds mount state checks to both actions:
1. `DrinkAction::Execute()` now checks `bot->IsMounted()` and returns false if mounted
2. `EatAction::Execute()` now checks `bot->IsMounted()` and returns false if mounted

This prevents bots from attempting to consume food or drinks while mounted, improving bot behavior realism and preventing unnecessary action attempts that would fail anyway.

### DIFF
--- a/src/strategy/actions/NonCombatActions.cpp
+++ b/src/strategy/actions/NonCombatActions.cpp
@@ -13,8 +13,10 @@ bool DrinkAction::Execute(Event event)
     if (bot->IsInCombat())
         return false;
 
-    // Don't drink while mounted
     if (bot->IsMounted())
+        return false;
+
+    if (botAI->HasAnyAuraOf(GetTarget(), "dire bear form", "bear form", "cat form", "travel form", "aquatic form","flight form", "swift flight form", nullptr)) 
         return false;
 
     if (botAI->HasCheat(BotCheatMask::food))
@@ -73,8 +75,10 @@ bool EatAction::Execute(Event event)
     if (bot->IsInCombat())
         return false;
 
-    // Don't eat while mounted
     if (bot->IsMounted())
+        return false;
+
+    if (botAI->HasAnyAuraOf(GetTarget(), "dire bear form", "bear form", "cat form", "travel form", "aquatic form","flight form", "swift flight form", nullptr)) 
         return false;
 
     if (botAI->HasCheat(BotCheatMask::food))
--- a/src/strategy/actions/NonCombatActions.cpp
+++ b/src/strategy/actions/NonCombatActions.cpp
@@ -10,13 +10,8 @@
 
 bool DrinkAction::Execute(Event event)
 {
-    if (bot->IsInCombat())
-        return false;
-
-    if (bot->IsMounted())
-        return false;
-
-    if (botAI->HasAnyAuraOf(GetTarget(), "dire bear form", "bear form", "cat form", "travel form", "aquatic form","flight form", "swift flight form", nullptr)) 
+    // gatekeeper
+    if (!isPossible())
         return false;
 
     if (botAI->HasCheat(BotCheatMask::food))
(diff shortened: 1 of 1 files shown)


src/strategy/actions/NonCombatActions.cpp
```diff
@@ -13,6 +13,10 @@ bool DrinkAction::Execute(Event event)
     if (bot->IsInCombat())
         return false;
 
+    // Don't drink while mounted
+    if (bot->IsMounted())
+        return false;
```
kadeshar: That should be in isUseful

src/strategy/actions/NonCombatActions.cpp
```diff
@@ -66,6 +70,10 @@ bool EatAction::Execute(Event event)
     if (bot->IsInCombat())
         return false;
 
+    // Don't eat while mounted
+    if (bot->IsMounted())
+        return false;
```
kadeshar: That should be in isUseful

nl-saw:
I think it would need to take shapeshifts into account as well.

src/strategy/actions/NonCombatActions.cpp
hermensbas: You sure, check mana class check agree. But the combat en isMounted check?

hermensbas (approved):

hermensbas:
> I think it would need to take shapeshifts into account as well.

Travelforms i assume?

gacuna89:
I saw in Stormwind a draenei eating on his mount :P

src/strategy/actions/NonCombatActions.cpp
hermensbas: You sure? check mana class check agree placing that into isUsefull function. But the combat en isMounted check?

gacuna89: Yeah, my bad, I messed that up

hermensbas:
I've added druid shapedshift forms, and moved around where the checks are defined and checked.

hermensbas (approved):

src/strategy/actions/NonCombatActions.cpp
```diff
@@ -10,7 +10,8 @@
 
 bool DrinkAction::Execute(Event event)
 {
-    if (bot->IsInCombat())
+    // gatekeeper
+    if (!isPossible())
```
kadeshar: That is checked in Engine::DoNextAction. No need to call it manually.

src/strategy/actions/NonCombatActions.cpp
```diff
@@ -52,21 +53,23 @@ bool DrinkAction::Execute(Event event)
...
 
 bool DrinkAction::isPossible()
 {
-    return !bot->IsInCombat() && (botAI->HasCheat(BotCheatMask::food) || UseItemAction::isPossible());
+    return !bot->IsInCombat() && 
+        !bot->IsMounted() &&
+        AI_VALUE2(bool, "has mana", "self target") &&
+        !botAI->HasAnyAuraOf(GetTarget(), "dire bear form", "bear form", "cat form", "travel form", 
+            "aquatic form","flight form", "swift flight form", nullptr) &&
+        (botAI->HasCheat(BotCheatMask::food) || UseItemAction::isPossible());
```
kadeshar, line 63: isUseful is checked first and have condition 
AI_VALUE2(uint8, "mana", "self target") < 100
isPossible is checked as second and have condition
AI_VALUE2(bool, "has mana", "self target")
This order dont have sense and i mentioned in previous review that this condition should be in isUseful

hermensbas: The intend of:
isUseful() → “do I even care about this?”, most likely and kinda intended for cheap checks
isPossible() → “am I allowed to?”, prolly more detailed and thereby more expensive checks.

The general implementation idea, do first cheap checks then expensive then execute. I kinda agree on AI_VALUE2(bool, "has mana", "self target") makes more sense adding to isUseful() from technical perspective not from intend/functional perspective. For the others combat, mount, shapeshifted, etc really belong in isPossible() from functional and technical perspective.

What do you think @kadeshar?

I've moved ' AI_VALUE2(bool, "has mana", "self target")' to isUseful() and removed the additional gatekeep check.

kadeshar: I only quote more experienced developer (mostlikely from cmangos playerbots):

Ispossible: can we do this at all (physically possible)
Isusefull: is it smart to do this?
Execute: do this (and return if we fail to do it)
Can also look it in order of complexity.
First easy fast checks (trigger)
Then easy check (ispossible)
Then harder (isUseful)
Then full hard checks and execution code (execute)

For me only important thing is that isPossible and isUseful conditions are not exclude each other

hermensbas: You are correct, just double checked the code; ispossible is called before isUseful in the strategy pattern executor

```c++
ActionResult Engine::ExecuteAction(std::string const name, Event event, std::string const qualifier)
{
    bool result = false;

    ActionNode* actionNode = CreateActionNode(name);
    if (!actionNode)
        return ACTION_RESULT_UNKNOWN;

    Action* action = InitializeAction(actionNode);
    if (!action)
    {
        delete actionNode;
        return ACTION_RESULT_UNKNOWN;
    }

    if (!qualifier.empty())
    {
        if (Qualified* q = dynamic_cast<Qualified*>(action))
            q->Qualify(qualifier);
    }

    if (!action->isPossible())
    {
        delete actionNode;
        return ACTION_RESULT_IMPOSSIBLE;
    }

    if (!action->isUseful())
    {
        delete actionNode;
        return ACTION_RESULT_USELESS;
    }

    action->MakeVerbose();

    result = ListenAndExecute(action, event);
    MultiplyAndPush(action->getContinuers(), 0.0f, false, event, "default");

    delete actionNode;

    return result ? ACTION_RESULT_OK : ACTION_RESULT_FAILED;
}
```

I will revaluate the code

hermensbas: Even though the HasAnyAuraOf is abit more expensive placing the check in isPossible makes more sense to me. IsMounted and InCombat are merely 'cheap' HasUnitFlag checks.

hermensbas: If you agree, resolve if not lemme know :)

kadeshar: You have right. I checked DoNextAction instead of ExecuteAction...

PR #1665, opened 2025-09-25
feat: Enhanced AiFactory with improved spec detection

gacuna89:
- Enhanced Druid spec detection with talent verification
- Added proper fallback mechanisms for unclear specs
- Improved Balance Druid detection (tab == 0)
- Enhanced Feral Druid detection with talent point validation
- Added safe fallback to caster strategies for edge cases
- Better handling of multi-spec scenarios
- Fixed operator spacing in conditional statements
- Added complete Shaman spec logic (Elemental, Enhancement, Restoration)
- Added complete Druid spec logic for non-combat strategies
- Improved spec detection for Shaman, Paladin and Priest classes

### DIFF
--- a/src/AiFactory.cpp
+++ b/src/AiFactory.cpp
@@ -353,25 +353,48 @@ void AiFactory::AddDefaultCombatStrategies(Player* player, PlayerbotAI* const fa
 
             break;
         case CLASS_DRUID:
-            if (tab == 0)
+        {
+            std::map<uint8, uint32> tabs = GetPlayerSpecTabs(player);
+            
+            if (tab == 0)  // Spec Balance
             {
                 engine->addStrategiesNoInit("caster", "cure", "caster aoe", "dps assist", nullptr);
                 engine->addStrategy("caster debuff", false);
             }
-            else if (tab == 2)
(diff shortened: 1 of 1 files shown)


src/AiFactory.cpp
```diff
@@ -475,13 +511,20 @@ void AiFactory::AddDefaultCombatStrategies(Player* player, PlayerbotAI* const fa
             }
         }
     }
-    if (sRandomPlayerbotMgr->IsRandomBot(player))
+    // For classes with specific logic, don't override assigned strategies
+    if (player->getClass() != CLASS_DRUID && 
+        player->getClass() != CLASS_SHAMAN && 
+        player->getClass() != CLASS_PRIEST && 
+        player->getClass() != CLASS_PALADIN)
```
kadeshar, line 518: Personally i dont agree that this classes should ignore config

gacuna89: The issue is that, for example, a Balance Druid will always behave as Feral. This is the same behavior we currently see with Elemental Shamans, who are treated as Enhancemen

kadeshar: That dont explain why config file should be ignored for this classes.

PR #1671, opened 2025-09-27
Fix warning: delete called on non-final that has virtual functions but non-virtual destructor

privatecore:
Add virtual destructor for the NamedObjectFactory class + small codestyle corrections based on: https://www.azerothcore.org/wiki/cpp-code-standards


privatecore:
@kadeshar forget to disable personal code format based on azerothcore's code standarts - will keep it in mind for the future.

hermensbas (approved):

PR #1675, opened 2025-09-28
insert of german chatter texts into text_loc3

Raz0r1337:
The original files from the base directory have been overwritten somewhere, and therefore they can be removed from there as well.

If possible revert #961


hermensbas:
We cant just revert #961, you need to add revert PR and solve merge-conflicts or remove it manually in this PR

This is sha related to #961: https://github.com/liyunfan1223/mod-playerbots/commit/d7b3843024479ea74b490a7087720d4a108b4a1b

Raz0r1337:
All right, then I'll get to work on it. 👍

Raz0r1337:
If I do this for German, should I also remove all other languages except English from the base file and prepare update sqls for them? That would clean up the base file a bit.

Raz0r1337:
I extracted all the text from the updated database and translated it freshly again.
Now everything should be correct.

Raz0r1337:
Am I doing something wrong here? The structure is exactly the same as all the other language files, and when I run the SQL file, it executes without errors, but nothing is written to the database.

Raz0r1337:
Sorry, I'm a bit tired; I overlooked that I was inserting the text into the wrong column (I should have used "text_loc3" instead of "text").

It's fixed now.

hermensbas:
this might be sufficient; overwriting base script in combination with german update script should end up in valid state again.

hermensbas:
loc0 → usually English (default client strings).
loc1 → Korean.
loc2 → French.
loc3 → German.
loc4 → Chinese (traditional).
loc5 → Chinese (simplified).
loc6 → Spanish.
loc7 → Mexican Spanish.
loc8 → Russian.

:)

hermensbas (dismissed):

Raz0r1337:
I decided to delete the German localization from the base file because it was a topic of discussion at one point, and it was suggested to me that it makes more sense to add new languages as updates rather than editing the base file. It will be merged at some point, and all updates, including the new languages, will be contained in the base file. It seemed like a good idea, and so I thought that I colud remove my old localization, which I had added to the base file myself.

hermensbas:
agree on that for whats its worth

hermensbas:
Can someone else approve? So we can extend of playerbase with german players? :)

noisiver (dismissed):
I don't speak german and can read only a little but I'll approve this since I trust @hermensbas and it appears to look good code-wise

It0w:
Thanks for the translation :)

I think the following IDs have errors in their variable names.

```
178
187
190
198
201
205
206
207
225
230
```

Some of the translations are a bit strange, but for MTL, considering the goal is to stay as close to the original as possible, they're pretty good.

Raz0r1337:
@It0w
Thx i've overseen that '-' on a placeholder blocks it.
I removed all '-' to make them work again.

I translated all English chatter texts 1-1 and only adjusted minor details that didn't make any sense in German.
I would have liked to have taken a more creative approach, but then the PR would have been rejected for not being close enough to the original.

kadeshar:
Hmmm now i notice that creation script dont guarantees text ids. I made test and run ai_playerbot_texts.sql

Example result:
obraz

Example update:
obraz

I think this text dont fit (i dont know german but based on translator)
obraz

That also applies to other language translations applied to database as update.
Maybe that is only my feeling but script which create table  ai_playerbot_texts should keep proper id's for texts.

Additionally update script SET id. I saw same bug in Wishmaster translation script. 

EDIT:
My table have 1707 records. Update scripts affect 853 ids. That is on purpose?

It0w:
@kadeshar
I checked my SQL table, and it turns out that my IDs don't match either.
Are the IDs assigned randomly when the table is created? I looked through the SQL file (```data/sql/playerbots/base/ai_playerbot_texts.sql```), but I couldn't find any IDs.

If the IDs are different for each instance, it would probably be better to refer to the text field rather than the IDs for match and replace.
Another option is to create the base table with uniform IDs so that they can be reused later. 

My table contains 1716 records.  ```¯\_(ツ)_/¯```

> I translated all English chatter texts 1-1 and only adjusted minor details that didn't make any sense in German. I would have liked to have taken a more creative approach, but then the PR would have been rejected for not being close enough to the original.

I know so good work :)

kadeshar:
About records difference i only applied creation scripts. I didnt reapplied updates which adds new rows. But yea, best that translation contains all rows also this extra one.

Raz0r1337:
Great, so that means all localization updates that are already in there and those that are still pending as PRs are useless because nothing is sorted correctly except for the ones that were created because the IDs match there.

I don't think it's possible to sort it using the original English text, because there are a few duplicates in the texts that make it impossible to uniquely identify a line, so SQL will just spit out errors.

The most sensible thing to do would be to assign unique IDs to the entries in the base file and then have the PR owners realign the translations with them.

Raz0r1337:
So my ai_playerbot_texts database has exactly 858 lines, no more, no less.
I had everything popularized from the core and all updates installed. At last i imported my translation SQL to check if the bots are using this. I don't know where the over 1700 lines come from.
Screenshot 2025-10-06 212628

Or do you mean the basefile in the source code?
There are still many comments and further instructions for the db in there, but they don't play a role in the translations.

Raz0r1337:
The strange thing here is that I deleted my database several times for testing purposes and had it refilled from the server, and my translations are exactly right. So at least I haven't had any ID shifts.

But I haven't yet adapted the base file to the one from my PR; I only tested it briefly once and saw that everything seemed to fit. Could it be that I made a mistake somewhere and that's why nothing fits anymore?

If you have only tested the PR with my base file so far, please try it with the original, which still contains my old German translation. Download all updates and then my German translation. If it fits, then we have found the culprit and I will delete the base file from my PR.

I will test the same thing here.

kadeshar:
Odd because based on history that file always had so much records.
Re-create table made same amount of records
obraz

Raz0r1337:
And on your screenshot the IDs and the translation are matching now. Very weird. 😂

kadeshar:
Matching?
For me dummy_end = 1707. For you 849

Raz0r1337:
I've read the entrys that are shown on your screenshot, the English texts are matching the Germans now.

Raz0r1337:
How can that be?
My translations go up to ID 853, your database goes up to over 1700, and on your screenshot where you can see entry 1700, my German strings are there and they even fit.

kadeshar:
Yeap but that is existing script which you cleaning up in this PR

Raz0r1337:
Oh well, that explains it.
In my case, however, if I let the core install all updates, the German translation is then deleted.

That's why I created the PR in the first place.
And I am using the same Basefile. 🤔

Raz0r1337:
I had a look in the update folder and looked at the two translations that have already been installed. All translations work with the IDs as mapping and they all have about 850 lines. For my database, I loaded both the original and the base file from my pr and loaded the core and had the DB filled in two cloned VMs. I always end up with 853 lines and all translations fit perfectly as far as I can see. My test VM is just the current PB fork of AzerothCore and mod-playerbots.

I'm not deep enough in the matter to be able to say more precisely. Do you have any idea what's going wrong here?

kadeshar:
I see reason 
obraz

900 lines was applied to base file but update script wasnt created. Databases created before this commit have less records.

We need to:

- modify base file to include id`s in INSERT statement
obraz

- create update script containing whole base file

- personally im think that german translation should be included in base and update script

If you will take care that let me know. If not, i will create issue or fix that myself.

Raz0r1337:
Thank you for looking at this.

I can insert the ids, but I don't think I can adapt my translation so that the IDs fit. And I don't think I can integrate my translation into the base file that the IDs fit.

noisiver:
Making a copy of `ai_playerbot_texts.sql` and placing it in the updates folder is not something I think should be done. The updates are for updates, not replacements. People who have modified the table will suffer for it. The update should include only the new entries using delete/insert statements. Adding the same lines to the base file I think is ok.

noisiver:
Sorry, I mean update queries just like you have done. I think that is the right way of doing it and you have done so. I didn't look closely enough before making the last comment. I don't think the update file should contain the entire text file that is in the base, which it looks like @kadeshar is suggesting.

kadeshar:
I agree with you @noisiver but this table in current shape is not updateable by id's. Im almost made fix for that only i have to move old comment from base file. My fix reapply whole table and i dont see other way than that. After that fix other changes should be applied via update script with id (both in INSERT and UPDATE)

noisiver:
I see. Yeah, that is a tough one. I don't think that should be done in this PR though. It should be a dedicated PR to restructure the texts and this PR should be what it is but of course either merged before that restructure or wait until the restructure is merged and then update this PR. I believe option one is the easiest.

hermensbas:
> I agree with you @noisiver but this table in current shape is not updateable by id's. Im almost made fix for that only i have to move old comment from base file. My fix reapply whole table and i dont see other way than that. After that fix other changes should be applied via update script with id (both in INSERT and UPDATE)

That should be fine, taking in regard these are statics values to begin with. So having a complete base file and taking updates from there sounds more logical then update doing more then just a little update.

kadeshar:
I made PR. Finally i made only update scripts which reapply whole table. I included all addiotional translations and texts

kadeshar:
@Raz0r1337 could you check that now some german translation is required? If yes then fix that PR if not close.

Raz0r1337:
Of course, when I have time again in the next few days, I'll get to it.

Thank you for doing that. 👍

Raz0r1337:
Unfortunately, my old translation was taken from the base file.
It contained some logical and spelling errors, which I found over the last months and were corrected with the PR.
I will adjust and update it over the next few days.

Thanks again for doing the work. ;-)

Raz0r1337:
I am currently correcting and completing the German localisation of the chatter texts.

I have a question about placeholders such as %item_link. I have already noticed that a - directly in front or back of the placeholder ensures that it is not replaced but written out in full.

My question is: does this apply to all symbols that appear directly before or after a placeholder? For example: %item_link, is BiS for hunters or an ? after the placeholders.

Raz0r1337:
@kadeshar

I took another week and extracted all the original English strings again and had them machine-translated into German. Then I went through every single line several times and adjusted everything so that it really makes sense for us Germans.

Now everything is fine with the localization.

Raz0r1337:
This was just a rename, i couldnt find any option on the github website to rename the file to fix a typo in the name.

icemansparks:
First of all, thank you for all the effort that has been put into this. 
I've gone through all of the translations as well now, and could not find any unfitting translations or wording (in case another German proofreader was needed) ;-)

Raz0r1337:
I am happy that I can contribute to this wonderful project in this way.
Thank you for your kind words.😊

And thank you for reviewing it. Additional reviewers and proofreaders are always welcome. 👍

kadeshar:
Thanks for effort :) Now german language should be great. Restore default base file. Any modification of this file doesnt have sense because is entire override in update script.
Also update script update id i dont know why. Update should look like 
```
UPDATE `ai_playerbot_texts` SET `text_loc3`='am Arsch der Welt' WHERE `id`=1;
```

Raz0r1337:
Thank you.
That's a remnant from the export that the IDs are in there. I overlooked that.
I'll correct the error right away.

And thank you for your help with your fix. 👍

Raz0r1337:
@kadeshar 
Now everything should be fine. ;-)

kadeshar:
> @kadeshar Now everything should be fine. ;-)

Base file still have changes.

kadeshar:
Update script should contain only rows which are real updated.

icemansparks:
> Update script should contain only rows which are real updated.

But since some of the german entries had been added to the base file some time ago (therefore they're missing in existing installations from before that) the proper solution now would be to all have them in the update file - wouldn't it?

kadeshar:
> > Update script should contain only rows which are real updated.
> 
> But since some of the german entries had been added to the base file some time ago (therefore they're missing in existing installations from before that) the proper solution now would be to all have them in the update file - wouldn't it?

Not if 2025_10_09_00_ai_playerbot_texts_fix.sql fixed that ealier.

Raz0r1337:
Again, just to clarify why I find it ok to replace all in the first time.

I had all the English texts from the DB that I had previously updated from the core fully updated. I first had these texts translated automatically with a script using Deepl. Then I went through each line twice manually and corrected everything that didn't fit in German and wasn't well formulated until it sounded good and made sense. If lines correspond to the text in the base file, it's more or less a coincidence. Since this is supposed to be a complete localisation, it makes sense to translate all lines once, especially since there was no German in the original base file and I had added it there. However, that was a purely machine-generated translation from chatGPT3 and these lines are in your text fix, and now it is a localisation made fully by humans. Any further changes, if anything else needs to be fixed, I will only change the respective line, but as an initial translation, this makes sense.

@kadeshar , can you please show me which line(s) you found that differs from the base file? I'll take another look at it today and that would speed up the process.

Raz0r1337:
@kadeshar 

I have now compared the current database imported from the core with all updates with my localisation.
I could not find any places where anything is missing or needs to be fixed.
If you mean that my translation does not match the base file, then that is correct, but it is also exactly as intended.

My localisation is based on your text fix, which is essentially a new base file that is applied over the old original as an update.
And my localisation fits your text fix.

Raz0r1337:
I removed 179 lines that were identical to the text-fix-update.
It should now be mergeable.

@kadeshar @noisiver 
Can you plz review it again if you have the time?

kadeshar:
PR still contains changes in ai_playerbot_texts.sql. This file should not be touched because whole its content is override by update script.

Raz0r1337:
Ok, i reuploaded the original file ai_playerbot_texts.sql.
So now everything is correct?

kadeshar:
> Ok, i reuploaded the original file ai_playerbot_texts.sql. So now everything is correct?

Yeap, i think is ok. Still someone else need to look on that.

nl-saw:
I've verified that the SQL applies as intended.

nl-saw (approved):

PR #1678, opened 2025-09-28
Added method to get translation bot text or default

kadeshar:


hermensbas:
I kinda expected such a function already existed, good call.

PR #1681, opened 2025-09-28
Config option to set max number of guild members in random bot guilds

IainD92:
Quick one to add a config option to set the max size for randombot guilds (currently hardcoded to 15 in master)

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -159,6 +159,9 @@ AiPlayerbot.RandomBotGuildNearby = 0
 # Number of guilds created by randombots
 AiPlayerbot.RandomBotGuildCount = 20
 
+# Maximum number of members in randombot guilds (minimum is hardcoded to 10)
+AiPlayerbot.RandomBotGuildSizeMax = 15
+
 # Delete all randombot guilds if set to 1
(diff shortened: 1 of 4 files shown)


brighton-chi:
I assume whatever you set the config to will result in all randombot guilds having that amount of members, correct? I don't know if that's upon creation or later (though I had thought currently all bots are added to a guild upon creation).

IainD92:
> I assume whatever you set the config to will result in all randombot guilds having that amount of members, correct? I don't know if that's upon creation or later (though I had thought currently all bots are added to a guild upon creation).

As far as I can tell through testing (and as the config comments for RandomBotGuildNearby suggest) bots aren't currently inviting other bots to guilds (although that may be from the guilds getting filled very quickly).

This branch changes the hardcoded limit in InitGuild(), which is called from PlayerbotFactory::Randomize, (if RandomBotGuildCount > 0). So on creation and when bots are manually initialised I think. There may be other situations where Randomise is called though….

Still familiarising myself with the project structure, but keen to get the city idling experience feeling more blizz-like and guilds are a big part of that so I’m planning to hunt down, become familiar with, and expand on guild logic to have them slowly grow over time by recruiting in trade chat…. hopefully I’ll have a definitive answer that’s intuitive to configure soon!

PR #1684, opened 2025-09-29
Fix spacing issue and removed unnecessary spaces

noisiver:
We use 4 spaces. Not 2 and not tabs.

### DIFF
--- a/src/PlayerbotTextMgr.cpp
+++ b/src/PlayerbotTextMgr.cpp
@@ -102,7 +102,7 @@ std::string PlayerbotTextMgr::GetBotText(std::string name, std::map<std::string,
 }
 
 std::string PlayerbotTextMgr::GetBotTextOrDefault(std::string name, std::string defaultText,
-std::map<std::string, std::string> placeholders)
+    std::map<std::string, std::string> placeholders)
 {
     std::string botText = GetBotText(name, placeholders);
     if (botText.empty())
--- a/src/strategy/actions/GenericBuffUtils.cpp
+++ b/src/strategy/actions/GenericBuffUtils.cpp
@@ -79,11 +79,11 @@ namespace ai::buff
     }
 
     std::string UpgradeToGroupIfAppropriate(
-    Player* bot,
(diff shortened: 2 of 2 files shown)


src/PlayerbotTextMgr.cpp
```diff
@@ -102,7 +102,7 @@ std::string PlayerbotTextMgr::GetBotText(std::string name, std::map<std::string,
 }
 
 std::string PlayerbotTextMgr::GetBotTextOrDefault(std::string name, std::string defaultText,
-    std::map<std::string, std::string> placeholders)
+std::map<std::string, std::string> placeholders)
```
privatecore: It's a bit hard to read the function/method definitions in this format, don't you think? I'm not sure which code style guide the maintainers are using, but one of the most common is the Google C++ Style Guide: https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions

src/strategy/actions/GenericBuffUtils.cpp
```diff
@@ -21,122 +21,125 @@
...
+        }
+        return false;
+    }
+
+    std::string UpgradeToGroupIfAppropriate(
+    Player* bot,
+    PlayerbotAI* botAI,
+    std::string const& baseName,
+    bool announceOnMissing,
+    std::function<void(std::string const&)> announce)
```
privatecore: Same here...

kadeshar:
I think to be perfect parameters should be moved 4 spaces to the right (comparing to other files in module)

noisiver:
Good call. I wasn't looking too clearly, I just did this not long ago so I wasn't expecting the same mistake to occur already.

hermensbas:
We can do an suto format the entire code when the PR list is low in future?

noisiver:
> We can do an suto format the entire code when the PR list is low in future?

I'll limit it to every month, so at the start of each month I'll go through and do this. That way it shouldn't constantly cause trouble for PRs.

Issue #1172, opened 2025-04-06
Keeping getting SQL error - fresh server


nugsforthugs:
**Bug Description**
Everytime it gets to the spawning bot section for the first time to randomize gear / location, it keeps spamming in console

"SQL(p): DELETE FROM playerbots_custom_strategy WHERE name = 6 AND owner = false AND idx = false
 [ERROR]: [1292] Truncated incorrect DOUBLE value: 'say'
Unhandled MySQL errno 1292. Unexpected behaviour possible."

Bots will spawn in world, but won't be moved around to different locations, they all stay in the starting zones.

**Commit Hash**
Lastest

**How To Reproduce Bug**
Just load up a fresh server in my case.

**Expected Behavior**

**Screenshots**

**Modules**
mod-playerbots
mod-learnspells

**Playerbot Settings**
1000 bots, 1 player

**System**
Windows 10 - Beelink mini PC

**Additional Context**
None.

NoxMax:
A thousand bots might be a bit too much for your mini pc, when they are all in the starting zones. So they get stuck.
If you are starting your bots at level 1, you probably need to drop the number of bots until your PC can manage them. If the bots are level randomized. Try running the command `playerbot rndbot teleport` in the server console. This will teleport the bots to areas suitable for their levels, and once they are distributed around, your PC might handle them better.

kadeshar:
That seems wrong
![Image](https://github.com/user-attachments/assets/1ae7cb61-9c83-4f36-8252-4a3df13231ee)

evrydayjunglist:
It appears that PLAYERBOTS_DEL_CUSTOM_STRATEGY prepared statement is expecting 3 parameters but it's only getting one from that function. 

To me it seems like the prepared statement is not right for this function. I tried fixing it locally by adding a new prepared statement to the core, then using that in this function. Nothing has blown up so far but I have no idea whether this is the right approach. Here it is anyways...

Playerbots AzerothCore - 
src/server/database/Database/Implementation/PlayerbotsDatabase.cpp
Added to line 34:
PrepareStatement(PLAYERBOTS_DEL_CUSTOM_STRATEGY_BY_OWNER, "DELETE FROM playerbots_custom_strategy WHERE owner = ?", CONNECTION_ASYNC);

mod-playerbots -
src/PlayerbotDbStore.cpp
Modified lines 85, 86, 87
[...]

PR #1688, opened 2025-09-30
Fix wrong prepared statement used for the PlayerbotDbStore::Reset issue #1172

privatecore:
PlayerbotDbStore has nothing to do with playerbots' custom strategies. Use the properly prepared statement PLAYERBOTS_DEL_DB_STORE.


Tierisch:
Is this the whole reason those entries weren't getting cleared out? Such a simple fix.

PR #1693, opened 2025-10-02
Maintenance config for altbots

IainD92:
Adds some boolean options to configure how the maintenance command affects altBots.

Stops players from skipping tradeskills, buying spells, working for bags, crafting and trading for glyphs, potions, consumables, etc

All options disabled by default, enable them to make life easier.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -489,21 +489,21 @@ AiPlayerbot.RPWarningCooldown = 30
 AiPlayerbot.MaintenanceCommand = 1
 
 # Enable/Disable specific maintenance command functionality for alt bots
-# Disable to prevent players from giving free bags, spells, skill levels, etc to their Altbots
+# Disable to prevent players from giving free bags, spells, skill levels etc to their alt bots
 # Default: 1 (enabled)
 AiPlayerbot.AltMaintenanceAmmo = 1
 AiPlayerbot.AltMaintenanceFood = 1
+AiPlayerbot.AltMaintenanceReagents = 1
 AiPlayerbot.AltMaintenanceConsumables = 1
 AiPlayerbot.AltMaintenancePotions = 1
-AiPlayerbot.AltMaintenanceReagents = 1
 
 AiPlayerbot.AltMaintenanceBags = 1
 AiPlayerbot.AltMaintenanceMounts = 1
+AiPlayerbot.AltMaintenanceSkills = 1
 
 AiPlayerbot.AltMaintenanceClassSpells = 1	# Spells from quests (tame/summon pets, totems, druid forms, etc)
 AiPlayerbot.AltMaintenanceAvailableSpells = 1	# Spells learnable from trainer
 AiPlayerbot.AltMaintenanceSpecialSpells = 1	# Leave DK starting area
-AiPlayerbot.AltMaintenanceSkills = 1
 AiPlayerbot.AltMaintenanceTalentTree = 1
 AiPlayerbot.AltMaintenanceGlyphs = 1
 AiPlayerbot.AltMaintenanceGemsEnchants = 1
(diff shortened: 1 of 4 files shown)


privatecore:
Is it really necessary to separate those maintenance options for the alt bots?

IainD92:
> Is it really necessary to separate those maintenance options for the alt bots?

A couple of people on discord mentioned not liking all the maintenance options on altbots for cheat/balance reasons. I had this branch lying around for personal use (I like food and ammo on to save on some micromanagement, class spells on to save lots of running back and forth in the first few levels, and everything else disabled to encourage crafting/trading and to not let players get free skill levels/spells/mounts/24 slot bags/enchants etc)


Left everything individually configurable so people can dial in exactly which features they want on/off for personal taste and vision of what is balanced vs what is worth being able to cheat on to avoid a certain level of tedium when managing a large number of characters.

privatecore:
I see now. Maybe leave all these settings enabled by default to maintain the old behavior and allow others who need to customize it?

IainD92:
> I see now. Maybe leave all these settings enabled by default to maintain the old behavior and allow others who need to customize it?

Good point! I’ll get that changed.

conf/playerbots.conf.dist
```diff
@@ -488,6 +488,29 @@ AiPlayerbot.RPWarningCooldown = 30
 # Default: 1 (enabled)
 AiPlayerbot.MaintenanceCommand = 1
 
+# Enable/Disable specific maintenance command functionality for alt bots
+# prevent players from getting free spells, bags, mounts, etc
+# Default: 0 (disabled)
```
privatecore: This also needs to be enabled by default if you want to preserve the old behavior ;)

IainD92: Whoops, updated my working server's copy of the conf instead of the one in this branch! xD

Dreathean:
Very nice! I know a lot of people will love having these options. Thoughts on changing 
image
to "Disable to prevent players from giving free bags, spells, skill levels, etc to their Altbots"?

IainD92:
> Very nice! I know a lot of people will love having these options. Thoughts on changing image to "Disable to prevent players from giving free bags, spells, skill levels, etc to their Altbots"?

Absolutely! Good catch

IainD92:
Made the suggested change to conf, and rearranged the order the options are presented (tried to group them by vibe/the type of gameplay they impact)

nl-saw:
Please make sure the order of the called factory.  is identical as the original, as it can cause issues otherwise.

IainD92:
> Please make sure the order of the called factory. is identical as the original, as it can cause issues otherwise.

No problem

hermensbas:
Set to draft, change to ready for review when its ready to be reviewed.

conf/playerbots.conf.dist
```diff
@@ -488,6 +488,34 @@ AiPlayerbot.RPWarningCooldown = 30
...
+AiPlayerbot.AltMaintenanceConsumables = 1
+AiPlayerbot.AltMaintenancePotions = 1
+
+AiPlayerbot.AltMaintenanceBags = 1
+AiPlayerbot.AltMaintenanceMounts = 1
+AiPlayerbot.AltMaintenanceSkills = 1
+
+AiPlayerbot.AltMaintenanceClassSpells = 1	# Spells from quests (tame/summon pets, totems, druid forms, etc)
+AiPlayerbot.AltMaintenanceAvailableSpells = 1	# Spells learnable from trainer
+AiPlayerbot.AltMaintenanceSpecialSpells = 1	# Leave DK starting area
```
nl-saw, line 506: Actually the description is wrong. Specialspells are defined in the config as well.

IainD92: Ah I see, that's very useful.
These descriptions have been removed in a more recent pull anyway - putting comments on the end of lines was causing the values to not be read correctly and reverting to default.

nl-saw: Ah had not noticed that, gladly then :+1:

PR #1691, opened 2025-10-01
Random bot guild totals fix

IainD92:
### **Issue:**
#1667 
-Bot guilds would keep being created on server restarts, eventually running out of guild names and hitting an infinite loop.

Suspect this was from the logic where random bot guilds would be tallied by running through a list of randomBots and checking if that bot is a guild leader. I'm not familiar enough with the project to know for sure, but I guessing that the list of randomBots only includes bots that are currently active as online players.

### **Fix:**
-begin the random bot guild check/creation process by querying the guilds table in acore_characters, then determining which ones are bot guilds. This gives us an absolute total of randomBot guilds, regardless of whether or not the leaders are active.

### **Testing:**
I've tested this with 200 guilds. Numerous bot inits, and numerous server restarts both with and without AiPlayerbot.DisabledWithoutRealPlayer = 1.
guild table in acore_characters seems to reach the config defined number of guilds and not randomly grow anymore.

### **Todo:**
This needs some formatting, and potential removal of an unnecessary std::find().
I'll tidy it up at some point over the next couple of days.

Could potentially add a system to abort guild creation, when the total number of guilds reaches the number of available names in the playerbots guild names table.

PR #1699, opened 2025-10-03
Random bot guild totals fix

IainD92:
Republish of #1691

Messed up my fork, started with a fresh one. Had to republish this (got automatically closed when old fork was deleted)

### DIFF
--- a/src/RandomPlayerbotFactory.cpp
+++ b/src/RandomPlayerbotFactory.cpp
@@ -902,25 +902,14 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
         } while (guildTableResults->NextRow());
     }
 
+    // Get a list of bots that are logged in and available to lead new guilds
     GuidVector availableLeaders;
     for (std::vector<uint32>::iterator i = randomBots.begin(); i != randomBots.end(); ++i)
     {
         ObjectGuid leader = ObjectGuid::Create<HighGuid::Player>(*i);
         if (Guild* guild = sGuildMgr->GetGuildByLeader(leader))
         {
-            uint32 guildID = guild->GetId();
-            if(std::find(sPlayerbotAIConfig->randomBotGuilds.begin(), sPlayerbotAIConfig->randomBotGuilds.end(), guildID) !=sPlayerbotAIConfig->randomBotGuilds.end())
-            {
-                // this randombot guild has already been counted, the leader is not availble
-            }
-            else
-            {
-                // somehow we missed this guild when checking the guild table
-                // THIS SHOULDN'T RUN.
-                // THIS ENTIRE FIND() IS A WASTE OF RESOURCES (in theory)
-                ++guildNumber;
-                sPlayerbotAIConfig->randomBotGuilds.push_back(guildID);
-            }
+            // Bot is already a GM
         }
         else
         {
@@ -929,6 +918,7 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
                 availableLeaders.push_back(leader);
         }
     }
+    LOG_INFO("playerbots", "randomBotGuilds - {} available leaders for new guilds found", availableLeaders.size());
 
     // Create up to randomBotGuildCount by counting only EFFECTIVE creations
     uint32 createdThisRun = 0;
@@ -936,10 +926,10 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
     {
         std::string const guildName = CreateRandomGuildName();
         if (guildName.empty())
-            continue;
+            break; // no more names available in playerbots_guild_names
 
         if (sGuildMgr->GetGuildByName(guildName))
-            continue;
+            continue; // name already taken, skip
 
         if (availableLeaders.empty())
         {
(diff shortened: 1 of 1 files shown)


IainD92:
I mentioned making a random name generator for when the table of random guild names runs out, but I'll do that as a separate PR.

For now, this does the job of stopping the server from continuing to create random bot guilds after reaching the desired number set in the config, and prevents the server from entering an infinite loop once it uses up all the names in the table.

Zerathane:
I've pulled and tested this PR which is a fix to the current build issue of the world server being stuck in an infinite loop when the maximum amount of guild names have been used from the playerbots_guild_names table. I am using the default amount of 400 names. 

First, on the most recent branch, I replicated the current issue that is present. I increased my bot count to 3000 from the default 500 and incrementally increased "AiPlayerbot.RandomBotGuildCount" from the default 20, all the way to 500. I ran the world server multiple times to allow for the creation of the guilds. Once the limit had been reached, this following image shows the error.

worldserverrndguild

I pulled PR #1699 onto a fresh build and followed the testing steps as detailed by the author. I cleared all current guilds and bots and reinitialized 5000 bots onto the server. Then in steps I increased my ""AiPlayerbot.RandomBotGuildCount" to 200, and then 500.  When restarting the world server, we're now presented with the (expected) warning of no more names left for random guilds. However the server does not get stuck into a loop and carries on logging in bots.

pr1699guildnamefix

src/RandomPlayerbotFactory.cpp
```diff
@@ -875,34 +875,78 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
...
+                    sPlayerbotAIConfig->randomBotGuilds.push_back(guildId);
+                }
+            } while (res->NextRow());
+        }
+    }
+
+    LOG_INFO("playerbots", "{}/{} random bot guilds exist in guild table",guildNumber, sPlayerbotAIConfig->randomBotGuildCount);
+    if (guildNumber >= sPlayerbotAIConfig->randomBotGuildCount)
+    {
+        LOG_INFO("playerbots", "No new random guilds required");
```
hermensbas: 914: I would prefer debug level

src/RandomPlayerbotFactory.cpp
```diff
@@ -875,34 +875,78 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
...
-            if (player && !player->GetGuildId())
-                availableLeaders.push_back(leader);
+            if (Player* player = ObjectAccessor::FindPlayer(leader))
+            {
+                if (!player->GetGuildId())
+                    availableLeaders.push_back(leader);
+            }
         }
     }
+    LOG_INFO("playerbots", "{} available leaders for new guilds found", availableLeaders.size());
```
hermensbas: 938: I would prefer debug level

src/RandomPlayerbotFactory.cpp
IainD92: Done!

src/RandomPlayerbotFactory.cpp
IainD92: Also, done!

src/RandomPlayerbotFactory.cpp
```diff
@@ -875,34 +875,78 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
...
+            continue;
         }
         else
         {
-            Player* player = ObjectAccessor::FindPlayer(leader);
-            if (player && !player->GetGuildId())
-                availableLeaders.push_back(leader);
+            if (Player* player = ObjectAccessor::FindPlayer(leader))
+            {
+                if (!player->GetGuildId())
```
kadeshar, line 933: I think still that can be one if

IainD92: Like this?
`if (!ObjectAccessor::FindPlayer(leader)->GetGuildId())`

I thought that might return a false positive if FindPlayer(leader) fails to find the specified bot .

What have I missed? Happy to learn!

kadeshar: Nah, my bad. Waiting for some tests before merging

hermensbas (approved):

src/RandomPlayerbotFactory.cpp
```diff
@@ -875,34 +875,78 @@ void RandomPlayerbotFactory::CreateRandomGuilds()
...
+                "SELECT g.guildid, c.account "
+                "FROM guild g JOIN characters c ON g.leaderguid = c.guid"))
+        {
+            do
+            {
+                Field* f = res->Fetch();
+                const uint32 guildId   = f[0].Get<uint32>();
+                const uint32 accountId = f[1].Get<uint32>();
+
+                // Boss considered 'bot' if his account is in botAccounts
```
nl-saw: This comment is very vague.

Thanks for the contribution!

IainD92: Thanks for reviewing.
I’ve updated the comment in the latest commit

kadeshar (approved):

PR #1701, opened 2025-10-03
Update method in QueryItemUsageForEquip

kadeshar:


Zerathane:
This pull request appears to fix the issue of Shamans continuously trying to equip an idol after looting.

This image is prior to the PR. Both characters are Shamans and have various idols in their inventory which I added.

shamanidol

The following image shows the Shaman looting items but not attempting to equip an Idol. Also, the Shaman will no longer /roll on an idol thinking its an upgrade.

shamanidolfixed

kadeshar:
@hermensbas can you take a look?

PR #1702, opened 2025-10-03
Fixes Issue #1695

### LINKED ISSUE #1695: Worldserver.exe crash due to ACCESS_VIOLATION
**Bug Description**
3  crashes now due to ACCESS_VIOLATION. All time have been on empty player server. 4 real players on server total. All either use rndbot or altbot parties. 

**Commit Hash**
9d82a3567

**How To Reproduce Bug**
Run worldserver.exe for 14+ hours

**Expected Behavior**
No crash

**Screenshots**
N/A

**Modules**
mod-ghost-speed
mod-quick-respawn
mod-flightmaster-whistle
mod-dungeonmaster
mod-no-hearthstone-cooldown
mod-ah-bot
mod-autobalance
mod-rdf-expansion
mod-player-bot-level-brackets
mod-playerbots

**Playerbot Settings**
500 bots
AiPlayerbot.botActiveAloneSmartScale = 1
AiPlayerbot.botActiveAloneSmartScaleWhenMinLevel = 1
AiPlayerbot.botActiveAloneSmartScaleWhenMaxLevel = 80

**System**
OS: Windows 10 Professional (Version 10.0, Build 19045)
Hardware:
Processor: Intel(R) Core(TM) i5-7600 CPU @ 3.50GHz
Number Of Processors: 4
Physical Memory: 33493436 KB (Available: 23496860 KB)
Commit Charge Limit: 38474172 KB

Wishmaster117:
Fix: Guard RemoveFromGroup() calls to prevent crashes

This PR adds lightweight guards around all Player::RemoveFromGroup() calls in the module (e.g., LeaveGroupAction, RandomPlayerbotMgr, PlayerbotMgr).
Before removing a bot from a group, we now verify:

Group* g = X->GetGroup() is not null, and

g->IsMember(X->GetGUID()) is true.

I see  issue #1695 reporting frequent crashes when bots leave groups.
In servers with many bots, group membership can change between decision and execution (or be touched by other threads). Calling RemoveFromGroup() on a non-member or a stale group can corrupt internal group state and lead to crashes in the core .

No behavior change when the bot is actually in a valid group.
If the bot is already not a member (or has no group), we skip the removal safely.

Fixes #1695.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -38,6 +38,7 @@
 #include "WorldSessionMgr.h"
 #include "DatabaseEnv.h"        // Added for gender choice
 #include <algorithm>            // Added for gender choice
+#include "Group.h"
 
 class BotInitGuard
 {
@@ -512,10 +513,16 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
             }
         }
 
-        if (!groupValid)
+    if (!groupValid)
+    {
+        if (Group* g = bot->GetGroup())
         {
-            bot->RemoveFromGroup();
+            if (g->IsMember(bot->GetGUID()))
+            {
+                bot->RemoveFromGroup();
+            }
         }
+    }
     }
 
     group = bot->GetGroup();
--- a/src/RandomPlayerbotMgr.cpp
+++ b/src/RandomPlayerbotMgr.cpp
@@ -53,6 +53,7 @@
 #include "Unit.h"
 #include "UpdateTime.h"
 #include "World.h"
+#include "Group.h"
 
 struct GuidClassRaceInfo
 {
@@ -1636,8 +1637,17 @@ bool RandomPlayerbotMgr::ProcessBot(Player* player)
     Group* group = player->GetGroup();
     if (group && !group->isLFGGroup() && IsRandomBot(group->GetLeader()))
     {
-        player->RemoveFromGroup();
-        LOG_INFO("playerbots", "Bot {} remove from group since leader is random bot.", player->GetName().c_str());
+        if (group->IsMember(player->GetGUID()))
+        {
+            LOG_DEBUG("playerbots", "RandomPlayerbotMgr: {} leaving group (leader is random bot, members before: {}).",
+                     player->GetName().c_str(), uint32(group->GetMembersCount()));
+            player->RemoveFromGroup();
+            LOG_DEBUG("playerbots", "RandomPlayerbotMgr: {} left group successfully.", player->GetName().c_str());
+        }
+        else
+        {
+            LOG_DEBUG("playerbots", "RandomPlayerbotMgr: {} not a member anymore, skip removal.", player->GetName().c_str());
+        }
     }
 
     // only randomize and teleport idle bots
@@ -2540,8 +2550,13 @@ void RandomPlayerbotMgr::RandomizeFirst(Player* bot)
     if (GET_PLAYERBOT_AI(bot))
         GET_PLAYERBOT_AI(bot)->Reset(true);
 
-    if (bot->GetGroup())
-        bot->RemoveFromGroup();
+    if (Group* g = bot->GetGroup())
+    {
+        if (g->IsMember(bot->GetGUID()))
+        {
+            bot->RemoveFromGroup();
+        }
+    }
 
     if (pmo)
         pmo->finish();
@@ -2581,8 +2596,13 @@ void RandomPlayerbotMgr::RandomizeMin(Player* bot)
     if (GET_PLAYERBOT_AI(bot))
         GET_PLAYERBOT_AI(bot)->Reset(true);
 
-    if (bot->GetGroup())
-        bot->RemoveFromGroup();
+    if (Group* g = bot->GetGroup())
+    {
+        if (g->IsMember(bot->GetGUID()))
+        {
+            bot->RemoveFromGroup();
+        }
+    }
 
     if (pmo)
         pmo->finish();
@@ -2663,8 +2683,13 @@ void RandomPlayerbotMgr::Refresh(Player* bot)
     uint32 money = bot->GetMoney();
     bot->SetMoney(money + 500 * sqrt(urand(1, bot->GetLevel() * 5)));
 
-    if (bot->GetGroup())
-        bot->RemoveFromGroup();
+    if (Group* g = bot->GetGroup())
+    {
+        if (g->IsMember(bot->GetGUID()))
+        {
+            bot->RemoveFromGroup();
+        }
+    }
 
     if (pmo)
         pmo->finish();
--- a/src/strategy/actions/LeaveGroupAction.cpp
+++ b/src/strategy/actions/LeaveGroupAction.cpp
@@ -8,6 +8,7 @@
 #include "Event.h"
 #include "PlayerbotAIConfig.h"
 #include "Playerbots.h"
+#include "Group.h"
 
 bool LeaveGroupAction::Execute(Event event)
 {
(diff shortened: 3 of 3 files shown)


noisiver:
This needs to be tested by the users who experience these crashes. A "possible fix" shouldn't be merged if it doesn't actually fix the issue so converted to draft until it's confirmed.

Wishmaster117:
> This needs to be tested by the users who experience these crashes. A "possible fix" shouldn't be merged if it doesn't actually fix the issue so converted to draft until it's confirmed.

I tested it but don't have a big server to test but il solve the problem,  so if the review it's ok this can be merged like that.  Waiting for @kadeshar opinion.

noisiver:
It has to be tested regardless

Celandriel:
It's not clear to me how this resolves the issue referenced in the crash. That error points to the IsRandomBot accessing invalid memory

Wishmaster117:
> It's not clear to me how this resolves the issue referenced in the crash. That error points to the IsRandomBot accessing invalid memory

This is why:

Server crash because bots leave groups via LeaveGroupAction, which triggers Player::RemoveFromGroup while another thread or path is already touching the same group/member. Result: corrupted member list / invalid iterator > crash in Group::_getMemberWSlot (03/10 01:44), or corrupted CharacterCache singleton set to NULL > crash in CharacterCache::UpdateCharacterGroup (03/10 01:42). The log from 09/29 is consistent with the same pattern.

Crash on 03/10/2025 01:44
Exception: ACCESS_VIOLATION in Group::_getMemberWSlot (Group.cpp: 2510) during a Group::RemoveMember, itself called by Player::RemoveFromGroup, following LeaveGroupAction::Leave (playerbots).
Main stack: LeaveGroupAction::Leave > PartyCommandAction::Execute > Engine::ListenAndExecute > Player::RemoveFromGroup > Group::RemoveMember > Group::_getMemberWSlot (crash).

RemoveMember's local variables show inconsistent values ​​(unreadable guid/name/flags), typical of an invalid iterator / corrupted member slot.

Conclusion: during a bot's group exit, the internal member list is already invalid (member already removed, double deletion, or thread race), and _getMemberWSlot dereferences something invalid = crash.

Crash on 03/10/2025 01:42
Exception: ACCESS_VIOLATION in CharacterCache::UpdateCharacterGroup (CharacterCache.cpp: 244) called by Group::Disband, following Group::RemoveMember < Player::RemoveFromGroup < LeaveGroupAction::Leave (playerbots).

The local report indicates CharacterCache* this = NULL at the time of the call—very abnormal for a singleton—which points to trampled memory rather than a simple uninitialized pointer.

Conclusion: Same "bot group exit" scenario, but this time the corruption occurs during the character cache update at group disband > call via null instance > crash.

noisiver:
Did the AI/LLM tell you that?

kadeshar:
I analyzed crash and i think problem is this PR https://github.com/liyunfan1223/mod-playerbots/pull/1499 which was merged 2 days ago

EDIT: 
Or not...

Wishmaster117:
> Did the AI/LLM tell you that?

It’s based on real crash logs and call stacks from live realms with many bots. The guards are basic defensive programming for code callable from multiple contexts. Either post stacks or a minimal repro to refute it otherwise, stop wasting everyone’s time (I know it’s difficult for you because it’s not what you do best.).

Abrou LLM, plenty of people use LLM, the difference is having the brains to analyze instead of parroting. You’re not showing that...

So let people like @kadeshar talk, they’re the ones who know what’s going on.

src/strategy/actions/LeaveGroupAction.cpp
```diff
@@ -84,7 +85,13 @@ bool LeaveGroupAction::Leave(Player* player)
...
     if (!shouldStay)
     {
-        bot->RemoveFromGroup();
+        if (Group* g = bot->GetGroup())
+        {
+            if (g->IsMember(bot->GetGUID()))
+            {
+                bot->RemoveFromGroup();
+            }
+        }
```
kadeshar, line 94: This new if is redundant

src/PlayerbotMgr.cpp
```diff
@@ -512,11 +513,17 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
...
+        if (Group* g = bot->GetGroup())
         {
-            bot->RemoveFromGroup();
+            if (g->IsMember(bot->GetGUID()))
+            {
+                bot->RemoveFromGroup();
+            }
         }
     }
+    }
```
kadeshar, line 526: This new if is redundant

src/RandomPlayerbotMgr.cpp
```diff
@@ -1636,8 +1637,17 @@ bool RandomPlayerbotMgr::ProcessBot(Player* player)
...
+        {
+            LOG_DEBUG("playerbots", "RandomPlayerbotMgr: {} leaving group (leader is random bot, members before: {}).",
+                     player->GetName().c_str(), uint32(group->GetMembersCount()));
+            player->RemoveFromGroup();
+            LOG_DEBUG("playerbots", "RandomPlayerbotMgr: {} left group successfully.", player->GetName().c_str());
+        }
+        else
+        {
+            LOG_DEBUG("playerbots", "RandomPlayerbotMgr: {} not a member anymore, skip removal.", player->GetName().c_str());
+        }
```
kadeshar, line 1650: This new if is redundant

src/RandomPlayerbotMgr.cpp
```diff
@@ -2540,8 +2550,13 @@ void RandomPlayerbotMgr::RandomizeFirst(Player* bot)
...
 
-    if (bot->GetGroup())
-        bot->RemoveFromGroup();
+    if (Group* g = bot->GetGroup())
+    {
+        if (g->IsMember(bot->GetGUID()))
+        {
+            bot->RemoveFromGroup();
+        }
+    }
```
kadeshar, line 2559: Duplicate if

src/RandomPlayerbotMgr.cpp
```diff
@@ -2663,8 +2683,13 @@ void RandomPlayerbotMgr::Refresh(Player* bot)
...
```
kadeshar, line 2692: Why leave group in refresh method?

noisiver:
I am not wasting anyones time, I am trying to determine if the pull request is valid and should be merged. We have been merging stuff without knowing that a pull request is working as expected and that it doesn't cause issues.

We have to change the rules and be more strict before merging pull requests because the quality of this module is rapidly degrading with the current logic for pull requests.

If you have a problem with that at the end of the day it is your problem.

kadeshar:
Sorry but i agree with Revision. This code doesnt make sense when you spend 2 minutes to analyze it. For me also look like generated by some tool which dont have access to whole context.

obraz

First you get group which bot belong and then you checking that this bot is member of this group. 
Lastly this checks are also checked in RemoveFromGroup.
We dont have time to review every line.

hermensbas:
I dont wanna 'bash' your attempt, but these type of errors seems indeed to be thread related.  In my understanding they can only be solved in few ways; using existing/designed thread handling mechanisms (e.g. events queueing) and so on, worldpackets for the client, or possible a some kind of locking mechanism (hard lock or pass-through).  

I am not entirely sure if we can fix them, since we altering the game state in multiple worker threads. I dont know enough about the core to make such a statement though.

I would be very surprised if this would fix the problem in all honesty.

hermensbas:
As in those additional conditional checks could possible lower the chance, which is worth trying i guess. But if its thread issue those methods are not threadsafe, and even if we would make them threadsafe its very likely they bubble up with same kinda issues. But thats just my gut feeling and slighty based on the logs and situation.

PR #1704, opened 2025-10-03
Fixe issue #1700

### LINKED ISSUE #1700: [Bug] Bots can pickup the flag in Eye of the Strom instantly, without cast time from the spawn point
**Bug Description**
When picking up the flag from the starting point in the Eye of the storm Battleground, the player has a casting animation / cast time to do so. Bots are currently able to pick up this flag without any delay, making it impossible to compete against a bot (regardless if its one of the own faction or of the opposing faction).

**Commit Hash**
fc69dd5

**How To Reproduce Bug**
- queue for an eye of the storm battleground
- wait for the battel to start
- watch bots insta-pickup the flag from the flag spawn point
- try to pickup a flag yourself (from the starting point of the flag in the middle of the BG)

**Expected Behavior**
Other than in warsong, the eye of the storm flag has a channeling time before it can be picked up from the spawn point. Bots should (like players) need to cast to pickup the flag from the spawn point. Flag pickups from the ground (after it was dropped by another player) should still be instant.

**Modules**
- mod-autobalance
- mod-playerbots
- mod-eluna (disabled)
- mod-dungeonrespawn
- mod-ahbot

**System**
OS: win 11 pro

Wishmaster117:
Fix EotS center flag pickup: cast instead of instant use
Bots were picking up the Eye of the Storm center flag instantly, skipping the required capture cast. This change makes bots cast SPELL_CAPTURE_BANNER when the flag is at the spawn  and avoids GAMEOBJ_USE in that case. For dropped flags, GAMEOBJ_USE remains unchanged. The wait time follows the spell timing from SpellInfo (channeled duration or calculated cast time).

Fixes #1700.

### DIFF
--- a/src/strategy/actions/BattleGroundTactics.cpp
+++ b/src/strategy/actions/BattleGroundTactics.cpp
@@ -3810,32 +3810,52 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
                     return MoveTo(bot->GetMapId(), go->GetPositionX(), go->GetPositionY(), go->GetPositionZ());
                 }
             }
+
             case BATTLEGROUND_EY:
-            { // Issue: Currently bots in EY take flag instantly without casttime
+            {
                 if (dist < INTERACTION_DISTANCE)
                 {
                     // Dismount before interacting
                     if (bot->IsMounted())
+                    {
                         bot->RemoveAurasByType(SPELL_AURA_MOUNTED);
+                    }
 
                     if (bot->IsInDisallowedMountForm())
+                    {
                         bot->RemoveAurasByType(SPELL_AURA_MOD_SHAPESHIFT);
+                    }
 
                     // Handle center flag differently (requires spell cast)
                     if (atBase)
                     {
                         SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(SPELL_CAPTURE_BANNER);
                         if (!spellInfo)
+                        {
+                            LOG_DEBUG("playerbots",
+                                "EotS: {} SpellInfo not found for SPELL_CAPTURE_BANNER, we do not interact instantly",
+                                bot->GetName());
                             return false;
+                        }
 
+                        // Cast the capture spell on the GameObject
                         Spell* spell = new Spell(bot, spellInfo, TRIGGERED_NONE);
                         spell->m_targets.SetGOTarget(go);
+                        LOG_DEBUG("playerbots", "EotS: {} lance SPELL_CAPTURE_BANNER sur GO guid={}", bot->GetName(), go->GetGUID().ToString());
                         spell->prepare(&spell->m_targets);
-                        botAI->WaitForSpellCast(spell);
-                        //return true; Intended to make a bot cast SPELL_CAPTURE_BANNER and wait for spell finish, but doesn't work and causes infinite loop
+
+                        // Set the wait to the actual duration of the spell
+                        int32 castMs = spellInfo->IsChanneled() ? spellInfo->GetDuration() : spellInfo->CalcCastTime(bot);
+                        if (castMs < 0) castMs = 0;
+                        botAI->SetNextCheckDelay(uint32(castMs));
+                        LOG_DEBUG("playerbots",
+                            "EotS: {} are capturing the flag wait {} ms ({})",
+                            bot->GetName(), castMs, spellInfo->IsChanneled() ? "channeled duration" : "cast time");
+                        return true;
                     }
 
                     // Pick up dropped flag
+                        bot->GetName(), go->GetGUID().ToString());
                     WorldPacket data(CMSG_GAMEOBJ_USE);
                     data << go->GetGUID();
                     bot->GetSession()->HandleGameObjectUseOpcode(data);
@@ -3848,7 +3868,7 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
                     // Move to flag if not in range
                     return MoveTo(bot->GetMapId(), go->GetPositionX(), go->GetPositionY(), go->GetPositionZ());
                 }
-            }
+             }
         default:
             break;
         }
--- a/src/strategy/actions/BattleGroundTactics.cpp
+++ b/src/strategy/actions/BattleGroundTactics.cpp
@@ -3855,7 +3855,6 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
                     }
 
                     // Pick up dropped flag
-                        bot->GetName(), go->GetGUID().ToString());
                     WorldPacket data(CMSG_GAMEOBJ_USE);
                     data << go->GetGUID();
                     bot->GetSession()->HandleGameObjectUseOpcode(data);


src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -3810,29 +3810,48 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
...
+                            LOG_DEBUG("playerbots",
+                                "EotS: {} SpellInfo not found for SPELL_CAPTURE_BANNER, we do not interact instantly",
+                                bot->GetName());
                             return false;
+                        }
 
+                        // Cast the capture spell on the GameObject
                         Spell* spell = new Spell(bot, spellInfo, TRIGGERED_NONE);
                         spell->m_targets.SetGOTarget(go);
+                        LOG_DEBUG("playerbots", "EotS: {} lance SPELL_CAPTURE_BANNER sur GO guid={}", bot->GetName(), go->GetGUID().ToString());
```
kadeshar: Im against log_debug every line

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -3810,29 +3810,48 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
...
-                        //return true; Intended to make a bot cast SPELL_CAPTURE_BANNER and wait for spell finish, but doesn't work and causes infinite loop
+
+                        // Set the wait to the actual duration of the spell
+                        int32 castMs = spellInfo->IsChanneled() ? spellInfo->GetDuration() : spellInfo->CalcCastTime(bot);
+                        if (castMs < 0) castMs = 0;
+                        botAI->SetNextCheckDelay(uint32(castMs));
+                        LOG_DEBUG("playerbots",
+                            "EotS: {} are capturing the flag wait {} ms ({})",
+                            bot->GetName(), castMs, spellInfo->IsChanneled() ? "channeled duration" : "cast time");
+                        return true;
```
kadeshar, line 3958: That is not fix but workaround shading real problem. You identified this problem and why we cant solve that properly?

Also i assume that normal casting can be break when character get damage.

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -3810,29 +3810,48 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
                     return MoveTo(bot->GetMapId(), go->GetPositionX(), go->GetPositionY(), go->GetPositionZ());
                 }
             }
+
```
kadeshar: Delete this empty line

src/strategy/actions/BattleGroundTactics.cpp
icemansparks: Yes, the cast should break when the caster receives damage.

Wishmaster117: In my PR i replace GAMEOBJ_USE by the true SPELL_CAPTURE_BANNER, i user the cast duration provided by SpellInfo.
This is what the official client does, the flag in the center requires a holding time of 8sec and any damage taken or action interrupts it, no pushback, all progress lost.

About "That is not fix but workaround shading real problem." you are true, it's a functionnal solution, i'me traying to manage how to solve that, but i posted that because for now till i manage  a lasting solution.

I make a mistake naming this Fixe, i change by provisory patch.

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Sorry, removed

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Empty line removed

Wishmaster117:
@kadeshar what do you think about :
Move all related battlegroud files to a proper battlegrond folder in stragtegy folder.
After that split each battelegroud tactics to his own file?

src/strategy/actions/BattleGroundTactics.cpp
kadeshar: For me is simple bot delay equal of cast duration. There is nothing related with taken actions or/and damage taken in this delay.

kadeshar:
First did you check that uncomment return true still cause infinite loop?
Second, how that look with your fix. Bot getting instant flag and then standing 8 sec without any action?

About BG refactoring Im not sure. Now code is optimal because reuse similar things between them. Refactoring which you propose may not have any value. Hard to tell without proper analysis.

Wishmaster117:
I spent a good amount of time reading the core files:
BattlegroundEY.h and BattlegroundEY.cpp: All the EotS-specific logic (central flag, base delivery, timers, worldstates/UI). Entities/GameObject/GameObject.cpp: Clicking on the central flag gameobject redirects to the BG method via EventPlayerClickedOnFlag.

Battleground.h and Spells/SpellDefines.h: Common constants (e.g., SPELL_RECENTLY_DROPPED_FLAG, aura interrupt flags).

In BattlegroundEY.h, we have the following enumeration:

enum BG_EY_FlagState { BG_EY_FLAG_STATE_ON_BASE = 1, // Flag present in the center on its GO BG_EY_FLAG_STATE_ON_PLAYER = 2, // Flag carried by a player BG_EY_FLAG_STATE_ON_GROUND = 3 // Flag on the ground (dropped) };

The current state is stored in BattlegroundEY::_flagState (initialized in Init() to ON_BASE).

 The logic when player pickup of the central flag is this one (i think):

The player clicks the "Netherstorm Flag" GO.
In Entities/GameObject/GameObject.cpp, under Use, we see:
// 184142: Netherstorm Flag if (bg->GetBgTypeID(true) == BATTLEGROUND_EY) bg->EventPlayerClickedOnFlag(player, this);
BG processing: BattlegroundEY::EventPlayerClickedOnFlag(Player*, GameObject*) (in BattlegroundEY.cpp)
Main effects:
Checks that no other carriers exist and that the player is in range.
Sets the state to ON_PLAYER.
Despawn of the central GO: SpawnBGObject(BG_EY_OBJECT_FLAG_NETHERSTORM, RESPAWN_ONE_DAY);  (set to "one day" = hide it until respawned).
Remembers the carrier: SetFlagPicker(player->GetGUID());
Applies the carrier's aura: BG_EY_NETHERSTORM_FLAG_SPELL (34976) via player->CastSpell(...).
Clears any "interrupt-on-PvP" auras (bandage, drink, etc.):
player->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_ENTER_PVP_COMBAT);
Updates worldstates/sounds/messages ("flag picked up" announcement).

There is no "channeling" for flag picking: it's instantaneous on click, no cast bar to interrupt.
I can't figue out where is this f***** thing....

Any idea?

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Now the bots stop casting when take damage

Wishmaster117:
Ok, now it's ok

noisiver:
The PR should be renamed to explain what it does, not just that it (maybe) fixes a certain issue. It's easier to understand when you read the commit history, otherwise you have to look at the issue itself to know what was fixed.

kadeshar:
Ehhhhh whole file refactoring with fix is messy. That should be 2 separate PR's. One with fix and one with refactoring.
And i dont understand refactoring commented code...

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -4,10 +4,12 @@
  */
 
 #include "BattleGroundTactics.h"
-#include "BattleGroundJoinAction.h"
+
+#include <algorithm>
```
kadeshar, line 8: Its used?

Wishmaster117:
Sorry i don't do anything else that put the file in visual studio and make a ctrl+k and ctrl+d to format it. 
I will pick the original file add my changes and upload it again without code format by visual studio. 

My changes starts line 3523

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -3811,28 +3861,51 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
...
 
                         Spell* spell = new Spell(bot, spellInfo, TRIGGERED_NONE);
                         spell->m_targets.SetGOTarget(go);
+
+                        bot->StopMoving();
                         spell->prepare(&spell->m_targets);
+
                         botAI->WaitForSpellCast(spell);
-                        //return true; Intended to make a bot cast SPELL_CAPTURE_BANNER and wait for spell finish, but doesn't work and causes infinite loop
+                        // return true; Intended to make a bot cast SPELL_CAPTURE_BANNER and wait for spell finish, but
```
kadeshar: this comment after changes should be deleted no splitted into separate lines

kadeshar:
Mostly i dont understand this code :D Waiting for tests or some more experienced devs reviews

Wishmaster117:
> Mostly i dont understand this code :D Waiting for tests or some more experienced devs reviews

You will understand when i upload the file only with my changes, don't spend time trying to review this one.

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Yes. The file uses it:
std::find is used in several places (e.g., to avoid duplicate objectives in AB, to check flags/paths).
std::sort is used to sort allies by distance in Eye of the Storm.

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Deleted

Wishmaster117:
> Mostly i dont understand this code :D Waiting for tests or some more experienced devs reviews

Done i revreted all other code formated by visual studio

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -3698,15 +3769,27 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
...
+    if (!eyeBg && bgType == BATTLEGROUND_EY)
+    {
+        eyeBg = static_cast<BattlegroundEY*>(bg);
+    }
+
+    if (!eyCenterFlag && eyeBg)
+    {
+        eyCenterFlag = eyeBg->GetBGObject(BG_EY_OBJECT_FLAG_NETHERSTORM);
+    }
+
```
kadeshar: This part is not clean for me. This variables are already declared in same way in line 3772-3780. Why in same action iteration you trying do same?

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -3811,28 +3922,49 @@ bool BGTactics::atFlag(std::vector<BattleBotPath*> const& vPaths, std::vector<ui
...
 
                     // Handle center flag differently (requires spell cast)
-                    if (atBase)
+                    if (isEyCenterFlag)
                     {
+                        for (uint8 type = CURRENT_MELEE_SPELL; type <= CURRENT_CHANNELED_SPELL; ++type)
+                        {
+                            if (Spell* currentSpell = bot->GetCurrentSpell(static_cast<CurrentSpellTypes>(type)))
+                            {
+                                if (currentSpell->m_spellInfo->Id == SPELL_CAPTURE_BANNER)
```
kadeshar: Missing null check for currentSpell->m_spellInfo

src/strategy/actions/BattleGroundTactics.cpp
```diff
@@ -4289,9 +4421,15 @@ bool ArenaTactics::moveToCenter(Battleground* bg)
             {
                 // they like to hang around at the tip of the pipes doing nothing, so we just teleport them down
                 if (bot->GetDistance(1333.07f, 817.18f, 13.35f) < 4)
+                {
+                    bot->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_TELEPORTED | AURA_INTERRUPT_FLAG_CHANGE_MAP);
```
kadeshar, line 4416: I dont understand this part. Why you remove this flags. Is that necessary only in battleground or always?

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Good point, eyeBg/and CenterBlog are solved once, but the section can be confusing. I'll simplify but I won't have access to my PC until Monday, I'm traveling tomorrow.

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: This is the original code — I didn’t change it.
Why remove these “flags”?
It’s not removing flags from the player; it’s removing auras that would be interrupted anyway by a teleport or a map change. We do it proactively to ensure no aura (e.g., mount/channeling) prevents or interferes with the fallback teleport when a bot gets stuck at the top of the pipe in Dalaran Sewers. The teleport happens immediately after via TeleportTo(...). You can see the exact sequence in the BATTLEGROUND_DS case.

Is this always necessary, or only in battlegrounds?
Only here — specifically in the Dalaran Sewers arena at the start (spawn movement logic). It’s not global behavior: it’s confined to the case BATTLEGROUND_DS and only runs if the bot is exactly near the two pipe positions (checked by distance to those coordinates).

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: Done

src/strategy/actions/BattleGroundTactics.cpp
Wishmaster117: You are right, eyeBg and eyCenterFlag are already initialized at the top of atFlag() when bgType == BATTLEGROUND_EY. The later conditional re-initialization in the same action iteration is redundant. I’ve removed that duplicate block to keep the control flow clean and avoid confusion.

src/strategy/actions/BattleGroundTactics.cpp
nl-saw: It looks like the Spell check is sufficient. It's done like that in all places, including your changes on YoggSaronLunaticGazeTrigger. If it were a problem, it would likely have been known by now.

nl-saw (approved):

Zerathane:
I have pulled this PR for testing.

Based on the current (PR) behaviours outlined by the author I observed the bots doing the following during a few Eye of the storm battlegrounds I ran.

- Upon approaching the flag they dismount and come to a standstill
- They channel capturing the flag  when they are inside the circle. I timed this to be roughly 8 seconds give or take.
- They are interrupted if they receive damage during this cast.
- Druids cancel their current shapeshift form to channel their flag capture.

Regarding druids. If they are interrupted by an enemy they will engage in combat with that enemy but not cast their shapeshift form for some time. Also they will not shapeshift once they have the flag. Which leaves them more vulnerable. (Not suggesting this is a bug with this PR just a behaviour I thought worth mentioning)

I didn't notice any strange or adverse behaviours of the bots during Eots when using this PR - The PR allows for more realistic combat at the flag now that the capture is not instant, and is interruptable.

Wishmaster117:
> I have pulled this PR for testing.
> 
> Based on the current (PR) behaviours outlined by the author I observed the bots doing the following during a few Eye of the storm battlegrounds I ran.
> 
> * Upon approaching the flag they dismount and come to a standstill
> * They channel capturing the flag  when they are inside the circle. I timed this to be roughly 8 seconds give or take.
> * They are interrupted if they receive damage during this cast.
> * Druids cancel their current shapeshift form to channel their flag capture.
> 
> Regarding druids. If they are interrupted by an enemy they will engage in combat with that enemy but not cast their shapeshift form for some time. Also they will not shapeshift once they have the flag. Which leaves them more vulnerable. (Not suggesting this is a bug with this PR just a behaviour I thought worth mentioning)
> 
> I didn't notice any strange or adverse behaviours of the bots during Eots when using this PR - The PR allows for more realistic combat at the flag now that the capture is not instant, and is interruptable.

Thank you very much for your testing and feedback.
About druid shapeshift i think this was the case before this changes can you confirm?
Anyway after this PR merge i will make another one to correct that and correct waypoints of the bot flag carrier and add some ai stuffs because I think that flag carrier are a little dumb. 🤣

nl-saw:
Tested, behavior is a lot better than before.

Wishmaster117:
> Tested, behavior is a lot better than before.

It's trange before bots take flag instantly, with this changes they cast the spell to take the flag, and they only have it after 8 secondes (blizzlike).
So or you have made something wrong or you should realy try the before.
Others who tested report that's better. 
Pay attention i just corrected the flag capture, not the waypoints and other that are same code as before.

Wishmaster117:
> Tested, behavior is a lot better than before.

Lol i read : Tested behavivior is not better than before => Note to mysel avoid read github in phone....


**EoS with thiis fix :** https://youtu.be/mNU5YT1O81w
Upon approaching the flag they dismount and come to a standstill
They channel capturing the flag when they are inside the circle. 8 sec (blizzlike)
They are interrupted if they receive damage during this cast.

**EoS without this fix:** https://youtu.be/euDe0sbBRVM
Flag instant capture

I wil work in bots waypoints after merged.

PR #1708, opened 2025-10-05
[fix crash] Crash on removeFromGroup

hermensbas:
# Replaced Directly player->removeFromGroup with client worldPacket handler

Certain calls are not intended to be called directly from the mapThread, which is where the playerbots mod 'lives'. Actually the majority of classes we use of the core has been designed to be called from the main world thread not from external threads. 

Since we live in both worlds (client and server) and in certain world-state changing events we needs to simulate a real player. Which means we pretend to be a player from a wow-client perspective, which uses world-packets to communicate with the server. Those world-packets are picked up by internal core handlers which execute the code from the world-thread, in some cases its oneway road towards child threads like mapThreads and never intended for the other way around.

Surely we could create thread dispatcher to mitigate that, which we are trying to avoid since we want to use the original logic executes in those handlers and thereby use core code/logic as intended.

test-branch:  
```bash
git clone https://github.com/hermensbas/mod-playerbots.git --branch=feature/removeFromGroup_replaced_with_worldpackets
```

possibly fix: https://github.com/liyunfan1223/mod-playerbots/issues/1695.  Possibly since i cant recreate the issue solo with some bots it seems.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -519,7 +519,7 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
     }
 
     group = bot->GetGroup();
-    if (!group)
+    if (group)
     {
         botAI->ResetStrategies();
     }
--- a/src/RandomPlayerbotMgr.cpp
+++ b/src/RandomPlayerbotMgr.cpp
@@ -1521,7 +1521,7 @@ bool RandomPlayerbotMgr::ProcessBot(Player* bot)
 {
 
     PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
-    if (botAI)
+    if (!botAI)
         return false;
 
     if (bot->InBattleground())
@@ -2379,7 +2379,7 @@ void RandomPlayerbotMgr::IncreaseLevel(Player* bot)
 void RandomPlayerbotMgr::RandomizeFirst(Player* bot)
 {
     PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
-    if (botAI)
+    if (!botAI)
         return;
 
     uint32 maxLevel = sPlayerbotAIConfig->randomBotMaxLevel;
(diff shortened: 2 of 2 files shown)


hermensbas:
There also seems a bug when bots leave group they keep following you around. I will look into that another time. Cant imagine thats by design, but not sure. For another time.

hermensbas:
I havent detected any delta fuctionality in leaving group, which is good obviously. Whether it fixes the threading problem entirely remains to be seen since adding to group in some situation still happens imperative (in code). But it might sufficient thought.

Whats important to test? That leaving group still works functionally as expected. As the ones who are running servers with mulitple real players with bots dont get access violations when bots leaving groups.

src/strategy/actions/LeaveGroupAction.cpp
```diff
@@ -84,7 +86,7 @@ bool LeaveGroupAction::Leave(Player* player)
     bool shouldStay = randomBot && bot->GetGroup() && player == bot;
     if (!shouldStay)
     {
-        bot->RemoveFromGroup();
+        botAI->RemoveFromGroup();
```
Celandriel: Would it be helpful to rename this LeaveOrDisbandGroup since the OPCODE you are invoking will do both depending on whether groupleader or not?

hermensbas: Not bad idea actually

hermensbas:
Waiting to be confirmed by someone who is running a server with multiple real players and bots encountering this issue. Although the fix is valid even without. But lets see if it fixes the related issue first.

Regrad:
I started testing. If there are no messages from me within a couple of days, then everything is fine and I just forgot to to report here. If there are any problems, I will definitely remember when I see a crash report.

hermensbas:
> I started testing. If there are no messages from me within a couple of days, then everything is fine and I just forgot to to report here. If there are any problems, I will definitely remember when I see a crash report.

Great, we will wait till your confirmation; your one of the few who is running a larger server.

Regrad:
It seems that this patch breaks the creation and initialization of bots in the world. All bots are level one.

hermensbas:
> It seems that this patch breaks the creation and initialization of bots in the world. All bots are level one.

aah fuck i know whats going wrong, early return. Will fix it when i have time.

hermensbas:
> It seems that this patch breaks the creation and initialization of bots in the world. All bots are level one.

Fixed, i messed up the early returns condition and caught little bug along the way.

Regrad:
So far, so good all day. No regressions were noticed, and there were no crashes for this reason either.

kadeshar:
@hermensbas @Regrad You feel that is stable to be good to merge to module?

Regrad:
Yes. So far, I have not found any problems. There are about 60 online players.

PR #1709, opened 2025-10-05
Updated the spell id for Spirit of Redemption

noisiver:
Replaced the spell id for Spirit of Redemption to allow a Holy Priest to drink

It needs some testing to make sure it always returns false when Spirit of Redemption is active


hermensbas:
Wonderful catch! Will check it later what you actually done hehe

kadeshar (approved):

hermensbas (approved):

PR #1713, opened 2025-10-06
Balance-Druid-improve-Starfall-usage-and-add-CC-safety

gacuna89:
- Move Starfall from default actions to AOE strategy only
- Require 2+ enemies for Starfall usage (prevents single-target casting)
- Add CC safety: avoid casting Starfall near current CC targets
- Prioritize Starfall over Hurricane in medium AOE situations
- Remove Starfall from pull/opener rotation to prevent early single-target usage

This prevents Balance druids from wasting Starfall on single targets
and breaking crowd control effects in group content.


Dohtt:
I do feel like starfall has a crucial place in a druids normal kit for bossfights.
image


It's also listed as a high prio spell, besides during eclipse procs and should be on permanent cooldown. 

I think this might reduce the balance druids dps capabilities.

gacuna89:
> I do feel like starfall has a crucial place in a druids normal kit for bossfights. image
> 
> It's also listed as a high prio spell, besides during eclipse procs and should be on permanent cooldown.
> 
> I think this might reduce the balance druids dps capabilities.

You're right, I was being too strict with Starfall.
After checking the rotation guide again, it should just be used on cooldown as a high-priority spell outside Eclipse.

I tweaked the logic a bit to make it more balanced:

Kept:
- Starfall back in the main actions with high priority (keeps DPS flow).
- Still avoids casting near CC’d targets to not break anything.

Changed:
- Removed the AOE-only restriction that was hurting damage.
- Only blocks single-target use for the first few seconds of combat (so it’s not wasted right at the start).
- After that, it just works normally on cooldown.

That should give a solid rotation without breaking CC or wasting Starfall on single-target fights.

nl-saw (approved):

Zerathane:
I've pulled this PR for testing but it's failing to compile with the following errors.

starfal

kadeshar:
Branch is compiling without issue. 
Use current azerothcore branch

Zerathane:
> Branch is compiling without issue. Use current azerothcore branch

Ah I'll make sure core is upated perhaps I'm a week or two behind on updates.

Apologies @gacuna89 Ill try again and feed back.

hermensbas:
Does this still require testing?

nl-saw:
@gacuna89 Could you or anyone else let us know in what scenario's this has been tested? Like BG/PvP/Raids/etc.

hermensbas (approved):

hermensbas:
I am not druid player, so if someone can verify the changes ingame and run some tests. So we can merge this?

Issue #1677, opened 2025-09-28
[Bug/AI] Multiple Paladins in a party with different specs all compete using their greater blessings.


Badgermilk0:
**Bug Description**
Paladins compete using greater buffs until they run out of reagents.

**Commit Hash**
0b74820f80042f95be236449ddf0ec89d5dc3eaa

**How To Reproduce Bug**
Have multiple Paladins in the party that triggers different needed buffs

**Expected Behavior**
Correct buffs on correct classes

**Screenshots**
<img width="822" height="279" alt="Image" src="https://github.com/user-attachments/assets/72546243-d49b-47b3-b1c2-25dfad5f6208" />

**Modules**
N/A

**Playerbot Settings**
N/A

**System**
Windows

**Additional Context**
The party where this appeared had:
1 Paladin - Protection
1 Paladin - Holy
1 Paladin - Retribuition
1 Shaman - Enhancement
1 Hunter - Beast Master

Badgermilk0:
To add, the issue then is that if one paladin triggers the requirement to put, for example, Might on the hunter, it will first try to use a Greater blessing.

But then the holy paladin suddenly have Might and casts greater blessing of wisdom, and so on..

I did not have this issue in the past it has appeared a few commits back.

Wishmaster117:
> To add, the issue then is that if one paladin triggers the requirement to put, for example, Might on the hunter, it will first try to use a Greater blessing.
> 
> But then the holy paladin suddenly have Might and casts greater blessing of wisdom, and so on..
> 
> I did not have this issue in the past it has appeared a few commits back.

How do you configure the paladins buff strategy?

Badgermilk0:
> > To add, the issue then is that if one paladin triggers the requirement to put, for example, Might on the hunter, it will first try to use a Greater blessing.
> > But then the holy paladin suddenly have Might and casts greater blessing of wisdom, and so on..
> > I did not have this issue in the past it has appeared a few commits back.
> 
> How do you configure the paladins buff strategy?

Default config - I did not make any alterations

*Edit*

These are the settings : 
.# Min group size to use Greater buffs (Paladin, Mage, Druid)
.# Default: 3
.AiPlayerbot.MinBotsForGreaterBuff = 3

.# Cooldown (seconds) between reagent-missing RP warnings, per bot & per buff
.# Default: 30
.AiPlayerbot.RPWarningCooldown = 30

Wishmaster117:
[...]

PR #1715, opened 2025-10-07
 Fix Multiple Paladins in a party with different specs all compete using their greater blessings. Issue #1677

Wishmaster117:
This PR attempt to fixe : https://github.com/liyunfan1223/mod-playerbots/issues/1677

These paladin buffs are a real headache. Here’s an explanation of the logic I used, could someone test it and let me know if everything’s okay now?

1) The common “buff framework” used in all cases

Mono/Greater detection + anti-double-buff
MakeAuraQualifierForBuff() builds a “mono+greater” qualifier for the target selector (“party member without aura”): e.g., "blessing of wisdom,greater blessing of wisdom" => prevents re-buffing if either aura is already present.

Automatic switch to group (“Greater”) buff
UpgradeToGroupIfAppropriate() upgrades to the group version if: 
a) you’re in a group with at least AiPlayerbot.MinBotsForGreaterBuff, 
b) the “greater” rank is known, 
c) reagents are available. Otherwise it keeps the mono version. If reagents are missing, the bot sends an RP warning (anti-spam via AiPlayerbot.RPWarningCooldown).

Generic actions side
BuffOnPartyAction picks targets via “party member without aura” using the (mono+greater) qualifier, then calls UpgradeToGroupIfAppropriate() before casting. So: if the group is large enough AND the bot has reagents => cast Greater (else mono, with RP announcement if desired).

2) Paladin blessing distribution the “who casts what” logic

The code has two layers: strategy triggers (which decide what actions to attempt) and actions (which decide “what” to cast on “whom”).

a) Triggers/Strategies
bmana: triggers Wisdom on party then Kings on party (Wisdom has priority).
bdps: triggers Might on party.
bstats: first Sanctuary on party (priority), then Kings on party.

The base “Blessing” trigger ensures no spam if a paladin blessing is already present (it checks Might/Wisdom/Kings/Sanctuary).

b) Coordination between paladins, avoid stepping on each other

Per “buff role” election (IsDesignatedRolePaladin): for the same role (“bdps”, “bmana”, “bstats”), if several paladins have the strategy, the “winner” is deterministic (lowest GUID). Others abstain => fewer collisions.

“Solo paladin” detection and group counting: utilities to adapt behavior (e.g., avoid overwriting Sanctuary, favor Greater when alone, etc.).

c) Per-blessing rules

Sanctuary (tank priority)
CastBlessingOfSanctuaryOnPartyAction looks for a tank without Sanctuary (self included if the paladin is a tank) and deliberately does not use Greater to reduce collisions. If there’s nothing to do, it stops.

Kings on party
Allowed if the paladin is elected for bstats or bmana.

“bmana” case: Mono Kings on tanks only (don’t break Sanctuary, no group-wide cast).
“bstats” solo-paladin case: avoids targeting itself (so as not to remove its own Sanctuary), and favors Greater to cover non-tanks when possible.
Multiple paladins in “bstats”: may disable switching to Greater for certain classes (war/paladin/druid/dk) to avoid overwriting Might at group scale. Then UpgradeToGroupIfAppropriate() decides based on reagents & group size when allowGreater remains true.

Wisdom on party (bmana)
Only the elected “bmana” paladin acts. It skips tanks (Kings/Sanct have priority for them). Casts “blessing of wisdom”, upgrading to Greater if eligible (reagents + size).

Might on party (bdps)
Only the elected “bdps” paladin acts. It retargets if needed to a Might-eligible player (physical dps, hunter, enh shaman, feral, ret, … prot only if there are ≥ 3 paladins in the group). Then casts “blessing of might”, upgrading to Greater if eligible.

The target selector for Might/Wisdom excludes people who already have Sanctuary (mono or greater) by including Sanctuary in the aura qualifier; this avoids overwriting Sanctuary.

d) “Targeted cast” cases (non–party-wide)

When casting Might/Wisdom on a specific target, the helpers GetActualBlessingOfMight/Wisdom automatically redirect to the other blessing based on class/spec (e.g., a caster receives Wisdom even if you call “Might”). It’s convenient and consistent.

3) Paladin triggers

BlessingTrigger::IsActive() prevents launching a new blessing if the target already has Might/Wisdom/Kings/Sanctuary.

BlessingOnPartyTrigger may attempt several blessings (kings, might, wisdom) depending on active strategies and timing, but the actions then filter via the per-role election.

@kadeshar , sory about all logs but i need to keep it for now


kadeshar:
- PR mean you ended your work and no longer debugging is needed. If not make that as draft
- make this code user-friendly. Variables like "a" or "ok0" means nothing while reading code
- recheck all added/affected conditions because at least few are useless

Wishmaster117:
> * PR mean you ended your work and no longer debugging is needed. If not make that as draft
> * make this code user-friendly. Variables like "a" or "ok0" means nothing while reading code
> * recheck all added/affected conditions because at least few are useless

Ok but before spending more time on this need testers and if it's ok and tests are concluentsi  revise the code.

kadeshar:
You can try asking someone to help in testing channel on discord

Celandriel:
As a note, you should also test for functionality, and if possible describe to the testers how to test it. What are the things they should be looking for? etc.

Wishmaster117:
> As a note, you should also test for functionality, and if possible describe to the testers how to test it. What are the things they should be looking for? etc.

I tested it but i don't have a server with multiple users I just hava a test server with some bots, so for me it's work but before put this in ready to review need more feedbacks in live server. 

The thinks they have to look are simple, if all in raid have the right buff and if Paladins still compete to put his buff like described in issue.

noisiver:
The PR should be renamed to explain what it does, not just that it (maybe) fixes a certain issue. It's easier to understand when you read the commit history, otherwise you have to look at the issue itself to know what was fixed.

Wishmaster117:
I messed up i make a new one

PR #1719, opened 2025-10-08
[fix crash] check bot / target world-state before cast

hermensbas:
https://github.com/liyunfan1223/mod-playerbots/issues/1718

Lowers the scope of the target/bot is in an invalid (world-)state  just before casting or receiving.

Kinda forced the error myself by cheating to debug the exact code path, but eitherway it seems the crash comes from the usage of invalid target when casted to be exact a target interally translates into UnitTarget, ItemTarget or GoTarget those targets are not checked before handing it over to the core, which can result into invalidObject at the time we call the prepare cast.

And added some additional checks when entering the method, for valid target en bot to begin with.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -3498,22 +3498,22 @@ bool PlayerbotAI::CastSpell(uint32 spellId, Unit* target, Item* itemTarget)
 
 
     // check if the actual target is valid, spell system uses targets.GetUnitTarget(), GetItemTarget(), GetGOTarget() internally.
-    Unit* unitTarget = targets.GetUnitTarget();
-    if (unitTarget && (!unitTarget->IsInWorld() || unitTarget->IsDuringRemoveFromWorld()))
+    Unit* uTarget = targets.GetUnitTarget();
+    if (uTarget && (!uTarget->IsInWorld() || uTarget->IsDuringRemoveFromWorld()))
     {
         // fail if the spell has a unit target that is invalid or being removed from the world
         delete spell;
         return false;
     }
-    Item* itemTarget = targets.GetItemTarget();
-    if (!itemTarget && (spellInfo->Targets & (TARGET_FLAG_ITEM | TARGET_FLAG_GAMEOBJECT_ITEM)))
+    Item* iTarget = targets.GetItemTarget();
+    if (!iTarget && (spellInfo->Targets & (TARGET_FLAG_ITEM | TARGET_FLAG_GAMEOBJECT_ITEM)))
     {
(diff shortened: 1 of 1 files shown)


src/PlayerbotAI.cpp
```diff
@@ -3491,6 +3497,15 @@ bool PlayerbotAI::CastSpell(uint32 spellId, Unit* target, Item* itemTarget)
     // if (spellSuccess != SPELL_CAST_OK)
     //     return false;
 
+
+    // check bot/target world-state just before preparing the actual spell
+    Unit* unitTarget = targets.GetUnitTarget();
+    if (!bot->IsInWorld() || bot->IsDuringRemoveFromWorld() ||
+        (unitTarget && (!unitTarget->IsInWorld() || unitTarget->IsDuringRemoveFromWorld())))
```
kadeshar: Duplicate if condition. We cant make that as method?

hermensbas: Dont think this condition will be used in other functions then in this function itself. Creating a private new function for this scope alone in a class with 6608 lines of code seems a bit counter productive? What do you think?

kadeshar: Im not sure this code at all. You checking same twice in same method.
Ofc i can accept that because some protection is better than nothing but its good way to check bot world state?
First i mean way when we always checking state and dont stop all actions while logging out/despawning (like CancellationToken in c#)?
Second this specific check shouldnt also contains !bot->GetSession() || bot->GetSession()->isLogingOut() like in PlayerbotAI::UpdateAI ?

hermensbas: There is no need for those additional checks, we are not interested whether the bot in the process of logging out, we only care for object still being there so the underlaying core code doesnt crash. 

We cant completely remove it without make core changes but i dont wanna do that. But this will lower the scope, the early return only prevents expensive spell logic when we already know we dont need it, under the pet logic handling.

You can still hit a player when in process of logging out, but it can also mean the bot already disconnected which is fine as long we check the object is likely to still exists hence bot->IsDuringRemoveFromWorld() which is on unit level is not mistaken.

hermensbas: I will double check though

hermensbas:
Kinda forced the error myself by cheating to debug the exact code path, but eitherway it seems the crash comes from the usage of invalid target when casted to be exact a target interally translates into UnitTarget, ItemTarget or GoTarget those targets are not checked before handing it over to the core, which can result into invalidObject at the time we call the prepare cast.

And added some additional checks when entering the method, for valid target en bot to begin with.

hermensbas:
Still have todo some additional test somewhere this week before setting on review again, and debugging is slooooooooooooow

PR #1720, opened 2025-10-09
Cleanups to config and source

brighton-chi:
Mostly non-substantive clean-up changes, but also with two proposed changes to config defaults:
- HealerDPSMapRestriction enabled by default. I think most players would not expect healers to be wasting mana attacking in instances so this is the better default.
- SightDistance increased to 100 yards from 75 yards. This is the value used by bots to find npcs. The corresponding distance is 100 yards for players by default in the Core .dist (with defaults of 250 yards in BGs and 170 yards in instances, actually). I don't think bots should have lesser sight distance than players. 

Other stuff:
- Revised defaults in PlayerbotAIConfig.cpp to match defaults in the .dist
- Modified super outdated log for when the module is disabled
- Fixed a couple of typos in the source
- Made some clarifying edits in the .dist (e.g., LimitEnchantExpansion doesn't just limit enchants but gems also)
- Removed SayWhenCollectingItems from the .dist because it doesn't exist

One question--does anybody have a sense as to what ShootDistance should be by default? The .dist has 5 yards, but the source had 25 yards. I'm not sure what it actually has to do with shooting; it seems to relate to calculating movement distance with respect to MovementAction::Flee, which I think is called in various other movement methods (and is not related to the strategy "flee"). But I don't really get it.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1059,7 +1059,8 @@ AiPlayerbot.ZoneBracket.4197 = 79,80
 #
 #
 
-# Maps where bots can be teleported to
+# Map IDs where bots can be teleported to
+# Defaults: 0 = Eastern Kingdoms, 1 = Kalimdor, 530 = Outland, 571 = Northrend
 AiPlayerbot.RandomBotMaps = 0,1,530,571
 
 # Probabilty bots teleport to banker (city)


Dreathean:
Changes and cleanup look great to me. Agreed on the changes with healer dps and sight distance. Don't know on the shoot distance personally.

brighton-chi:
I don't think this requires a Wiki update.

kadeshar:
I see distance changes which are part of performance guide.
There is even issue on that (but i dont remember that for this specific distance types)

brighton-chi:
I don't see it: https://github.com/liyunfan1223/mod-playerbots/wiki/Playerbot-Configuration#recommended-config

I mean it's included in the screenshot of the recommended config, but if we're going to have the recommended config still say 75 yards, then the default config should just say 75 yards, and this PR shouldn't change it.

I think it's weird to have bots not be able to see as far as players, but if there's a reason that is problematic then we don't have to change it. I will say 75 yards is a problem for some TBC instances, but I have boss strats temporarily modify sight distance to 150 yards in Gruul's Lair and Magtheridon's Lair anyway.

kadeshar:
Im dont understand why recommended distance config are not default thats why i marked that PR to solve that finally.

conf/playerbots.conf.dist
NoxMax: May I also recommend adding here
`# Default: 0(Eastern Kingdoms),1(Kalimdor),530(Outland),571(Northrend)`
I was looking at this config today and thought adding some names to the numbers might be helpful for people who don't recognize these main maps IDs.

noisiver: I don't think the map names should be a part of the default description. I can guarantee you that it will confuse some people. Adding information just under the `# Maps where bots can be teleported to` to include that information (map ID = map name) would make sense though.

NoxMax: I see your point that they might think the word "default" means they can copy/paste that entire line with the names. Then maybe something like this?
```
# Map IDs where bots can be teleported to
# 0 = Eastern Kingdoms, 1 = Kalimdor, 530 = Outland, 571 = Northrend
AiPlayerbot.RandomBotMaps = 0,1,530,571
```

noisiver: That looks good

PR #1721, opened 2025-10-09
Implement Magtheridon strategy

brighton-chi:
I'm marking this as a draft for now because I haven't done a detailed review of the code, but I'm posting it now in case anybody wants to give it a try.

Here's what the strategy (should) do.

### **Channeler Phase**

While you can probably AoE down all five Hellfire Channelers, that's more dicey with IP nerfs and it's no fun, so the strategy takes what would have still been considered an aggressive approach back in the day by (1) assigning the Main Tank to the first Channeler, (2) having Hunters misdirect two more Channelers to the MT, and (3) one Off Tank picking up each of the fourth and fifth Channelers and dragging them out of Shadow Volley Range from the main group. Sometimes the pull gets a little wonky and one of the OTs might end up with one of the Channelers that was intended for the MT, but it should work out in the end.

DPS will move through Channelers from Square -> Star -> Circle -> Diamond -> Triangle. Once Square, Star, and Circle are down, the MT will go sit by Magtheridon and wait for him to become active instead of helping with the last two Channelers. I could have made the MT help with the fourth Channeler too, but it's not needed, and positioning to pick up Magtheridon after the third Channeler is a failsafe for low DPS groups that aren't able to get four down before Magtheridon breaks free.

The top priority for Warlocks is to banish/fear the Burning Abyssals, and they will continue to do so even after Magtheridon becomes active (you are not supposed to kill the Abyssals; they have a gazillion HP and automatically despawn after a minute). Their next priority is to put Curse of Tongues on the Channelers.

### **Magtheridon Positioning**

The MT will pick up Magtheridon and pull him (moving backwards because Magtheridon kind of hits like a bus) to a position up against the far wall. Ranged DPS will spread out from a point roughly in the center of the room, and Healers will spread out from a point that is a little closer to Magtheridon. I have not built in aoe avoidance (except for cube clickers, see below) because the general avoid aoe strategy seems to work fine for this fight.

### **Clicking Manticron Cubes**

Now, the fun part. Bots will be assigned to clicking cubes by standard group selection order (reverse join order), but assignment is done via two passes. The first pass will look to select five ranged DPS bots that are _not_ Warlocks. This is because Warlocks may still be needed to keep Abyssals banished/feared and because Warlocks of all three specs put out by far the most damage of all ranged DPS at level 70 in pre-raid/T4 gear. If there are not five non-Warlock ranged DPS bots available, then the logic goes to the second pass, which can pick any bot that is not a Tank.

Cube clicking works on a timer:

1. 48 seconds after Magtheridon breaks free, assigned cube clickers move near their cubes (but a few yards outside of interact distance). During this time, they should move around still to avoid Debris (by maintaining distance from the triggering NPC) and Conflagration (by maintaining distance from the triggering gameobject). Blast Nova is on a 54.35s to 55.4s timer, and I found 48s to always be ample time to get to the cubes, but YMMV so this is a good thing to test. Going to a cube too early not only takes away DPS but also risks more hazards appearing on/around the cube that will then cause problems when the cube needs to be clicked.
2. Blast Nova is a 2s cast, followed by a 10s channel (if not interrupted by the cubes). As soon as the cast begins, bots will move into interaction range and click the cube. Well, there is a randomized delay between 200ms (about the fastest possible human reaction time to visual stimuli) and 1500ms. It didn’t happen to me in a few runs, but it may be possible that the delay causes the raid to eat one tick of Blast Nova (I’m not sure if the first blast comes as soon as the channel starts). Again, another good thing to test, but also one tick is not going to kill anybody, and it’s arguably good to introduce some degree of imperfection.
3. Once Blast Nova stops channeling (i.e., all five cubes have been clicked and channeled simultaneously), bots will interrupt their cube clicking and go back to regularly scheduled activities. Again, I’ve introduced a randomized delay, this time between 200ms and 3000ms. Note that bots can easily be perfect at this—if I don’t do the randomized delay, they click and let go so fast that you can barely even see the beams appear. It’s so atrocious for immersion that I consider the lack of any randomization to be totally unacceptable for this strategy.
4. 48s after Blast Nova, bots will go back to their near-cube positions, rinse and repeat.

If an assigned cube clicker dies, another bot should be immediately assigned. All bots in the raid track the same timer so the new bot should step into the prior bot’s shoes. Of course, if Blast Nova is about to go off and a clicker dies next to a cube, you’re probably wiping because I didn’t assign backups to stand in place. That’s too much of a dad guild-level strategy even for me.

And that’s about it. Figuring out the cubes was a tremendous pain in the ass, but I’ve really enjoyed the learning process.


Zerathane:
Awesome work, i'll be able to pull this tomorrow afternoon and give it a try.

Dreathean:
This is awesome! Just wanted to add that I totally agree with including some degree of imperfection, it goes such a long way for immersion.

brighton-chi:
I need to refactor this significantly, but for now I've made one update as there is a risk of null pointer deference upon a wipe because Magtheridon disappears so just so quick fixes for that.

I noticed the ranged/healer spread action is trying to send them TO Magtheridon lol. So yeah that needs to be fixed.

brighton-chi:
Completely reworked this to make a separate trigger for each method, split up multipliers, and redo pretty much all of the movement logic.

I don't think I'm going to break up action methods--that will be something I'll do differently for the next strat (if there is a next one, hah).

src/strategy/raids/magtheridon/RaidMagtheridonHelpers.cpp
```diff
@@ -0,0 +1,191 @@
...
+            return false;
+        }
+    }
+
+    // Conflagration
+    GuidVector gos = *botAI->GetAiObjectContext()->GetValue<GuidVector>("nearest game objects");
+    for (const auto& goGuid : gos)
+    {
+        GameObject* go = botAI->GetGameObject(goGuid);
+        if (!go || go->GetEntry() != 181832) // Invisible game object at the center of Conflagration
```
kadeshar: Consider not using magic numbers

src/strategy/raids/magtheridon/RaidMagtheridonMultipliers.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "Playerbots.h"
+#include "WarlockActions.h"
+
+using namespace MagtheridonHelpers;
+
+// Don't do anything other than clicking cubes when Magtheridon is casting Blast Nova
+float MagtheridonUseManticronCubeMultiplier::GetValue(Action* action)
+{
+    Unit* magtheridon = AI_VALUE2(Unit*, "find target", "magtheridon");
+    if (magtheridon->HasUnitState(UNIT_STATE_CASTING) &&
```
kadeshar, line 21: magtheridon can be nullptr

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include "RaidMagtheridonHelpers.h"
+#include "Playerbots.h"
+
+using namespace MagtheridonHelpers;
+
+// Temporarily set bot sight distance to 150 to see the entire room
+bool MagtheridonSetBotSightTrigger::IsActive()
+{
+    float originalSightDistance = sPlayerbotAIConfig->sightDistance;
+    sPlayerbotAIConfig->sightDistance = 150.0f;
```
kadeshar: Im almost sure that shouldnt be done via config file. That is not global config for all bots?
Check RangeAction::Execute(Event event) i think this action dynamic change range via chat. You can analyze this code.

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    Unit* magtheridon = AI_VALUE2(Unit*, "find target", "magtheridon");
+    Unit* channeler = AI_VALUE2(Unit*, "find target", "hellfire channeler");
+    Group* group = bot->GetGroup();
+
+    bool result = group && (magtheridon && magtheridon->IsAlive() ||
+                  channeler && channeler->IsAlive());
+
+    if (!result)
+        sPlayerbotAIConfig->sightDistance = originalSightDistance;
```
kadeshar: Such restore dont cover all situations like teleporting out from magtheridon fight

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,234 @@
...
+        auto timerIt = magtheridonBlastNovaTimer.find(bot->GetMapId());
+        if (timerIt != magtheridonBlastNovaTimer.end())
+        {
+            time_t lastBlastNova = timerIt->second;
+            if (now - lastBlastNova >= 49)
+            {
+                return false;
+            }
+        }
+    }
```
kadeshar: MagtheridonSpreadRangedTrigger have similar code. There is no way to unified that?

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
brighton-chi: Hm, you might be right. I'll need to check this (for Gruul too).

nl-saw:
Thanks for the contribution, one request tho on first glance:

Please try and stick to braceless if statements where applicable. Braced returns hurt to look at.

i.e.

```
        if (since < spreadWaitSeconds)
        {
            return false;
        }
```

to

```
        if (since < spreadWaitSeconds)
            return false;
```

brighton-chi:
> Thanks for the contribution, one request tho on first glance:
> 
> Please try and stick to braceless if statements where applicable. Braced returns hurt to look at.
> 
> i.e.
> 
> ```
>         if (since < spreadWaitSeconds)
>         {
>             return false;
>         }
> ```
> 
> to
> 
> ```
>         if (since < spreadWaitSeconds)
>             return false;
> ```

Sure, I don't have a strong opinion one way or another. I floated the question on Discord a month ago actually and mostly got indifference with perhaps some slight preference to opt for braces.

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
brighton-chi: I'm not sure how RangeAction::Execute logic might apply here because honestly I don't know what exactly was causing the issue with bots no longer participating when too far away. Increasing the sight distance fixed the problem, and I don't know if RangeAction would do the same since it seems to be somewhat different and for modifying different interaction distances (but I don't really understand it).

However, with the default in the config modified to 100 yards to match normal player sight distance, there seems to no longer be an issue (so I think 100 is enough, even if 75 wasn't always enough). So I'll just remove this trigger (and from High King Maulgar too--I checked, and 100 works fine).

nl-saw:
The difference is readability. See the below suggestion from another PR as example.

Cheers

```
    {
        // Heirloom items scale with player level
        if (proto->Quality == ITEM_QUALITY_HEIRLOOM)
        {
            // Use player level as effective item level for heirlooms - Quality EPIC
            weight_ *= PlayerbotFactory::CalcMixedGearScore(lvl, ITEM_QUALITY_EPIC);
        }
        else
        {
            // Blend with item quality and level for normal items
            weight_ *= PlayerbotFactory::CalcMixedGearScore(proto->ItemLevel, proto->Quality);
        }
        return weight_;
    }
```
vs

```
    {
        // Heirloom items scale with player level
        // Use player level as effective item level for heirlooms - Quality EPIC
        // Else blend with item quality and level for normal items

        if (proto->Quality == ITEM_QUALITY_HEIRLOOM)
            weight_ *= PlayerbotFactory::CalcMixedGearScore(lvl, ITEM_QUALITY_EPIC);

        else
            weight_ *= PlayerbotFactory::CalcMixedGearScore(proto->ItemLevel, proto->Quality);

        return weight_;
    }
```

brighton-chi:
Oh I'm not arguing, and I get what you're saying. I'm fine with whatever so can take that approach.

nl-saw:
> Oh I'm not arguing, 

Oh, I did not take it as such. Cheers

brighton-chi:
I modified the code to leverage helpers a lot more and also redid a lot of the logic surrounding timers, which I'm trying to get more used to. I improved some other stuff too.

Sorry, I should I kept this on draft longer. I don't plan to work any more on this PR though, except to address comments and any issues from testing.

Zerathane:
I've pulled the most recent updates to this PR and attempted the boss a dozen or so times.
The PR was pulled onto a clean build with the most recent Core and Playerbots module updates.
Raid average ilv <105


<ins> Channeler Phase </ins>

The strategy does a really good, consistent job of assigning the various channeler's their marks and the tanks pick them up and move them to their locations, hunters help this with misdirection which I observed them casting. I've noticed if an offtank assigned to green is a feral druid, he will move to his position before casting bear form, so his aggro is a few seconds delayed but this hasn't caused my raid any issues.

I tried to slow this phase down to allow for as many Burning Abyssals to spawn as possible. With 3 Abyssals present my 3 warlocks are quick to banish them and reapply banish once it drops off. The PR writeup states the warlocks will then cast curse of tongues on the channelers to reduce their cast speed, however my damage and activity reports show only curse of elements being cast by all 3 warlocks.

One minor behaviour I noticed, which is inconsistent and hard to replicate is sometimes the offtanks will stay attacking the Burning Abyssals despite being banished and the fight moving to phase 2 and Magtheridon being free. This hasn't caused any issues as the main tank still aggros and positions Mag, but just a little behaviour I'm not sure if it's intentional or not. 

The below image shows my 2 offtanks and hunter pets attacking a banished Abyssal despite being in P2 with Magtheridon active.

WoWScrnShot_101925_115231

<ins> Magtheridon positioning </ins>

I observed no problems with the positioning of Magtheridon when entering P2. The main tank moves backwards towards the northern most wall of the room and Mag stays there for the duration of the fight. The rest of the raid do stay spread and use most of the space available in the room.

<ins> Manticron Cubes </ins>

This is definitely the most fun part of the strategy to observe and I've found the bots to an excellent job of moving to their assigned cubes with roughly 10 seconds left on the Blast Nova timer. I also tried to stress test this phase by randomly killing cube clickers and watching a new clicker be assigned and take its place. The following image shows the bots still executing this mechanic perfectly, even with the half of the raid, and original cube clickers dead.

WoWScrnShot_101925_120337

brighton-chi:
Thanks a bunch! You're right about Curse of Tongues--I ended up removing the mechanic because it seemed unnecessary due to how quickly bots react and was just giving Warlocks another thing to worry about. With how strong their DPS is at level 70 (the highest ranged DPS, and it's not even close) and the fact that they already had to control Abyssals, it felt like a waste of their capabilities.

hermensbas:
I will review soon, thanks for the amazing work

nl-saw (approved):

src/strategy/raids/magtheridon/RaidMagtheridonHelpers.cpp
```diff
@@ -0,0 +1,233 @@
...
+    Map* map = bot->GetMap();
+    if (!map)
+        return nullptr;
+
+    auto bounds = map->GetCreatureBySpawnIdStore().equal_range(dbGuid);
+    if (bounds.first == bounds.second)
+        return nullptr;
+
+    Creature* creature = bounds.first->second;
+    return creature;
```
kadeshar: Unneccesary variable
`return bounds.first->second;`

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+bool MagtheridonBurningAbyssalWarlockCCTrigger::IsActive()
+{
+    Unit* magtheridon = AI_VALUE2(Unit*, "find target", "magtheridon");
+    if (!magtheridon || bot->getClass() != CLASS_WARLOCK)
+        return false;
+
+    const GuidVector& npcs = AI_VALUE(GuidVector, "nearest hostile npcs");
+        return std::any_of(npcs.begin(), npcs.end(), [&](const ObjectGuid& npc)
```
kadeshar: Can we pass to lambda only what is required like

`return std::any_of(npcs.begin(), npcs.end(), [botAI](const ObjectGuid& npc) `
?

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
```diff
@@ -0,0 +1,706 @@
...
+    }
+
+    return false;
+}
+
+bool MagtheridonDPSPriorityAction::Execute(Event event)
+{ 
+    // Listed in order of priority
+    Creature* channelerSquare   = GetChanneler(bot, SOUTH_CHANNELER);
+    Creature* channelerStar = GetChanneler(bot, WEST_CHANNELER);
```
kadeshar: Introduce this variable after checking channelerSquare because there is possibility that will be not required at all. Similar with channelerCircle, channelerDiamond , channelerTriangle

src/strategy/raids/magtheridon/RaidMagtheridonHelpers.cpp
```diff
@@ -0,0 +1,233 @@
...
+}
+
+void MarkTargetWithIcon(Player* bot, Unit* target, uint8 iconId)
+{
+    if (!target)
+        return;
+
+    if (Group* group = bot->GetGroup())
```
kadeshar: Personally im not fun of such condition (im see few such if's in this PR). Code after that must assume that can be nullptr. Consider in such if's early return from method.

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
brighton-chi: I tried botAI and got a compile error. I changed & to this, which I think is more limited, and that worked.

hermensbas (dismissed):

hermensbas:
Leaving the final say to @noisiver

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
```diff
@@ -0,0 +1,717 @@
...
+    if (currentTarget && bot->GetVictim() != currentTarget)
+    {
+        return Attack(currentTarget);
+
+        if (!bot->IsWithinMeleeRange(currentTarget))
+        {
+            return MoveTo(currentTarget->GetMapId(), currentTarget->GetPositionX(), 
+                          currentTarget->GetPositionY(), currentTarget->GetPositionZ(), 
+                          false, false, false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
+        }
```
kadeshar: If is unreachable because return Attack

src/strategy/raids/magtheridon/RaidMagtheridonStrategy.cpp
```diff
@@ -0,0 +1,42 @@
...
+    triggers.push_back(new TriggerNode("magtheridon need to manage timers and assignments", NextAction::array(0,
+        new NextAction("magtheridon manage timers and assignments", ACTION_EMERGENCY + 1), nullptr)));
+
+    triggers.push_back(new TriggerNode("magtheridon burning abyssal spawned", NextAction::array(0,
+        new NextAction("magtheridon warlock cc burning abyssal", ACTION_RAID + 3), nullptr)));
+
+    triggers.push_back(new TriggerNode("magtheridon boss engaged by ranged", NextAction::array(0,
+        new NextAction("magtheridon spread ranged", ACTION_RAID + 2), nullptr)));
+
+    triggers.push_back(new TriggerNode("magtheridon need to pull west and east channelers", NextAction::array(0,
```
kadeshar: I think should be "magtheridon pulling west and east channelers" to work.

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,128 @@
...
+bool MagtheridonPullingWestAndEastChannelersTrigger::IsActive()
+{
+    Unit* magtheridon = AI_VALUE2(Unit*, "find target", "magtheridon");
+
+    Creature* channelerStar = GetChanneler(bot, WEST_CHANNELER);
+    Creature* channelerCircle = GetChanneler(bot, EAST_CHANNELER);
+
+    return magtheridon && bot->getClass() == CLASS_HUNTER &&
+           (channelerStar && channelerStar->IsAlive() || 
+            channelerCircle && channelerCircle->IsAlive());
```
kadeshar: Should be 
return magtheridon && bot->getClass() == CLASS_HUNTER &&
           ((channelerStar && channelerStar->IsAlive()) || (channelerCircle && channelerCircle->IsAlive()));
?

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,128 @@
...
+    Creature* channelerDiamond  = GetChanneler(bot, NORTHWEST_CHANNELER);
+    Creature* channelerTriangle = GetChanneler(bot, NORTHEAST_CHANNELER);
+
+    if (!magtheridon || botAI->IsHeal(bot) || botAI->IsMainTank(bot) || 
+        botAI->IsAssistTankOfIndex(bot, 0) && channelerDiamond && channelerDiamond->IsAlive() ||
+        botAI->IsAssistTankOfIndex(bot, 1) && channelerTriangle && channelerTriangle->IsAlive())
+        return false;
+
+    return channeler && channeler->IsAlive() || magtheridon && 
+           !magtheridon->HasAura(SPELL_SHADOW_CAGE);
```
kadeshar: Should be
return (channeler && channeler->IsAlive()) || (magtheridon && 
           !magtheridon->HasAura(SPELL_SHADOW_CAGE));
?

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
```diff
@@ -0,0 +1,128 @@
...
+    Unit* magtheridon = AI_VALUE2(Unit*, "find target", "magtheridon");
+    Unit* channeler = AI_VALUE2(Unit*, "find target", "hellfire channeler");
+
+    Creature* channelerDiamond  = GetChanneler(bot, NORTHWEST_CHANNELER);
+    Creature* channelerTriangle = GetChanneler(bot, NORTHEAST_CHANNELER);
+
+    if (!magtheridon || botAI->IsHeal(bot) || botAI->IsMainTank(bot) || 
+        botAI->IsAssistTankOfIndex(bot, 0) && channelerDiamond && channelerDiamond->IsAlive() ||
+        botAI->IsAssistTankOfIndex(bot, 1) && channelerTriangle && channelerTriangle->IsAlive())
+        return false;
```
kadeshar, line 56: Same question about this conditions

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
brighton-chi: Don't need this anyway. Removing.

src/strategy/raids/magtheridon/RaidMagtheridonStrategy.cpp
brighton-chi: Thank you. Changed the name in the triggers and didn't update this.

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
brighton-chi: I think it is the same either way but I can make this change if it's easier to read.

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
brighton-chi: Same as above, don't need the extra parens but I'll add if easier to read

src/strategy/raids/magtheridon/RaidMagtheridonTriggers.cpp
brighton-chi: The meaning is the same but I can change if easier to read.

brighton-chi:
Main tank movement seems kind of screwy all of a sudden. Maybe it's related to all the Core changes, which still are pretty busted for me with mobs taking weird winding paths and mobs and bots going through the floor and through walls. Gonna try to see if I can figure out what's going on here specifically later.

Edit: OK, fine now.

I forgot one important thing that I'll fix now (and for Gruul too). Casters can't be made to return true from Attack (or to return Attack). It screws with their DPS and in the case of Moonkins and Ele Shamans actually gets them stuck doing nothing. I think it's because if they're not close enough, they do not have the ability to autoattack (no wands or bows like other ranged and no logic to move to melee range to attack like melee).

So basically anytime Attack is called, unless it is only for melee, it can't return true. I use Attack to basically set a target and commence combat; if there is a better way to do it then that would be ideal, but I haven't figured it out yet.

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
```diff
@@ -305,7 +280,7 @@ bool MagtheridonAssignDPSPriorityAction::Execute(Event event)
         SetRtiTarget(botAI, "cross", magtheridon);
 
         if (bot->GetVictim() != magtheridon)
-            return Attack(magtheridon);
+            Attack(magtheridon);
```
kadeshar: I dont understand all this removals. Now action never return true which is wrong.

brighton-chi: Attack() does not work for moonkin druids and elemental shamans if you return true after it. They will freeze in place and do nothing because they will run that same loop repeatedly. My guess is it is because they cannot autoattack at range (like hunters and other casters) and also do not have logic to get into melee range to autoattack (like melee bots).

kadeshar: Still is wrong flow. If you dont want execute that by this class filter them in isUseless. If you want them to execute that action make own flag which change on true when one of attack is executed and return this flag.

Action always should have flow where Execute return true.

brighton-chi: I don't really know how to make it work. There should probably be a better way to get a bot to start attacking a target without forcing them to autoattack. It doesn't make sense for casters who can autoattack, either.

If this has to return true, then probably I'll have to filter them out, which makes the strategy worse since then you have two specs that are going to behave randomly.

brighton-chi: Well, I guess I could make them move into melee range to go whack the target if they are out of range and return true on the movement. It's kind of a stupid result but would be better than them doing nothing or not following any strategy.

I haven't tried that but if I'm right about what is causing the problem then that should get them acting again.

kadeshar: Simply mark target + change rti target wouldnt solve a problem?

brighton-chi: I'll give it a try and let you know. I did that before when I was first working on Maulgar strategies and it would not cause bots to switch targets properly, but I have more experience now so I definitely could have been doing something else back then to cause the problem.

Edit: Deleted the other stuff since it's not related, I'll bring it up separately if needed. Getting too verbose, sorry.

brighton-chi: Figured out how to make it work. Instead of using GetVictim() as the condition, use GetTarget() and then SetSelection before returning Attack(). Then even if the bot can't autoattack, the function won't enter the loop again so the bot doesn't lock up.

brighton-chi:
Looks like a lot happened but it's due to needing an extra level of indents inside the helper namespaces. I made this commit just to move the enums in the helper header inside the helper namespace because I am running into issues with conflicting definitions in other strategies I am working on.

brighton-chi:
Oh also this one needs a wiki update. I will coordinate with Dreathean on that after merger.

kadeshar:
Please next time avoid such refactoring in approved PR.

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
```diff
@@ -0,0 +1,703 @@
...
+    if (channelerCircle && channelerCircle->IsAlive())
+        MarkTargetWithCircle(bot, channelerCircle);
+
+    // After first three channelers are dead, wait for Magtheridon to activate
+    if ((!channelerSquare || !channelerSquare->IsAlive()) &&
+        (!channelerStar   || !channelerStar->IsAlive()) &&
+        (!channelerCircle || !channelerCircle->IsAlive()))
+    {
+        const Location& position = MagtheridonsLairLocations::WaitingForMagtheridonPosition;
+        if (!bot->IsWithinDist2d(position.x, position.y, 2.0f))
```
noisiver, line 34: Brackets used for single line if

brighton-chi:
> Please next time avoid such refactoring in approved PR.

As I mentioned, it just looks like a bunch changed because I fixed one indent. Or are you saying if something looks like it changes a lot even if it doesn't, just don't make the change?

Beyond that, all I did was move enums into the helper namespace, which was necessary because it is causing me build errors for other stuff I'm working on (and presumably could do the same for other people too).

kadeshar:
If its refactoring then it can wait for next PR. Now we need make all proccess again (2 approves + tags)

brighton-chi:
> If its refactoring then it can wait for next PR. Now we need make all proccess again (2 approves + tags)

Yeah, I preferred not to but thought if something causes compile issues with future boss strategies it should be addressed now. I get what you mean though.

Silly that the whole thing shows up changed with one indent. You would think the comparison could just show the indent instead of making it look like the file was basically deleted and remade.

kadeshar:
Even single line change reverting approves.

brighton-chi:
Yeah I hear you.

hermensbas (approved):

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
hermensbas: There is line split though due the length of the statement, i dont mind the curly braces in this situation. Not sure AC guidelines say in this matter.

noisiver: Code in other places (in playerbots) do not use curly brackets in this instance though, which is why I requested that change.

noisiver: If you think it's best to leave it like this I am fine with that, I just based it on the rest of the code we use. Let me know.

noisiver:
The AzerothCore codestyle workflow threw this at me:
```
Starting AzerothCore CPP Codestyle check...

Please read the C++ Code Standards for AzerothCore:
https://www.azerothcore.org/wiki/cpp-code-standards

Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 136
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 141
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 148
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 164
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 203
Please use the 'auto const&' syntax instead of 'const auto&': /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 182
Please use the 'auto const&' syntax instead of 'const auto&': /root/test/src/magtheridon/RaidMagtheridonHelpers.cpp at line 198
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 36
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 95
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 160
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 178
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 193
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 215
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 353
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 386
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 387
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 411
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 543
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 577
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 590
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 600
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 612
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 614
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 635
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 643
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 645
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 682
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 685
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 693
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 696
Please use the 'auto const&' syntax instead of 'const auto&': /root/test/src/magtheridon/RaidMagtheridonActions.cpp at line 318
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonHelpers.h at line 64
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 11
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 20
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 41
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 53
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 58
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 69
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 89
Please use the 'auto const&' syntax instead of 'const auto&': /root/test/src/magtheridon/RaidMagtheridonTriggers.cpp at line 103
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonMultipliers.cpp at line 49
Trailing whitespace found: /root/test/src/magtheridon/RaidMagtheridonMultipliers.cpp at line 66

Multiple blank lines check : Passed
Trailing whitespace check : Failed
GetCounter() check : Passed
Misc codestyle check : Failed
GetTypeId() check : Passed
NpcFlagHelpers check : Passed
ItemFlagHelpers check : Passed
ItemTemplateFlagHelpers check : Passed

Please fix the codestyle issues above.
```

src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
brighton-chi: However this plays out is pretty relevant to me because this scenario comes up throughout my PRs. I was doing these without brackets because I was interpreting no brackets around a single statement, regardless of number of lines. Gromleq wanted this to be bracketed so I changed throughout. I will comport with whatever you guys want but just need a consistent approach--LMK.

nl-saw: https://www.azerothcore.org/wiki/cpp-code-standards

hermensbas (approved):

PR #1732, opened 2025-10-13
Heirloom quality auto-equip calculation implemented

kikiviki:
Edited quality_blend code to include a special case for heirloom items

This change sets the heirloom item to the player level, and assigns it an arbitrary quality of EPIC for purposes of calculation.
If the item is not an heirloom, the default calculation is run which assigns proto->ItemLevel and proto->Quality instead, using the default calculation values.

This may need tweaked slightly based on heirloom item scaling, but **should** prevent bots from equipping a green item over an heirloom item.

### DIFF
--- a/src/factory/StatsWeightCalculator.cpp
+++ b/src/factory/StatsWeightCalculator.cpp
@@ -94,16 +94,14 @@ float StatsWeightCalculator::CalculateItem(uint32 itemId, int32 randomPropertyId
     if (enable_quality_blend_)
     {
         // Heirloom items scale with player level
+        // Use player level as effective item level for heirlooms - Quality EPIC
+        // Else - Blend with item quality and level for normal items
         if (proto->Quality == ITEM_QUALITY_HEIRLOOM)
-        {
-            // Use player level as effective item level for heirlooms - Quality EPIC
             weight_ *= PlayerbotFactory::CalcMixedGearScore(lvl, ITEM_QUALITY_EPIC);
-        }
+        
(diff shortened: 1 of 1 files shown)


nl-saw:
Thanks for the contribution, one request tho on first glance:

Please try and stick to braceless if statements where applicable. Braced returns hurt to look at.

i.e.

```

    {
        // Heirloom items scale with player level
        // Use player level as effective item level for heirlooms - Quality EPIC
        // Else blend with item quality and level for normal items

        if (proto->Quality == ITEM_QUALITY_HEIRLOOM)
            weight_ *= PlayerbotFactory::CalcMixedGearScore(lvl, ITEM_QUALITY_EPIC);

        else
            weight_ *= PlayerbotFactory::CalcMixedGearScore(proto->ItemLevel, proto->Quality);

        return weight_;
    }
```

kikiviki:
Thank you for the feedback.

My code style was based off a function from the same file, StatsWeightCalculator::CalculateRandomProperty, which has the braced ifs / returns.

I am happy to comply moving forward.

nl-saw:
Thanks, you mind making the changes here then? Feel free to amend what u used as base then as well, as there is no need to copy inferior code and use that as an excuse ;)

kikiviki:
@nl-saw for my reference in the future, is there a style guide for best practices I can follow ?

nl-saw:
@kikiviki no, not yet. We should define it at some point tho.

PR #1737, opened 2025-10-16
FIX: ICC - default return position for BQL

nl-saw:
as otherwise the code after the original return 1.0f; will never be executed.


hermensbas:
Can you add description of intend?

nl-saw:
Sure, it's pretty self-explanatory tho - the code after the original return 1.0f; will never be executed.

PR #1739, opened 2025-10-18
Added shaman default combat strategy names on top of custom

hermensbas:
We should stick with the default names or should be at least part of the list. These default names are there for a certain level of abstraction and unity between classes. Adding addtional specialization shouldn't be issue as long the default strategies name such as caster, dps, melee, healer, offhealer are supported out of the box.


hermensbas:
Tagged @Wishmaster117 (MultiBot)

Wishmaster117:
> Tagged @Wishmaster117 (MultiBot)

What part i have to modify on Multibot?

hermensbas:
> > Tagged @Wishmaster117 (MultiBot)
> 
> What part i have to modify on Multibot?

More headsup then anything else. If you make use of things like "co +heal" that those commands should work for each class. The shaman overhaul introduced removing those and replacing them with e.g. "co +resto". This PR ensures shaman supports those default strategies again.

PR #1742, opened 2025-10-19
Fix typo in server loading message

brighton-chi:
Damn it.


hermensbas:
^^ ik3 would be mad :P

brighton-chi:
> ^^ ik3 would be mad :P

ike3?

😂

Dreathean:
Typo fix aside, it is kinda funny too how outdated the server message is when compared to the readme also - that server message must be from Celguar's cmangos version. (and it is also not lost on me that there is a typo in the SS here for the readme lol).
image

PR #1745, opened 2025-10-19
Fix: Prevent addClass bots from getting realm firsts

NoxMax:
Follow up to #1487, which targeted randombots only and addClass were still able to get realm firsts. This fix targets ~~anything that's not a real player~~ both.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -220,7 +220,8 @@ public:
 
     bool OnPlayerBeforeAchievementComplete(Player* player, AchievementEntry const* achievement) override
     {
-        if ((sRandomPlayerbotMgr->IsRandomBot(player) || sRandomPlayerbotMgr->IsAddclassBot(player)) && (achievement->flags == 256 || achievement->flags == 768))
+        if ((sRandomPlayerbotMgr->IsRandomBot(player) || sRandomPlayerbotMgr->IsAddclassBot(player)) &&
+            (achievement->flags & (ACHIEVEMENT_FLAG_REALM_FIRST_REACH | ACHIEVEMENT_FLAG_REALM_FIRST_KILL)))
         {
             return false;
         }


noisiver:
This looks like it'll stop altbots and not just rndbots and addclass bots from getting these realm first achievements, which they should be able to.

NoxMax:
I did think about alts, and wondered if I should simply make it (sRandomPlayerbotMgr->IsRandomBot(player) || sRandomPlayerbotMgr->IsAddclassBot(player)) but I considered bots are bots, including alts. But is the general consensus that altbots should be treated as real players achievement-wise?

noisiver:
Yes, altbots are characters created by the player and shouldn't be included in this. They should be able to get realm first achievements.

NoxMax:
Then explicit random and addclass targeting only it is.

noisiver:
Fantastic, thank you very much. I think it makes sense because you have to actually create the characters and more or less manage them yourself (depending on how detailed you want to be) so they shouldn't be seen the same way as rndbots or addclass bots.

I think it's makes sense to add addclass bots though because they are really just a variation of rndbots.

src/Playerbots.cpp
```diff
@@ -220,7 +220,7 @@ class PlayerbotsPlayerScript : public PlayerScript
 
     bool OnPlayerBeforeAchievementComplete(Player* player, AchievementEntry const* achievement) override
     {
-        if (sRandomPlayerbotMgr->IsRandomBot(player) && (achievement->flags == 256 || achievement->flags == 768))
+        if ((sRandomPlayerbotMgr->IsRandomBot(player) || sRandomPlayerbotMgr->IsAddclassBot(player)) && (achievement->flags == 256 || achievement->flags == 768))
```
hermensbas: -Magic numbers

-The condition checks; cheap checks first expensive last. So achievement checks first.

src/Playerbots.cpp
```diff
@@ -220,8 +220,7 @@ class PlayerbotsPlayerScript : public PlayerScript
 
     bool OnPlayerBeforeAchievementComplete(Player* player, AchievementEntry const* achievement) override
     {
-        PlayerbotAI* botAI = GET_PLAYERBOT_AI(player);
-        if (botAI && !botAI->IsRealPlayer() && (achievement->flags == 256 || achievement->flags == 768))
+        if ((sRandomPlayerbotMgr->IsRandomBot(player) || sRandomPlayerbotMgr->IsAddclassBot(player)) && (achievement->flags == 256 || achievement->flags == 768))
```
privatecore:
```cpp
achievement->flags & (ACHIEVEMENT_FLAG_REALM_FIRST_REACH | ACHIEVEMENT_FLAG_REALM_FIRST_KILL)
```
Please update those flags as well; it's painful to see so many magic numbers. ;)

NoxMax: I see their enums in AC. Got it!

nl-saw (approved):

PR #1752, opened 2025-10-20
FIX: part of logic in picking quest reward

nl-saw:
Avoids dereferencing begin() after confirming the set is empty.


hermensbas:
Good catch

Issue #1689, opened 2025-10-01
RandomBot stuck in ring equip loop.


Bryoal3k3:
**Bug Description**
Ive started a new realm with all bots starting from level 1 and noticed a bug around the deadmines levels (About 20). The randombot in group would have 1 ring with stats (Int, Spr etc.) and 1 ring with just +Resist. When the bot would loot another stat ring they would just get into a loop of equiping the second stat ring over the first one instead of upgrading the resist ring. Noticed it with stat ring being in the first slot and resist ring in the second.

**Commit Hash**
Current.

**How To Reproduce Bug**
Put a ring with stats in first ring slot, and a ring with resist in the second ring slot. Give the randombot a better stat ring and they will just keep flipping the two in the first slot.

**Expected Behavior**
Bot equips both stat rings and not the resist

**Screenshots**

**Modules**
Please list all modules used as many are known to cause conflicts with Playerbots.

**Playerbot Settings**
Number of bots, scaling settings, etc if performance related.

**System**
OS: [e.g. Windows, Linux]
Hardware: [e.g. CPU if performance related]

**Additional Context**

kadeshar:
Provide rings ids + bot class/spec which had problem

Bryoal3k3:
![Image](https://github.com/user-attachments/assets/4c10228c-6da0-4578-ace1-daaf9941d525)

![Image](https://github.com/user-attachments/assets/f086a32b-9699-48e5-b5ce-ee7e5bb3d7a9)

Been a really busy day. Havent had a chance to try other characters or items. So far I have a Destro Lock and the issue is with Clay "Of the .... " rings (item 11993) and Quartz "Of the ...." rings (item 11965).

I tried in a isolated environment. Gave the lock the first ring "Clay of the Gorilla", he equiped it. Next I gave him "Quartz ring of Shadow Resistance" and he put it in the second slot. After I gave him a second "Clay ring of the Bear" and he just equips it over the ring in slot 1. After a minute or so I get the whisper he is equiping the other "Clay Ring" but again all he does is just put it over the other Clay ring. 

Im assuming the bot knows that the Clay rings with actual stats on them is better than the resist ring, but it doesnt know that its in Ring Slot 2

If you look at the screenshot, after casting create soulstone he tries to equip it. Not sure what bot logic there is but it seems after certain actions he tries to equip it which is why I was getting spammed whispers in the dungeons run last night.

brighton-chi:
[...]

PR #1757, opened 2025-10-22
Fixes equip bug with random suffix rings

avirar:
Fixes issue #1689 by correctly checking the random properties of ring (and trinket, but there aren't any that I'm aware of) slot items

### DIFF
--- a/src/strategy/actions/EquipAction.cpp
+++ b/src/strategy/actions/EquipAction.cpp
@@ -290,29 +290,19 @@ void EquipAction::EquipItem(Item* item)
                     bool betterThanFirst = newItemScore > firstItemScore;
                     bool betterThanSecond = newItemScore > secondItemScore;
 
+                    // Early return if new item is not better than either equipped item
+                    if (!betterThanFirst && !betterThanSecond)
+                        return;
+
                     if (betterThanFirst && betterThanSecond)
                     {
                         // New item is better than both - replace the worse of the two equipped items
                         if (firstItemScore > secondItemScore)
-                        {
                             dstSlot++; // Replace second slot (worse)
-                        }
                         // else: keep dstSlot as-is (replace first slot)
                     }
-                    else if (betterThanSecond && !betterThanFirst)
-                    {
-                        // Only better than second slot - replace it
-                        dstSlot++;
-                    }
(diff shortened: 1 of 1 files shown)


src/strategy/actions/EquipAction.cpp
```diff
@@ -271,19 +271,48 @@ void EquipAction::EquipItem(Item* item)
...
                     }
+                    else if (betterThanFirst && !betterThanSecond)
+                    {
+                        // Only better than first slot - replace it (dstSlot already points to it)
+                    }
+                    else
+                    {
+                        // New item is not better than either equipped item - don't equip
+                        return;
+                    }
```
kadeshar: If first if would be 
```
if (!betterThanFirst && !betterThanSecond)
    return;
```
then empty else if is redundant. Also good practice is early exit method.

src/strategy/actions/EquipAction.cpp
```diff
@@ -271,19 +271,48 @@ void EquipAction::EquipItem(Item* item)
...
+                    bool betterThanFirst = newItemScore > firstItemScore;
+                    bool betterThanSecond = newItemScore > secondItemScore;
+
+                    if (betterThanFirst && betterThanSecond)
+                    {
+                        // New item is better than both - replace the worse of the two equipped items
+                        if (firstItemScore > secondItemScore)
+                        {
+                            dstSlot++; // Replace second slot (worse)
+                        }
```
kadeshar: Single line statement in if should be without {}. 
https://www.azerothcore.org/wiki/cpp-code-standards section Brackets

src/strategy/actions/EquipAction.cpp
avirar: Done

src/strategy/actions/EquipAction.cpp
avirar: and done!

nl-saw (approved):

biship:
I've been using this for 4 days and no issues to report.

Dreathean:
> I've been using this for 4 days and no issues to report.

Hey Biship! Thank you for checking out this PR, would you be able to elaborate on any testing done with this PR? Confirmed functionality with regards to the random properties on rings?

PR #1758, opened 2025-10-22
Rewrite RandomPlayerbotFactory for improved maintainability and future updates

privatecore:
- Rewrite the `RandomPlayerbotFactory` constructor and remove `availableRaces` from scope.
- Add a new utility method, `IsValidRaceClassCombination` -- check for proper race and class combinations.
- Rewrite the `CombineRaceAndGender` utility -- keep it DRY and easy to read.
- Remove unused attributes from the `RandomPlayerbotFactory` class and the `CreateRandomBot` method.

### DIFF
--- a/src/RandomPlayerbotFactory.cpp
+++ b/src/RandomPlayerbotFactory.cpp
@@ -61,17 +61,20 @@ Player* RandomPlayerbotFactory::CreateRandomBot(WorldSession* session, uint8 cls
 {
     LOG_DEBUG("playerbots", "Creating a new random bot for class: {}", cls);
 
-    const uint8 expansion = sWorld->getIntConfig(CONFIG_EXPANSION);
     const bool alliance = static_cast<bool>(urand(0, 1));
 
     std::vector<uint8> raceOptions;
     for (uint8 race = RACE_HUMAN; race < MAX_RACES; ++race)
     {
+        // skip disabled with config races
+        if ((1 << (race - 1)) & sWorld->getIntConfig(CONFIG_CHARACTER_CREATING_DISABLED_RACEMASK))
+            continue;
+
         // Try to get 50/50 faction distribution for random bot population balance.
         // Without this check, races from the faction with more class options would dominate.
         if (alliance == IsAlliance(race))
         {
-            if (IsValidRaceClassCombination(race, cls, expansion))
+            if (IsValidRaceClassCombination(race, cls, sWorld->getIntConfig(CONFIG_EXPANSION)))
                 raceOptions.push_back(race);
         }
     }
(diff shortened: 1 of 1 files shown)


NoxMax:
If I understand correctly, now every account is garneted 5A/5H bots, correct? Also have you checked that it plays nice with `EnablePeriodicOnlineOffline` and non 50/50 `RandomBotAlliance/HordeRatio`?

privatecore:
> If I understand correctly, now every account is garneted 5A/5H bots, correct? Also have you checked that it plays nice with `EnablePeriodicOnlineOffline` and non 50/50 `RandomBotAlliance/HordeRatio`?

No, this logic was from the original author -- I just added a comment to clarify why this check is there (`alliance == IsAlliance(race)`). The main difference from the old one is the simplified logic for `raceOptions` and the check for `_playerInfo` versus hardcoded `availableRaces` values.

privatecore:
@NoxMax  Updated the comment -- hope its now clearer, cuz at first, it seems like its not needed here since we already pick the race randomly later.

NoxMax:
I see. I did some stuff around that code few months ago and had to work around its faction distribution. So when I read "Force 50/50" I thought that the system now does actually do that, but you are right, on closer look `const bool alliance = static_cast<bool>(urand(0, 1));` is still there.

To be clear the system does not create an exact 50/50 ratio, it creates an almost 50/50 ratio. So maybe the word "force" can give a false illusion of exactness. CreateRandomBot is more like tossing a coin 1000 times, you will most likely get something close to 500/500, but not necessarily exactly that.

src/RandomPlayerbotFactory.cpp
```diff
@@ -19,188 +19,96 @@
...
-        case RACE_BLOODELF:
-            return static_cast<NameRaceAndGender>(static_cast<uint8>(NameRaceAndGender::BloodelfMale) + gender);
         default:
-            LOG_ERROR("playerbots", "The race with ID %d does not have a naming category", race);
-            return static_cast<NameRaceAndGender>(static_cast<uint8>(NameRaceAndGender::GenericMale) + gender);
+            baseIndex = NameRaceAndGender::GenericMale;
+            break;
     }
+
+    return static_cast<NameRaceAndGender>(static_cast<uint8>(baseIndex) + ((gender > 1) ? 0 : gender));
```
kadeshar: Avoid using magic numbers. Maybe something like

    return static_cast<NameRaceAndGender>(static_cast<uint8>(baseIndex) + ((gender >= GENDER_NONE) ? GENDER_MALE : gender));

src/RandomPlayerbotFactory.cpp
```diff
@@ -19,188 +19,96 @@
...
-    availableRaces[CLASS_WARRIOR].push_back(RACE_UNDEAD_PLAYER);
-    availableRaces[CLASS_WARRIOR].push_back(RACE_TAUREN);
-    availableRaces[CLASS_WARRIOR].push_back(RACE_TROLL);
-    if (expansion >= EXPANSION_THE_BURNING_CRUSADE)
-    {
-        availableRaces[CLASS_WARRIOR].push_back(RACE_DRAENEI);
-    }
+    // skip expansion races if not playing with expansion
+    if (expansion < EXPANSION_THE_BURNING_CRUSADE && (race == RACE_BLOODELF || race == RACE_DRAENEI))
+        return false;
```
kadeshar, line 50: I know its not part of PR but you added support for config. Consider support condition for CharacterCreating.Disabled.RaceMask config

src/RandomPlayerbotFactory.cpp
privatecore: Nice catch!

src/RandomPlayerbotFactory.cpp
privatecore: There’s just one point: it seems better to add the check in `IsValidRaceClassCombination`, but the loops for iterating through classes in `CreateRandomBots` and the loop for races in `CreateRandomBot` also look more logical, as they would reduce the number of iterations. What do you think?

kadeshar (dismissed):

kadeshar (approved):

hermensbas:
Didnt notice any regression during testing

Issue #1753, opened 2025-10-21
Playerbots won't fly


grace-of-hainaut:
**Bug Description**
Since the latest update, playerbots will no longer fly. They'll mount onto the expected mount/transform into flight form, but they'll path along the ground underneath the player character rather than following them directly.

**Commit Hash**
7e810f8174b2e70d300310c0d740e52ebea655a2 (azerothcore/Playerbots [`4f9b4d5`](https://github.com/mod-playerbots/azerothcore-wotlk/commit/4f9b4d5470b754933049d3f037fb8c880bfebc66))

**How To Reproduce Bug**
Have a character and at least one playerbot with the appropriate Riding skill and an appropriate mount/skill. Ensure the bot is following, then mount and leave the ground. The bot will mount but only follow along the ground.

**Expected Behavior**
The bots should follow through the air.

**Modules**
None

**Playerbot Settings**
1-4 playerbots

**System**
OS: Linux
Hardware: Intel Core i5 13th gen

**Additional Context**
I've update the mapping data to v17, and have tried with a fresh playerbots.conf file

hermensbas:
Can anyone confirm?

nl-saw:
Confirmed.

grace-of-hainaut:
I know corroboration to my own report is of limited use, but for completeness I just tested with a completely fresh repo clone, build and database. I created a couple of test characters, set them to 77, gave them flying and mounts and am still seeing the same thing:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b12a4226-da4e-49ea-87dc-0299448edf37" />


Worldserver: AzerothCore rev. fc1d8e2a64c8 2025-10-21 10:38:27 +0200 (Playerbot branch) (Unix, RelWithDebInfo, Static)
Playerbots commit: 6e1c9114df9be02acad3accca0df575b491eaa53

hermensbas:
Does this happen on every map? And you are absolute sure this wasnt the case before the core merge? It would help to narrow the issue down.

grace-of-hainaut:
> Does this happen on every map? And you are absolute sure this wasnt the case before the core merge? It would help to narrow the issue down.

[...]

PR #1761, opened 2025-10-22
Fix Playerbots won't fly #1753

Wishmaster117:
### Summary
Playerbots now correctly follow their master in the air. The regression introduced after the core merge (module 50ac6e5 + core f96e7c5) caused bots to mount/shape-shift but stay glued to the ground.

### What’s changed

Set all required flight flags when a flight aura is present (flying mount or druid flight form): CAN_FLY, DISABLE_GRAVITY, FLYING.

Don’t early-return from Follow(...) on 2D distance only; also require small ΔZ to consider “in range”.

When bot/master/vehicle are flying, disable mmaps ground pathing (allowPath=false) for MovePoint/MovePointBackwards to allow true 3D movement.

Keep existing APIs and style; no new functions.

### Why
Ground pathing (mmaps) and 2D-only range checks prevented ascent and produced “flying mount stuck on the ground”.

### Testing

Mount a flying mount (or druid flight form), take off, /follow.

Bot should take altitude and maintain distance in 3D.

Ground behavior unchanged; follow works as before on foot.

Vehicle flying: follows in 3D; passengers without control remain unchanged.

This solve issue: https://github.com/mod-playerbots/mod-playerbots/issues/1753

### DIFF
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -199,7 +199,8 @@ bool MovementAction::MoveTo(uint32 mapId, float x, float y, float z, bool idle, 
     {
         VehicleSeatEntry const* seat = vehicle->GetSeatForPassenger(bot);
         Unit* vehicleBase = vehicle->GetBase();
-        generatePath = vehicleBase->CanFly();
+        // If the mover (vehicle) can fly, we DO NOT want an mmaps path (2D ground) => disable pathfinding
+        generatePath = !vehicleBase || !vehicleBase->CanFly();
         if (!vehicleBase || !seat || !seat->CanControl())  // is passenger and cant move anyway
             return false;
 
@@ -207,14 +208,20 @@ bool MovementAction::MoveTo(uint32 mapId, float x, float y, float z, bool idle, 
         if (distance > 0.01f)
         {
             MotionMaster& mm = *vehicleBase->GetMotionMaster();  // need to move vehicle, not bot
+            // Disable ground pathing if the bot/master/vehicle are flying
+            auto isFlying = [](Unit* u){ return u && (u->HasUnitMovementFlag(MOVEMENTFLAG_FLYING) || u->IsInFlight()); };
+            bool allowPathVeh = generatePath;
+            Unit* masterVeh = botAI ? botAI->GetMaster() : nullptr;
+            if (isFlying(vehicleBase) || isFlying(bot) || isFlying(masterVeh))
+                allowPathVeh = false;
             mm.Clear();
             if (!backwards)
             {
-                mm.MovePoint(0, x, y, z, FORCED_MOVEMENT_NONE, 0.0f, 0.0f, generatePath);
+                mm.MovePoint(0, x, y, z, FORCED_MOVEMENT_NONE, 0.0f, 0.0f, allowPathVeh);
             }
             else
             {
-                mm.MovePointBackwards(0, x, y, z, generatePath);
+                mm.MovePointBackwards(0, x, y, z, allowPathVeh);
             }
             float speed = backwards ? vehicleBase->GetSpeed(MOVE_RUN_BACK) : vehicleBase->GetSpeed(MOVE_RUN);
             float delay = 1000.0f * (distance / speed);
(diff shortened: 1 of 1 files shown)


hermensbas:
You of a bitch you beat me to it! haha 

I was rounding up the PR lel, but great work. And ur solutions pretty much same as mine. So i will and i prolly will add some comments in the code once merged.

Great job

hermensbas:
Alright, we almost there. I have add some changes though. Still testing, as ground calculations still have to be added which prevent bots from flying while they should walk.

But the switch map MMAP en VMAP en liftoff and landing works.

hermensbas:
But we do that after, i will finished my other test and code review. When good enough i will merge. And improve it from there where needed. i like your base better then mine tbh i did some spline calculations which i wasnt happy about.

hermensbas:
initial test are green, druids, master or slave, groups, solo, swimming, bg, i will do quick PR but we gonna improvement once merged.

hermensbas:
I am kinda suprised though that when enforced splines(vmap) calcs instead in path (mmap), you dont force the destination path. But it seems its not needed . 

I will experiment with it abit later and add the ground calculations so they start walking when close to the ground. But great work, atleast i can sleep now and stop coding haha

Wishmaster117:
> I am kinda suprised though that when enforced splines(vmap) calcs instead in path (mmap), you dont force the destination path. But it seems its not needed .
> 
> I will experiment with it abit later and add the ground calculations so they start walking when close to the ground. But great work, atleast i can sleep now and stop coding haha

Haha same,  I also expected we’d need to "force a dest spline".
Turned out the only real blocker was mmaps flattening everything to 2D.
Once we stop generating a path while flying and let MoveSpline do its thing, it naturally tracks Z, so no extra “force destination path” needed.

Switching to ground when you’re close to terrain, love it. If you try that, maybe add a bit of hysteresis to avoid flip-flop (e.g. walk when |ΔZ| < 4m, go back to flying when |ΔZ| > 7m) plus a tiny cooldown (300–500ms). That keeps it smooth.

And yeah, get some sleep, we’ve all been there 😄

brighton-chi:
Is this a bot problem or a fix that is being implemented to address a Core movement issue? I noticed that Nightbane no longer has a flying animation and now walks in midair--I'm wondering if that issue is related.

Wishmaster117:
> Is this a bot problem or a fix that is being implemented to address a Core movement issue? I noticed that Nightbane no longer has a flying animation and now walks in midair--I'm wondering if that issue is related.

This looks like a Core movement regression, not a Playerbots change.
The PR here only touches bot-side actions (MovementActions.cpp) and doesn’t modify Core creature movement.
Nightbane walking midair instead of playing a flying animation usually means Core flags aren’t set as expected (CAN_FLY / DISABLE_GRAVITY / FLYING) or a recent change is forcing ground pathing (mmaps) instead of a 3D spline.
Could you share your Core commit hash and data extract version? If you can reproduce it on a clean Core (without the module), that would confirm it’s Core-related. Also worth checking creature_template.InhabitType (AIR) and what flags Nightbane has during the event.
Our bot fix merely sets the proper flight flags for bots only and disables ground pathing while flying; it doesn’t affect bosses.

Take a look at official azerothcore git you would see lot of problems since they change movements stuffs.

hermensbas:
Bot problem due a refactoring the core MotionMaster PointMoventGenerator which is used throughout the core. This change only relates to bot movement action that used certain functions and mechanics of that MotionMaster class. And TakeOff and moveland are not auto initialized anymore. So the triggers had to be added manual now together with vmap/mmap movement switching

vmap are spline movement for 3d env so used when flying and mmap pathing is 2d terrain paths so when walking.

Handling flying was mainly done by the core after setting a certain trigger.

Nightbane has to be addressed in the core though.

hermensbas:
I still have to cleanup the code a little bit, but also retest BG. But after a few good nights of sleep

Wishmaster117:
Yes, but the fixes will have to be made on the core side, and to avoid breaking the behavior of all the creatures, all the scripts will have to be modified. They completely messed up on the Azerothcore side in this refactor.

brighton-chi:
Hm okay, seems they've got some things to sort out on the main repo.

Nightbane's issue is visual. I don't know how that relates to flags, though IsFlying() still returns true for him when he's supposed to be flying (but appears to be walking in midair).

Wishmaster117:
> Hm okay, seems they've got some things to sort out on the main repo.
> 
> Nightbane's issue is visual. I don't know how that relates to flags, though IsFlying() still returns true for him when he's supposed to be flying (but appears to be walking in midair).

In fact, on the azerothcore side, their intention is good, the MotionMaster / PointMovementGenerator refactor was designed to make the calls explicit (who decides to fly/walk, who chooses 3D spline "vmap" vs. 2D path "mmap").
The goal of such a refactor is to avoid automatic "magic" (auto-TakeOff/MoveLand) that cause desyncs and costly re-planes, and to centralize consistent behaviors for everyone (creatures, players, vehicles). Doxygen reflects this evolution with new parameters (e.g., ForcedMovement) and direct use of MoveSpline.
But this brings a side effect: what relied on automatic triggers (e.g., a script that simply performed MovePoint assuming the Core "takeoff/land" and chose the correct mode) no longer works unless you manually add the flight flags and the correct path mode.  So you see Nightbane walking in the air.
This is a Core issue that's already been reported: there have been reports on Azerothcore of Nightbane descending/walking in the air instead of staying in the air. So now, scripts must explicitly set CAN_FLY/DISABLE_GRAVITY/FLYING and disable ground path generation while flying.

nl-saw:
Nice work  :+1:

PR #1762, opened 2025-10-23
Continuation to fix issue: Playerbots won't fly #1753 : Keep bots on the ground when the master is grounded on a flying mount

Wishmaster117:
### Keep bots on the ground when the master is grounded on a flying mount

Okay, this was a little more complicated to set up than I expected...

**Problem:**
After the Core movement refactor, bots could “skim” or hover at ground level and flip between fly/walk when the master was on a flying mount but still on the ground. They also sometimes kept flight flags briefly after dismounting.

**What this PR changes (bot-side only):**
Ground lock by master: Bots only start flying if the master is actually flying (FLYING / IsInFlight). We ignore CAN_FLY (because a flying mount on the ground sets it).

Force 2D pathing on ground: While the master is grounded, we force mmaps (2D) in all MoveTo(...) branches (including vehicles). In flight, we switch to vmap (3D).

Immediate landing clamp: If the master is grounded, bots land immediately (cooldown bypassed) to prevent ground-level hovering.

Flag cleanup on ground: When not flying and the master is grounded, we clear residual CAN_FLY / DISABLE_GRAVITY to avoid “hover” or tiny hops.

Hysteresis + small cooldown: Keep anti-ping-pong (walk if |ΔZ_ground| < 4 m, fly if > 7 m) and a short toggle cooldown (200 ms) for stability.

**Now:**

The bot no longer treats a grounded flying mount as “master is flying.”

Path generation matches the mode: 2D when the master walks, 3D only when the master actually flies.

Residual flight flags are cleaned up, so bots don’t stay suspended after dismount.

Tested scenarios: walking with a flying mount, takeoff/landing transitions, druid flight form, vehicles, slopes/bridges.

**Tuning knobs :**
MASTER_GROUND_LOCK (2.5f) — tolerance for “master is grounded”.
Hysteresis thresholds (4 m / 7 m) and toggle cooldown (200 ms).

### DIFF
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -42,6 +42,11 @@
 #include "Vehicle.h"
 #include "WaypointMovementGenerator.h"
 #include "Corpse.h"
+#include "GameTime.h"     // GameTime::GetGameTimeMS()
+#include <map>            // cooldown per bot for toggle fly/ground
+
+// // Per bot flight/ground toggle cooldown (ms)
+static std::map<uint32 /*guidLow*/, uint32 /*lastToggleMs*/> s_flightToggleMs;
 
 MovementAction::MovementAction(PlayerbotAI* botAI, std::string const name) : Action(botAI, name)
 {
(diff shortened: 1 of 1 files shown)


hermensbas:
Lemme a have look into it first

nl-saw:
Hey guys, I appreciate the efforts.

jorgeEF:
> You have #1761? They should be flying now just have some lingering issues.

I've just updated to last commit and they are flying again, thanks

hermensbas:
Still working on solution which mitigates and actual core changes, but it takes time. But lets keep this PR on draft meanwhile. It might be helpful i wont able to solve it in the "correct" way

PR #1763, opened 2025-10-23
Fix PlayerbotAI constructors' members order and wrong type comparison (uint32 -> int32)

privatecore:
Fix warnings: -Wsign-compare and -Wsign-compare.

I would also like to mention that there are issues with the following methods:

`IsMainTank` -- in the last loop, it returns the result of comparing the first alive tank it finds with the current bot, which seems odd.
`IsBotMainTank` -- it is used in only two places in the code (Yogg-Saron strategy) and has a completely different logic for checking. In the last loop, it only checks for a suitable tank-bot with a lower index (if it's not the bot itself). Additionally, there is a logic issue in the loop: if none of the conditions are met after the first iteration, the method returns `false`.

Can someone from the maintainers take a look at this section of the code for possible errors and logic issues?

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -106,14 +106,14 @@ void PacketHandlingHelper::AddPacket(WorldPacket const& packet)
 PlayerbotAI::PlayerbotAI()
     : PlayerbotAIBase(true),
       bot(nullptr),
+      master(nullptr),
+      accountId(0),
       aiObjectContext(nullptr),
       currentEngine(nullptr),
+      currentState(BOT_STATE_NON_COMBAT),
       chatHelper(this),
       chatFilter(this),
-      accountId(0),
-      security(nullptr),
-      master(nullptr),
-      currentState(BOT_STATE_NON_COMBAT)
+      security(nullptr)
 {
     for (uint8 i = 0; i < BOT_STATE_MAX; i++)
         engines[i] = nullptr;
@@ -128,9 +128,9 @@ PlayerbotAI::PlayerbotAI()
 PlayerbotAI::PlayerbotAI(Player* bot)
     : PlayerbotAIBase(true),
       bot(bot),
+      master(nullptr),
       chatHelper(this),
       chatFilter(this),
-      master(nullptr),
       security(bot)  // reorder args - whipowill
 {
     if (!bot->isTaxiCheater() && HasCheat((BotCheatMask::taxi)))
@@ -1865,11 +1865,11 @@ int32 PlayerbotAI::GetAssistTankIndex(Player* player)
     {
         return -1;
     }
+
     int counter = 0;
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-
         if (!member)
         {
             continue;
@@ -1879,11 +1879,13 @@ int32 PlayerbotAI::GetAssistTankIndex(Player* player)
         {
             return counter;
         }
+
         if (IsTank(member, true) && group->IsAssistant(member->GetGUID()))
         {
             counter++;
         }
     }
+
     return 0;
 }
 
@@ -2196,14 +2198,15 @@ bool PlayerbotAI::IsMainTank(Player* player)
             break;
         }
     }
+
     if (mainTank != ObjectGuid::Empty)
     {
         return player->GetGUID() == mainTank;
     }
+
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-
         if (!member)
         {
             continue;
@@ -2214,6 +2217,7 @@ bool PlayerbotAI::IsMainTank(Player* player)
             return player->GetGUID() == member->GetGUID();
         }
     }
+
     return false;
 }
 
@@ -2235,8 +2239,7 @@ bool PlayerbotAI::IsBotMainTank(Player* player)
         return true;  // If no group, consider the bot as main tank
     }
 
-    uint32 botAssistTankIndex = GetAssistTankIndex(player);
-
+    int32 botAssistTankIndex = GetAssistTankIndex(player);
     if (botAssistTankIndex == -1)
     {
         return false;
@@ -2250,8 +2253,7 @@ bool PlayerbotAI::IsBotMainTank(Player* player)
             continue;
         }
 
-        uint32 memberAssistTankIndex = GetAssistTankIndex(member);
-
+        int32 memberAssistTankIndex = GetAssistTankIndex(member);
         if (memberAssistTankIndex == -1)
         {
             continue;


brighton-chi:
I always have to reset my main tank whenever I log off and back in. I'm not sure if this weirdness is unique to me or if there is an issue with the code surrounding main tank determination.

privatecore:
> I always have to reset my main tank whenever I log off and back in. I'm not sure if this weirdness is unique to me or if there is an issue with the code surrounding main tank determination.

Can you please provide more detailed information on how you create a group/raid, whats group/raid composition (classes+specs), how you set the main tank, and how you reset the main tank? This will help me dig deeper into the logic and try to identify the issue.

brighton-chi:
> > I always have to reset my main tank whenever I log off and back in. I'm not sure if this weirdness is unique to me or if there is an issue with the code surrounding main tank determination.
> 
> 
> 
> Can you please provide more detailed information on how you create a group/raid, whats group/raid composition (classes+specs), how you set the main tank, and how you reset the main tank? This will help me dig deeper into the logic and try to identify the issue.

Sure.

I create a raid using Multibot, which I believe effectively is just using buttons to process .playerbots bot add. My composition has all classes and most specs, including all tank specs. I set the main tank by typing /mt botname. When I log back in, the main tank will still have the main tank icon but will not be treated as such (e.g., other tanks will fight with the main tank for aggro). I type /mt botname again and that fixes it.

Sometimes, when I log back in, there will be two tanks with the main tank icon (the original main tank and another bot that has seemingly been set as the main tank). I think the original main tank's persisting icon is just visual at that point. If I do /mt botname, it will clear the extraneous mark. 

I have even had the main tank icon given to a bot that was not a tank (but had been at an earlier point in time). My latest raid has been in place for a few days, and a couple of days ago, I had a warlock set as a tank for Leotheras. Yesterday, on two logins, the warlock had the main tank icon even though it had long been switched back to DPS spec and had the "tank" strategy removed. On another occasion, the warlock did not have the main tank icon but was treated as such following a relog (I realized it because the warlock was executing actions coded for IsMainTank() in the Magtheridon strategies).

To be clear, I've never had any problems if I do /mt botname again after logging back in. Somehow I fail to remember to do that most of the time despite constantly recompiling and relogging as I work on boss strategies so I end up with the issue screwing things up a lot.

brighton-chi:
> Fix warnings: -Wsign-compare and -Wsign-compare.
> 
> I would also like to mention that there are issues with the following methods:
> 
> `IsMainTank` -- in the last loop, it returns the result of comparing the first alive tank it finds with the current bot, which seems odd. `IsBotMainTank` -- it is used in only two places in the code (Yogg-Saron strategy) and has a completely different logic for checking. In the last loop, it only checks for a suitable tank-bot with a lower index (if it's not the bot itself). Additionally, there is a logic issue in the loop: if none of the conditions are met after the first iteration, the method returns `false`.
> 
> Can someone from the maintainers take a look at this section of the code for possible errors and logic issues?

I think @kadeshar should look at IsBotMainTank, but I think you are right, the second-to-last return false should be removed.

For IsMainTank, I think it works OK. In that case, "player" is not the bot that runs the function necessarily. So the function can return any member of the group. I know that's how it works in practice. Specifically, it will return the player or bot with the main tank flag or, if there isn't one, the last player or bot tank that joined the group (because GetFirstMember iterates through members in reverse join order). Well, the exception is the weirdness I mentioned above about what happens if I log off and back on.

privatecore:
Hmm, the only place where `IsMainTank()` and `MEMBER_FLAG_MAINTANK` interact is in `void RazorscaleBossHelper::AssignRolesBasedOnHealth()`:
```cpp
    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
    {
        Player* member = ref->GetSource();
        if (member && botAI->IsMainTank(member))
            group->SetGroupMemberFlag(member->GetGUID(), false, MEMBER_FLAG_MAINTANK);
    }
```
This may be the issue if the check fails and the flag is not removed from a previously set member. However, inside `SetGroupMemberFlag`, this flag is still removed from each member of the group. So, it’s kind of hard to say what’s causing the issue.

For testing purposes, try executing the MySQL query when the bug occurs:
`SELECT * FROM group_member WHERE memberGuid IN (<BOTS_GUIDS_SEPARATED_WITH_COMMAS>);`
The most interesting part will be the `memberFlags` field, which stores all the member flags if they are set. This will help determine if it’s some sort of visual bug or if the member flags are messed up.

> For IsMainTank, I think it works OK. In that case, "player" is not the bot that runs the function necessarily. So the function can return any member of the group. I know that's how it works in practice. Specifically, it will return the player or bot with the main tank flag or, if there isn't one, the last player or bot tank that joined the group (because GetFirstMember iterates through members in reverse join order). Well, the exception is the weirdness I mentioned above about what happens if I log off and back on.

Maybe you’re right; the second part looks like a fallback if `MEMBER_FLAG_MAINTANK` is not set.

brighton-chi:
> Hmm, the only place where `IsMainTank()` and `MEMBER_FLAG_MAINTANK` interact is in `void RazorscaleBossHelper::AssignRolesBasedOnHealth()`:
> 
> ```c++
>     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
>     {
>         Player* member = ref->GetSource();
>         if (member && botAI->IsMainTank(member))
>             group->SetGroupMemberFlag(member->GetGUID(), false, MEMBER_FLAG_MAINTANK);
>     }
> ```
> 
> This may be the issue if the check fails and the flag is not removed from a previously set member. However, inside `SetGroupMemberFlag`, this flag is still removed from each member of the group. So, it’s kind of hard to say what’s causing the issue.
> 
> For testing purposes, try executing the MySQL query when the bug occurs: `SELECT * FROM group_member WHERE memberGuid IN (<BOTS_GUIDS_SEPARATED_WITH_COMMAS>);` The most interesting part will be the `memberFlags` field, which stores all the member flags if they are set. This will help determine if it’s some sort of visual bug or if the member flags are messed up.
> 
> > For IsMainTank, I think it works OK. In that case, "player" is not the bot that runs the function necessarily. So the function can return any member of the group. I know that's how it works in practice. Specifically, it will return the player or bot with the main tank flag or, if there isn't one, the last player or bot tank that joined the group (because GetFirstMember iterates through members in reverse join order). Well, the exception is the weirdness I mentioned above about what happens if I log off and back on.
> 
> Maybe you’re right; the second part looks like a fallback if `MEMBER_FLAG_MAINTANK` is not set.

How curious, the flags are definitely messed up. I logged out and back in and ran the SQL query for all of my tanks and also my warlock who I previously used as a tank (note I switched it to a tank a while after creating the group and have long since switched it back to DPS). The results indicate there are two bots with the main tank flag, the warrior that I had set as the main tank before I logged out, and the warlock. The Blizzard UI does show both of them as the main tank.

privatecore:
> How curious, the flags are definitely messed up. I logged out and back in and ran the SQL query for all of my tanks and also my warlock who I previously used as a tank (note I switched it to a tank a while after creating the group and have long since switched it back to DPS). The results indicate there are two bots with the main tank flag, the warrior that I had set as the main tank before I logged out, and the warlock. The Blizzard UI does show both of them as the main tank.

So, I checked the original behavior from AzerothCore, and it seems that it's normal for `memberFlags` not to be updated when changing group flags (such as assist, maintank, etc.), since `SendUpdate` is called for the group every time a player logout/login, where each member gets updated. I tried changing the maintank and assist roles with the UI and commands, logout/login, but I never encountered two tanks...

You can try this:
Shut down the server, truncate the `groups` and `group_member` tables, avoid the Ulduar raid, and just play while keeping an eye on the member roles. When you first find two tanks, stop and try to remember what happened before, such as: you added a new bot, removed a bot, the bot logged out, the bot logged in, the server crashed, the raid wiped, the raid was recreated, etc.

The most likely reason is a server crash and the subsequent loading of the group and members with incorrect flags.

kadeshar:
Simple no crash on startup test should be enough

Issue #1768, opened 2025-10-25
Bots keep whispering last master when removed from group


nl-saw:
**Bug Description**
I noticed that a bot keeps whispering me after being removed from the group that they were were drinking stuff.

**Commit Hash**
[The hash of the current commit.](https://github.com/mod-playerbots/mod-playerbots/commit/7237b154e0e57a3e9c06414da4d55e7e31ad22a0)

**How To Reproduce Bug**
Have a bot in your team that is low health and drinking/eating. Then remove the bot from the group. Observe it keeps whispering you what it is consuming.

**Expected Behavior**
No whisper comms after being removed from group, except for Bye!

**Modules**
None

**Playerbot Settings**
Number of bots, scaling settings, etc if performance related.

Zerkenn:
I can confirm this happens to me all the time with RDF groups. If I leave a RDF group most of the time the random bots don't whisper "Goodbye!". They keep randomly whispering me while I play even though we are no longer in the same group. Sometimes they even teleport out with me after I leave the group and follow me around as if we are in a group together.

I ether have to message my old group members the leave command or relog.

If I join a new RDF group and then leave that new one sometimes the bots from the first group then finally whisper "Goodbye!".

nl-saw:
Fixed with https://github.com/mod-playerbots/mod-playerbots/pull/1774

nl-saw:
@Zerkenn Can you confirm on your end?

Zerkenn:
> [@Zerkenn](https://github.com/Zerkenn) Can you confirm on your end?


Test passed:

### Raid Groups With 2 Real Players
- Created a raid group and invited another real player and added some random bots.
- Left group on the party leader and all bots transferred ownership to the other real player left in the party.
- Left the group on the second real player. All the random bots whispered goodbye and went on about their business.

### RDF 1 Real Player
- Joined a random dungeon finder group.
- Left the group and all the bots whispered goodbye.

### RDF 2 Real Players
- Joined a random dungeon finder group with another real player.
- Left the dungeon finder group and the random bots transferred ownership to the real player.
- Left the dungeon on the second real player and they all whispered goodbye to the second player.

### Battlegrounds
- Joined several different battlegrounds and then left.
- The battleground continued when the real player left.

### Arena
- Joined several arena skirmishes.
- No issues found.


All seems to be working properly after #1774 . 
Great work! 
Thank you all for your contributions.

PR #1774, opened 2025-10-25
Fix. Leave group actions 

Celandriel:
Fix for issue #1768 where the bot master was not getting reset. 
I also cleaned up leave group action to focus up function scope.
I moved the resets from #612 to the actual leaving function.

Disclosure: LLMs were NOT used in authoring this PR.

Test cases to consider for testers.
1. Master disbands group with random bots. Bots should go about their business. 
2. If you can dual box, create a raid where two real player both invite random bots. One player leaves group, their bots should also leave. (edge case, if a random bot that is supposed to leave the group becomes leader they may disband the whole group.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -800,17 +800,11 @@ void PlayerbotAI::Reset(bool full)
 
 void PlayerbotAI::LeaveOrDisbandGroup()
 {
-    if (!bot || IsRealPlayer())
+    if (!bot || !bot->GetGroup() || IsRealPlayer())
         return;
-    if (bot->GetGroup())
-    {
-        WorldPacket* packet = new WorldPacket(CMSG_GROUP_DISBAND);
-        bot->GetSession()->QueuePacket(packet);
-    }
-    if (sRandomPlayerbotMgr->IsRandomBot(bot))
-        SetMaster(nullptr);
-    Reset(true);
-    ResetStrategies();
+
+    WorldPacket* packet = new WorldPacket(CMSG_GROUP_DISBAND);
+    bot->GetSession()->QueuePacket(packet);
 }
 
 bool PlayerbotAI::IsAllowedCommand(std::string const text)


Celandriel:
I'm not sure what I did to the code_format file, but I can't reset it

Zerkenn:
So far this is what I have found from testing:

The issue with player bots not resetting when leaving RDF groups seems to be fixed. They no longer whisper "Goodbye!" when you leave the group but they are no longer in control of the player after leaving.

There is another issue I found.

1. Invite another real player to a group and convert to raid.
2. On the raid leader account invite some randombots
3. leave the raid on the raid leader account
4. the entire raid is disbanded but the 2nd account now is in control of some if not all of the bots while not being in a group. Whispering them leave has them respond "Goodbye!" but they still dont reset. They keep following you and responding "Goodbye!" Every single time you whisper them. You have to logout of the 2nd account to reset the bots.


video of issue
https://www.youtube.com/watch?v=DP1Kx_MKUow

Celandriel:
I thought the two real players could be a problem.
Some clarifying questions. 
Is this with only random bots or altbots as well?
Alt bots and addClass bots are intentionally excluded from leaving and must be logged out by command. 

Did the first leader invite all random bots? Or did the first add some and the second more? And then the seconds bots remain in group? 

Edit: Just saw the video. 
It looks like the master is being passed on and not reset correctly. Almost as if they do not all receive the same message. 
Will investigate triggers further.

Celandriel:
So what I think is happening, and the main impact of the change #1708 is that by not directly removing the bot from the group the way it used to be, on the next DoNextAction the current code identifies that the bot is in a group with no master, and then sets a new master. See https://github.com/mod-playerbots/mod-playerbots/blob/7237b154e0e57a3e9c06414da4d55e7e31ad22a0/src/PlayerbotAI.cpp#L1368 

My idea is to introduce a LeavingGroup flag that causes it to ignore the behavior and then resets when the bot is out of group.

Edit: A better way may actually be to not reset the master in the leave action but make it happen in the DoNextAction to prevent conflicts.

Celandriel:
I'm going to stop trying to fix this from my phone.
This commit should be the one to consider if anyone feels like testing before I get to it.
https://github.com/mod-playerbots/mod-playerbots/pull/1774/commits/a9662f710f3d9ad4e9109aac4de7d9e153c8d95e

src/PlayerbotAI.cpp
```diff
@@ -1363,16 +1362,15 @@ void PlayerbotAI::DoNextAction(bool min)
     PlayerbotAI* masterBotAI = nullptr;
     if (master)
         masterBotAI = GET_PLAYERBOT_AI(master);
-
+    
+    PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
+    if (!botAI)
+    {
```
hermensbas: Even though i personally prefer curly braces the code guideline states no curly brackets in this scenario.

hermensbas:
> So what I think is happening, and the main impact of the change #1708 is that by not directly removing the bot from the group the way it used to be, on the next DoNextAction the current code identifies that the bot is in a group with no master, and then sets a new master. See
> 
> https://github.com/mod-playerbots/mod-playerbots/blob/7237b154e0e57a3e9c06414da4d55e7e31ad22a0/src/PlayerbotAI.cpp#L1368
> 
> My idea is to introduce a LeavingGroup flag that causes it to ignore the behavior and then resets when the bot is out of group.
> 
> Edit: A better way may actually be to not reset the master in the leave action but make it happen in the DoNextAction to prevent conflicts.

Hmmmz possible down-side of using worldpackets ("client/bot" -> core) is that the actual leaving of the group is being delegated to the mapThread is queued and is handled possibly a server tick (or a few) later along the road. DoNextaAction suggest its a chain of responsibility pattern so basically a follow up could handled in same server tick can cause problems when not checked proper. So certain states that are being checked in same server-tick might give a false/positive of very short lived state e.g. whether the bots is still in group meanwhile in best case next server-tick its the bot might have left the group. Not sure its related, But i forgot about the facts the actions are executed in the chained context.  


I think in most situations this is handled with flags states to prevent this from happening.

Celandriel:
I 

> > So what I think is happening, and the main impact of the change #1708 is that by not directly removing the bot from the group the way it used to be, on the next DoNextAction the current code identifies that the bot is in a group with no master, and then sets a new master. See
> > https://github.com/mod-playerbots/mod-playerbots/blob/7237b154e0e57a3e9c06414da4d55e7e31ad22a0/src/PlayerbotAI.cpp#L1368
> > 
> > My idea is to introduce a LeavingGroup flag that causes it to ignore the behavior and then resets when the bot is out of group.
> > Edit: A better way may actually be to not reset the master in the leave action but make it happen in the DoNextAction to prevent conflicts.
> 
> Hmmmz possible down-side of using worldpackets ("client/bot" -> core) is that the actual leaving of the group is being delegated to the mapThread is queued and is handled possibly a server tick (or a few) later along the road. DoNextaAction suggest its a chain of responsibility pattern so basically a follow up could handled in same server tick can cause problems when not checked proper. So certain states that are being checked in same server-tick might give a false/positive of very short lived state e.g. whether the bots is still in group meanwhile in best case next server-tick its the bot might have left the group. Not sure its related, But i forgot about the facts the actions are executed in the chained context.
> 
> If that makes any sense, hehe

It does make sense and I think its something we need to be aware of with how to handle any destructive event. nl-saw noticed that even while forcing a group reset the next donextaction call that happened almost simultaneously would reset the master. This code only manages the master state based on the server side state. Logging out is another one I want to take a look at next and likely needs another approach to handle.

hermensbas:
Ye its very tricky situation, u want world thread to handle the leaveGroup but you want  botAI handle/act on most recent state changes. So i guess the usage of flags comes into play, but that being that. Its very difficult when the rest of code is inconsistent. Maybe i should make some documentation about threading models and how to deal with them in a correct way?

nl-saw:
Did some testing with this just now. Seems the one problem to tackle with it is to have bots elect a new human player as master, if one is present in current group.

hermensbas:
> Did some testing with this just now. Seems the one problem to tackle with it is to have bots elect a new human player as master, if one is present in current group.

The electing should be done by checking the bots in group and not having a flag which states its flagged to leave the group? Maybe that flag already already exists, and can be added manually when calling leaveOrDisbandGroup or we need to introduce one? This prevents wierd situations where you elect new master while next server tick the elected master is kicked, or logged out. Its not directly related to this particular issue, but could introduce new wierd situation if not dealt with proper.

hermensbas:
> > > Did some testing with this just now. Seems the one problem to tackle with it is to have bots elect a new human player as master, if one is present in current group.
> > 
> > 
> > The electing should be done by checking the bots in group and not having an flag which states its flagged to leave the group? Maybe that flag already already exists, and be added manually or need to introduce one? This prevents wierd situations where you elect new master while next server tick the elected master is kicked, or logged out.
> 
> Maybe we should take one step back. Did we ever get confirmation changing leaves via worldpackets solved anything? I'm concerned it could otherwise be more confusing/harder to deal with than necessary due to the 'split logic'

Yes and its necessary to prevent race conditions. The whole model of the core is arranged around the fact world state has to be changed/handled by the world thread which will delegate the events if needed to one or more mapThreads. When playing solo its very likely the whole group is in the same map, if not, or playing with multiple real players u will get race conditions due invalid objects which causses the server to crash.

E.g. movement on the map can handled in map thread itself since its only relevant for that state of that map, a local teleport from A to B on same map could executed on the mapThread. A map to map teleport has to be executed through the world-thread since multiple threads are involved which are handled by the world thread. 

So dealing with state changes of a unit/player which are queued for an action (and likely will handled very soon), are being flagged so you know whats planned to happen and can act on those flags. Atleast thats my understanding of the core, but i am not expert on the matter either. But i do know for a fact cross state maps changed having to run against the world-thread to prevent ACCESS exception and server crashed. Since the bots live in both worlds, server AI side and mimic client actions can be a bit complicated. And the fact that world packets are design for the communication between client and server, which can be annoying when wanting to deal with live state for AI decisions 

So the general guideline would be:

-If your action has impact or could be relevant for the world-state (aka multiple maps) then it should be executed through the world thread through the usage of world packets.
-If the action only impact or is only relevant for the current map it should be save to directly call code changing the states of objects.

Not entirely sure about this but:
When dealing with queued tasks while we never know exactly when the actual task is being executed on the map(Threads), as in could be next server tick but could also be like 20 ticks later for example. So for that reason i assume we see a lot of flags being added to units/players which indicates the current or expected state change so we can act on those flags in combination of the current object states.

nl-saw:
I saw it, too late, tired. Going to sleep now

Celandriel:
This isnt working correctly right now.

Celandriel:
It seems ok now. Setting a new master only works when the group leader transfers  leadership to another player. Then on leaving or the DoNextAction should find a new master.
If you don't pass master the group gets destroyed too quickly for the bots to select a new master and end up leaving.

Depending on where you are the bots may take time to reactivate. If im in a low level area with high level bots they will take time to reset and teleporting elsewhere.

nl-saw:
I gave this some testing, and functionally it seems to be working pretty well.

Celandriel:
I just found another conflict. 
By doing some logging there are three ways that bots end up leaving groups,
Receiving and processing the 1. Party packet, 2. UpdateAIGroupMembership, 3 DoNextAction
The one the order is random based on ai timings server messages etc. 

the main UpdateAI triggers the following commands
https://github.com/mod-playerbots/mod-playerbots/blob/5d3e64800f5281b4cf9d3fc4f25b27bc6a46d1dd/src/PlayerbotAI.cpp#L358-L361

and UpdateAIInternal then calls
https://github.com/mod-playerbots/mod-playerbots/blob/5d3e64800f5281b4cf9d3fc4f25b27bc6a46d1dd/src/PlayerbotAI.cpp#L482

I think this section

https://github.com/Celandriel/mod-playerbots/blob/aa6a7bfc997f7d654253e4cfc9ce81e8868df289/src/PlayerbotAI.cpp#L1369-L1400
Should be merged with UpdateAIGroupMembership to have a single manager of groups and masters instead of competing actions trying to modify the master/group state

src/PlayerbotAI.cpp
```diff
@@ -1363,67 +1365,15 @@ void PlayerbotAI::DoNextAction(bool min)
     PlayerbotAI* masterBotAI = nullptr;
     if (master)
         masterBotAI = GET_PLAYERBOT_AI(master);
-
+
```
nl-saw: Sneaky tab

nl-saw:
> I think this section
> 
> https://github.com/Celandriel/mod-playerbots/blob/aa6a7bfc997f7d654253e4cfc9ce81e8868df289/src/PlayerbotAI.cpp#L1369-L1400
> Should be merged with UpdateAIGroupMembership to have a single manager of groups and masters instead of competing actions trying to modify the master/group state

Yeah I think that would make sense.

Also given that in
```cpp
    // Test BG master set
    if (group && (!master || (masterBotAI && !masterBotAI->IsRealPlayer())))
```

this may have been some test at some point, didn't dive in the full history, but indeed having things compete in this manner is not great at all.

Celandriel:
Just pushed the changes. 
From the testing ive done it seems acceptable. 
I have not specifically tested the Alterac Valley case that the code considers, or the LFG, but from a simple invite/uninvite/leave perspective it seems appropriate.

One big change in the group logic was that previously the UpdateAIGroupMembership only iterated through all bots when in LFG not normal groups. This changes the logic to always call search for another real player. (Except in AV where other bots are acceptable masters) and relegates all logic to finding a new master to the new helper FindNewMaster. I did this so that I can check on triggering a group leave action and deciding if a bot should leave. 
This must be done in both places because you cannot predict which event happens first (processing the message, or update cycle for the bot), so you need redundancy to not leave a group prematurely. In the future this could be extended to logout non alts and addclass, but that is out of scope.

ScoobyPwnsOnU:
Tested in multiple scenarios.

1 player and 2 rndbots:
They followed me -> i left party -> they both stood there for about 5 seconds then took off in 2 different directions

1 player and 4 altbots:
They followed me -> i left party -> they kept following me

1 player and 4 addclassbots:
They followed me -> i left party -> they kept following me

1 player, 2 rndbots, 1 altbot, and 1 addclassbot:
They followed me -> i left party -> the altbot and addclassbot kept following me, the 2 rndbots wandered off

3 players multiboxing, 1 altbot, and 1 rndbot:
They followed me -> i left party, the altbot dropped party also -> the party is now 2 players and an rnd bot -> the rnd bot starts following the new party leader -> i leave party on that new leader -> the party is now 1 player and 1 rndbot -> the rnd bot starts following the last player in the party and newest party leader -> i leave party on that player, leaving the rndbot with nobody left -> the rnd bot wanders off

All of these scenarios sound like they played out correctly. ✅

nl-saw:
PlayerbotAI::FindNewMaster() could certainly use a refactor, but we can do that at a later stage with another PR and for now keep the code as it was.

nl-saw (dismissed):

Wishmaster117:
@Celandriel , I'm not sure I fully understand the logic, but it seems to me there's some confusion in "Player* PlayerbotAI::FindNewMaster()".

GET_PLAYERBOT_AI(groupLeader) returns nullptr for a real (non-bot) player.
So if the leader is a real player, leaderBotAI == nullptr, the condition (!leaderBotAI || !leaderBotAI->IsRealPlayer()) is true, the leader isn't assigned as the new master, and we go looking for someone else.

Conversely, we only take the leader if leaderBotAI exists and leaderBotAI->IsRealPlayer() is true, which is strange (a "real" AI bot?).

As a result, the real leader is almost never chosen, which will cause erratic transitions (or a lack of reassignment) in two-player raids.

After that, it seems to me there's some confusion between `playerMaster` and `newMaster`. In the loop, we have:

* If `memberBotAI` exists (meaning `member` is a bot), we test `memberBotAI->IsRealPlayer()` to set `playerMaster`. This will never trigger (a bot is not `IsRealPlayer()`), so `playerMaster` remains useless/unassigned in these branches.

* For real players, where `memberBotAI == nullptr`, we directly assign `newMaster = member; break;` (outside of Battlegrounds), without prioritizing the leader or checking the lead transfer. We then end up taking the first real member found, which doesn't seem to be the desired policy.

The AV (Battlegrounds) case is too nested; okay for now, but the leader priority should remain clear.
The AV block is specific and can remain, but the priority rule outside of Battlegrounds should be:
"If the leader is a valid real player, we take the leader otherwise, no opportunistic reassignment." .

Am I wrong? I haven't tested the modifications, just read the code.

nl-saw:
Overall, the function works for most cases, but there are a few areas that might benefit from some refactoring and clearer intent. None of these are blockers, just things to consider for maintainability and readability (with summary of LLM):

Null Safety:
Some paths (e.g., groupLeader, group, or GET_PLAYERBOT_AI()) might dereference nulls under rare conditions. Adding early returns or explicit null checks would make the logic more robust.

Flow Complexity:
The nested if/continue structure makes it hard to follow the master selection flow. Refactoring into smaller helper functions (like FindBestRealPlayerCandidate() or FindBestBotCandidate()) could make it easier to reason about.

Master Selection Logic:
playerMaster and newMaster are both used as potential results but overlap in purpose. Consolidating them into a single bestMaster variable might reduce confusion.

Battleground Conditions:
The battleground-specific checks (especially AV) are quite specialized. It might be worth centralizing this into a dedicated helper for BG selection, or at least documenting why AV is treated differently.

Performance Considerations:
For large groups (like in AV), repeatedly calling functions like GET_PLAYERBOT_AI() and IsInSameRaidWith() for every member might add up. We could cache or reorder some checks for efficiency.

IsInSameRaidWith reliability:
Depending on the core version, this function can behave inconsistently for non-raid parties. Might be worth confirming or simplifying the check (e.g., comparing GetGroup() pointers directly).

Readability:
Adding some comments or clearer naming (like bestRealPlayer, bestBotFallback) would make intent more obvious for future maintainers.

None of these are urgent, but addressing them gradually could make this logic easier to extend and debug later on.

Zerathane:
I ran through some scenarios similar to ScoobyPwnsOnU and found the same results. The PR looks to resolve the issue of Rndbots following the player when leaving group/raid, Furthermore if a second real player is in the party the bots will switch to player to as their master if player one leaves.

I tested a few instances of Alterac Valley and found the bug seems to persist however. When the player leaves the battleground using the leave battleground button on the minimap, bots in the players party of the BG will still see the player as their master, despite being in AV and not being grouped with the player. 

The image below shows a bot identifying the player as a master despite being in AV. I was not grouped with the bot.
botpartypr

Celandriel:
Downgrade to draft while completing refactor based on comments. 
Thank you all for the comments.

src/PlayerbotAI.cpp
```diff
@@ -4130,6 +4057,69 @@ bool IsAlliance(uint8 race)
...
+Player* PlayerbotAI::FindNewMaster()
+{
+    // Ideally we want to have the leader as master.
+    Player* newMaster = nullptr;
+    Player* playerMaster = nullptr;
+
+    // Only allow real players as masters unless in battleground.
+    if (Group* group = bot->GetGroup())
+    {
+        Player* groupLeader = GetGroupMaster();
```
nl-saw: Should verify that groupLeader exists before proceeding

Celandriel: So I thought about it, but getplayerbotai is guarded against nullptr and it complicated the code with another if statement.

hermensbas:
Can some describe the exact issues we trying to solve. Just to ensure we all on same page and looking in the same direction. I see alot of issues described which are not linked to world packets leaving group and some might be? Bots leaving a party and still following you around for example is not, this was case before and after the world packets impl. 

So which exact issues are trying to solve?

Celandriel:
So the issue that prompted this was random bots not resetting properly on leaving group.

Alts and addclass are not considered because with no master they just get stuck. 

Resolving this issue identified a series of conflicting events related bots receiving packets, game state changes, and internal Ai updates.

A player leaving a group triggers the server sending a PARTY_OP_LEAVE message to be handled by the bot, while also destroying the group if that player is the leader. So what can happen is the internal update process could cause group updates at the updateaigroupmembership function or donextaction, or handling the world packet.

Then as I was editing I thought the scope of the Leave in leavegroupaction was too broad, because it was adding on another layer of verifications that could fail in a function that is used in multiple places. I simplified it to it handles random bots, and in the future we can also handle alts and addclass logging out. The conditional checking is moved to the main members of the class and depends on the message received whether the bot should follow it. One such condition was trying to find a new master before leaving the group, which led to a new handler FindNewMaster being created from existing code in DoNextAction. 

For the leaveordisbandgroup the only issue here was that sometimes by the time this function was processed the bots were no longer in the group, and the resets wouldn't happen because the first conditional was a reset check.
So I split it off into three sections leave group, reset master, general resetting.

Then for the final refactoring it didn't makes sense to me to have master and group checking occur twice in every Ai tick in both updateaiinternal and donextaction so the logic got merged in a unified place and I renamed the helper UpdateAiGroupAndMaster.

hermensbas:
Thanks for the summary, gonna review and test tonight aswell.


EDIT: will be tomorrow i am afraid had some other things i had todo IRL.

src/PlayerbotAI.cpp
nl-saw: Ok it was checked in other places, which is what triggered me. Thanks for checking.

nl-saw:
Did some testing and the good news is that they now choose human leader if one present in the group, that is good. However, on Leave command, most of them now still stand still, and few just walk around a little. I'm pretty convinced this is due to all ```botAI->Reset();``` were removed. I think we need to keep it in one or more places.

Celandriel:
Where were you when testing, what level were the bots and how long did you wait? 
What I found was that if you are in a zone not appropriate for the bot level the rpg system really struggles to figure out what to do for a while, until eventually they go on their way.|

Edit: Also the resets werent exactly removed, just moved. If the bots no longer identify you as master then the reset command would have been executed as well.

nl-saw:
image
They are standing there now for at least 20 minutes. Few others ultimately left, but they used to pretty quickly run off.
I was in a raid group with 1 other real player when I issues the Leave command. ```botAI->Reset();``` has been removed in this PR, I just double checked it.

Celandriel:
If you whisper 'who' to them, do they identify a master?

Celandriel:
The issue is that by not destroying the group on leaving when strategies are reset, for some bots are still in the group, and so the strategies that get applied are  in group strategies instead of open world strategies.
Thinking about how to solve.

hermensbas:
shall be make it a draft meanwhile?

Celandriel:
Fixed the issue that nl-saw observed. 
The core problem was the leave group was prematurely resetting the master state before confirming that the bot is out of group. This caused the wrong set of strategies to be applied to the bot making them think they stayed in group on the strategy reset. The lack of master in this case limited the cleanup function UpdateAIGroupAndMaster to properly reset the master state once no longer in group. 

I confirmed random bots will swap masters to other real players as well.

hermensbas:
Seems alright to me, havent seen unexpected behaviour.

nl-saw:
Tested, I believe you nailed it now.

hermensbas:
gonna do additional test and then merge if ok

PR #1778, opened 2025-10-27
Use actual item stats for scaling items

Quit:
# Issue

Currently, items with scaling stats (e.g. heirlooms) are evaluated at their minimum level, meaning they are often weighted much lower than even e.g. gray items. This leads to `autogear` replacing heirlooms with whatever equipment the bots can pick up.

# Changes

`StatsCollector::CollectItemStats` gets a second parameter for the player level and then attempts to evaluate the stats according to the passed level. This results in the item having the proper stats that the player/bot should actually have when wearing the item, rather than using some level 0/1 kind of value which is likely worse than most gray items.

The code for the scaling has more-or-less been lifted from Player.cpp, except for the scripting bits. In order to make it work, I had to rework the loop to always iterate over the all possible props, because for the heirlooms I've tested, `StatsCount` is 0.

# Open Questions

There's a few things that this PR does not address because it's a bit more than I currently know about playerbots, so I'd like to mention them. Any guidance is appreciated.

1. `StatsCollector::CollectItemStats` gets another parameter, which is technically an public interface change. Should a default parameter value (likely `0`) be specified, or an overload created so that the old method still exists (and just calls the method with the parameter and a default value)? What are the backwards compatibility concerns for e.g. third parties that may use this code?
2. The original code in `Player.cpp` also allowed scripts to change these values. I'm not doing that here because I don't have a reference to the player, and since the Player variant seems to be in a method that is called when the player stats are refreshed, I'm not sure if there might be side-effects if we call these scripts just for comparison's sake. So while this works for unscripted items, scripted items with scaling would still generate the wrong values.
3. ~~Values that aren't iterated upon in the stats list (e.g. attack damage) aren't scaled yet. I would need to dig deeper to figure out where those scales are set/applied.~~ - attack damage, spell power and armor should be done now, as should be scaled attributes in general.
4. ~~I'm not 100% certain on the way the item templates (`proto`) works, and as such, I'm not entirely sure if `proto->StatsMod[i]` and `proto->ItemStat[i]` are even related.~~ As far as I can tell from the few heirlooms I've tried, there's no strict relation - if an SSD/SSV exists, it will take precedence and the normal stat is completely ignored

# Testing

Limited testing has been done from my side, but usually something like this:

1. Create a bot character of a significantly high level (e.g. 20) and equip it with a heirloom
2. Give it a random gray item that matches its level
3. Run `autogear`
5. Prior to this patch, the bot replaces all heirlooms with gray-or-better items if available; after this patch, they seem to stick around.

I haven't really done any in-depth testing with broad ranges of equipment/player levels/similar because I lack the knowledge to do so, but I thought I would share what I have so far. :)

### DIFF
--- a/src/factory/StatsCollector.cpp
+++ b/src/factory/StatsCollector.cpp
@@ -43,19 +43,51 @@ void StatsCollector::CollectItemStats(ItemTemplate const* proto, uint8 playerLev
     ScalingStatValuesEntry const* ssv =
         proto->ScalingStatValue ? sScalingStatValuesStore.LookupEntry(ssdLevel) : nullptr;
 
-    if (proto->IsRangedWeapon())
+    for (uint8 i = 0; i < MAX_ITEM_PROTO_DAMAGES; ++i)
     {
-        float val = (proto->Damage[0].DamageMin + proto->Damage[0].DamageMax) * 1000 / 2 / proto->Delay;
-        stats[STATS_TYPE_RANGED_DPS] += val;
+        float minDamage = proto->Damage[i].DamageMin;
+        float maxDamage = proto->Damage[i].DamageMax;
+
+        if (ssv && i == 0)
+        {
+            int extraDPS = ssv->getDPSMod(ScalingStatValue);
+            if (extraDPS)
+            {
+                float average = extraDPS * proto->Delay / 1000.0f;
+                float mod = ssv->IsTwoHand(proto->ScalingStatValue) ? 0.2 : 0.3f;
+
+                minDamage = (1.0 - mod) * average;
+                maxDamage = (1.0 + mod) * average;
+            }
+        }
+
+        if (proto->IsRangedWeapon())
+        {
+            float val = (minDamage + maxDamage) * 1000 / 2 / proto->Delay;
+            stats[STATS_TYPE_RANGED_DPS] += val;
+        }
+        else if (proto->IsWeapon())
+        {
+            float val = (minDamage + maxDamage) * 1000 / 2 / proto->Delay;
+            stats[STATS_TYPE_MELEE_DPS] += val;
+        }
     }
-    else if (proto->IsWeapon())
-    {
-        float val = (proto->Damage[0].DamageMin + proto->Damage[0].DamageMax) * 1000 / 2 / proto->Delay;
-        stats[STATS_TYPE_MELEE_DPS] += val;
-    }
-    stats[STATS_TYPE_ARMOR] += proto->Armor;
+
     stats[STATS_TYPE_BLOCK_VALUE] += proto->Block;
-    for (int i = 0; i < MAX_ITEM_PROTO_STATS; i++)
+
+    uint32 armor = proto->Armor;
+    if (ssv)
+    {
+        uint32 ssvarmor = ssv->getArmorMod(ScalingStatValue);
+        if (proto->ScalingStatValue > 0 || ssvarmor < proto->Armor)
+            armor = ssvarmor;
+    }
+    else if (armor && proto->ArmorDamageModifier)
+        armor -= uint32(proto->ArmorDamageModifier);
+
+    stats[STATS_TYPE_ARMOR] += armor;
+
+    for (uint8 i = 0; i < MAX_ITEM_PROTO_STATS; i++)
     {
         const _ItemStat& stat = proto->ItemStat[i];
         uint32 statType = stat.ItemStatType;
@@ -69,9 +101,19 @@ void StatsCollector::CollectItemStats(ItemTemplate const* proto, uint8 playerLev
         }
         else if (i < proto->StatsCount)
             val = stat.ItemStatValue;
+        else
+            continue;
 
         CollectByItemStatType(statType, val);
     }
+
+    uint32 spellBonus = ssv ? ssv->getSpellBonus(ScalingStatValue) : 0;
+    if (ssv && spellBonus)
+    {
+        stats[STATS_TYPE_SPELL_POWER] += spellBonus;
+        stats[STATS_TYPE_HEAL_POWER] += spellBonus;
+    }
+
     for (uint8 j = 0; j < MAX_ITEM_PROTO_SPELLS; j++)
     {
         switch (proto->Spells[j].SpellTrigger)
--- a/src/factory/StatsCollector.cpp
+++ b/src/factory/StatsCollector.cpp
@@ -25,8 +25,24 @@ void StatsCollector::Reset()
     }
 }
 
-void StatsCollector::CollectItemStats(ItemTemplate const* proto)
+void StatsCollector::CollectItemStats(ItemTemplate const* proto, uint8 playerLevel)
 {
+    ScalingStatDistributionEntry const* ssd =
+        proto->ScalingStatDistribution ? sScalingStatDistributionStore.LookupEntry(proto->ScalingStatDistribution)
+                                       : nullptr;
+
+    uint32 ScalingStatValue = proto->ScalingStatValue > 0 ? proto->ScalingStatValue : 0;
+    uint32 ssdLevel = playerLevel;
+
+    if (ssd != nullptr)
+    {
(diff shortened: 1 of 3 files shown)


Quit:
An update on some of the points:

- Attack, armor and spellpower are indeed separate on ssv. I've done some more copying work here but need to polish it up, especially because it means iterating on the damages rather than taking "just" the first slot like it's currently done.
- I've added a check for the index matter, and it does seem to trigger: `index 0, proto type 4 does not match ssv type 38` (i.e. `stat.ItemStatType` is `ITEM_MOD_STRENGTH`, while `ssd->StatMod[i]` seems to be `ITEM_MOD_ATTACK_POWER`.

I've not logged which item produces the error, so I can't get into it in detail. It's possible that the value isn't configured and therefore the ssv isn't "overwriting" anything of value, so it would be less neat, but doable. Player.cpp seems to ignore this edge case currently, and if an ssv is set and available for a slot, it'll take it and completely ignore whatever might be configured in the non-scaling way.

By now, the code has become a bit unwieldy, but I'm not sure what a decent outsourcing could look like. Helper functions would either need to get ssd/ssv for themselves (leading to redundant lookups in every helper method), or have them passed in - neither is really ideal.

**Edit:** Added some more logging, the issue becomes more difficult it seems:

`Item [42945] Venerable Dal'Rend's Sacred Charge index 0, proto type ITEM_MOD_STRENGTH does not match ssv type ITEM_MOD_ATTACK_POWER. Constant value 4 would be replaced by SSV value 7`

The weapon doesn't have any strength on any level, and the SSV of 7 for attack power would be correct for level 24. I guess this means that it is a "data error", i.e. we can/should have the SSVs take hold and ignore whatever the stat says in this case.

Quit:
Two points are still open, but considering scripts are already ignored in the current approach, I think this is ready for a review.

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
+                                       : nullptr;
+
+    uint32 ScalingStatValue = proto->ScalingStatValue > 0 ? proto->ScalingStatValue : 0;
+    uint32 ssdLevel = playerLevel;
+
+    if (ssd != nullptr)
     {
-        float val = (proto->Damage[0].DamageMin + proto->Damage[0].DamageMax) * 1000 / 2 / proto->Delay;
-        stats[STATS_TYPE_RANGED_DPS] += val;
+        if (ssdLevel > ssd->MaxLevel)
```
kadeshar, line 39: That if should be merge with previous

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
-void StatsCollector::CollectItemStats(ItemTemplate const* proto)
+void StatsCollector::CollectItemStats(ItemTemplate const* proto, uint8 playerLevel)
 {
-    if (proto->IsRangedWeapon())
+    ScalingStatDistributionEntry const* ssd =
+        proto->ScalingStatDistribution ? sScalingStatDistributionStore.LookupEntry(proto->ScalingStatDistribution)
+                                       : nullptr;
+
+    uint32 ScalingStatValue = proto->ScalingStatValue > 0 ? proto->ScalingStatValue : 0;
+    uint32 ssdLevel = playerLevel;
```
kadeshar, line 35: there is point of rewriting this variable?

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
 
-void StatsCollector::CollectItemStats(ItemTemplate const* proto)
+void StatsCollector::CollectItemStats(ItemTemplate const* proto, uint8 playerLevel)
 {
-    if (proto->IsRangedWeapon())
+    ScalingStatDistributionEntry const* ssd =
+        proto->ScalingStatDistribution ? sScalingStatDistributionStore.LookupEntry(proto->ScalingStatDistribution)
+                                       : nullptr;
+
+    uint32 ScalingStatValue = proto->ScalingStatValue > 0 ? proto->ScalingStatValue : 0;
```
kadeshar, line 34: That ternary operation is redundant. uint32 is 0 or greater. Assigning negative still equal 0. Also name should be starting from lower case.

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
-    stats[STATS_TYPE_ARMOR] += proto->Armor;
+
     stats[STATS_TYPE_BLOCK_VALUE] += proto->Block;
-    for (int i = 0; i < proto->StatsCount; i++)
+
+    uint32 armor = proto->Armor;
+    if (ssv)
+    {
+        uint32 ssvarmor = ssv->getArmorMod(ScalingStatValue);
+        if (proto->ScalingStatValue > 0 || ssvarmor < proto->Armor)
```
kadeshar, line 82: proto->Armor you have that already in armor variable.

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
+        float minDamage = proto->Damage[i].DamageMin;
+        float maxDamage = proto->Damage[i].DamageMax;
+
+        if (ssv && i == 0)
+        {
+            int extraDPS = ssv->getDPSMod(ScalingStatValue);
+            if (extraDPS)
+            {
+                float average = extraDPS * proto->Delay / 1000.0f;
+                float mod = ssv->IsTwoHand(proto->ScalingStatValue) ? 0.2 : 0.3f;
```
kadeshar, line 57: 0.2f?
This values are something real or you you came up with that?

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
+
+                minDamage = (1.0 - mod) * average;
```
kadeshar, line 59: 1.0f?

src/factory/StatsCollector.cpp
```diff
@@ -25,26 +25,95 @@ void StatsCollector::Reset()
...
+                maxDamage = (1.0 + mod) * average;
```
kadeshar, line 60: 1.0f?

Quit:
@kadeshar: I'll write this here instead of every single comment, because most have the same answer:

It's this way in AzerothCore's Player.cpp, as mentioned in the PR description. See https://github.com/mod-playerbots/azerothcore-wotlk/blob/Playerbot/src/server/game/Entities/Player/Player.cpp#L6629-L6686

As mentioned, I've tried to keep the implementation of this method as close to the Player.cpp's as possible, to avoid future maintenance overhead if it should change, or the script hooks are implemented. Naming schemes can be adjusted, of course, but I can't explain to you what the magic numbers mean, nor if the formulas are correct. It's just this way in AC itself.

kadeshar:
Copying code introduce maintenance overhead and avoid nothing. None one will know that you copied code from other place. Especial such quality.

Quit:
One can argue about the code quality, and it's possible to change the code reasonably. However, if Player.cpp changes - and Player.cpp is the actual code that determines the stats of the item - and you wish to update the logic for the item comparison, you now have to figure out what the "cleaned" code does, and how to apply that change to it.

I can apply the requested changes so it's more in-line with whatever style you're using, but I cannot explain the magic numbers, or why the logic is that way. I can only point at Player.cpp and say "Well, that's how the server does it."

kadeshar:
About magic number explanation is in line 6961 Player.cpp. You can copy that comment with addition that is defined already in core

Celandriel:
I concur with Kadeshars review comments. Can look over again when addressed as best as possible.

ThePenguinMan96:
Was this ever implemented or merged to the test/master branch?

Quit:
Sorry, I kind of forgot about this issue, and since then moved on from Warcraft once again. I'm therefore not going to do the necessary changes to get the PR fixed, but if somebody wants to do it, they are of course welcome to use whatever they find necessary from mine.

PR #1779, opened 2025-10-27
README.md edits

Vortikai:
I've noticed some sprawl in the README.md. The following edits have been made around the goal of clarity, brevity, and emphases when needed:

- Removed dead link to the Spanish README.md at the top. The README_ES.md file itself is not removed from the repo, but is however out of date and does not seem useful at this time
- Removed the addons section. This seems to be covered by the Wiki in the Documentation section unless I'm mistaken? We can add these back if needed
- Reorganized installation instructions to emphasize the importance of using the Playerbots branch of AzerothCore for all installations
- Moved the platform support from the Frequently Asked Questions section to the installation instructions
- Modified punctuation of bulleted list. These can return if it irks someone, but it seems easier to read without them
- Added a encouragement to star the project for updates and gain more visibility
- Removed the Frequently Asked Questions section. I encourage this to be covered in a wiki page but it can return if deemed necessary. I think a lot of it can be moved to other sections if we really need them in there
- Added a "Contributing" section that replaces the Coding Standards section
- Coding Standards are included in the Contributing section
- Added a second link to the Discord server in the Contributing section
- Removed a link to the main Playerbots branch from the introduction section. This is covered and emphasized in the installation section instead
- Migrated the encouragements to make bug reports from the introduction, along with the message that this is still in active development, to the contributing section
- Reduced redundant language when not necessary (e.g. "As noted above,")
- Updated language around custom branch and require branch for uniformity. This will make it more clear to the users about the subject
- Removed "Classic" from the "Classic Installation" terminology for being inaccurate. The subject is now known simply as "Cloning the Repositories" while "Docker Installation" still remains distinct. This will make what was formerly known as "Classic Installation" as de facto default
- Appended the "Documentation" section with add-on information
- Unified all instances of AddOn to the Blizzard spelling without a hyphen following the WoW 3.3.5a client spelling
- Appended "AzerothCore" to instances of "branch" to ensure readers they are, in fact, installing AzerothCore here from our required branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
     <a href="https://github.com/mod-playerbots/mod-playerbots/blob/master/README.md">English</a>
     |
     <a href="https://github.com/mod-playerbots/mod-playerbots/blob/master/README_CN.md">中文</a>
+    |
+    <a href="https://github.com/mod-playerbots/mod-playerbots/blob/master/README_ES.md">Español</a>
 </p>
 
 
@@ -65,7 +67,7 @@ services:
       - ./modules:/azerothcore/modules:ro
 ```
 
-itionally, this override file can be used to set custom configuration settings for `ac-worldserver` and any modules you install as environment variables:
+Additionally, this override file can be used to set custom configuration settings for `ac-worldserver` and any modules you install as environment variables:
 
 ```yml
 services:


Vortikai:
Looking for comments and suggestions.

I've been thinking about more suggestions, though I'm on the fence with these:

- For README.md, migrate installation instructions to a wiki page that offers setup instructions. This may help emphasize to new users the need to install from the Playerbots branch of AC instead of using AC's, and also gives a bit of unity with AC as they also have installation instructions linked in the readme as well. The downside is more clicking to find relevant info.
- Migrate Addons to the installation instructions wiki to promote them there as well

I'd do this myself but have no rights to edit the wiki, and this would mean fewer contributions to that section if they ever needed updates.

NoxMax:
One thing to keep in mind about moving information from README to the wiki page, is that a lot of people coming here might be very new to GitHub, and at most have only downloaded precompiled release files before that need no instructions. They don't have intuitive knowledge of the layout and might not notice the wiki section. Maybe the README FAQ should keep the questions but link the answers to the wiki, maybe the rest of the team have better ideas. Regardless, just keep in mind that not everyone who ends up here is familiar in navigating GH.

README.md
```diff
@@ -18,25 +16,27 @@
...
 
 ## Installation
 
-### Classic Installation
+Supported platforms are Ubuntu, Windows, and macOS. Other Linux distributions may work, but may not receive support. 
+
+**All `mod-playerbots` installations require a custom branch of AzerothCore: [mod-playerbots/azerothcore-wotlk/tree/Playerbot](https://github.com/mod-playerbots/azerothcore-wotlk/tree/Playerbot).** This branch allows the playerbots module to build and function. Updates from the upstream are implemneted regularly to this branch. Instructions for installing this required branch and this module are provided below.
 
-As noted above, `mod-playerbots` requires a custom branch of AzerothCore: [mod-playerbots/azerothcore-wotlk/tree/Playerbot](https://github.com/mod-playerbots/azerothcore-wotlk/tree/Playerbot). To install the module, simply run:
+### Installation
```
kadeshar: Doubled Installation section

README.md
```diff
@@ -2,8 +2,6 @@
     <a href="https://github.com/mod-playerbots/mod-playerbots/blob/master/README.md">English</a>
     |
     <a href="https://github.com/mod-playerbots/mod-playerbots/blob/master/README_CN.md">中文</a>
-    |
-    <a href="https://github.com/brighton-chi/mod-playerbots/blob/readme/README_ES.md">Español</a>
```
kadeshar, line 6: Shouldnt be changed on https://github.com/mod-playerbots/mod-playerbots/blob/master/README_ES.md ?

Vortikai: As noted in the notes, this file hasn't been updated in over a year and doesn't seem to fall in line with the current Readme.mds on the English or Chinese versions. I recommend removing it for now until someone updates it, but keeping readme_es.md in the repo.

README.md
Vortikai: Good catch. I'll rename the second installation "Cloning the Repository"

README.md
Vortikai: I can simply add this back if we want to though. It just doesn't seem like it's useful is all in its current state.

Vortikai:
> One thing to keep in mind about moving information from README to the wiki page, is that a lot of people coming here might be very new to GitHub, and at most have only downloaded precompiled release files before that need no instructions

True, keeping the audience in mind is important. I assumed anyone having to work with this module would be familiar with or should be getting familiar with GitHub, but that might not be the case. We can leave the install instructions on the main page then.

I'm still not sure if the FAQ section is needed on the readme. A link to a wiki with FAQ questions would be more prudent. Looking for others to chime in.

README.md
nl-saw: ES readme hasn't been updated in a long while and I think it's safe to say it is no longer of much relevance. Still, we could leave a link to it, maybe someday someone will give it an update.

README.md
```diff
@@ -65,7 +65,7 @@ services:
       - ./modules:/azerothcore/modules:ro
 ```
 
-Additionally, this override file can be used to set custom configuration settings for `ac-worldserver` and any modules you install as environment variables:
+itionally, this override file can be used to set custom configuration settings for `ac-worldserver` and any modules you install as environment variables:
```
kadeshar: Typo

README.md
Vortikai: Added it back

kadeshar (approved):

kadeshar:
Merging because PR dont affect module working

PR #1780, opened 2025-10-27
FIX: Refactor/restore ..AssistOfIndex Functions

nl-saw:
FIX: Restore original behavior before it was touched 9mo ago.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -2112,7 +2112,6 @@ bool PlayerbotAI::IsMainTank(Player* player)
     }
 
     ObjectGuid mainTank = ObjectGuid();
-    Group::MemberSlotList const& slots = group->GetMemberSlots();
 
     for (Group::member_citerator itr = slots.begin(); itr != slots.end(); ++itr)
     {
@@ -2228,20 +2227,17 @@ bool PlayerbotAI::IsHealAssistantOfIndex(Player* player, int index)
 {
     Group* group = player->GetGroup();
     if (!group)
-    {
         return false;
-    }
-    Group::MemberSlotList const& slots = group->GetMemberSlots();
+
     int counter = 0;
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-        if (group->IsAssistant(member->GetGUID()) && IsHeal(member))
+        if (member && group->IsAssistant(member->GetGUID()) && IsHeal(member))
         {
             if (index == counter)
-            {
                 return player == member;
-            }
+
             counter++;
         }
     }
@@ -2249,12 +2245,11 @@ bool PlayerbotAI::IsHealAssistantOfIndex(Player* player, int index)
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-        if (!group->IsAssistant(member->GetGUID()) && IsHeal(member))
+        if (member && !group->IsAssistant(member->GetGUID()) && IsHeal(member))
         {
             if (index == counter)
-            {
                 return player == member;
-            }
+
             counter++;
         }
     }
@@ -2267,20 +2262,17 @@ bool PlayerbotAI::IsRangedDpsAssistantOfIndex(Player* player, int index)
 {
     Group* group = player->GetGroup();
     if (!group)
-    {
         return false;
-    }
-    Group::MemberSlotList const& slots = group->GetMemberSlots();
+
     int counter = 0;
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-        if (group->IsAssistant(member->GetGUID()) && IsRangedDps(member))
+        if (member && group->IsAssistant(member->GetGUID()) && IsRangedDps(member))
         {
             if (index == counter)
-            {
                 return player == member;
-            }
+
             counter++;
         }
     }
@@ -2288,12 +2280,11 @@ bool PlayerbotAI::IsRangedDpsAssistantOfIndex(Player* player, int index)
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-        if (!group->IsAssistant(member->GetGUID()) && IsRangedDps(member))
+        if (member && !group->IsAssistant(member->GetGUID()) && IsRangedDps(member))
         {
             if (index == counter)
-            {
                 return player == member;
-            }
+
             counter++;
         }
     }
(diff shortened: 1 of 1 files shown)


nl-saw:
~~Tested some affected mobs/bosses in ICC. Seems all good with no change in behavior.~~

src/PlayerbotAI.cpp
```diff
@@ -2298,54 +2224,107 @@ uint32 PlayerbotAI::GetGroupTankNum(Player* player)
...
-            if (index == counter)
-            {
-                return player == member;
-            }
-            counter++;
+            if (group->IsAssistant(member->GetGUID()))
+                assistantHealers.push_back(member);
+            else
+                nonAssistantHealers.push_back(member);
         }
```
kadeshar: Consider using early exit rather than nested if's

like
```
if (!IsHeal(member))
    continue;
```

src/PlayerbotAI.cpp
```diff
@@ -2298,54 +2224,107 @@ uint32 PlayerbotAI::GetGroupTankNum(Player* player)
...
     }
-    return false;
+
+    // Combine: assistants before non-assistants
+    allTanks.insert(allTanks.end(), assistTanks.begin(), assistTanks.end());
+    allTanks.insert(allTanks.end(), nonAssistTanks.begin(), nonAssistTanks.end());
+
+    // Return true if index is valid and matches tank at that index, otherwise false
+    return (index >= 0 && index < static_cast<int>(allTanks.size())) ? player == allTanks[index] : false;
```
kadeshar: Im not sure that is good change. Previous implementation had early exit when found proper index. You compared old and new implementation performance in different group compositions?

src/PlayerbotAI.cpp
```diff
@@ -2298,54 +2224,107 @@ uint32 PlayerbotAI::GetGroupTankNum(Player* player)
...
+                assistTanks.push_back(member);
+            else
+                nonAssistTanks.push_back(member);
         }
     }
-    return false;
+
+    // Combine: assistants before non-assistants
+    allTanks.insert(allTanks.end(), assistTanks.begin(), assistTanks.end());
+    allTanks.insert(allTanks.end(), nonAssistTanks.begin(), nonAssistTanks.end());
```
kadeshar: I dont know how much that comment is accurate but will paste screen for check
obraz

src/PlayerbotAI.cpp
nl-saw: We're no longer working with counters, but vectors. There is no perf impact. The groups are small and not many. New way is actually better and more robust, instead of relying on counters. Previous implementation was wrong and is what I corrected. You can verify this with LLM.

src/PlayerbotAI.cpp
nl-saw: Bit of a LLM slop response. Implementation of reserve is negligible with these kind of low numbers < 40.

src/PlayerbotAI.cpp
nl-saw: Does not matter in this case. It is doing functionally the same, if not healer nothing else is done and it iterates to next person.

kadeshar: It does matter in code readbility. Good practise is reduce nested if's.

src/PlayerbotAI.cpp
kadeshar: 1. We? I dont see any announcement about that.
2. You measured that to confirm that?
3. Agree with statement that maybe is irrelevant. 
4. What was wrong in previous implementation (i mean tank method) except refactoring from counter on vectors?
5. I dont need to verify nothing via LLM.

nl-saw: Erm, I meant no offense..

1- We, as in the changes in this PR - it is indeed not something we spoke of or agreed upon.
2- No need, for such small numbers, this won't have any impact. For comparison, I removed a unnecessary iteration in the AC Core which literally iterated over **all** players and creatures **twice**, and that fix was barely noticeable. We talk many thousands there, compared to 40 tops here...
4- It did double group iteration, as mentioned in the notes "Removed redundant double group iteration"
5- I meant you could verify the functional changes of what I broke earlier

Hope that clarifies :)

src/PlayerbotAI.cpp
nl-saw: It actually could be better here, so I pushed a new commit.

kadeshar (approved):

src/PlayerbotAI.cpp
privatecore: Actually, counter is better in this case:
https://en.cppreference.com/w/cpp/container/vector.html

Also, in the previous implementation space comp O(1) vs O(N) in vector based. Time comp should be the same, if Im not mistaken = O(N).

brighton-chi:
Something to raise--these functions do not (now or before this PR) exclude players from consideration. Obviously, including them allows players to take on whatever specialized role is being assigned, but players could also be thrust into specific roles without being aware of them having that responsibility. That being said, the likelihood here is low given how GetFirstMember works and that there is no filtering beyond assistant vs. non-assistant status (the raid would need to have more than one player, with the applicable player being the last group member to join that would fit the criteria for the particular role).

My personal view is to exclude players in scenarios where it is likely that they would be chosen unintentionally (e.g., High King Maulgar strategy, where a solo player may very well be the best geared mage in the raid) but include them in situations where that would be unlikely (such as with these functions). I just want to make sure consideration is given to this by those who contribute/review/merge.

brighton-chi:
Also, the penultimate return false in IsBotMainTank needs to be removed because it is busting that function. I'll defer on whether to delete that as part of this PR too or address it separately, though it seems to me that it's cleanest to just delete it with this PR since the relevant function is contained within the same cpp and conceptually has commonality with the AssistOfIndex functions.

nl-saw:
> Also, the penultimate return false in IsBotMainTank needs to be removed because it is busting that function. I'll defer on whether to delete that as part of this PR too or address it separately, though it seems to me that it's cleanest to just delete it with this PR since the relevant function is contained within the same cpp and conceptually has commonality with the AssistOfIndex functions.

Thanks a lot for the feedback. I am in favor of creating a new PR for that, as it is technically a separate issue than what I am restoring here. Would you mind creating an issue for it so we can track/action it later?

brighton-chi:
> > Also, the penultimate return false in IsBotMainTank needs to be removed because it is busting that function. I'll defer on whether to delete that as part of this PR too or address it separately, though it seems to me that it's cleanest to just delete it with this PR since the relevant function is contained within the same cpp and conceptually has commonality with the AssistOfIndex functions.
> 
> Thanks a lot for the feedback. I am in favor of creating a new PR for that, as it is technically a separate issue than what I am restoring here. Would you mind creating an issue for it so we can track/action it later?

I can just PR it later today.

nl-saw:
Decided to roll the code back to before it was touched 9mo ago.

src/PlayerbotAI.cpp
```diff
@@ -2326,16 +2284,49 @@ bool PlayerbotAI::IsAssistTankOfIndex(Player* player, int index)
...
+    Group* group = player->GetGroup();
+    if (!group)
+    {
+        return false;
+    }
+    Group::MemberSlotList const& slots = group->GetMemberSlots();
+    int counter = 0;
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
```
privatecore, line 2305: nullptr check?

PR #1782, opened 2025-10-28
important bugfix for the last PR #1675

Raz0r1337:
important bugfix for the last PR #1675

This fixes the deletion of the complete text_loc3 column before filling them.

My bad srry.


Raz0r1337:
@kadeshar 
@nl-saw 

please review this as fast as possible there is a bug that adds holes in the german text_loc3 collumn

srry i oversee this

nl-saw (approved):

PR #1784, opened 2025-10-29
CORE - Improved language detection for bots

Wishmaster117:
I've had this problem for a long time, my bots only speak English even though I'm playing on a French client.
I suppose this must be the case for some other people who do not have a large number of players with the same local client.
If we use French DBCs, the bots bug because they only recognize US DBCs.
From what I understand, the language is chosen as follows:
On load, the module reads the entire `ai_playerbot_texts` table and stores each text variant in a dictionary indexed by the locale ID: the `text` column remains the default value (English), and the `text_loc1` to `text_loc8` columns fill slots 1 through 8.

Whenever a real player connects, the module increments a counter for that player's DBC locale using `AddLocalePriority(player->GetSession()->GetSessionDbcLocale())`.

When a bot needs a text, `GetLocalePriority()` returns the most frequently used locale index among currently connected players. The corresponding string is then retrieved. if the box is empty, we fall back to the English version (text[0]).

### This PR improve language detection.
**Summary**

- log both the client DBC locale and the account database locale when a player logs in
- fall back to the account locale when the client reports enUS but the account is configured for another locale
- keep the existing vote-based selection so bots always speak the majority language among connected players

**Therefore, the original behavior is maintained. Bots still choose the most represented language among connected players (the counter is simply more efficient by prioritizing the account's locale when it differs from the client's English). For example, if more English-speaking players are connected, the language will revert to English, as the bots always share the majority locale.**

### DIFF
--- a/.github/workflows/codestyle_cpp.yml
+++ b/.github/workflows/codestyle_cpp.yml
@@ -5,10 +5,15 @@ on:
       - opened
       - reopened
       - synchronize
+      - ready_for_review
     paths:
       - src/**
       - "!README.md"
       - "!docs/**"
+
+concurrency:
+  group: "codestyle-cppcheck-${{ github.event.pull_request.number }}"
+  cancel-in-progress: true
 
 jobs:
   triage:
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ We also have a **[Discord server](https://discord.gg/NQm5QShwf9)** where you can
 
 Supported platforms are Ubuntu, Windows, and macOS. Other Linux distributions may work, but may not receive support. 
 
-**All `mod-playerbots` installations require a custom branch of AzerothCore: [mod-playerbots/azerothcore-wotlk/tree/Playerbot](https://github.com/mod-playerbots/azerothcore-wotlk/tree/Playerbot).** This branch allows the playerbots module to build and function. Updates from the upstream are implemneted regularly to this branch. Instructions for installing this required branch and this module are provided below.
+**All `mod-playerbots` installations require a custom branch of AzerothCore: [mod-playerbots/azerothcore-wotlk/tree/Playerbot](https://github.com/mod-playerbots/azerothcore-wotlk/tree/Playerbot).** This branch allows the `mod-playerbots` module to build and function. Updates from the upstream are implemented regularly to this branch. Instructions for installing this required branch and this module are provided below.
 
 ### Cloning the Repositories
 
@@ -50,7 +50,7 @@ For more information, refer to the [AzerothCore Installation Guide](https://www.
 
 ### Docker Installation
 
-Docker installations are considered experimental (unofficial with limited support), and previous Docker experience is recommended. To install the `mod-playerbots` on Docker, first clone the required branch of AzerothCore and this module:
+Docker installations are considered experimental (unofficial with limited support), and previous Docker experience is recommended. To install `mod-playerbots` on Docker, first clone the required branch of AzerothCore and this module:
 
 ```bash
 git clone https://github.com/mod-playerbots/azerothcore-wotlk.git --branch=Playerbot
@@ -103,7 +103,7 @@ Please click on the "⭐" button to stay up to date and help us gain more visibi
 
 ## Acknowledgements
 
-`mod-playerbots` is is based off [ZhengPeiRu21/mod-playerbots](https://github.com/ZhengPeiRu21/mod-playerbots) and [celguar/mangosbot-bots](https://github.com/celguar/mangosbot-bots). We extend our gratitude to [@ZhengPeiRu21](https://github.com/ZhengPeiRu21) and [@celguar](https://github.com/celguar) for the continued efforts in maintaining the module.
+`mod-playerbots` is based on [ZhengPeiRu21/mod-playerbots](https://github.com/ZhengPeiRu21/mod-playerbots) and [celguar/mangosbot-bots](https://github.com/celguar/mangosbot-bots). We extend our gratitude to [@ZhengPeiRu21](https://github.com/ZhengPeiRu21) and [@celguar](https://github.com/celguar) for their continued efforts in maintaining the module.
 
 Also, a thank you to the many contributors who've helped build this project:
 
--- a/src/strategy/actions/AreaTriggerAction.cpp
+++ b/src/strategy/actions/AreaTriggerAction.cpp
@@ -40,7 +40,14 @@ bool ReachAreaTriggerAction::Execute(Event event)
         return true;
     }
 
-    bot->GetMotionMaster()->MovePoint(at->map, at->x, at->y, at->z);
+    bot->GetMotionMaster()->MovePoint(
+        /*id*/ at->map,
+        /*coords*/ at->x, at->y, at->z,
+        /*forcedMovement*/ FORCED_MOVEMENT_NONE,
+        /*speed*/ 0.0f,             // default speed (not handled here)
+        /*orientation*/ 0.0f,       // keep current orientation of bot
+        /*generatePath*/ true,      // true => terrain path (2d mmap); false => straight spline (3d vmap)
+        /*forceDestination*/ false);
 
     float distance = bot->GetDistance(at->x, at->y, at->z);
     float delay = 1000.0f * distance / bot->GetSpeed(MOVE_RUN) + sPlayerbotAIConfig->reactDelay;
--- a/src/strategy/actions/CheckMountStateAction.cpp
+++ b/src/strategy/actions/CheckMountStateAction.cpp
@@ -81,7 +81,7 @@ bool CheckMountStateAction::isUseful()
     // to mostly be an issue in tunnels of WSG and AV)
     float posZ = bot->GetPositionZ();
     float groundLevel = bot->GetMapWaterOrGroundLevel(bot->GetPositionX(), bot->GetPositionY(), posZ);
-    if (!bot->IsMounted() && posZ < groundLevel)
+    if (!bot->IsMounted() && !bot->HasWaterWalkAura() && posZ < groundLevel)
         return false;
 
     // Not useful when bot does not have mount strat and is not currently mounted
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -10,6 +10,7 @@
 #include <iomanip>
 #include <string>
 
+#include "Corpse.h"
 #include "Event.h"
 #include "FleeManager.h"
 #include "G3D/Vector3.h"
@@ -41,7 +42,6 @@
 #include "Unit.h"
 #include "Vehicle.h"
 #include "WaypointMovementGenerator.h"
-#include "Corpse.h"
 
 MovementAction::MovementAction(PlayerbotAI* botAI, std::string const name) : Action(botAI, name)
 {
@@ -192,14 +192,15 @@ bool MovementAction::MoveTo(uint32 mapId, float x, float y, float z, bool idle, 
     {
         return false;
     }
+
     bool generatePath = !bot->IsFlying() && !bot->isSwimming();
-    bool disableMoveSplinePath = sPlayerbotAIConfig->disableMoveSplinePath >= 2 ||
-                                 (sPlayerbotAIConfig->disableMoveSplinePath == 1 && bot->InBattleground());
+    bool disableMoveSplinePath =
+        sPlayerbotAIConfig->disableMoveSplinePath >= 2 ||
+        (sPlayerbotAIConfig->disableMoveSplinePath == 1 && bot->InBattleground());
     if (Vehicle* vehicle = bot->GetVehicle())
     {
         VehicleSeatEntry const* seat = vehicle->GetSeatForPassenger(bot);
         Unit* vehicleBase = vehicle->GetBase();
-        // If the mover (vehicle) can fly, we DO NOT want an mmaps path (2D ground) => disable pathfinding
         generatePath = !vehicleBase || !vehicleBase->CanFly();
         if (!vehicleBase || !seat || !seat->CanControl())  // is passenger and cant move anyway
             return false;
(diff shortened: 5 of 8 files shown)


nl-saw (changes requested):
Please remove the brackets for single if statements/returns throughout the files.

Wishmaster117:
> Please remove the brackets for single if statements/returns throughout the files.

Done

src/PlayerbotMgr.cpp
```diff
@@ -1602,8 +1602,34 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
...
+
+    LocaleConstant locale = dbcLocale;
+    if (dbcLocale >= MAX_LOCALES && dbLocale < MAX_LOCALES)
+        locale = dbLocale;
+    else if (dbLocale < MAX_LOCALES && dbcLocale < MAX_LOCALES && dbcLocale != dbLocale)
+        if (dbcLocale == LOCALE_enUS && dbLocale != LOCALE_enUS)
+            locale = dbLocale;
+
+    if (locale >= MAX_LOCALES)
+        locale = LOCALE_enUS;
```
privatecore:
```cpp
if (dbLocaleIndex < MAX_LOCALES)
{
    if (dbcLocale >= MAX_LOCALES || (dbcLocale == LOCALE_enUS && dbLocaleIndex != LOCALE_enUS))
        locale = dbLocaleIndex;
}

if (locale >= MAX_LOCALES)
    locale = LOCALE_enUS;
```
Will this cover your original logic?

src/PlayerbotTextMgr.cpp
```diff
@@ -190,29 +192,61 @@ bool PlayerbotTextMgr::GetBotText(std::string name, std::string& text, std::map<
...
+    uint32 topLocaleCount = 0;
+
+    auto const buildLocaleCounts = [this]() -> std::string
+    {
+        std::ostringstream stream;
+        stream << '[';
+        for (uint8 i = 0; i < MAX_LOCALES; ++i)
+        {
+            if (i)
+            {
```
nl-saw: Last bracket to remove. Otherwise code looks ok to me

Wishmaster117: Done

src/PlayerbotMgr.cpp
```diff
@@ -1602,8 +1602,34 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
...
+        return;
+    }
+
+    LocaleConstant const dbcLocale = session->GetSessionDbcLocale();
+    LocaleConstant const dbLocale = session->GetSessionDbLocaleIndex();
+
+    LocaleConstant locale = dbcLocale;
+    if (dbcLocale >= MAX_LOCALES && dbLocale < MAX_LOCALES)
+        locale = dbLocale;
+    else if (dbLocale < MAX_LOCALES && dbcLocale < MAX_LOCALES && dbcLocale != dbLocale)
```
nl-saw: You could combine these if statements to avoid some nesting

Wishmaster117: Too late you approved the changes 1 hour ago :D 
I do this tomorrow, i'm tired.

src/PlayerbotMgr.cpp
```diff
@@ -1602,8 +1602,34 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
...
+
+    WorldSession* session = player->GetSession();
+    if (!session)
+    {
+        LOG_WARN("playerbots", "Unable to register locale priority for player {} because the session is missing", player->GetName());
+        return;
+    }
+
+    LocaleConstant const dbcLocale = session->GetSessionDbcLocale();
+    LocaleConstant const dbLocale = session->GetSessionDbLocaleIndex();
```
privatecore: `LocaleConstant const dbLocaleIndex = session->GetSessionDbLocaleIndex();`
Please, its hard to read...

src/PlayerbotMgr.cpp
privatecore:
```cpp
if (dbLocaleIndex < MAX_LOCALES)
{
    if (dbcLocale >= MAX_LOCALES || (dbcLocale == LOCALE_enUS && dbLocaleIndex != LOCALE_enUS))
        locale = dbLocaleIndex;
}

if (locale >= MAX_LOCALES)
    locale = LOCALE_enUS;
```
Will this cover your original logic?

src/PlayerbotMgr.cpp
Wishmaster117: Okay, I've changed the variable names for better readability.

src/PlayerbotMgr.cpp
Wishmaster117: Okay, I've changed the variable names for readability, and I've nested the if statements and added comments for clarity.

Wishmaster117:
I tested it with a WoW client in English and French and it works, the bots speak the language selected by the client correctly. I don't have other languages ​​to test with, but if it works with these two, there's no reason why it shouldn't work with other languages.

src/PlayerbotMgr.cpp
```diff
@@ -1602,8 +1602,39 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
...
+    WorldSession* session = player->GetSession();
+    if (!session)
+    {
+        LOG_WARN("playerbots", "Unable to register locale priority for player {} because the session is missing", player->GetName());
+        return;
+    }
+
+    // Explicit retrieval of the two locale indices (DBC and DB)
+    LocaleConstant const dbcLocaleIndex = session->GetSessionDbcLocale();
+    LocaleConstant const dbLocaleIndex  = session->GetSessionDbLocaleIndex();
```
privatecore:
```cpp
    LocaleConstant const dbcLocale = session->GetSessionDbcLocale();
    LocaleConstant const dbLocaleIndex  = session->GetSessionDbLocaleIndex();
```
Hmm, having only one character different for the variable is kind of a challenge ;)

Wishmaster117: Sorry, this is getting too picky, looking at the code we can clearly see that these are two different variables.

src/PlayerbotMgr.cpp
```diff
@@ -1602,8 +1602,39 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
...
+    // - prefer DB if DBC is invalid and DB is valid
+    // - or if both are valid but DBC == enUS and DB != enUS
+    bool const dbcValid = dbcLocaleIndex < MAX_LOCALES;
+    bool const dbValid  = dbLocaleIndex  < MAX_LOCALES;
+    if ((!dbcValid && dbValid) ||
+        (dbcValid && dbValid && dbcLocaleIndex == LOCALE_enUS && dbLocaleIndex != LOCALE_enUS))
+        usedLocale = dbLocaleIndex;
+
+    if (usedLocale >= MAX_LOCALES)
+        usedLocale = LOCALE_enUS;
```
privatecore: Now I see your logic, take a look:
```cpp
// If the DBC locale is invalid, use the database locale.
// Or (DBC locale valid) if the DBC locale is equal to enUS and the database locale is not equal to enUS.
// Set locale to the database locale.
if (dbcLocale >= MAX_LOCALES || (dbcLocale == LOCALE_enUS && dbLocaleIndex != LOCALE_enUS))
    usedLocale = dbLocaleIndex;

// This check for the database locale is valid, so you don't need to check it in the previous if..else statements
if (usedLocale >= MAX_LOCALES)
    usedLocale = LOCALE_enUS;
```

src/PlayerbotMgr.cpp
privatecore: Okay, then there was no point in renaming them earlier; they were just as "clearly" to read.

src/PlayerbotMgr.cpp
Wishmaster117: Okay, is that clearer now?

privatecore: Dude, srsly, I didn’t mean to offend or upset you in any way. Understand me correctly, not everyone has the opportunity to read code in a convenient IDE on a computer; some might not use an IDE at all. When you see two variables in the code that differ by only one character, followed by a large block of if..else, it’s really hard to read, believe me.

Wishmaster117: > Dude, srsly, I didn’t mean to offend or upset you in any way. Understand me correctly, not everyone has the opportunity to read code in a convenient IDE on a computer; some might not use an IDE at all. When you see two variables in the code that differ by only one character, followed by a large block of if..else, it’s really hard to read, believe me.

Lol (the problem with communicating via keyboards is that you don't get people's feelings) I'm not offended at all or anything. When we code, we have our own perspective, and you're absolutely right, we don't necessarily think about the other people who will read the code.

If it can improve things, all the better :)

nl-saw (approved):

gacuna89:
With this change I fixed the locales issue.

Looks like the issue is in the code, not the translations. Playerbots should use `GetSessionDbLocaleIndex()` (DB texts) instead of `GetSessionDbcLocale()` (DBC files). This is consistent with other parts of the core (`Player.h`, `WorldSession.cpp`, `Unit.cpp`). That’s why bots keep defaulting to English even with `esMX` set.

`uint32 playerLocale = player->GetSession()->GetSessionDbLocaleIndex();`

Wishmaster117:
> With this change I fixed the locales issue.
> 
> Looks like the issue is in the code, not the translations. Playerbots should use `GetSessionDbLocaleIndex()` (DB texts) instead of `GetSessionDbcLocale()` (DBC files). This is consistent with other parts of the core (`Player.h`, `WorldSession.cpp`, `Unit.cpp`). That’s why bots keep defaulting to English even with `esMX` set.
> 
> `uint32 playerLocale = player->GetSession()->GetSessionDbLocaleIndex();`

Thanks for the feedback, you’re absolutely right that GetSessionDbLocaleIndex() should be used for bot texts since they come from the database, not DBC files.

However, your suggested change alone isn’t fully sufficient because it doesn’t handle invalid locale values (>= MAX_LOCALES) or provide a proper fallback to LOCALE_enUS. Without that safeguard, bots might still default incorrectly in some edge cases.

I agree the logic can be simplified though, and I’ll update the code to use the DB locale with a safe fallback.

I’m currently on vacation and don’t have access to my PC until Sunday, but I’ll push the fix as soon as I’m back.

Thanks again for catching this!

src/PlayerbotMgr.cpp
```diff
@@ -1602,8 +1602,29 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
...
+
+    WorldSession* session = player->GetSession();
+    if (!session)
+    {
+        LOG_WARN("playerbots", "Unable to register locale priority for player {} because the session is missing", player->GetName());
+        return;
+    }
+
+    // Locales: DBC (client files) vs DB (translations for bot texts)
+    LocaleConstant const clientLocale   = session->GetSessionDbcLocale();
```
kadeshar: This variable is not used to any calculation. Why even mention that in logs.
Even that getting that variable in such situation should have condition its log_debug enabled.

src/PlayerbotMgr.cpp
kadeshar: Now session is in variable.

src/PlayerbotTextMgr.cpp
```diff
@@ -190,29 +192,58 @@ bool PlayerbotTextMgr::GetBotText(std::string name, std::string& text, std::map<
...
 {
-    if (!locale)
+    if (locale >= MAX_LOCALES)
+    {
+        LOG_WARN("playerbots", "Ignoring locale {} for bot texts because it exceeds MAX_LOCALES ({})", locale, MAX_LOCALES - 1);
         return;
+    }
 
-    botTextLocalePriority[locale]++;
+    ++botTextLocalePriority[locale];
```
kadeshar: Explain why this is changed

src/PlayerbotTextMgr.cpp
```diff
@@ -190,29 +192,58 @@ bool PlayerbotTextMgr::GetBotText(std::string name, std::string& text, std::map<
...
+        uint32 const localeCount = botTextLocalePriority[i];
+        if (localeCount > topLocaleCount)
+        {
             topLocale = i;
+            topLocaleCount = localeCount;
+        }
     }
 
+    LOG_DEBUG("playerbots", "Locale priority check: active sessions {}, counts {}, chosen locale {} with {} votes", activeSessions,
+        buildLocaleCounts(), topLocale, topLocaleCount);
```
kadeshar: Sorry im totally dont agree with putting everywhere LOG_DEBUG which are calculated even where logging disabled. Especial using lamda,  topLocaleCount calculations etc.

src/PlayerbotMgr.cpp
Wishmaster117: You’re absolutely right, clientLocale is not used for any decision here, and I was only logging it for debugging while testing the change.
I’ve removed the unused clientLocale variable and the debug log, and kept only the DB locale selection with a safe fallback to LOCALE_enUS.

src/PlayerbotTextMgr.cpp
Wishmaster117: Changing it was not intentional from a semantic point of view, it just slipped in while I was editing the function.
I’ve reverted it back to botTextLocalePriority[locale]++ and also removed the extra LOG_DEBUG to keep this function simple and avoid unnecessary logging overhead.

src/PlayerbotTextMgr.cpp
Wishmaster117: Fair point about the debug logging overhead here.
The lambda and the detailed LOG_DEBUG calls were only meant for troubleshooting while I was validating the behavior, but they are not required for normal operation and do add extra cost even when logging is disabled.
I’ve removed the verbose debug logs (and the lambda) from GetLocalePriority() and kept only the minimal logic

src/PlayerbotMgr.cpp
Wishmaster117: Good catch, thanks.
I already store the WorldSession* in session, so there’s no reason to call player->GetSession() again here.
I’ve updated this to use session->GetAccountId() instead.

src/PlayerbotTextMgr.cpp
```diff
@@ -190,27 +192,36 @@ bool PlayerbotTextMgr::GetBotText(std::string name, std::string& text, std::map<
...
 
     for (uint8 i = 0; i < MAX_LOCALES; ++i)
     {
-        if (botTextLocalePriority[i] > topLocale)
+        uint32 const localeCount = botTextLocalePriority[i];
+        if (localeCount > topLocaleCount)
+        {
             topLocale = i;
+            topLocaleCount = localeCount;
+        }
```
kadeshar: topLocaleCount is not used.
Tbh this changes should be reverted.

Wishmaster117: You’re right that topLocaleCount is not needed anymore here.
It was originally introduced together with some debug logging, and once the logs were removed it became just an extra accumulator.
This removes topLocaleCount entirely and keeps the function focused on its original behavior.

Also removed #include <sstream> not used anymore here

src/PlayerbotTextMgr.cpp
```diff
@@ -190,26 +190,29 @@ bool PlayerbotTextMgr::GetBotText(std::string name, std::string& text, std::map<
 
 void PlayerbotTextMgr::AddLocalePriority(uint32 locale)
 {
-    if (!locale)
```
kadeshar, line 193: You sure about deleting this check?

Wishmaster117: Yes, 
In AzerothCore, 0 is a valid locale (LOCALE_enUS), not an invalid or “empty” value.
Keeping that check would incorrectly ignore enUS players and prevent their language from contributing to the priority system.
The only invalid values are >= MAX_LOCALES, and that validation is still in place.
OnPlayerLogin also normalizes the locale and falls back to LOCALE_enUS when needed, so the extra check was redundant and actually harmful.
That’s why it was removed.

nl-saw (approved):

PR #1790, opened 2025-10-31
Feature - Add pvp / arena chat command for Playerbots

Wishmaster117:
### Summary:
This PR adds a new whisper command "pvp stats"  that allows players to ask a bot to report its current Arena Points, Honor Points, and Arena Teams (name and team rating).

**Reason:**
Due to a client limitation in WoW 3.3.5a, the inspection window does not display another player's Arena or Honor points , only team data.
This command provides an easy in-game way to check a bot’s PvP currencies without modifying the client or core packets.

**Example usage:**

`/w BotName pvp stats`

**Bot reply:**

```
[PVP] Arena Points: 302 | Honor Points: 11855
[PVP] 2v2: <The Fighters> (rating 2000)
[PVP] 3v3: <The Trio> (rating 573)
```

**Implementation notes:**

Uses existing core getters (GetArenaPoints, GetHonorPoints, GetArenaTeamId, etc.).
Fully integrated into the chat command system (ChatTriggerContext, ChatActionContext).
Safe, no gameplay changes, purely informational.
No harcoded texts, use database local instead

### If merged, Multibot need update

### DIFF
--- a/src/strategy/actions/ChatActionContext.h
+++ b/src/strategy/actions/ChatActionContext.h
@@ -67,7 +67,7 @@
 #include "TellItemCountAction.h"
 #include "TellLosAction.h"
 #include "TellReputationAction.h"
-#include "TellPvpAction.h"
+#include "TellPvpStatsAction.h"
 #include "TellTargetAction.h"
 #include "TradeAction.h"
 #include "TrainerAction.h"
@@ -98,7 +98,7 @@ public:
         creators["quests"] = &ChatActionContext::quests;
         creators["leave"] = &ChatActionContext::leave;
         creators["reputation"] = &ChatActionContext::reputation;
-        creators["pvp"] = &ChatActionContext::pvp;
+        creators["tell pvp stats"] = &ChatActionContext::tell_pvp_stats;
         creators["log"] = &ChatActionContext::log;
         creators["los"] = &ChatActionContext::los;
         creators["rpg status"] = &ChatActionContext::rpg_status;
@@ -282,7 +282,7 @@ private:
     static Action* quests(PlayerbotAI* botAI) { return new ListQuestsAction(botAI); }
     static Action* leave(PlayerbotAI* botAI) { return new LeaveGroupAction(botAI); }
     static Action* reputation(PlayerbotAI* botAI) { return new TellReputationAction(botAI); }
-    static Action* pvp(PlayerbotAI* botAI) { return new TellPvpAction(botAI); }
+    static Action* tell_pvp_stats(PlayerbotAI* botAI) { return new TellPvpStatsAction(botAI); }
     static Action* log(PlayerbotAI* botAI) { return new LogLevelAction(botAI); }
     static Action* los(PlayerbotAI* botAI) { return new TellLosAction(botAI); }
     static Action* rpg_status(PlayerbotAI* botAI) { return new TellRpgStatusAction(botAI); }
--- a/src/strategy/actions/TellPvpStatsAction.cpp
+++ b/src/strategy/actions/TellPvpStatsAction.cpp
@@ -3,7 +3,7 @@
  * and/or modify it under version 3 of the License, or (at your option), any later version.
  */
 
-#include "TellPvpAction.h"
+#include "TellPvpStatsAction.h"
 
 #include <map>
 
@@ -29,7 +29,7 @@ namespace
     }
 }
 
-bool TellPvpAction::Execute(Event /*event*/)
+bool TellPvpStatsAction::Execute(Event /*event*/)
 {
     Player* const master = GetMaster();
     if (!master || !bot)
--- a/src/strategy/generic/ChatCommandHandlerStrategy.cpp
+++ b/src/strategy/generic/ChatCommandHandlerStrategy.cpp
@@ -27,7 +27,7 @@ void ChatCommandHandlerStrategy::InitTriggers(std::vector<TriggerNode*>& trigger
     PassTroughStrategy::InitTriggers(triggers);
 
     triggers.push_back(new TriggerNode("rep", NextAction::array(0, new NextAction("reputation", relevance), nullptr)));
-    triggers.push_back(new TriggerNode("pvp stats", NextAction::array(0, new NextAction("pvp", relevance), nullptr)));
+    triggers.push_back(new TriggerNode("pvp stats", NextAction::array(0, new NextAction("tell pvp stats", relevance), nullptr)));
     triggers.push_back(new TriggerNode("q", NextAction::array(0, new NextAction("query quest", relevance),
                                                               new NextAction("query item usage", relevance), nullptr)));
     triggers.push_back(new TriggerNode("add all loot", NextAction::array(0, new NextAction("add all loot", relevance),
@@ -119,7 +119,7 @@ ChatCommandHandlerStrategy::ChatCommandHandlerStrategy(PlayerbotAI* botAI) : Pas
     supported.push_back("stats");
     supported.push_back("leave");
     supported.push_back("reputation");
-    supported.push_back("pvp stats");
+    supported.push_back("tell pvp stats");
     supported.push_back("log");
     supported.push_back("los");
     supported.push_back("rpg status");
--- a/src/strategy/actions/TellPvpStatsAction.h
+++ b/src/strategy/actions/TellPvpStatsAction.h
@@ -3,17 +3,17 @@
  * and/or modify it under version 3 of the License, or (at your option), any later version.
  */
 
-#ifndef _PLAYERBOT_TELLPVPACTION_H
-#define _PLAYERBOT_TELLPVPACTION_H
+#ifndef _PLAYERBOT_TELLPVPSTATSACTION_H
+#define _PLAYERBOT_TELLPVPSTATSACTION_H
 
 #include "Action.h"
 
 class PlayerbotAI;
 
-class TellPvpAction : public Action
+class TellPvpStatsAction : public Action
 {
 public:
-    TellPvpAction(PlayerbotAI* botAI) : Action(botAI, "pvp") {}
+    TellPvpStatsAction(PlayerbotAI* botAI) : Action(botAI, "tell pvp stats") {}
 
     bool Execute(Event event) override;
 };
--- a/src/strategy/triggers/ChatTriggerContext.h
+++ b/src/strategy/triggers/ChatTriggerContext.h
@@ -24,7 +24,7 @@ public:
         creators["leave"] = &ChatTriggerContext::leave;
         creators["rep"] = &ChatTriggerContext::reputation;
         creators["reputation"] = &ChatTriggerContext::reputation;
-        creators["pvp stats"] = &ChatTriggerContext::pvp;
+        creators["pvp stats"] = &ChatTriggerContext::pvp_stats;
         creators["log"] = &ChatTriggerContext::log;
         creators["los"] = &ChatTriggerContext::los;
         creators["rpg status"] = &ChatTriggerContext::rpg_status;
@@ -226,7 +226,7 @@ private:
     static Trigger* stats(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "stats"); }
     static Trigger* leave(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "leave"); }
     static Trigger* reputation(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "reputation"); }
-    static Trigger* pvp(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "pvp stats"); }
+    static Trigger* pvp_stats(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "pvp stats"); }
     static Trigger* log(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "log"); }
     static Trigger* los(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "los"); }
     static Trigger* rpg_status(PlayerbotAI* botAI) { return new ChatCommandTrigger(botAI, "rpg status"); }


src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,71 @@
...
+            case 2: return "5v5";
+        }
+        return "?v?";
+    }
+}
+
+bool TellPvpAction::Execute(Event /*event*/)
+{
+    Player* const master = GetMaster();
+    Player* const bot = botAI->GetBot();
```
kadeshar: Action have own bot variable

src/strategy/actions/TellPvpStatsAction.cpp
```diff
@@ -0,0 +1,71 @@
...
+            std::ostringstream line;
+            line << "[PVP] " << BracketName(slot) << ": <" << team->GetName() << "> (rating " << team->GetRating() << ")";
+            botAI->TellMaster(line.str());
+        }
+    }
+
+    if (!anyTeam)
+    {
+        botAI->TellMaster("[PVP] I have no Arena Team.");
+    }
```
kadeshar, line 83: single line if should be without {}

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,71 @@
...
+        uint32 const teamId = bot->GetArenaTeamId(slot);
+        if (!teamId)
+            continue;
+
+        if (ArenaTeam* team = sArenaTeamMgr->GetArenaTeamById(teamId))
+        {
+            anyTeam = true;
+            std::ostringstream line;
+            line << "[PVP] " << BracketName(slot) << ": <" << team->GetName() << "> (rating " << team->GetRating() << ")";
+            botAI->TellMaster(line.str());
```
kadeshar: Every text send to player should be translatable

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,71 @@
...
+    Player* const bot = botAI->GetBot();
+    if (!master || !bot)
+        return false;
+
+    // Currencies
+    {
+        std::ostringstream line;
+        line << "[PVP] Arena points: " << bot->GetArenaPoints()
+             << " | Honor Points: " << bot->GetHonorPoints();
+        botAI->TellMaster(line.str());
```
kadeshar: Every text send to player should be translatable

src/strategy/actions/ChatActionContext.h
```diff
@@ -97,6 +98,9 @@ class ChatActionContext : public NamedObjectContext<Action>
         creators["quests"] = &ChatActionContext::quests;
         creators["leave"] = &ChatActionContext::leave;
         creators["reputation"] = &ChatActionContext::reputation;
+        // PVP currencies + teams
+        creators["pvp"] = &ChatActionContext::pvp;
+        creators["arena"] = &ChatActionContext::pvp;
```
kadeshar: Same chat action with multiple commands. Im not sure that should exists at all. Even that why "arena" exists and "battleground" or "bg" not

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+bool TellPvpAction::Execute(Event /*event*/)
+{
+    Player* const master = GetMaster();
+    Player* const bot = botAI->GetBot();
+    if (!master || !bot)
+        return false;
+
+    // Currencies
+    {
```
kadeshar: Why this brackets

src/strategy/generic/ChatCommandHandlerStrategy.cpp
```diff
@@ -27,6 +27,9 @@ void ChatCommandHandlerStrategy::InitTriggers(std::vector<TriggerNode*>& trigger
     PassTroughStrategy::InitTriggers(triggers);
 
     triggers.push_back(new TriggerNode("rep", NextAction::array(0, new NextAction("reputation", relevance), nullptr)));
+    // PVP currencies + teams
+    triggers.push_back(new TriggerNode("pvp",   NextAction::array(0, new NextAction("pvp", relevance), nullptr)));
+    triggers.push_back(new TriggerNode("arena", NextAction::array(0, new NextAction("pvp", relevance), nullptr)));
```
kadeshar: Same here.

src/strategy/actions/TellPvpAction.cpp
Wishmaster117: You’re right sorry.
I removed the local Player* bot and now use the inherited bot pointer instead.

src/strategy/actions/TellPvpStatsAction.cpp
Wishmaster117: The original if (!anyTeam) was a single-line body indeed.
After switching to PlayerbotTextMgr and placeholders, the body now has multiple statements, so braces are required and consistent with the code style.

src/strategy/actions/TellPvpAction.cpp
Wishmaster117: Good point, I switched all messages to use PlayerbotTextMgr so they can be localized.
pvp_currency = [PVP] Arena points: %arena_points | Honor Points: %honor_points
pvp_arena_team = [PVP] %bracket: <%team_name> (rating %team_rating)
pvp_no_arena_team = [PVP] I have no Arena Team.
The defaults keep the previous English strings, but they’re now overridable per locale via the DB.

src/strategy/actions/TellPvpAction.cpp
Wishmaster117: solved

src/strategy/actions/TellPvpAction.cpp
Wishmaster117: solved

src/strategy/actions/ChatActionContext.h
Wishmaster117: The idea was to players have the two possibilities using pvp or arena, which feels natural from a player’s perspective.
But thinking about that that's useless, i removed it.

src/strategy/generic/ChatCommandHandlerStrategy.cpp
Wishmaster117: Arena removed

Wishmaster117:
Here is a sample of multibot when merged:
image

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+    inline char const* BracketName(uint8 slot)
+    {
+        // Slots order in AzerothCore: 2v2, 3v3, 5v5
+        switch (slot)
+        {
+            case 0: return "2v2";
+            case 1: return "3v3";
+            case 2: return "5v5";
+        }
```
Celandriel, line 28: Style guide says we should have a default here.

Wishmaster117: All branches are actually covered here: if slot  {0, 1, 2}, execution just falls through the switch and we return "?v?".
So there is no “control reaches end of non-void function” risk and the function is completely safe as-is,  you can think of that final return as an implicit default case.

That said, I’m fine making it an explicit default if you prefer it for style/readability.

Celandriel (dismissed):

src/strategy/actions/TellPvpAction.cpp
noisiver: It should probably be
```cpp
default: return "5v5";
```

instead of

```cpp
case 2: return "5v5";
```

That's the typical standard.

Wishmaster117: > It should probably be
> 
> ```c++
> default: return "5v5";
> ```
> 
> instead of
> 
> ```c++
> case 2: return "5v5";
> ```
> 
> That's the typical standard.

Using default: return "5v5", here would mean that any unexpected slot value (e.g. 42, 255, corrupted data, etc.) would be treated as “5v5”, which is simply wrong.
The intention of BracketName is to map the known indices 0/1/2 to 2v2/3v3/5v5, and treat everything else as “unknown”. That’s why I keep:
case 0: return "2v2";
case 1: return "3v3";
case 2: return "5v5";
default: return "?v?";
This still follows the style guide (explicit default), but keeps the mapping correct instead of silently turning invalid values into “5v5”.

noisiver: Why would that ever happen? Please explain how or why that could happen.

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "Playerbots.h"
+#include "SharedDefines.h"
+
+namespace
+{
+    inline char const* BracketName(uint8 slot)
+    {
+        switch (slot)
+        {
+            case 0: return "2v2"; // 2v2
```
noisiver: 0 should be ARENA_SLOT_2v2

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,87 @@
...
+            case 1: return "3v3"; // 3v3
```
noisiver: 1 should be ARENA_SLOT_3v3

src/strategy/actions/TellPvpAction.cpp
```diff
@@ -0,0 +1,87 @@
...
+            case 2: return "5v5"; // 5v5
```
noisiver: If you actually plan to keep this then 2 should be ARENA_SLOT_5v5 - but it should be managed to not require this case.

src/strategy/actions/TellPvpAction.cpp
noisiver: Seeing as you run this check in the for loop:
```cpp
        uint32 const teamId = bot->GetArenaTeamId(slot);
        if (!teamId)
            continue;
```
There's no possibility that it will check an invalid arena team and can't ever return anything but 2v2, 3v3 and 5v5. You're running with flawed logic.

noisiver (changes requested):

src/strategy/actions/TellPvpAction.cpp
noisiver: These 3 **should** look like this:
```cpp
switch (slot)
{
    case ARENA_SLOT_2v2: return "2v2";
    case ARENA_SLOT_3v3: return "3v3";
    default: return "5v5";
}
```

Celandriel: Is this the right way? I'm not saying you can ever have a different number from the 0-2, but wouldn't it better to have some other default value?

noisiver: With the current code in this PR you can't ever have anything other than 0-2. It checks if the slot ID is a valid arena team and if not it ignores it. That's why it can't ever be wrong.

noisiver: If it ever went wrong then something is horribly wrong in AzerothCore and this would be the least of our concern.

noisiver: Just to prove what I'm saying, it checks not only once but twice here to make sure there is actually an arena team in that slot id
{9E8A2779-6337-4953-BAEF-F6D5BF8A9A8D}

Celandriel: That's fair. I requested the default value according to style guide. Is it better to have the case or default for 5v5?
I do agree with the magic  numbers change in the code.

Celandriel: That's fair. I requested the default value according to style guide. Is it better to have the case or default for 5v5?
I do agree with the magic  numbers change in the code.

noisiver: It should be ARENA_SLOT_2v2, ARENA_SLOT_3v3 and default (which would be 5v5) since it can never, **ever**, receive any other number so adding extra checks is just bloat for no reason.

data/sql/playerbots/updates/2025_11_16_00_ai_playerbot_multi_pvp_text_variables.sql
```diff
@@ -0,0 +1,101 @@
...
+    '[PVP] 我沒有競技場隊伍。',
+    -- esES
+    '[PVP] No tengo equipo de arena.',
+    -- esMX
+    '[PVP] No tengo equipo de arena.',
+    -- ruRU
+    '[PVP] У меня нет команды арены.'
+WHERE NOT EXISTS (
+    SELECT 1 FROM `ai_playerbot_texts` WHERE `name` = 'pvp_no_arena_team'
+);
```
noisiver: No newline at the end of the file

src/strategy/actions/TellPvpAction.h
```diff
@@ -0,0 +1,21 @@
...
+
+class TellPvpAction : public Action
+{
+public:
+    TellPvpAction(PlayerbotAI* botAI) : Action(botAI, "pvp") {}
+
+    bool Execute(Event event) override;
+};
+
+#endif
```
noisiver: No newline at the end of the file

noisiver (changes requested):

src/strategy/actions/TellPvpAction.cpp
Wishmaster117: I see the point thanks for pointing this

Wishmaster117:
Done, not easy to edit in phone

Celandriel:
Just one more thought. Maybe the command should be `pvp stats` instead of just pvp. It makes it more descriptive?

Celandriel:
Hold on, this was basically ready to go. I was just wondering about a command change?

Wishmaster117:
Updated the user-facing chat command to "pvp stats".

Celandriel (dismissed):

Wishmaster117:
Multibot already updates by the way.

hermensbas (dismissed):

Celandriel (approved):

Celandriel:
Was the intention that this command would work with all bots or only those who see you as master? Currently its limited to bots who see you as master.

Wishmaster117:
> Was the intention that this command would work with all bots or only those who see you as master? Currently its limited to bots who see you as master.

For now, the command only works for the bots for which you are the master.
It’s possible to extend it to all bots, but that will be done in another PR.
Let’s merge this one first.

Wishmaster117:
> Was the intention that this command would work with all bots or only those who see you as master? Currently its limited to bots who see you as master.

I made another PR (https://github.com/mod-playerbots/mod-playerbots/pull/1938) to extend the command to All Bots so or you choose to merge this one or the other one :)

Wishmaster117:
New PR who extend command to all bots : https://github.com/mod-playerbots/mod-playerbots/pull/1938

PR #1793, opened 2025-11-01
Codestyle cpp azerothcore

kadeshar:
I cant test that PR without merging that to mod-playerbots/mod-playerbots master branch :(

### DIFF
--- a/var/.suppress.cppcheck
+++ b/var/.suppress.cppcheck
@@ -1,0 +1,1 @@
+cppcheckError


noisiver:
We need checks for SQL as well, not just C++.

[sql-codestyle.yml](https://github.com/azerothcore/azerothcore-wotlk/blob/master/.github/workflows/sql-codestyle.yml) and [codestyle-sql.py](https://github.com/azerothcore/azerothcore-wotlk/blob/master/apps/codestyle/codestyle-sql.py) are used for that.

You forgot about [.suppress.cppcheck](https://github.com/azerothcore/azerothcore-wotlk/blob/master/.suppress.cppcheck)

apps/codestyle-cpp.py
```diff
@@ -0,0 +1,263 @@
...
+                f"Tab found! Replace it to 4 spaces: {file_path} at line {line_number}")
+            check_failed = True
+
+    # Handle the script error and update the result output
+    if check_failed:
+        error_handler = True
+        results["Misc codestyle check"] = "Failed"
+
+# Main function
+parsing_file(src_directory)
```
noisiver, line 263: Missing newline at the end of the file

kadeshar:
Added var. I think every workflow need to be added by separate PR (at least via github interface)

apps/codestyle-cpp.py
kadeshar: Im only copying, orginal also dont have it
obraz
And codestyle is applying only to c++ files

noisiver:
The SQL part is also about codestyle, I think they should be added together.

kadeshar:
Its other workflow. Im adding currently codestyle for cpp. I will not include sql until this workflow dont start working.

hermensbas (approved):

Issue #1800, opened 2025-11-02
Create a robot arena team. The 5v5 team and the 3v3 team are dissatisfied, causing the queue time to be too long.


ramblerSY:
**Bug Description**
After using .playerbots rndbot init to initialize the team,    select count(*) from arena_team where type=5   The number is 200, but the sql 

SELECT
    at.arenateamid AS TeamID,
    at.name AS TeamName,
    at.rating AS TeamRating,
    p.name AS CaptainName,
    COUNT(atm.guid) AS MemberCount
FROM
    arena_team at
JOIN
    characters p ON at.captainGuid = p.guid
LEFT JOIN
    arena_team_member atm ON atm.arenateamid = at.arenateamid
WHERE
    at.type = 5
GROUP BY
    at.arenateamid, at.name, at.rating, p.name
HAVING
    COUNT(atm.guid) < 5
ORDER BY
    MemberCount ASC, at.rating DESC;

          the number    reaches 167. The same situation is true for 3v3.

advice:  in void RandomPlayerbotFactory::CreateRandomArenaTeams(ArenaType type, uint32 count)     
When creating a team, the number of inspection robots was increased. After testing, the waiting time reached within 1 minute.

test config is
worldserver.conf:
Arena.MaxRatingDifference = 550
Arena.RatingDiscardTimer = 30000
Arena.ArenaStartRating = 1850
Arena.LegacyArenaStartRating = 1850
Arena.ArenaStartPersonalRating = 1700
Arena.ArenaStartMatchmakerRating = 1850

playerbots.conf:
AiPlayerbot.MinRandomBots = 2500
AiPlayerbot.MaxRandomBots = 2500
AiPlayerbot.RandomBotMinLevel = 80
AiPlayerbot.RandomBotMaxLevel = 80
AiPlayerbot.RandomBotJoinBG = 1
AiPlayerbot.RandomBotAutoJoinBG = 1
AiPlayerbot.RandomBotAutoJoinArenaBracket = 14
AiPlayerbot.RandomBotAutoJoinBGRatedArena2v2Count = 10
AiPlayerbot.RandomBotAutoJoinBGRatedArena3v3Count = 200
AiPlayerbot.RandomBotAutoJoinBGRatedArena5v5Count = 200
AiPlayerbot.RandomBotArenaTeam2v2Count = 10
AiPlayerbot.RandomBotArenaTeam3v3Count = 200
[...]

PR #1801, opened 2025-11-02
Update RandomPlayerbotFactory.cpp

ramblerSY:
https://github.com/mod-playerbots/mod-playerbots/issues/1800


privatecore:
Arena teams are initialized/populated inside this method:
https://github.com/mod-playerbots/mod-playerbots/blob/43164e74e1f10ede8bb19231fe7629bec764e5d5/src/factory/PlayerbotFactory.cpp#L4147

PR #1805, opened 2025-11-04
Brace structure formatting

NoxMax:
There is no functional change to the code here.

I am working on this file part of a longer PR and this block was painful to look at. It's got a very long line and braces inside braces inside braces. I'm not sure what the best way to structure it. Here is my suggestion, but I am not sure if it's actually even better now. We got wrapped lines in this block that have 1 space indentation instead of 4 (which is kinda what's implied we should stick to according to the AC standard)

So if anything, I am making this PR to discuss and establish some precedence about how such long lines should be formatted in the future.


src/strategy/values/AttackersValue.cpp
```diff
@@ -166,33 +166,53 @@ bool AttackersValue::IsPossibleTarget(Unit* attacker, Player* bot, float range)
...
+        !(attacker->GetCreatureType() == CREATURE_TYPE_CRITTER && !attacker->IsInCombat()) &&
+        !attacker->HasUnitFlag(UNIT_FLAG_IMMUNE_TO_PC) &&
+        !attacker->HasUnitFlag(UNIT_FLAG_NOT_SELECTABLE) &&
+        bot->CanSeeOrDetect(attacker) &&
+        !(sPlayerbotAIConfig->IsPvpProhibited(attacker->GetZoneId(), attacker->GetAreaId()) &&
+            (attacker->GetGUID().IsPlayer() || attacker->GetGUID().IsPet())) &&
+        !(attacker->IsPlayer() && !attacker->IsPvP() && !attacker->IsFFAPvP() &&
+            (!bot->duel || bot->duel->Opponent != attacker)) &&
+        (
+            !c ||
```
NoxMax: Or maybe keeping most of the original vertical spacing, but just making sure to always use 4 space indentation
```
        (!c ||
            (!c->IsInEvadeMode() &&
                ((!isMemberBotGroup && botAI->HasStrategy("attack tagged", BOT_STATE_NON_COMBAT)) ||
                    leaderHasThreat ||
                        (!c->hasLootRecipient() &&
                            (!c->GetVictim() ||
                                (c->GetVictim() &&
                                    ((!c->GetVictim()->IsPlayer() || bot->IsInSameGroupWith(c->GetVictim()->ToPlayer())) ||
                                        (botAI->GetMaster() && c->GetVictim() == botAI->GetMaster()))))) ||
                        c->isTappedBy(bot))));
```

src/strategy/values/AttackersValue.cpp
```diff
@@ -166,33 +166,53 @@ bool AttackersValue::IsPossibleTarget(Unit* attacker, Player* bot, float range)
...
+                                    (!c->GetVictim()->IsPlayer() || bot->IsInSameGroupWith(c->GetVictim()->ToPlayer())) ||
+                                    (botAI->GetMaster() && c->GetVictim() == botAI->GetMaster())
+                                )
+                            )
+                        )
+                    ) ||
+                    c->isTappedBy(bot)
+                )
+            )
+        );
```
NoxMax: Or maybe keeping most of the original vertical spacing, but just making sure to always use 4 space indentation
```
        (!c ||
            (!c->IsInEvadeMode() &&
                ((!isMemberBotGroup && botAI->HasStrategy("attack tagged", BOT_STATE_NON_COMBAT)) ||
                    leaderHasThreat ||
                    (!c->hasLootRecipient() &&
                        (!c->GetVictim() ||
                            (c->GetVictim() &&
                                ((!c->GetVictim()->IsPlayer() || bot->IsInSameGroupWith(c->GetVictim()->ToPlayer())) ||
                                    (botAI->GetMaster() && c->GetVictim() == botAI->GetMaster()))))) ||
                    c->isTappedBy(bot))));
```

hermensbas: You can assign the conditions to a variable if you look for more readability instead of having the entire thing inline.

NoxMax: I guess such example is approaching a limit that would require that. Maybe a standard to set is if a block requires some X number or more of indentations and/or conditions, then introducing extract variables would be best practice.

privatecore: Formatting won't help here; the entire `AttackersValue::IsPossibleTarget` method needs to be rewritten. ;)

icemansparks: I agree with that, the formatting makes it easier to look at, but the code is still very much a mess when it comes to understandable variable names 🙈

NoxMax:
Now edited with extract variables following Bash's suggestion. All in all, I think we should try to avoid going more than 8-10 indentations deep whenever possible, even 8 might be too much. And if something is returning more than 3 inline conditions, probably best to extract them.
Also I added comments because comments are always great, and removed returned deprecated code but kept the commented-out bools

privatecore:
Actually, there’s no need for any additional variables, and the method can be rewritten to exit earlier to skip unnecessary logic and calls. For ex.:
```cpp
bool AttackersValue::IsPossibleTarget(Unit* attacker, Player* bot, float range)
{
    if (!attacker)
        return false;

    if (!attacker->IsVisible() || !attacker->IsInWorld() || attacker->GetMapId() != bot->GetMapId())
        return false;

    if (attacker->isDead() || attacker->HasSpiritOfRedemptionAura())
        return false;
...
```
Start with general statements and then move to specific ones :wink: 
This will also fit the original codestyle from AzerothCore.

src/strategy/values/AttackersValue.cpp
```diff
@@ -144,67 +144,88 @@ bool AttackersValue::hasRealThreat(Unit* attacker)
...
+    if (!attacker)
+        return false;
+
     Creature* c = attacker->ToCreature();
+    PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
+
     bool rti = false;
-    if (attacker && bot->GetGroup())
+    if (bot->GetGroup())
         rti = bot->GetGroup()->GetTargetIcon(7) == attacker->GetGUID();
```
privatecore: You can freely comment out the entire block above and fix the warning:
```
#24 996.9 /azerothcore/modules/mod-playerbots/src/strategy/values/AttackersValue.cpp:148:10: warning: variable 'rti' set but not used [-Wunused-but-set-variable]
#24 996.9     bool rti = false;
```

NoxMax: Yeah I see now that it's use was already deprecated at `// attacker->isFeared()) && !rti) &&`  and likewise the usage of `Unit* victim = attacker->GetVictim();` has no trace left of it.
I will keep but comment out `float /*range*/` though for future reference.

privatecore:
Some mock tests: https://godbolt.org/z/ecEM6Po1W
Also, it seems we need a nullptr check against `botAI` to prevent any issues.

NoxMax:
It's a slightly functional change for this format only draft, but I am working on another PR that I will post either tomorrow or Friday. It modifies this file among several others and I plan to integrate all the changes made here there + the nullptr check. If that PR is not merged I will convert this draft to a PR and add the check here.

PR #1809, opened 2025-11-04
Simplify IsBotMainTank logic

brighton-chi:
IsBotMainTank() has an excess return false at the end that makes it not function properly. I was going to just delete that, but now I think the function as a whole is needlessly complex. It goes through a process of setting fallbacks based on assistant index if there is no main tank set; that is not something used for IsMainTank(), and I don't see the point in doing so as opposed to having the same general logic as just iterating through tanks generally if there is no main tank set.

Therefore, I'm proposing to greatly simplify IsBotMainTank() to just be IsMainTank(), with the additional requirement that the main tank be a bot, which is how I would expect IsBotMainTank() to work in practice.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -2219,58 +2219,7 @@ bool PlayerbotAI::IsMainTank(Player* player)
 
 bool PlayerbotAI::IsBotMainTank(Player* player)
 {
-    if (!player->GetSession()->IsBot() || !IsTank(player))
-    {
-        return false;
-    }
-
-    if (IsMainTank(player))
-    {
-        return true;
-    }
-
-    Group* group = player->GetGroup();
-    if (!group)
-    {
-        return true;  // If no group, consider the bot as main tank
-    }
-
-    uint32 botAssistTankIndex = GetAssistTankIndex(player);
-
-    if (botAssistTankIndex == -1)
-    {
-        return false;
-    }
-
-    for (GroupReference* gref = group->GetFirstMember(); gref; gref = gref->next())
-    {
-        Player* member = gref->GetSource();
-        if (!member)
-        {
-            continue;
-        }
-
-        uint32 memberAssistTankIndex = GetAssistTankIndex(member);
-
-        if (memberAssistTankIndex == -1)
-        {
-            continue;
-        }
-
-        if (memberAssistTankIndex == botAssistTankIndex && player == member)
-        {
-            return true;
-        }
-
-        if (memberAssistTankIndex < botAssistTankIndex && member->GetSession()->IsBot())
-        {
-            return false;
-        }
-
-        return false;
-    }
-
-    return false;
+    return player->GetSession()->IsBot() && IsMainTank(player);
 }
 
 uint32 PlayerbotAI::GetGroupTankNum(Player* player)


privatecore:
If the entire logic from `IsBotMainTank` is dropped, it becomes redundant, as it's only used in two places:
https://github.com/mod-playerbots/mod-playerbots/blob/master/src/strategy/raids/ulduar/RaidUlduarTriggers.cpp#L1883
https://github.com/mod-playerbots/mod-playerbots/blob/master/src/strategy/raids/ulduar/RaidUlduarTriggers.cpp#L1980
And it’s already manipulating the bot object, so check against `player->GetSession()->IsBot()` is unnecessary...

brighton-chi:
I don't know what the purpose of IsBotMainTank is where it is used.

@kadeshar wrote the strategy, can you advise?

I don't really think the function needs to exist, considering it's very easy to check for bots with IsMainTank (just using GET_PLAYERBOT_AI in addition) if needed. I'm fine to remove it entirely, but that's not my call obviously.

src/PlayerbotAI.cpp
```diff
@@ -2219,58 +2219,7 @@ bool PlayerbotAI::IsMainTank(Player* player)
 
 bool PlayerbotAI::IsBotMainTank(Player* player)
```
kadeshar, line 2220: This method return that bot is mainTank without taking into consideration real player as main tank.
Its important when we want trigger something only once time best by bot tank. I used that twice but such usage is used multiple time in code without reusable method which is wrong. I dont agree with this changes.

brighton-chi: I'm not understanding. If you want the function to also apply to players, why not just use IsMainTank? There can only ever be one main tank.

kadeshar: If player is marked as main tank that method ignore that and select next tank in order.

brighton-chi: I thought that was the point of IsBotMainTank. If you want players to be included, why not just use IsMainTank?

hermensbas: Whats it intended difference between IsBotMainTank in IsMainTank? The naming is kinda confusing. IsMainTank also accepts bots as player but shouldnt be used on bots? or? or should logic be merged?

brighton-chi: The way I interpret the code is that IsBotMainTank will exclude any human player that is the main tank. I don't know why such a function is useful--I made the PR because the function is busted (extra return false at the end). IsBotMainTank is used only in Ulduar strategies so we need Gromleq to advise on the purpose. I'm not understanding his explanation so far.

hermensbas: Could you try and explain abit more o rephrase what you mean Gromleg?

kadeshar: I want reusable method which make this:
obraz

Code from print screen is not mine and is duplicated multiple time across module.

But looking carefully on IsBotMainTank you have right that is something not right with this method.
You can fix it. If not i can make it.

brighton-chi: So this is supposed to go down the list of tanks from the main tank through each assisttankofindex until it gets to the first one that is a bot?

kadeshar: Sounds good :)

brighton-chi:
Alright, will close this and do a new one in the future maybe that changes the function. For now though, we just need to keep in mind it's broken.

Issue #1803, opened 2025-11-03
Initializing the number of players in the arena team


ramblerSY:
**Bug Description**
1. I use AiPlayerbot.DeleteRandomBotAccounts = 1 to delete all bots to init all bots
2. i edit AiPlayerbot.DeleteRandomBotAccounts = 0,and start worldserver.exe,
 Here I want to talk about my configuration
AiPlayerbot.MinRandomBots = 2500
AiPlayerbot.MaxRandomBots = 2500
AiPlayerbot.RandomBotAutoJoinBGRatedArena3v3Count = 300
AiPlayerbot.RandomBotAutoJoinBGRatedArena5v5Count = 240
AiPlayerbot.RandomBotArenaTeam3v3Count = 300
AiPlayerbot.RandomBotArenaTeam5v5Count = 240

I use many bots and numerous 5v5 and 3v3 teams to reduce the queue time in high-score arenas

i add code in void RandomPlayerbotFactory::CreateRandomArenaTeams(ArenaType type, uint32 count)

        arenateam->SaveToDB();
        
        if (arenateam->GetMembersSize() < (uint32)type)
        {
            LOG_INFO("playerbots", "---------------------Arena team '{}' has insufficient members: {}/{}", arenaTeamName.c_str(),
                     arenateam->GetMembersSize(), (uint32)type);
        }

        sArenaTeamMgr->AddArenaTeam(arenateam);

To check if the number of people in each team is full

3. I found that when creating arena teams, due to an insufficient number of bots, many 5v5 and 3v3 teams are understaffed

4.log:
237/2500 Bot Zovderl logged in
238/2500 Bot Thrudu logged in
239/2500 Bot Sigaunu logged in
240/2500 Bot Wardan logged in
20 random bot guilds available (created this run: 20)
---------------------Arena team 'Shadow Pack' has insufficient members: 1/2
Arena: Player Futleekick (GUID Full: 0x0000000000001329 Type: Player Low: 4905) already has an arena team of type 2
---------------------Arena team 'Wyrm Oath' has insufficient members: 0/2
Arena: Player Futleekick (GUID Full: 0x0000000000001329 Type: Player Low: 4905) already has an arena team of type 2
---------------------Arena team 'Arcane Accord' has insufficient members: 0/2
Arena: Player Futleekick (GUID Full: 0x0000000000001329 Type: Player Low: 4905) already has an arena team of type 2
[...]

PR #1814, opened 2025-11-06
Initializing the number of players in the arena team 

ramblerSY:
https://github.com/mod-playerbots/mod-playerbots/issues/1803

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -1477,6 +1477,7 @@ void PlayerbotAI::ApplyInstanceStrategies(uint32 mapId, bool tellMaster)
             break;
         case 544:
             strategyName = "magtheridon";  // Magtheridon's Lair
+            break;
         case 565:
             strategyName = "gruulslair";  // Gruul's Lair
             break;


privatecore:
Do you want to fill arena teams on each `PlayerbotFactory::Randomize` call?

icemansparks:
Two questions:
1) Wouldn't it be better to do this in `RandomPlayerbotFactory::CreateRandomArenaTeams` instead?
2) It seems like neither the `InitArenaTeam` nor the `CreateRandomArenaTeams` has a faction check for the bots that should be added to the team, which could result in failing to add a bot? (keeping `AllowTwoSide.Interaction.Arena` in mind)

ramblerSY:
> Two questions:
> 
> 1. Wouldn't it be better to do this in `RandomPlayerbotFactory::CreateRandomArenaTeams` instead?
> 2. It seems like neither the `InitArenaTeam` nor the `CreateRandomArenaTeams` has a faction check for the bots that should be added to the team, which could result in failing to add a bot? (keeping `AllowTwoSide.Interaction.Arena` in mind)

You're absolutely right!I think initializing the team every time isn't appropriate, but I'm not sure when the bots have finished logging in. Could you give me some advice? I believe the best time to initialize the team is after the bots have completed login.

ramblerSY:
> Do you want to fill arena teams on each `PlayerbotFactory::Randomize` call?

I tested it and found that the initialization    RandomPlayerbotFactory::CreateRandomArenaTeams(ARENA_TYPE_2v2, sPlayerbotAIConfig->randomBotArenaTeam2v2Count);
RandomPlayerbotFactory::CreateRandomArenaTeams(ARENA_TYPE_3v3, sPlayerbotAIConfig->randomBotArenaTeam3v3Count);
RandomPlayerbotFactory::CreateRandomArenaTeams(ARENA_TYPE_5v5, sPlayerbotAIConfig->randomBotArenaTeam5v5Count);
only happens once. The bots don't reinitialize their teams afterward.

ramblerSY:
When the number of bots is insufficient, the teams are initialized anyway. If there are too many teams, it leads to teams with incomplete members. Increasing the number of teams can reduce arena queue times. Based on my tests, the current queue time is generally under one minute.

icemansparks:
> When the number of bots is insufficient, the teams are initialized anyway. If there are too many teams, it leads to teams with incomplete members. Increasing the number of teams can reduce arena queue times. Based on my tests, the current queue time is generally under one minute.

Yes, i just verified and it seems bots need to be online so they can be added to an arena team.
This creates some challenges though - because with the option for bots to be only active with players - the bots only login after the server is started and a player connects. At this time, the init is already finished and we have incomplete arena teams?
(in theory - i had not time to do a test run)

ramblerSY:
> > 当机器人数量不足时,团队无论如何都会被初始化。如果团队太多,就会导致团队成员不完整。增加球队数量可以减少竞技场排队时间。根据我的测试,当前的排队时间一般在一分钟以内。
> 
> 是的,我刚刚验证过,似乎机器人需要在线才能被添加到竞技场团队中。 但这带来了一些挑战 - 因为机器人只能与玩家一起活动 - 机器人只能在服务器启动并且玩家连接后登录。此时,初始化已经完成,并且我们的竞技场团队尚未完成？ （理论上 - 我没有时间进行试运行）

The method I use is to start initializing the team when the last bot is initialized    at void PlayerbotFactory::Randomize(bool incremental)   
 //if (bot->GetLevel() >= 70)
 if (GetOnlineBotCount() >= sPlayerbotAIConfig->minRandomBots)
 {
     LOG_INFO("playerbots", "================playbotfactory=========Initializing arena teams...{}",
              GetOnlineBotCount());
     pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Arenas");
     // LOG_INFO("playerbots", "Initializing arena teams...");
     InitArenaTeam();
     if (pmo)
         pmo->finish();
 }

If online bot count >= sPlayerbotAIConfig->minRandomBots  then init team

privatecore:
The main problem with arena teams and guilds initialization is that someone placed this logic into the `PlayerbotFactory::Randomize` method, which is called on every AI tick. I understand that the original author may have wanted some randomization behind adding bots to the world and rotating arena teams/guilds, but this decision still seems odd.

Here's how I see it:
1. Arena teams should be initialized (created) after all random bots are added/created (at server startup?).
2. Arena teams should be populated with members in a random manner after the initialization process.
3. I'm not sure about class bots, but they currently also go through this randomization in the `PlayerbotHolder::OnBotLogin` method.

What do you think?

icemansparks:
> * Arena teams should be initialized (created) after all random bots are added/created (at server startup?).
> * Arena teams should be populated with members in a random manner after the initialization process.

I agree with that; the arena team init should not necessarily run on every update, but rather on server startup or even only once after the first bot creation. While it still should make sure that teams are:
- fully populated
- adhere to the level restrictions (it doesn't really make sense to have a randombot at level 10 being added to an arena team right now?)

Adding randomization will be ok for a later step (simulating teams quitting and new teams being formed over time).

For team composition, a dedicated function could be added later on as well, making sure you have some good compositions as well, not just "2 random bots in a 2vs2 team". However, this would also be more of an addition and not an immediately necessary fix.

> * I'm not sure about class bots, but they currently also go through this randomization in the `PlayerbotHolder::OnBotLogin` method.

I don't know much about addclass bots - since they're not part of the arena teams, this should be looked at separately.

ramblerSY:
> The main problem with arena teams and guilds initialization is that someone placed this logic into the `PlayerbotFactory::Randomize` method, which is called on every AI tick. I understand that the original author may have wanted some randomization behind adding bots to the world and rotating arena teams/guilds, but this decision still seems odd.
> 
> Here's how I see it:
> 
> 1. Arena teams should be initialized (created) after all random bots are added/created (at server startup?).
> 2. Arena teams should be populated with members in a random manner after the initialization process.
> 3. I'm not sure about class bots, but they currently also go through this randomization in the `PlayerbotHolder::OnBotLogin` method.
> 
> What do you think?
Yes! i think Arena teams should be initialized after all bots  log in

ramblerSY:
and i think is do this? after  Arena teams   https://github.com/mod-playerbots/mod-playerbots/pull/1789

kadeshar:
Codestyle failed. Please fix that.

privatecore:
> Codestyle failed. Please fix that.

The author is waiting for this PR: https://github.com/mod-playerbots/mod-playerbots/pull/1789. I think it would be better to put this into draft(?) state since this implementation still raises too many questions.

PR #1813, opened 2025-11-05
Threadsafety login/logout/cancellogout and auto logout for lost addclasss bots

hermensbas:
Modified version of https://github.com/mod-playerbots/mod-playerbots/pull/1722
Subreference: https://github.com/mod-playerbots/mod-playerbots/issues/1669

Should also fix multiple logout crashes.

All the credits to @Wishmaster117 

- No direct logouts then via queue to prevent faulty status during strategy handling downstream
- AddClass bots removed when no group or group without the master

We dont care whether they are in LF queues or BG for that matter, thats the responsibility of the owner/master

Tested:
- Removing addClass by leave command and whether the bots logs out
- Multiple addClass in group and raid format
- Leave command during LFG group, dungeon or battleground
- Add class bots and logout and login with master and whether they remain in the group 

(will update after i am done and done testing)

Issue #1812, opened 2025-11-05
Bug. PlayerbotMgr/Database Callback Crash (ACCESS_VIOLATION)


Celandriel:
Categorizing some of Regrads errors into more focused group. 
This is just a quick AI review and grouping of logs to kick off analysis. 

**Bug Description**

Log File | Date/Time | Faulting Function | Line | Cause
-- | -- | -- | -- | --
[29-10_13-29-55].txt [cite: 93614] | Oct 29, 13:29 | std::_Hash<...>::erase [cite: 93615] | xhash line 1144 | Attempting to erase an element from a hash map/set with an invalid pointer, originating from the PlayerbotHolder::HandlePlayerBotLoginCallback function[cite: 93616]. This is likely a use-after-free or invalid iterator/pointer access.
[25-10_21-53-18].txt [cite: 129032] | Oct 25, 21:53 | std::_Hash<...>::erase [cite: 129034] | xhash line 1144 | Identical stack to the crash above. The crash happens during the processing of ready SQL query callbacks (World::ProcessQueryCallbacks [cite: 129036]), suggesting a concurrent access or lifetime issue related to playerbot data structures.
[26-10_18-29-0].txt [cite: 147070] | Oct 26, 18:29 | std::_Hash<...>::erase [cite: 147071] | xhash line 1144 | Identical stack to the crashes above.
[26-10_18-33-1].txt [cite: 164732] | Oct 26, 18:33 | std::_Hash<...>::erase [cite: 164733] | xhash line 1144 | Identical stack to the crashes above.

**Commit Hash**
Unknown. but AzerothCore rev. 5bcd2d517165 2025-10-20 09:52:31 +0300 (Playerbot branch) so recent. 

Modules
Transmog

Playerbot Settings
200 bots

System
OS: windows
Hardware: Xeon

[5bcd2d517165_worldserver.exe_\[29-10_13-29-55\].txt](https://github.com/user-attachments/files/23372623/5bcd2d517165_worldserver.exe_.29-10_13-29-55.txt)

[5bcd2d517165_worldserver.exe_\[25-10_21-53-18\].txt](https://github.com/user-attachments/files/23372626/5bcd2d517165_worldserver.exe_.25-10_21-53-18.txt)

[5bcd2d517165_worldserver.exe_\[26-10_18-29-0\].txt](https://github.com/user-attachments/files/23372629/5bcd2d517165_worldserver.exe_.26-10_18-29-0.txt)

[...]

PR #1816, opened 2025-11-06
Add thread safety for group operations

### LINKED ISSUE #1124: Thread Safety Improvements for AzerothCore Playerbots
The playerbots currently don't correctly reflect the threading model of AzerothCore. We process everything in the map thread, which causes issues with thread-unsafe code. Critical operations like Group modifications, LFG, Auction, Guilds, global chat channels, friends functionality, battleground manager, and playerbot loading are thread-unsafe and should not be processed directly in map threads. This can lead to race conditions, crashes, and unpredictable behavior that's difficult to debug.

Implement a proper threading model for playerbots that respects the core's architecture:

1. Add a hook to WorldScript::OnUpdate to process all thread-unsafe operations
2. Refactor the playerbot code to clearly separate thread-safe and thread-unsafe operations
3. Implement proper synchronization mechanisms for data shared between threads

This would ensure that operations like group modifications, LFG interactions, and guild operations are handled correctly within the world thread context.

1. Implementing thread-safe versions of all the problematic APIs (too extensive, high risk of bugs)
2. Adding fine-grained locks around thread-unsafe code sections (performance concerns)
3. Restructuring the entire playerbot system to avoid thread-unsafe operations (too disruptive)

**Additional context**
This architectural improvement would align the playerbots code with AzerothCore's design principles. It would resolve numerous hard-to-diagnose stability issues and provide a more robust foundation for future playerbot features. The current approach leads to intermittent crashes that are especially noticeable under high server load or when many playerbots are active simultaneously.

blinkysc:
Fixes crashes and race conditions when bots perform group/guild/arena operations by moving thread-unsafe code to world thread.

Potentially fixes #1124

## Changes

- Added operation queue system that runs in world thread
- Group operations (invite, remove, convert to raid, set leader) now queued
- Arena formation refactored to use queue
- Guild operations changed to use packet queueing

## Testing

Set `MapUpdate.Threads` > 1 in worldserver.conf to enable multiple map threads, then test:
- Group formation and disbanding
- Arena team formation
- Guild operations (invite, promote, demote, remove)

- Run with TSAN
cmake ../ \
  -DCMAKE_CXX_FLAGS="-fsanitize=thread -g -O1" \
  -DCMAKE_C_FLAGS="-fsanitize=thread -g -O1" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=thread" \
  -DCMAKE_INSTALL_PREFIX=/path/to/install \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo

build

export TSAN_OPTIONS="log_path=tsan_report:halt_on_error=0:second_deadlock_stack=1"
./worldserver

The crashes/race conditions should no longer occur with concurrent map threads.

## New Files

- `PlayerbotOperation.h` - Base class defining the operation interface (Execute, IsValid, GetPriority)
- `PlayerbotOperations.h` - Concrete implementations: GroupInviteOperation, GroupRemoveMemberOperation, GroupConvertToRaidOperation, GroupSetLeaderOperation, ArenaGroupFormationOperation
- `PlayerbotWorldThreadProcessor.h/cpp` - Singleton processor with mutex-protected queue, processes operations in WorldScript::OnUpdate hook, handles batch processing and validation

### DIFF
--- a/src/PlayerbotOperation.h
+++ b/src/PlayerbotOperation.h
@@ -20,7 +20,7 @@
  * Thread Safety:
  * - The constructor and data members must be thread-safe (use copies, not pointers)
  * - Execute() is called in the world thread and can safely perform thread-unsafe operations
- * - Subclasses must not store raw pointers to game objects (use ObjectGuid instead)
+ * - Subclasses must not store raw pointers to (core/world thread) game object (use ObjectGuid instead)
  */
 class PlayerbotOperation
 {


hermensbas:
Are you adding thread dispatchers? Just peeking at the code?

blinkysc:
> Are you adding thread dispatchers? Just peeking at the code?

Yes, essentially. I added an operation queue system with a world thread processor.
Map threads queue operations (via PlayerbotWorldThreadProcessor::QueueOperation()), and the processor dispatches them in the world thread through WorldScript::OnUpdate(). It's a producer-consumer pattern where map threads are producers and the world thread consumes/executes the queued operations.

hermensbas:
So this will give us the option to stop abusing world packets or provide a way calling world-thread functions (in)directly through the thread queue directly

Because i was working around the topic in regard of login/logout etc. And kinda came to conclusion it should be fixed on a higher level then i am doing now. Was particular looking at the Playerbots.cpp hooks

blinkysc:
> Collaborator

Exactly. This gives you both options: Direct world-thread calls via the operation queue - create an operation class and queue it, no packet abuse needed. And/or QueuePacket still works but is now optional (I left guild operations using it since it was simpler)

The queue pattern is extensible - just add new operation classes for whatever world-thread functions you need. Way cleaner than packet workarounds.

hermensbas (approved):
Tested

-BattleGroundJoin
-Guild action
-InviteToGroup
-PassLeadershipToMaster

Looked at threadProperties, for now hardcoded seems fine until proven otherwise.

src/Playerbots.cpp
```diff
@@ -329,6 +331,12 @@ class PlayerbotsWorldScript : public WorldScript
 
         LOG_INFO("server.loading", ">> Loaded playerbots config in {} ms", GetMSTimeDiffToNow(oldMSTime));
         LOG_INFO("server.loading", " ");
+        LOG_INFO("server.loading", "Playerbots World Thread Processor initialized");
+    }
+
+    void OnUpdate(uint32 diff) override
+    {
```
hermensbas, line 338: Maybe add some explaining here for those who do not understand by simply reading the code?

hermensbas:
> > Collaborator
> 
> Exactly. This gives you both options: Direct world-thread calls via the operation queue - create an operation class and queue it, no packet abuse needed. And/or QueuePacket still works but is now optional (I left guild operations using it since it was simpler)
> 
> The queue pattern is extensible - just add new operation classes for whatever world-thread functions you need. Way cleaner than packet workarounds.

Agree, or at least have this as additional option in those grey areas where existing world packets simply not suited for the scenario. Although we have to remind ppl that working world-packets or related functions would have the preference in certain use cases to ensure core processing/details are handled as intended.

privatecore:
Looks promising! But wouldn’t it be better to create several levels of priority right away so others can use them in their code instead of relying on abstract values from 0 to 100? For example, 0 = lowest level, 1 = default (normal), 2 = high level, 3 = critical level?

hermensbas:
> Looks promising! But wouldn’t it be better to create several levels of priority right away so others can use them in their code instead of relying on abstract values from 0 to 100? For example, 0 = lowest level, 1 = default (normal), 2 = high level, 3 = critical level?

We can always add it later once tested? But i assume you mean?

```c++

enum PlayerbotOperationPriority : uint32
{
    PRIORITY_LOW        = 0,   // Logging, statistics
    PRIORITY_NORMAL     = 10,  // Background operations (default)
    PRIORITY_HIGH       = 50,  // Player-facing operations (group invites)
    PRIORITY_CRITICAL   = 100  // Crash prevention, cleanup
};

virtual PlayerbotOperationPriority GetPriority() const { return PRIORITY_NORMAL; }
```

?

privatecore:
Yep, but in a slightly different way since @blinkysc already mentioned in the description that it's about [priority levels](https://github.com/mod-playerbots/mod-playerbots/pull/1816/files#diff-b1e3d63cc67e0a7bf9dcd971e1004a08fecb956a69a855a71975691089a6e2e1R52):
```cpp
enum PlayerbotOperationPriority : uint8
{
    PRIORITY_LOW        = 0, // Logging, statistics
    PRIORITY_NORMAL     = 1, // Background operations (default)
    PRIORITY_HIGH       = 2, // Player-facing operations (group invites)
    PRIORITY_CRITICAL   = 3  // Crash prevention, cleanup
};
```

blinkysc:
> Yep, but in a slightly different way since @blinkysc already mentioned in the description that it's about [priority levels](https://github.com/mod-playerbots/mod-playerbots/pull/1816/files#diff-b1e3d63cc67e0a7bf9dcd971e1004a08fecb956a69a855a71975691089a6e2e1R52):
> 
> ```c++
> enum PlayerbotOperationPriority : uint8
> {
>     PRIORITY_LOW        = 0, // Logging, statistics
>     PRIORITY_NORMAL     = 1, // Background operations (default)
>     PRIORITY_HIGH       = 2, // Player-facing operations (group invites)
>     PRIORITY_CRITICAL   = 3  // Crash prevention, cleanup
> };
> ```

The enum @hermensbas prescribed would probably be better; this way you preserve the granularity

privatecore:
> The enum @hermensbas prescribed would probably be better; this way you preserve the granularity

Nah, that doesn’t make sense:
priority = 60 https://github.com/blinkysc/mod-playerbots/blob/threadSafety/src/PlayerbotOperations.h#L381
priority = 70 https://github.com/blinkysc/mod-playerbots/blob/threadSafety/src/PlayerbotOperations.h#L423

Or do you want to create a constant for each priority value? :confused:

blinkysc:
> > The enum @hermensbas prescribed would probably be better; this way you preserve the granularity
> 
> Nah, that doesn’t make sense: priority = 60 https://github.com/blinkysc/mod-playerbots/blob/threadSafety/src/PlayerbotOperations.h#L381 priority = 70 https://github.com/blinkysc/mod-playerbots/blob/threadSafety/src/PlayerbotOperations.h#L423
> 
> Or do you want to create a constant for each priority value? 😕

I just used the constant and applied the higher values based on perceived priority. The enum is fine either way, you can have 5,10,15 or 1,2,3,4. 0-100 just felt like it was ok, can be leveraged later with refactoring (ie need something between group and arena) and ~50 is pretty much your regular "group" actions. Logout has a higher priority because it should happen before others. Arena has a lot going on so wanted it slightly higher than group actions. Relative ordering matters, not absolute values

hermensbas:
Personally i would use default value which offers some space for adding more custom granularity. Regardless did anyone run tests, look more in depth the general idea.  And able to run some tests? i've running a dozen of test trying to invoke multiple actions on multi clients but its still very limited way to tests such things.

@Regrad server would be a better way to test the outcome of the intended behaviour

blinkysc:
@Regrad @here I included cmake options for a TSAN as well if you could add those (keep in mind this creates significant slowdown much like gdb). Even if you have no problems I wouldn't mind checking out the "tsan_report" if there is any type of lock contention

Wishmaster117:
> > > The enum @hermensbas prescribed would probably be better; this way you preserve the granularity
> > 
> > 
> > Nah, that doesn’t make sense: priority = 60 https://github.com/blinkysc/mod-playerbots/blob/threadSafety/src/PlayerbotOperations.h#L381 priority = 70 https://github.com/blinkysc/mod-playerbots/blob/threadSafety/src/PlayerbotOperations.h#L423
> > Or do you want to create a constant for each priority value? 😕
> 
> I just used the constant and applied the higher values based on perceived priority. The enum is fine either way, you can have 5,10,15 or 1,2,3,4. 0-100 just felt like it was ok, can be leveraged later with refactoring (ie need something between group and arena) and ~50 is pretty much your regular "group" actions. Logout has a higher priority because it should happen before others. Arena has a lot going on so wanted it slightly higher than group actions. Relative ordering matters, not absolute values

I find it a shame that after such a change (congratulations to the author), a priority system wasn't added immediately instead of continuing with "magic numbers"; if refactoring is going to happen, it would be better to start from scratch on a modular basis.

Something like:

/**
* Operation priority levels.
* Higher number == higher scheduling priority in the world-thread processor.
* Keep returning an int (ABI-stable) so existing overrides are not broken.
*/
enum class OpPriority : int32
{
        Low      = 0,
        Normal   = 1,
        High     = 2,
        Critical = 3
};

 // Default priority stays Normal for most operations unless overridden.
virtual int32 GetPriority() const { 
return static_cast<int32>(OpPriority::Normal);
 }

Etc....

Regrad:
> Personally i would use default value which offers some space for adding more custom granularity. Regardless did anyone run tests, look more in depth the general idea. And able to run some tests? i've running a dozen of test trying to invoke multiple actions on multi clients but its still very limited way to tests such things.
> 
> @Regrad server would be a better way to test the outcome of the intended behaviour

I am ready to test this PR if it is ready for testing. And do I need to include this: https://github.com/mod-playerbots/mod-playerbots/pull/1813 ? What is the expected result I should get from testing?

Wishmaster117:
> > Personally i would use default value which offers some space for adding more custom granularity. Regardless did anyone run tests, look more in depth the general idea. And able to run some tests? i've running a dozen of test trying to invoke multiple actions on multi clients but its still very limited way to tests such things.
> > @Regrad server would be a better way to test the outcome of the intended behaviour
> 
> I am ready to test this PR if it is ready for testing. And do I need to include this: #1813 ?

Just this one i think,  1813 is draft for now, waiting the merge of this one

privatecore:
> I find it a shame that after such a change (congratulations to the author), a priority system wasn't added immediately instead of continuing with "magic numbers"; if refactoring is going to happen, it would be better to start from scratch on a modular basis.

Hold your horses :wink: All these changes still need testing with a reasonable number of real players and activity. Since this PR created a queue and processes operations in batches, we don't know how it will impact performance in the case of long-running operations and/or frequent thread locks.

Regrad:
I started testing.

Wishmaster117:
> > I find it a shame that after such a change (congratulations to the author), a priority system wasn't added immediately instead of continuing with "magic numbers"; if refactoring is going to happen, it would be better to start from scratch on a modular basis.
> 
> Hold your horses 😉 All these changes still need testing with a reasonable number of real players and activity. Since this PR created a queue and processes operations in batches, we don't know how it will impact performance in the case of long-running operations and/or frequent thread locks.

Hopefully we have @Regrad for this !!!
My horses are already holded i have no pc here, just my phone 🤣

Acode is not the better way to do that...

hermensbas:
> > Personally i would use default value which offers some space for adding more custom granularity. Regardless did anyone run tests, look more in depth the general idea. And able to run some tests? i've running a dozen of test trying to invoke multiple actions on multi clients but its still very limited way to tests such things.
> > @Regrad server would be a better way to test the outcome of the intended behaviour
> 
> I am ready to test this PR if it is ready for testing. And do I need to include this: #1813 ? What is the expected result I should get from testing?

Skip 1813, its left in WIP (and broken state) due this change.

hermensbas:
> I started testing.

Take your time, the best outcome would be seeing an increase of stability. If possible some crash logs when it does crash so we can confirm what is working and whatnot. I tried testing is with multiple clients but in all honesty on way to really test is playing with real players and playing with bots in various ways. No better tests.

Thanks for testing @Regrad

nl-saw:
I've tested this now for about an hour. Albeit I am a single player, I haven;t observed anything out of the ordinary.

nl-saw (approved):

blinkysc:
Even if it works with regrad I wouldn’t mind testing by slamming it with bots performing all the actions queued to world thread (maybe like 500 bots) with the tsan and look for contention

hermensbas:
> Even if it works with regrad I wouldn’t mind testing by slamming it with bots performing all the actions queued to world thread (maybe like 500 bots) with the tsan and look for contention

Agree, that being said; I think Regrad has 50-100 real players using the bots. I will run some additional tests pushing it towards 5000 bots and see how the queue backpressure holdsup. Somewhere this weekend.

blinkysc:
> Agree, that being said; I think Regrad has 50-100 real players using the bots. I will run some additional tests pushing it towards 5000 bots and see how the queue backpressure holdsup.

I was thinking something like a branch with this code and hook and separate cpp to just make the bots constantly login logoff and join groups from different continents etc nonstop. Then just observe tsan.

nl-saw:
Might need some more testing for edge cases

Regrad:
So far, there has been one crash, but no crash reports have appeared. Even empty ones. Strange.

Regrad:
I do not know if this is related to this patch, but bots have stopped going to the arena. Endless waiting.

Regrad:
[ae28c3e1dcc4_worldserver.exe_[8-11_20-9-22].txt](https://github.com/user-attachments/files/23433919/ae28c3e1dcc4_worldserver.exe_.8-11_20-9-22.txt)

Regrad:
[ae28c3e1dcc4_worldserver.exe_[8-11_23-59-31].txt](https://github.com/user-attachments/files/23434469/ae28c3e1dcc4_worldserver.exe_.8-11_23-59-31.txt)

Regrad:
[ae28c3e1dcc4_worldserver.exe_[9-11_0-6-34].txt](https://github.com/user-attachments/files/23434484/ae28c3e1dcc4_worldserver.exe_.9-11_0-6-34.txt)

Wishmaster117:
> [ae28c3e1dcc4_worldserver.exe_[9-11_0-6-34].txt](https://github.com/user-attachments/files/23434484/ae28c3e1dcc4_worldserver.exe_.9-11_0-6-34.txt)

According to the logs, the three crashes have the same origin: `PlayerbotHolder::HandlePlayerBotLoginCallback => PlayerbotMgr.cpp:208`, during a `std::unordered_set::erase` on `botLoading`; the `_Keyval` argument is "<Unable to read memory>", typical of corruption/a race on the container.
The callback is triggered via `SQLQueryHolderCallback::InvokeIfReady` in the world thread (World::ProcessQueryCallbacks loop).
PR #1816 does not affect the login callback sequence. It introduces a `PlayerbotWorldThreadProcessor` and moves group operations (`ConvertToRaid`, `AddMember`, logout cleanup) to a queue executed in the world thread.
The crashes occur in the login query callback when clearing the `guid` from `botLoading`. PR #1816 deals with group operations, not the login path or botLoading.
The crash pattern (unreadable key during an erase on unordered_set) points to a data race or double botLoading manipulation between world threads (callback DB).
The symbols show the call from the world thread; if another thread accesses botLoading without protection, the hash table can be corrupted.
BotLoading should probably be synchronized, an ACE_Thread_Mutex should be added to PlayerbotHolder, and all accesses (insert/erase/contains) should be protected.
The erase operation must be locked, but the location where the GUID is inserted must also be locked.

@blinkysc , what do you think?

Celandriel:
That's the same issue from #1812 where  it's deleting an already deleted bot loading. Isn't it as simple as checking if the guid is in the variable?

blinkysc:
@Regrad pushed a fix. Keep rolling, and posting crashes

blinkysc:
> I do not know if this is related to this patch, but bots have stopped going to the arena. Endless waiting.

I reverted to master and tried arena with it still not working? Maybe this was bug before this?

Regrad:
> > I do not know if this is related to this patch, but bots have stopped going to the arena. Endless waiting.
> 
> I reverted to master and tried arena with it still not working? Maybe this was bug before this?

I can't figure out yet at what point the arena stopped working and which commit causes it. I can't check it yet because I'm busy testing this fix. Can you identify the cause, or the commit causing the crash? As far as I understand from your message, it doesn't work for you either?

Regrad:
New crash report
[ae28c3e1dcc4_worldserver.exe_[10-11_17-7-8].txt](https://github.com/user-attachments/files/23459318/ae28c3e1dcc4_worldserver.exe_.10-11_17-7-8.txt)

blinkysc:
> New crash report [ae28c3e1dcc4_worldserver.exe_[10-11_17-7-8].txt](https://github.com/user-attachments/files/23459318/ae28c3e1dcc4_worldserver.exe_.10-11_17-7-8.txt)

Did you git pull the recent commit?

Regrad:
> > New crash report [ae28c3e1dcc4_worldserver.exe_[10-11_17-7-8].txt](https://github.com/user-attachments/files/23459318/ae28c3e1dcc4_worldserver.exe_.10-11_17-7-8.txt)
> 
> Did you git pull the recent commit?

I updated it today as soon as it became available.
image

Regrad:
Can I start testing the new fix?

blinkysc:
> Can I start testing the new fix?

Before that can you rebuild? I see this in the crash log Revision: AzerothCore rev. ae28c3e1dcc4 2025-11-08 00:22:56 +0300 (Playerbot branch) (Win64, RelWithDebInfo, Static) unless your date is wrong?

Regrad:
> > Can I start testing the new fix?
> 
> Before that can you rebuild? I see this in the crash log Revision: AzerothCore rev. ae28c3e1dcc4 2025-11-08 00:22:56 +0300 (Playerbot branch) (Win64, RelWithDebInfo, Static) unless your date is wrong?

ok

Regrad:
I think this crash does not apply to this fix. But I'm throwing it here just in case.

[ae28c3e1dcc4_worldserver.exe_[10-11_22-53-46].txt](https://github.com/user-attachments/files/23471554/ae28c3e1dcc4_worldserver.exe_.10-11_22-53-46.txt)

Regrad:
[ae28c3e1dcc4_worldserver.exe_[11-11_15-57-46].txt](https://github.com/user-attachments/files/23480987/ae28c3e1dcc4_worldserver.exe_.11-11_15-57-46.txt)

Regrad:
I'll drop all the crashes here, just in case. If they are irrelevant, just ignore them.

Regrad:
The arena is still not working. Can someone check if it works without this PR?

Regrad:
At the moment, if we exclude the two crash reports that I dropped above, the server has become much more stable. I even put 44 threads on the ThreadPool and MapUpdate.Threads. The server performance has not deteriorated. The only negative I see so far is that bots have stopped registering for the arena for some reason unknown to me.
This currently applies to d8be6b860077a26132a7f6cf6a3cdb470244bdfd. I'll install the update tomorrow and test it for a couple of days.
It looks very promising at the moment.

hermensbas:
> At the moment, if we exclude the two crash reports that I dropped above, the server has become much more stable. I even put 44 threads on the ThreadPool and MapUpdate.Threads. The server performance has not deteriorated. The only negative I see so far is that bots have stopped registering for the arena for some reason unknown to me. This currently applies to [d8be6b8](https://github.com/mod-playerbots/mod-playerbots/commit/d8be6b860077a26132a7f6cf6a3cdb470244bdfd). I'll install the update tomorrow and test it for a couple of days. It looks very promising at the moment.

I can test that, you mean 2vs2 ranked arena for example?

hermensbas:
Character: lvl 80
conf:
```
2000 bots
level range 70-80
AiPlayerbot.RandomBotAutoJoinBGRatedArena2v2Count = 1
```

Without this PR:
--------------------------
BG: works
2vs2RankedArena: Works

With this PR: 
---------------------------
BG: Works
2vs2RankedArena:  Not working anymore (waited for 15min)

biship:
Been running for 3 days and have not seen any crashes. I've created multiple parties and raids and passed leadership and kicked players. I have not done any bg or arena.

Regrad:
[ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt)
[ae28c3e1dcc4_worldserver.exe_[13-11_2-39-26].txt](https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt)
[ae28c3e1dcc4_worldserver.exe_[13-11_4-12-13].txt](https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt)

Wishmaster117:
> [ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_2-39-26].txt](https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_4-12-13].txt](https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt)

**[ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt)**
Crash : ACCESS_VIOLATION in Item::CheckSoulboundTradeExpired (Item.cpp, AzerothCore Playerbot core).
Called by Player::UpdateSoulboundTradeItems in PlayerStorage.cpp during a player's tick.
Faulty thread : MapUpdater worker, without a direct call to the mod-playerbots module.
Most likely cause: null pointer due to an inconsistent state in the trade data of a soulbound item (BOP tradable) (item deleted or invalid but still present in the trade logic).
The playerbots module does not contain the faulty function, but can more easily trigger this case due to the massive looting/trading by bots.
@blinkysc : Perhaps post this crash report on the Azerothcore GitHub page?

**https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt:**
**Immediate cause of the crash:**
An RTSC "see spell" event arrives with an empty WorldPacket in SeeSpellAction::Execute.
The code reads from it without any checks → out-of-bounds reading of the ByteBuffer → C++ exception → uncaught → crash in the MapThread.
**Faulty area in the module:**
mod-playerbots/src/strategy/actions/SeeSpellAction.cpp, function SeeSpellAction::Execute(Event event) (line ~40 in your build).
**What needs to be fixed :**
Before running p >> ..., verify that the packet contains enough bytes for castCount + spellId + castFlags.
The same applies to targets.Read(...) (or wrap everything in a try { ... } catch (ByteBufferException const&) { ... } block on the AI ​​side, but AcerothCore prefers size checks before reading).
Verify that targets.GetDst() is not nullptr before using it.

**Weak points in this code**
**In SeeSpellAction::Execute:**
No size check:
```
p.rpos(0);
p >> castCount >> spellId >> castFlags; // assumes the packet is at least 6 bytes long
```

No SpellCastTargets check:
SpellCastTargets targets;

```
targets.Read(p, botAI->GetMaster()); // reads even more data
WorldPosition spellPosition(master->GetMapId(), targets.GetDst()->_position);
/ assumes GetDst() != nullptr
```
No exception handling around packet reading.
With a malformed packet (like in log: _wpos = 0), any of these reads can throw an exception, which is exactly what's happening.


**https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt**

**Threads involved:**
World thread
The call stack of the World thread (in the log) shows that it is blocked here:
```
World::Update(...)
-> MapMgr::Update(...)
-> MapUpdater::wait()
```
The World thread is waiting for pending_requests to return to 0.
As long as a MapUpdateRequest (or other UpdateRequest) has not finished, it remains stuck in wait().

Map (worker) thread blocking everything
In another call stack from the same crash, we see a MapUpdater worker thread stuck in the middle of the playerbots code:
Typical path according to the log:

MapUpdater::WorkerThread()
-> MapUpdateRequest::call()
-> Map::Update(...)
-> Player::Update(...)
-> ScriptMgr::OnPlayerAfterUpdate(...)
-> PlayerbotsPlayerScript::OnPlayerAfterUpdate(Player* player, uint32 diff)
-> botAI->UpdateAI(diff);
-> PlayerbotAIBase::UpdateAI(...)
-> PlayerbotAI::UpdateAIInternal(...)
-> PlayerbotAI::DoNextAction(...)
-> Engine::DoNextAction(...) (in src/strategy/Engine.cpp)
-> Engine::ListenAndExecute(...)
-> ListSpellsAction::Execute(Event)
-> ListSpellsAction::GetSpellList(...) (always in src/strategy/actions/ListSpellsAction.cpp)

**What happens in practice**

A master sends a `spells` command to a bot (event with source = "spells", param = "").
The bot's AI receives this event via `Engine::DoNextAction → Engine::ListenAndExecute → ListSpellsAction::Execute`.
`ListSpellsAction::Execute`: builds a list of all the bot's known spells (`GetSpellList`), 
sorts the list with `std::sort` + `CompareSpells` (which rescans every `skillLineAbilityStore` for each comparison),
then sends one chat line per spell to the master using `TellMasterNoFacing`.

All of this runs in a MapUpdater thread while `MapUpdater::wait()` on the World thread waits for the updates to finish.
The processing of this ListSpellsAction is so resource-intensive (especially if the bot knows many spells, and with the 50-line comment limit) that a Map::Update tick takes more than 60 seconds.

The FreezeDetector, which monitors the World thread's execution frequency, detects that it is stuck for more than 60 seconds in MapUpdater::wait() -> it triggers Acore::Abort and the world server crashes.

Why this code are commented?
        // if (++count >= 50)
        // {
        //     std::ostringstream msg;
        //     msg << (spells.size() - 50) << " more...";
        //     botAI->TellMasterNoFacing(msg.str());
        //     break;
        // }

And maybe:
```
Avoid heavy sorting/scanning on every call.
Hide the results of GetSpellList(filter) or pre-calculate/order them differently.
Simplify CompareSpells (avoid re-scanning the entire DBC in the comparator).
Split the sending into multiple ticks.
```

For example: store the list as a "state" on the AI ​​side, and send 20–50 rows per UpdateAI instead of all at once.

@blinkysc , @hermensbas what do you think?

blinkysc:
> [ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_2-39-26].txt](https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_4-12-13].txt](https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt)

All these don't relate to the PR and are separate issues

Regrad:
> > [ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_2-39-26].txt](https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_4-12-13].txt](https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt)
> 
> All these don't relate to the PR and are separate issues

I'm publishing all crash reports just in case. So far, the main problem is a non-functioning arena. Otherwise, everything is fine.

Wishmaster117:
> > [ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_2-39-26].txt](https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_4-12-13].txt](https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt)
> 
> All these don't relate to the PR and are separate issues

I know that I just asking in case of this PR can be improved adding other things

blinkysc:
> I'm publishing all crash reports just in case. So far, the main problem is a non-functioning arena. Otherwise, 
everything is fine.

~~I'm fairly confident the Arena wasn't working before this? I have tried on master branch with same result. Still wouldn't mind if others could confirm or just something with my DB maybe.~~ EDIT: Others able to get Arena without so I will try to make time to debug that

> I know that I just asking in case of this PR can be improved adding other things

Fixes should be in separate PR's

hermensbas:
You have my blessing atleast, nice work!

Regrad:
> You have my blessing atleast, nice work!

Need to fix the arena first.

blinkysc:
> > You have my blessing atleast, nice work!
> 
> Need to fix the arena first.

I have tested and Arena does work... This PR doesn't affect it. Rated Arena is the problem. If you que practice arena it works fine. The system has poor constraint on the rated system so just fails at finding a "suitable" rated team unless you have conditions just right (lots of lvl 80 bots with lots of arena teams)

This is also the case for the master branch

hermensbas:
> e system has poor constraint on the rated system so just fails at finding a "suitable" rated team unless you have conditions just right (lots of lvl 80 bots with lots of arena teams)
> 
> This is also the case for the master branch

Seems unrelated we just both tested it, as in current master shows same behaviour with current config atleast.

hermensbas:
Or do you see different behaviour current then master and this path @Regrad ?

Regrad:
> Or do you see different behaviour current then master and this path @Regrad ?

I checked it out. The arena works without this PR.

hermensbas:
> > Or do you see different behaviour current then master and this path @Regrad ?
> 
> I checked it out. The arena works without this PR.

Will test against once home, or unless someone does that meanwhile. Which arena related config do you have @Regrad to ensure we testing the same thing.

Regrad:
```
AiPlayerbot.RandomBotAutoJoinICBrackets = 1
AiPlayerbot.RandomBotAutoJoinEYBrackets = 2
AiPlayerbot.RandomBotAutoJoinAVBrackets = 3
AiPlayerbot.RandomBotAutoJoinABBrackets = 6
AiPlayerbot.RandomBotAutoJoinWSBrackets = 7

# Battlegrounds count (per bracket!):
AiPlayerbot.RandomBotAutoJoinBGICCount = 0
AiPlayerbot.RandomBotAutoJoinBGEYCount = 1
AiPlayerbot.RandomBotAutoJoinBGAVCount = 0
AiPlayerbot.RandomBotAutoJoinBGABCount = 1
AiPlayerbot.RandomBotAutoJoinBGWSCount = 1

# Arena configuration:
AiPlayerbot.RandomBotAutoJoinArenaBracket = 14

AiPlayerbot.RandomBotAutoJoinBGRatedArena2v2Count = 0
AiPlayerbot.RandomBotAutoJoinBGRatedArena3v3Count = 0
AiPlayerbot.RandomBotAutoJoinBGRatedArena5v5Count = 0

AiPlayerbot.RandomBotArenaTeam2v2Count = 100
AiPlayerbot.RandomBotArenaTeam3v3Count = 100
AiPlayerbot.RandomBotArenaTeam5v5Count = 50

AiPlayerbot.RandomBotArenaTeamMaxRating = 2000
AiPlayerbot.RandomBotArenaTeamMinRating = 1500
```

biship:
Been running for 7 days and have not seen any crashes. I've created multiple parties and raids and passed leadership and kicked players. I have not done any bg or arena. I've run about 10 dungeons with different parties.

Regrad:
@blinkysc is it ready for test?

Regrad:
Arena is now working with this PR

Regrad:
I'm currently testing 3 fixes at the same time, so I'm throwing off the crash report to everyone, just in case, because I can't determine which PR it's going to be.
[495d3d495f17_worldserver.exe_[18-11_23-28-27].txt](https://github.com/user-attachments/files/23618805/495d3d495f17_worldserver.exe_.18-11_23-28-27.txt)

Regrad:
> I'm currently testing 3 fixes at the same time, so I'm throwing off the crash report to everyone, just in case, because I can't determine which PR it's going to be. [495d3d495f17_worldserver.exe_[18-11_23-28-27].txt](https://github.com/user-attachments/files/23618805/495d3d495f17_worldserver.exe_.18-11_23-28-27.txt)

Apart from this error, which I cannot definitively identify as related to this PR, everything else is working very well right now. If this crash does not apply to this PR and @blinkysc has finished the job, then it is ready to merge. The same applies to https://github.com/mod-playerbots/mod-playerbots/pull/1843 and https://github.com/mod-playerbots/mod-playerbots/pull/1841, @hermensbas

hermensbas:
Thanks for the feedback and testing Regrad.

hermensbas:
> > I'm currently testing 3 fixes at the same time, so I'm throwing off the crash report to everyone, just in case, because I can't determine which PR it's going to be. [495d3d495f17_worldserver.exe_[18-11_23-28-27].txt](https://github.com/user-attachments/files/23618805/495d3d495f17_worldserver.exe_.18-11_23-28-27.txt)
> 
> Apart from this error, which I cannot definitively identify as related to this PR, everything else is working very well right now. If this crash does not apply to this PR and @blinkysc has finished the job, then it is ready to merge. The same applies to #1843 and #1841, @hermensbas

@blinkysc could you have a look at the crashlog given? Never saw that one before and since its tracked back to script hooks "OnPlayerAfterUpdate" as in another race condition

blinkysc:
> @blinkysc could you have a look at the crashlog given? Never saw that one before and since its tracked back to script hooks "OnPlayerAfterUpdate" as in another race condition

It's not related to this PR; this one is only for group operations.

src/strategy/actions/GuildManagementActions.cpp
```diff
@@ -135,8 +135,8 @@ bool GuildRemoveAction::isUseful()
 
 void GuildRemoveAction::SendPacket(WorldPacket packet)
 {
-    WorldPackets::Guild::GuildOfficerRemoveMember data = WorldPacket(packet);
-    bot->GetSession()->HandleGuildRemoveOpcode(data);
+    WorldPacket* data = new WorldPacket(packet);
+    bot->GetSession()->QueuePacket(data);
```
kadeshar: 4 method with same body? Sounds like something to improve

src/PlayerbotOperations.h
```diff
@@ -0,0 +1,500 @@
...
+public:
+    GroupInviteOperation(ObjectGuid botGuid, ObjectGuid targetGuid)
+        : m_botGuid(botGuid), m_targetGuid(targetGuid)
+    {
+    }
+
+    bool Execute() override
+    {
+        Player* bot = ObjectAccessor::FindPlayer(m_botGuid);
+        Player* target = ObjectAccessor::FindPlayer(m_targetGuid);
```
kadeshar, line 32: Consider passing this values from ealier IsValid

src/PlayerbotOperations.h
```diff
@@ -0,0 +1,500 @@
...
+
+        if (!bot || !target)
```
kadeshar, line 34: That is not already validated in IsValid?

src/PlayerbotOperations.h
```diff
@@ -0,0 +1,500 @@
...
+{
+public:
+    GroupRemoveMemberOperation(ObjectGuid botGuid, ObjectGuid targetGuid)
+        : m_botGuid(botGuid), m_targetGuid(targetGuid)
+    {
+    }
+
+    bool Execute() override
+    {
+        Player* bot = ObjectAccessor::FindPlayer(m_botGuid);
```
kadeshar, line 113: Same. Consider passing that from IsValid
Consider that in other classes.

src/PlayerbotOperations.h
```diff
@@ -0,0 +1,500 @@
...
+        Player* target = ObjectAccessor::FindPlayer(m_targetGuid);
+
+        if (!bot || !target)
```
kadeshar, line 116: Same. Its bot variable not already validated in IsValid?
Consider that in other classes.

Wishmaster117:
@hermensbas , @kadeshar , regarding to @Regrad feedback, all is green for this PR include arénas, maybe you can try to merge it?

src/PlayerbotWorldThreadProcessor.cpp
```diff
@@ -0,0 +1,217 @@
...
+        std::lock_guard<std::mutex> statsLock(m_statsMutex);
+        m_stats.totalOperationsSkipped++;
+        return false;
+    }
+
+    // Queue the operation
+    m_operationQueue.push(std::move(operation));
+
+    // Update statistics
+    {
```
kadeshar, line 75: This {} are neccesary?

src/PlayerbotWorldThreadProcessor.cpp
```diff
@@ -0,0 +1,217 @@
...
+    return true;
+}
+
+void PlayerbotWorldThreadProcessor::ProcessBatch()
+{
+    // Extract a batch of operations from the queue
+    std::vector<std::unique_ptr<PlayerbotOperation>> batch;
+    batch.reserve(m_batchSize);
+
+    {
```
kadeshar, line 90: This {} are neccesary?

src/PlayerbotOperations.h
```diff
@@ -0,0 +1,500 @@
...
+
+private:
+    ObjectGuid m_botGuid;
+    uint32 m_masterAccountId;
+};
+
+class OnBotLoginOperation : public PlayerbotOperation
+{
+public:
+    OnBotLoginOperation(ObjectGuid botGuid, PlayerbotHolder* holder)
```
kadeshar, line 471: PlayerbotHolder its not game object in terms defined in PlayerbotOperation.h?

Image

src/PlayerbotWorldThreadProcessor.cpp
```diff
@@ -0,0 +1,217 @@
...
+            uint32 startTime = getMSTime();
+
+            // Execute the operation
+            bool success = operation->Execute();
+
+            uint32 executionTime = GetMSTimeDiffToNow(startTime);
+            totalExecutionTime += executionTime;
+
+            // Log slow operations
+            if (executionTime > 100)
```
kadeshar, line 134: We shouldnt check LogLevel before any further operations?

src/PlayerbotWorldThreadProcessor.cpp
hermensbas: Thats to create local scope to limit life when using mutex.

src/PlayerbotWorldThreadProcessor.cpp
hermensbas: Same here local scope for mutex

src/PlayerbotOperations.h
hermensbas: Dont think isValid has been implemented in the processor for sake of simplicity, nevertheless worth considering.

src/PlayerbotWorldThreadProcessor.cpp
hermensbas: Never noticed any logging level checks in core or playerbots in all honesty, so if we talking about performance the only condition added is 'if (executionTime > 100)' so adding log check here seems a bit trivial? The other values needed for making that condition check are needed for the logic further down the path.

Regrad:
In order not to clutter up this topic, I will be dropping all new crash reports on this topic.:
https://github.com/mod-playerbots/mod-playerbots/issues/1840

src/PlayerbotOperations.h
hermensbas: Looked into it, they serve different purpose. Is valid is called in batch before the execute, the additional check in just defense programming and since we need the objects in the execute itself. So they can be equal, sometimes are equal but they dont have to have equal.

src/strategy/actions/GuildManagementActions.cpp
hermensbas: fixed moved to shared GuidManageAction and used reference as param.

@blinkysc hope you dont mind i did the commit myself.

src/PlayerbotOperations.h
hermensbas: Updated the comment which also answers your question:

 *  Subclasses must not store raw pointers to **(core/world thread)** game object (use ObjectGuid instead)

hermensbas:
Just retested the effected actions, all green on my screen.

src/PlayerbotWorldThreadProcessor.cpp
kadeshar: And string building which also including GetName

src/PlayerbotOperations.h
kadeshar: Execute is running right after IsValid. Thats mean that check + getting values is duplicated. Similar solution is in AiObject used in Action/Trigger/Multiplier/Value 

obraz

kadeshar:
Answered in previous comments + send PM

src/PlayerbotOperations.h
hermensbas: They are run twice indeed, but different intensions. But since isValid equals the info required to start the execute which is more coincidence then anything else and therefor it seems there is direct overlap and in this case there is. From a code structure perspective the isValid check can be totally different used info in the execute.

src/PlayerbotWorldThreadProcessor.cpp
hermensbas:
```c++
something like
if (!sLog->ShouldLog("playerbots", LogLevel::LOG_LEVEL_WARN)) prolly 
```

hermensbas: We agreed on adding this later

PR #1818, opened 2025-11-07
Add ability to prevent PVP near flight masters and innkeepers

### LINKED ISSUE #1354: PvpProhibitedZoneIds does not extend to summoned creatures
**Bug Description**
PvpProhibitedZoneIds (and I am guessing PvpProhibitedAreaIds) does not extend to pets and summoned creatures

**Commit Hash**
89556dafa1a3daa0f4954be14f332a8cb12afd3e

**How To Reproduce Bug**
1.Go to Zim'Torga in Zuldrak for example (which is prohibited zone 4323)
2.Do a "playerbot rndbot teleport" to quickly get bots in camping zones.
3.Notice as some of the bots summon creatures, the creatures aggro on PVP players.
4.The players being attacked often don't respond.

**Expected Behavior**
While in prohibited zones, summons should be PASSIVE. Once outside those zones, they can freely aggro.

**Modules**
ah-bot
bot-reset

**Playerbot Settings**
~7000

**System**
Debian

NoxMax:
I like PVP, but I don't like being ambushed the moment I hearthstone to an inn or land at a flight master. I hope this PR can be of use to others who feel the same.

## Current State: 
Right now we have PvpProhibitedZoneIds and PvpProhibitedAreaIds that allows for adding zones/areas one by one where PVP is prohibited. It works for the most part, but one exception I found months ago is that bot pets on aggro state will not respect the PVP prohibition

## Changes Proposed:
Allow the user to set a distance around all flight masters and/or innkeepers where bots would not engage in PVP. The new member, IsNearProtectedNPC, is checked at the same times as the pre-existing PvpProhibitedZoneIds and PvpProhibitedAreaIds, as part of IsPvpProhibited. So there is nothing new that gets checked more frequently than what already exists.

Fixed aggro pets starting fights in PVP prohibited areas. Their AI was doing its own thing regardless of their master bot. I tried stopping them at PetsAction.cpp, then I realized that's about giving them a command to act, not explicitly their act. So I stopped them at AttackersValue.cpp. Small caveat though: if a bot pet is attacking a player/bot, and their target enters a PVP prohibited area, they won't stop attacking until their master stops and gets out of the combat strat.

Refactored IsPossibleTarget in AttackersValue.cpp. I originally posted about this in the #1805 draft. That block was desperate for a better structure..

Reason For Change:
Flight master landings and inn areas have been the most common contentious areas for bot PVP. You can add them to the current PvpProhibitedAreaIds, but you have to do it for each area you would like that to apply. Also, areas can be a lot larger than just the immediate 50 or so yards around an innkeeper that you don't want fights in. 

Testing:
I did about 20 different tests of different situations to make sure everything works, but these are the main criteria that were checked and it would be appreciated if they (and anything else you can think of) can be tested to confirm.

1. That the default values of 0 make no effect on the current state.
2. That PvpProhibitedInnkeeperDistance and PvpProhibitedFlightMasterDistance are independent of each other.
3. That the new configs do not conflict with the old ones. If an area with an innkeeper is listed in PvpProhibitedAreaIds, then the entire area is PVP prohibited, even if PvpProhibitedInnkeeperDistance has small or 0 value.
4. That a bot would stop attacking once the target IsNearProtectedNPC. The attack may not stop in a split second the moment the threshold is crossed, as the PVP check is not done every single tick.

CLOSES #1354, CLOSES #1805

### DIFF
--- a/src/PlayerbotAIConfig.cpp
+++ b/src/PlayerbotAIConfig.cpp
@@ -696,12 +696,12 @@ bool PlayerbotAIConfig::IsInRandomQuestItemList(uint32 id)
     return find(randomBotQuestItems.begin(), randomBotQuestItems.end(), id) != randomBotQuestItems.end();
 }
 
-bool PlayerbotAIConfig::IsPvpProhibited(uint32 zoneId, uint32 areaId, Player* bot)
+bool PlayerbotAIConfig::IsPvpProhibited(uint32 zoneId, uint32 areaId, Player* player)
 {
     if (IsInPvpProhibitedZone(zoneId) || IsInPvpProhibitedArea(areaId) || IsInPvpProhibitedZone(areaId))
         return true;
 
-    if (bot && IsNearProtectedNPC(bot))
+    if (player && IsNearProtectedNPC(player))
         return true;
 
     return false;
@@ -717,16 +717,16 @@ bool PlayerbotAIConfig::IsInPvpProhibitedArea(uint32 id)
     return find(pvpProhibitedAreaIds.begin(), pvpProhibitedAreaIds.end(), id) != pvpProhibitedAreaIds.end();
 }
 
-bool PlayerbotAIConfig::IsNearProtectedNPC(Player* bot)
+bool PlayerbotAIConfig::IsNearProtectedNPC(Player* player)
 {
-    if (!bot)
+    if (!player)
         return false;
 
     if (pvpProhibitedInnkeeperDistance <= 0.0f && pvpProhibitedFlightMasterDistance <= 0.0f)
         return false;
 
-    WorldPosition botPos(bot);
-    uint32 mapId = bot->GetMapId();
+    WorldPosition botPos(player);
+    uint32 mapId = player->GetMapId();
 
     if (pvpProhibitedFlightMasterDistance > 0.0f)
     {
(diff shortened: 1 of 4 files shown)


nl-saw:
Thanks. I agree pets etc should respect the zone/area setting and not PvP.

I do not agree with setting some range on fly/inn keepers. Personally I like it the way it is, and if I am not mistaking it is blizzlike as-is. This is what PvpProhibitedAreaIds is for, if someone is bothered by it.

NoxMax:
Well naturally the default range is 0, so by default this PR only fixes the aggro pet issue. Beyond that, the range feature is so one doesn't have to lookup the area ID for each and every fly/inn area and add them to the conf. Plus, areas are kinda big sometimes, so this gives the option to the player to restrict PVP only in the immediate area of these NPCs

hermensbas:
I agree with nl-saw, you also have to consider this has to maintained and adds more complexity during problems. In the timeframe we trying to stabilize the code and behaviour. Adding feature on feature with the idea because i prefer my bots to be yellow but i make it configurable isnt good enough.

NoxMax:
If you say so. For now though I will make another PR that fixes the pet behavior along with the refactoring, but without the added feature.

PR #1819, opened 2025-11-07
FIX: Battlegrounds - Unset bot's master when current master left BG

nl-saw:
Adds a check for if current master left the BG and group,  if so release set master and carry on in BG.

This prevents multiple bots in (potentially multiple different) BG's to still consider you as their master when you yourself have left.

- Applies when you join BGs with a party of bots.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -374,10 +374,12 @@ void PlayerbotAI::UpdateAIGroupAndMaster()
 
     Group* group = bot->GetGroup();
 
+    bool IsRandomBot = sRandomPlayerbotMgr->IsRandomBot(bot);
+
     // If bot is not in group verify that for is RandomBot before clearing  master and resetting.
     if (!group)
     {
-        if (master && sRandomPlayerbotMgr->IsRandomBot(bot))
+        if (master && IsRandomBot)
         {
             SetMaster(nullptr);
             Reset(true);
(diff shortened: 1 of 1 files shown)


nl-saw:
Tested with both randombots / alt bots in group join BG, as well as mixture of bots and 2nd real player. Works as intended.

You can still also summon them after leaving a BG. However, if you leave the group after leaving the BG without doing so, they will continue the BG on their own.

src/PlayerbotAI.cpp
```diff
@@ -380,12 +386,9 @@ void PlayerbotAI::UpdateAIGroupAndMaster()
...
     }
 
-    if (bot->InBattleground() && bot->GetBattleground()->GetBgTypeID() != BATTLEGROUND_AV)
-        return;
-
-    PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
-    if (!botAI)
-        return;
+    // Bot in BG, but master no longer part of a group: release master
+    if (bot->InBattleground() && master && !master->GetGroup())
```
Celandriel: I wonder if we should be checking for IsRandomBot here. 
I can envision a way in which a mixed group of altbots from multiple players all get transfered to the group leader on exiting the battleground.

nl-saw: I think this is covered as you still want your altbot to continue if you leave. The assumption here is that different altbots have different (real) leader (not raid leader) when going into a BG as a group. Hence only the altbots of which the leader left the BG and group will be affected. Maybe give this change a go and test.

hermensbas (dismissed):

src/PlayerbotAI.cpp
nl-saw: Pushed new commit based on our chat and findings.

Issue #1817, opened 2025-11-07
Playerbots and NPCs walking through air/walls after latest update (AzerothCore rev. b5cdb0e63e4f, mod-playerbot 85c7009)


jsaxinger:
### Current Behaviour

After updating to the latest AzerothCore and mod-playerbot versions, playerbots and NPCs no longer properly follow pathfinding and instead walk through air and walls.

### Environment
AzerothCore revision: b5cdb0e63e4f (2025-11-05 23:18:17 +0100, Playerbot branch)
mod-playerbot revision: `85c7009fe1af8815c203ed963f431b5a08930f86 `(master branch)

### Data files tested:
Data v18 (maps, vmaps, mmaps)
Data v17 (maps, vmaps, mmaps)

### Expected Behaviour

Bots and NPCs should correctly use pathfinding and follow terrain, avoiding obstacles as usual.

### Steps to reproduce the problem

1. Use the above AzerothCore and mod-playerbot revisions.
2. Load either Data v17 or v18.
3. Spawn any playerbot or NPC.
4. Observe movement behavior — bots walk through geometry instead of using proper paths.

### Extra Notes

- The issue occurs with both fresh builds and updated servers.
- Rebuilding mmaps/vmaps does not fix the issue.
- Switching to official AzerothCore revision with Data v18 resolves the problem.

### AC rev. hash/commit

AzerothCore rev. b5cdb0e63e4f 2025-11-05 23:18:17 +0100 (Playerbot branch) (Unix, RelWithDebInfo, Static)
Connected players: 1. Characters in world: 1.
Connection peak: 1.
Server uptime: 11 minute(s) 4 second(s)
Update time diff: 1ms. Last 500 diffs summary:
|- Mean: 3ms
|- Median: 1ms
|- Percentiles (95, 99, max): 11ms, 12ms, 15ms
Using SSL version: OpenSSL 3.0.17 1 Jul 2025 (library: OpenSSL 3.0.17 1 Jul 2025)
Using Boost version: 1.74.0
Using CMake version: 3.25.1
Using MySQL version: 80407
Found MySQL Executable: /usr/bin/mysql
Compiled on: Linux 6.1.0-40-amd64
Worldserver listening connections on port 8085
Realmlist (Realm Id: 1) configured in port 8085
VMAPs status: Enabled. LineOfSight: true, getHeight: true, indoorCheck: true
MMAPs status: Enabled
[...]

PR #1822, opened 2025-11-09
Patch1-for issue #1817

Wishmaster117:
Here is a first patch for issue #1817
---
**Goal**

* Update `AreaTriggerAction` to the **new `MovePoint` signature** (with explicit booleans).
* Tighten the “flying patch” in `MovementActions` so **pathfinding (mmaps)** is disabled **only if the mover (vehicle/bot) can actually fly**, and **not** just because the master is flying (the previous scope was too broad and could produce straight-line splines).

---

**Changes in `AreaTriggerAction.cpp`**

**Objective:**
Call `MovePoint` using the new parameter order — **`forcedMovement, speed, orientation, generatePath, forceDestination`**.
We previously used the short form without explicit booleans. Making them explicit avoids ambiguity during core merges.

---

**Changes in `MovementActions.cpp`**

**Objectives:**

* Stop disabling mmaps due to the **master** being in-flight. Only disable ground pathfinding when the **mover** (vehicle/bot) is actually flying/can fly.
* Pass **`forceDestination` explicitly** to `MovePoint` to match the recent signature and avoid argument shifts.

---

**Summary / Rationale**

Previously, `allowPathVeh/allowPath` was turned off if the **master** was flying. That forced **direct splines** for bots on the ground ⇒ they could appear to “walk through walls/into the void.”
Now we restrict this to “**the mover is flying/can fly**” (vehicle or bot), which prevents the unintended behavior without breaking real flight.
We also make **`forceDestination=false`** explicit to align with the updated `MovePoint` signature (and avoid ABI/argument-order pitfalls), even where the code only used the single `generatePath` argument (8th parameter).

---

**Expected results (please help test):**

* **Ground bots:** resume correct **mmaps** paths even when the **master** is flying.
* **Flying bots/vehicles:** still use 3D splines (no ground PF), as intended by the “flying” patch.
* **Area triggers:** use the new API with explicit parameters, avoiding any ambiguity.

### DIFF
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -320,7 +320,6 @@ bool MovementAction::MoveTo(uint32 mapId, float x, float y, float z, bool idle, 
             bool allowPath = generatePath;
             if (isFlying(bot) || bot->HasUnitMovementFlag(MOVEMENTFLAG_CAN_FLY))
                 allowPath = false;
-            
             mm.Clear();
             if (!backwards)
             {


hermensbas:
Lemme verify the behaviour first with OG code and setting values hardcoded. To verify the effect is gone with those values.

Wishmaster117:
> Lemme verify the behaviour first with OG code and setting values hardcoded. To verify the effect is gone with those values.

Ok but the “flying patch” can disable pathfinding (mmaps) in cases where the bot/vehicle isn’t actually flying (e.g., master is flying => straight spline for a ground bot). Even if that alone doesn’t explain NPCs walking through everything, it’s a major confounding factor.
So in my opinion this changes are needed.

PR #1823, opened 2025-11-10
Feat. Enable Altbots to roam freely

Celandriel:
When a master summons their alts and then leaves the groups the bots will reset and gain the non-combat strategies that are proper to random bots. This makes them all the same. 

The bots are tied to the player session. 

This works, and the easiest way to test is to message alt bots 'rpg status' and it should say something other than IDLE. 

NOTICE. There may be unpredictable effects from enabling this, and is considered highly experimental at this time. 

Small refactor on the definition of altbots, and added a variable for altMaster as current logic depends heavily on master usage.


Celandriel:
One change I just realized isnthat by clearing a master you may be able to invite other players alts.

nl-saw:
> One change I just realized isnthat by clearing a master you may be able to invite other players alts.

There's a config option for that which should be respected.

Unrelated to this particular PR:  ::botAI->IsAlt() could need a refactor on its logic, its not a very robust check atm.

Celandriel:
I'm want to think more about the refactoring because I'm not super happy with how I implemented it. It was just the first thing that came to mind.

hermensbas:
Not testing yet but lets wait for https://github.com/mod-playerbots/mod-playerbots/pull/1816 before considering to merge.
Love the idea though, and i like how you implemented the solution.

Celandriel:
Yup! No hurry, I just had this in progress and wanted to close out the idea before I forgot.

Celandriel:
A note here, I likely need to modify security to use IsAltMaster and prevent other players from taking ones alts.

PR #1827, opened 2025-11-10
feat(randombot): Added true persistence option and teleportation disabling.

SmashingQuasar:
#  Description

This PR adds two new configuration options:

```
# This option will prevent ALL random bots from being randomized.
# It does not prevent periodic teleport.
# Without it, random bots will be periodically randomized unless they are in a guild
# with at least one character who is not a random bot.
AiPlayerbot.DisableRandomBotPeriodicRandomization = 1

# This option will prevent ALL random bots from being periodically teleported.
# THIS SHOULD NOT BE ENABLED AT ALL TIMES. It exists mostly for debugging purposes.
# Disabling periodic teleport will make the bots almost unable to reach new areas.
AiPlayerbot.DisableRandomBotPeriodicTeleportation = 0
```

**It also makes it so all random bots currently within a guild with at least one member who is not a random bot (so not a randombot account) will not be randomized.**

It uses a Meyer's singleton as a registry for in-memory cache.
It does not require any database change.
It uses SOLID principles and DDD patterns (repository, registry, etc.).

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -630,7 +630,7 @@ AiPlayerbot.RandomBotHordeRatio = 50
 AiPlayerbot.DisableDeathKnightLogin = 0
 
 # Enable simulated expansion limitation for talents and glyphs
-# If enabled, limits talent trees to 5 rows plus the middle talent of the 6th row for bots until level 61 
+# If enabled, limits talent trees to 5 rows plus the middle talent of the 6th row for bots until level 61
 # and 7 rows plus the middle talent of the 8th row for bots from level 61 until level 71
 # Default: 0 (disabled)
 AiPlayerbot.LimitTalentsExpansion = 0
@@ -641,6 +641,17 @@ AiPlayerbot.EnableRandomBotTrading = 1
 
 # Configure message prefixes which will be excluded in analysis in trade action to open trade window
 AiPlayerbot.TradeActionExcludedPrefixes = "RPLL_H_,DBMv4,{звезда} Questie,{rt1} Questie"
+
+# This option will prevent ALL random bots from being randomized.
+# It does not prevent periodic teleport.
+# Without it, random bots will be periodically randomized unless they are in a guild
+# with at least one character who is not a random bot.
+AiPlayerbot.DisableRandomBotPeriodicRandomization = 1
+
+# This option will prevent ALL random bots from being periodically teleported.
+# THIS SHOULD NOT BE ENABLED AT ALL TIMES. It exists mostly for debugging purposes.
+# Disabling periodic teleport will make the bots almost unable to reach new areas.
+AiPlayerbot.DisableRandomBotPeriodicTeleportation = 0
 
 #
 #
--- a/src/PlayerbotAIConfig.cpp
+++ b/src/PlayerbotAIConfig.cpp
@@ -647,6 +647,9 @@ bool PlayerbotAIConfig::Initialize()
 
     selfBotLevel = sConfigMgr->GetOption<int32>("AiPlayerbot.SelfBotLevel", 1);
 
+    disableRandomBotPeriodicRandomization = sConfigMgr->GetOption<bool>("AiPlayerbot.DisableRandomBotPeriodicRandomization", false);
+    disableRandomBotPeriodicTeleportation = sConfigMgr->GetOption<bool>("AiPlayerbot.DisableRandomBotPeriodicTeleportation", false);
+
     RandomPlayerbotFactory::CreateRandomBots();
     if (World::IsStopped())
     {
--- a/src/PlayerbotAIConfig.h
+++ b/src/PlayerbotAIConfig.h
@@ -302,6 +302,9 @@ public:
     std::vector<std::string> botCheats;
     uint32 botCheatMask = 0;
 
+	bool disableRandomBotPeriodicRandomization;
+	bool disableRandomBotPeriodicTeleportation;
+
     struct worldBuff
     {
         uint32 spellId;
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -5,18 +5,22 @@
 
 #include "PlayerbotMgr.h"
 
+#include <openssl/sha.h>
+
+#include <algorithm>  // Added for gender choice
 #include <cstdio>
 #include <cstring>
+#include <iomanip>
 #include <istream>
 #include <string>
-#include <openssl/sha.h>
 #include <unordered_set>
-#include <iomanip>
 
+#include "BroadcastHelper.h"
 #include "ChannelMgr.h"
 #include "CharacterCache.h"
 #include "CharacterPackets.h"
 #include "Common.h"
+#include "DatabaseEnv.h"  // Added for gender choice
 #include "Define.h"
 #include "Group.h"
 #include "GroupMgr.h"
@@ -32,12 +36,7 @@
 #include "RandomPlayerbotMgr.h"
 #include "SharedDefines.h"
 #include "WorldSession.h"
-#include "ChannelMgr.h"
-#include "BroadcastHelper.h"
-#include "PlayerbotDbStore.h"
 #include "WorldSessionMgr.h"
-#include "DatabaseEnv.h"        // Added for gender choice
-#include <algorithm>            // Added for gender choice
 
 class BotInitGuard
 {
@@ -73,6 +72,7 @@ class PlayerbotLoginQueryHolder : public LoginQueryHolder
 private:
     uint32 masterAccountId;
     PlayerbotHolder* playerbotHolder;
+
 public:
     PlayerbotLoginQueryHolder(PlayerbotHolder* playerbotHolder, uint32 masterAccount, uint32 accountId, ObjectGuid guid)
         : LoginQueryHolder(accountId, guid), masterAccountId(masterAccount), playerbotHolder(playerbotHolder)
@@ -109,20 +109,33 @@ void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId
     bool linkedAccount = sPlayerbotAIConfig->allowTrustedAccountBots && IsAccountLinked(accountId, masterAccountId);
 
     bool allowed = true;
+
     std::ostringstream out;
     std::string botName;
     sCharacterCache->GetCharacterNameByGuid(playerGuid, botName);
+
     if (!isRndbot && !sameAccount && !sameGuild && !addClassBot && !linkedAccount)
     {
         allowed = false;
+    }
+
+    if (!allowed && masterPlayer && masterPlayer->IsGameMaster())
+    {
+        allowed = true;
+    }
+
+    if (!allowed)
+    {
         out << "Failure: You are not allowed to control bot " << botName.c_str();
     }
+
     if (masterAccountId && masterPlayer)
     {
         PlayerbotMgr* mgr = GET_PLAYERBOT_MGR(masterPlayer);
         if (!mgr)
         {
-            LOG_DEBUG("playerbots", "PlayerbotMgr not found for master player with GUID: {}", masterPlayer->GetGUID().GetRawValue());
+            LOG_DEBUG("playerbots", "PlayerbotMgr not found for master player with GUID: {}",
+                      masterPlayer->GetGUID().GetRawValue());
             return;
         }
         uint32 count = mgr->GetPlayerbotsCount() + botLoading.size();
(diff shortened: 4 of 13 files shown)


biship:
So rndbots are periodically teleported around and periodically randomized? wow, didn't know that. I just assumed they went about their business questing & levelling like human players.

NoxMax:
> Without it, random bots will be periodically randomized unless they are in a guild

So now bots in guilds with players don't randomize regardless of the setting, correct?
If so now I don't need to set Min/MaxRandomBotRandomizeTime to 31104000 to stop my guild bots from randomizing, but still let the other randomize.

hermensbas:
solid en ddd principles in this type of project? Sorry but this PR makes no sense what so ever. I suggest u learn the project first and existing variables. 

And please go read DDD design principles first instead randomly calling something DDD.

The bots are being randomized for a reason, once the bots are smart and good enough to roam freely we could consider this type of change

SmashingQuasar:
> solid en ddd principles in this type of project? Sorry but this PR makes no sense what so ever. I suggest u learn the project first and existing variables.
> 
> And please go read DDD design principles first instead randomly calling something DDD.
> 
> The bots are being randomized for a reason, once the bots are smart and good enough to roam freely we could consider this type of change

I don't know why this response was needed.
I am literally trying to implement something that has been requested many times on the Discord. I've discussed this extensively on the discord on the design and approach with all positive reception. I am not sure why you would close this so abruptly without engaging in a discussion.

I also teach DDD at master's degree at the University and I am not sure what you don't like. It's far from being fully DDD refactored of course I am not going to turn a repository over on its head in one PR.

Anyway, I'll just keep my changes for myself then, it's unfortunate for the people who were looking forward to these changes.

hermensbas:
Then i dont need explain when you introduce new patterns that are not being used currently in whole codebase but implementing for them for single use case, regardless whether is good idea or not might need some discussion or approach guidelines. Whether your truth is better a truth or not is not relevant you simply added a new truth which needs be maintained, supported and carried. Imagine if every PR applies this approach? 

The reason i am directly closing it is that introduce new patterns but "halfway" implemented for a single use case without any plan going forward? 

If the talk is shall we implement thread locking, shall we move towards DDD, shall implement a data layer thats great. Introducing like this is not great regardless whether its good idea or not for a single small use case.

Your touching so much topics steering into direction which obviously needs some discussion before you make a PR like this. Since you added so much for so little i dont feel like going over ur PR in this state.   

In regards of DDD adding an repository/data layer has nothing todo about implementing DDD. DDD is something you consider when you start a project or maybe considering refactoring the whole towards DDD. 

So why i am being so harsh, because clearly you know what your doing, your clearly are aware of certain principles and therefor the impact using or not using certain principles. And you choose to introduce them like this, this is not the way to go and i think you know that. 

I am sorry for being so harsh, but also consider random ppl maintain the code, review, test the code every single time we make a change. And maybe i am reflecting my frustration towards you in regard lets add more features and more configurability instead of fixing the current issues.

Wishmaster117:
I’ll be honest about how I see this PR, both as a developer and as someone who knows the playerbots module a little.

On the functional idea
Conceptually, I actually like what this PR is trying to do.

Having real persistence for some random bots , especially bots tied to real players or guilds, makes a lot of sense. A lot of people clearly want bots that don’t get re-randomized every X days, particularly for “private server with my guild of bots” setups.

So in terms of user need, I’m on board:
exposing config options to disable:
periodic re-randomization, and
periodic teleportation,
is useful and very understandable from an admin/player point of view.

The logic “if a guild has at least one non-random player, don’t randomize its bots” is also smart for that kind of use.

Where it bothers me technically
However, in its current form, there are several things that would make me hesitate to merge this if I were maintaining the project.

1. Changing the default behavior
In the PR, the playerbots.conf.dist is updated with:

AiPlayerbot.DisableRandomBotPeriodicRandomization = 1
AiPlayerbot.DisableRandomBotPeriodicTeleportation = 0

So in the new .dist, periodic randomization is effectively disabled by default, while the actual config reader in PlayerbotAIConfig uses false as the default value.

That means this PR is not just “adding a toggle” ,  it is silently shifting the default design towards “no more periodic randomization” for people who copy/paste or follow the new dist file.

For a module where the core concept of random bots is built around:

randomization and redistribution in the world,
constant refreshing of profiles,
simulating a more “living” MMO,

this is a pretty big design shift, not just a quality-of-life toggle.

From that angle, I fully understand the reaction:
“The bots are being randomized for a reason, once the bots are smart and good enough to roam freely we could consider this type of change.”

Personally, I think this is the biggest mistake of the PR:
the options should be introduced, but with neutral defaults (0 / false), so the current behavior remains intact and server owners can explicitly opt into “true persistence”.

2. Architecture: DDD / registry / repository
The PR introduces:
registry/PlayerGuildRegistry.*
repository/PlayerGuildRepository.*
some utility code,
and a PlayerbotsGuildScript that acts as a bridge between guild events and the registry.

Taken in isolation, the design is clean. But for this project, it feels like we’re introducing a mini DDD/repository/registry architecture in just one area of the module, while the rest of the codebase doesn’t follow that pattern at all.

Playerbots is already complex, but it’s still in an AzerothCore-style C++:
managers,
scripts,
direct DB access,
relatively straightforward service objects,

not a heavy DDD / hexagonal architecture.

For something as simple as “this guild contains at least one non-random player, therefore its bots shouldn’t be randomized”, this adds quite a bit of extra structure to maintain in the future, especially for contributors who are not interested in (or used to) DDD.

So on that point, I do think the @hermensbas  has a valid concern:
the PR is pushing an architecture style that doesn’t match the rest of the project for a relatively small feature.

About the code itself
From what I can see, the code is not “bad code” at all:

The hooks on GuildScript to keep the registry in sync (create, add member, remove member, disband) are logically placed.
The checks using IsRandomBot make sense.
The extra query on OnGuildRemoveMember to decide whether to keep or drop the guild from the registry is justifiable since guild events are rare.

Overall, it’s well thought out. It’s just over-engineered compared to the surrounding code, in my opinion.

I think the main issue here is more about communication than about the idea itself. 🙂
From my point of view, @SmashingQuasar put genuine effort into this work to address a real need and improve the experience for server owners, and it looks like the concept was even discussed in detail on Discord beforehand. It would probably help everyone if we could focus the discussion on how (or if) this fits the current design of random bots and what adjustments would be needed, rather than on the way the proposal was initially structured.

As someone who try to actively works with this module, here’s where I land:

Conceptually

I like the idea:
Real persistence for certain random bots (especially guild-related).
Config toggles to disable periodic randomization and teleportation.
That matches real admin/player needs and fits very well with how people use playerbots on custom servers.

As a PR to merge as-is
I understand @hermensbas opinion of not merge this PR in its current form, because:
It changes behavior by default (randomization disabled in the dist file).
It introduces a relatively heavy DDD-style structure in a codebase that doesn’t use that style elsewhere.
It adds complexity in a subsystem (random bots) that is already fragile, while there are still crashes/edge cases to handle.

What I’d prefer to see instead:
Keep the config options, but default them to “off” to preserve current behavior.
Implement the “guild has at least one non-random player” logic in a simpler way that fits the existing architecture (re-using existing managers/data when possible, minimal extra layers).

Clearly document in the PR description that:
the defaults do not change anything for existing servers,
the new behavior is fully opt-in, and the design of random bots as “world fillers” is respected unless the admin explicitly decides otherwise.

For my own usage
For my own private setup, with a custom playerbots branch, I’d absolutely consider adopting something like this idea, adjusted to my needs, because I do want reliable, persistent bots tied to my characters/guild.

**But for upstream, I think it needs to be stripped down and aligned with the existing style and philosophy of the module before it can be realistically accepted.**

@hermensbas and @SmashingQuasar, would it be possible to reopen the discussion on this?
@SmashingQuasar, if you’re up for it, would you be willing to adjust the PR to align more closely with the current architecture and conventions of the module? 
I don’t mean that we should be “stuck in the past”, but for bigger design changes I think it would be great if we could collaborate on them and maybe split things into smaller, well-scoped refactors. And before going that far, it might be good to first stabilize the existing behavior and then iterate on top of it.

Just my two cents, I think there’s something valuable here if we can find a shape that works for everyone.

biship:
As a user of my own server full of bots, I went from excited to sad within 24 hrs. The more I use this module, the more I realize it's not a living world of bots simulating player behaviour. It's only that up close, when the bots are running around you. This type of PR will move this module towards being a must-have for Azerothcore.

brighton-chi:
I understand Bash and SmashingQuasar had a productive discussion, and the intent is to implement this type of feature, just not in the exact form of this PR.

Ultimately, I think we can also recognize that a solution can simultaneously be better in a vacuum but also not the best approach to apply in context.

SmashingQuasar:
So I think there a few things to be said here.

We did have a productive discussion with @hermensbas regarding this whole thing. Overall, I understand the concern about adding a new behaviour to the mod.
I'm not going to repeat the discussion we had with @hermensbas as we agreed on almost everything and we found a way to work with this concept.

With that said, I think one of the problems that this event highlights is that it is problematic to have long discussion about new developments and concepts on Discord, have them validated by several admins and contributors, only to face a lot of pushback from people not following those discussions. I understand not everyone is always following the discord; however, everyone should keep in mind that they may have missed a considerable amount of discussions when they read a PR and not jump in to conclusions too hastily.

Regarding what @Wishmaster117 said:
I'm going to address the design shift first. I agree that this changes the default design of the mod, and I am totally alright with making the configuration to false by default. In reality, I just committed what I had locally for now since it was a draft PR. Nevertheless, I think it's unreasonable to expect people to guess that the bots they invite to their guilds and raids and give loot to will eventually be entirely randomized. I haven't really seen anyone not be very astonished by this on the Discord. So even if it's a design shift, when it was not documented clearly, and nobody seems to find this to be the right approach, should this design persist just for the sake of it? I don't think it should.
It's also unreasonable to expect people using playerbots to level up 40 altbots  to raid with them. I understand some people don't mind running around with an entire raid of bots doing quests, but I have my doubts about this being the majority of people. One could argue that they could use GM commands to set them to level 60, but when you start justifying a certain design by asking people to cheat, I find this to be a slippery slope that is not going to yield any good result.

Regarding using DDD: I used a very mild (very, very mild) DDD approach. If I had to switch to full DDD, I would have to do a 1000 files PR turning the entire repository on its head and it would be unmergeable and break everything. So you can't just do all of it in one PR. I'm not saying we should do DDD. I'm fine not using it. What I am not fine with is keeping things the way they are.
There is no real structure right now. Almost all files are god-files with 3 000 lines of code mixing up several symbols (class, enum, constants) there are macros everywhere, there does not seem to be any use of real inlining. It's an OSS project that was passed over the years from several developers with different mindsets and approach and that's fine. I'm not blaming anyone about it. However, it cannot stay that way and that is one of the thing I feel like @hermensbas and I agree the most on.

So, it's fine not using DDD, it's not fine to keep cramming everything in `PlayerbotAI.cpp` .

Overall, I'm totally fine with letting this PR in draft for now until it is further tested. I can't reopen it because it was rejected so only the maintainers can reopen it. I think we can find a way to work on that subject without turning the repository on its head. There isn't that much code changed, and that includes the two small refactors I did. There is one thing that I want to work on before this PR is merged and it's inventory management for bots because without inventory cleaning they tend to get stuck on inventory which is one of the problem I identified.

As I said, @hermensbas and I agree on the fact that the core of the module requires more stabilization. It's something I am working on right now as you can see. I just think that this feature can stay in experimental mode for people wanting to use it, without any guarantee of support if it breaks. I'll keep rebasing it on `main` because this feature is critical to me. I am also totally fine using it only for myself but that does not seem to align with what has been repeatedly requested on the discord.

I would be super interested in discussing it more on Discord, this might be more convenient and could accelerate the debate. ❤️

Wishmaster117:
Thanks a lot for taking the time to write such a detailed reply, and for all the work you’re putting into this. It’s really appreciated. ❤️

First of all, I’m glad to read that you and @hermensbas already had a productive discussion and that you’ve found some common ground on the concept and on the need to stabilize and clean up the core of the module. That’s reassuring.

I also want to acknowledge something you pointed out very clearly: the gap between Discord discussions and what people see on GitHub. I’m actually one of those people who doesn’t follow Discord closely enough (mea culpa), so my first comment was mostly based on what I saw in the PR itself, not on the full context of the previous conversations. That definitely biases how things look “from the outside”.

On the design shift: what you say makes a lot of sense to me.  
I still think that, for upstream, having the new config options default to `false` is the safest way to avoid surprising existing servers, and I’m really happy to hear you’re already on board with that. At the same time, I completely agree that expecting admins/players to just “guess” that guild bots they invite, gear up and play with will later be fully re-randomized is not realistic. Most people I know would also be very surprised by that behaviour. So I don’t see any contradiction here: we can both (1) keep the current behaviour as the default via config, and (2) openly question whether that behaviour is actually what most people want in the long run, as long as we document it clearly.

Regarding the architecture/DDD point: my earlier comment was mostly about consistency with the rest of the module. When I see “DDD” mentioned in a PR description, my reflex is to worry about introducing a style that the rest of the codebase doesn’t follow. I understand much better now what you meant: you’re not trying to flip the entire repo into hardcore DDD, you’re just trying to introduce a bit of structure and avoid making the existing “god files” even worse. On that, I 100% agree with you. Keeping to shove everything into `PlayerbotAI.cpp` is not sustainable, and I think most of us feel the same pain when working in that file.

From my point of view, your answer actually fits very well with what I wrote earlier:
- I still like the **idea** of this feature (real persistence, guild-linked bots, config toggles, etc.).
- I still think the **safe** way for upstream is to keep the new behaviour opt-in via config defaults.
- And I completely agree that the current codebase needs more structure and stabilization, not just “more of the same” crammed into giant files.

Personally, I would really like to see this PR reopened as a **draft** once you’re ready:
- with the configs defaulting to `false` for upstream,
- clearly marked as experimental/opt-in for people who want to help test and give feedback,
- and evolving alongside the stabilization work you’re already doing (like inventory management and bots getting stuck on full bags).

That sounds like a very reasonable compromise between:
- respecting the existing design of random bots for current servers,
- and giving people who care about this feature (and there are many) a way to start using it and iterating on it in the open.

In any case, I just wanted to say that I appreciate your work and your willingness to discuss and adapt the approach instead of giving up on the idea. I’m definitely interested in seeing where this goes, and I’d be happy to test or give feedback when I can.

SmashingQuasar:
@Wishmaster117 I couldn't agree more.
I'm going to shift the configuration, and also add a configuration entirely disable this new feature (since the guild membership not being randomized is currently hardcoded). That way we can introduce this as experimental without messing up anything for people.
I would even be in favour of adding an "Experimental features" section of the configuration and tagging them within the code to indicate what is supposed to be very stable and what's not. Also to keep those features on an `experimental` branch or on their own branch until we are a bit more confident about making them more widely available.

Wishmaster117:
> I would even be in favour of adding an "Experimental features" section of the configuration and tagging them within the code to indicate what is supposed to be very stable and what's not. Also to keep those features on an `experimental` branch or on their own branch until we are a bit more confident about making them more widely available.

Agree with that i already talked about @hermensbas about that

PR #1433, opened 2025-07-10
Add Fishing action and fish with master. 

Celandriel:
With @notOrrytrout I started working on enabling bot fishing with master, but also on their own. 
The first commit didnt crash, showing that it was possible to have a bot cast when master does. Currently it compiles but crashes when you try to fish with a bot in the group, whether the bot has fishing or not. It makes me think that the check in FishingValues is broken somehow, but I cant figure out how.

PR #1829, opened 2025-11-11
Fix: Stop pets from fighting in PVP prohibited zones

NoxMax:
Stripped down version of #1818. No new features. Refactors IsPossibleTarget in AttackersValue.cpp to a better style and makes sure pets don't attack in prohibited zones.

Testing:
Confirmed that aggro pets no longer attack in PVP prohibited areas, but still do outside them. Zim'Torga in Zul'Drak is a good example to test this (ID 4323). Lookout for death knights with a Risen Ally (uncontrolled and naturally aggro) now they respect PVP prohibition like their master.

Note: If you manually teleport a bot that is in mid combat to a PVP prohibited area, its aggro pet might still attack, because its master is still in combat strategy. Otherwise the pet will not attack if its master has switched to non-combat.


hermensbas:
Could you add what you tested yourself?

NoxMax:
Updated. Though I would consider a second look at the original feature, particularly because it is less of a new independent code structure and more an amendment to the existing IsPvpProhibited. Still, if you think it's better left off then I understand.

src/strategy/values/AttackersValue.cpp
```diff
@@ -142,57 +141,106 @@ bool AttackersValue::hasRealThreat(Unit* attacker)
...
+                    (c->GetVictim() &&
+                        ((!c->GetVictim()->IsPlayer() || bot->IsInSameGroupWith(c->GetVictim()->ToPlayer())) ||
+                            (botAI->GetMaster() && c->GetVictim() == botAI->GetMaster()))))) ||
+            c->isTappedBy(bot);
+
+        if (!canAttack)
+            return false;
+    }
 
-    return attacker && attacker->IsVisible() && attacker->IsInWorld() && attacker->GetMapId() == bot->GetMapId() &&
```
nl-saw, line 169: I like the done refactor of this beast

hermensbas:
> Updated. Though I would consider a second look at the original feature, particularly because it is less of a new independent code structure and more an amendment to the existing IsPvpProhibited. Still, if you think it's better left off then I understand.

The orginal including the not fighting around flymasters and such u mean? Although i kinda agree its blizzard like so we should not fix that, but that being said. It seems sometimes bots just keep idleing around flymasters, i cant imagine thats intended behaviour. And if not we get as side-effect you getting ganked there alot. 

I would rather look for solution or cause whether this intended behaviour or simply side-effect because the bots aint doing that they suppose to be doing? Wouldnt you agree? but i am not sure whether its intended behaviour of the bots in all honesty

hermensbas:
Can someone else test this as well? And give some report back. Invisible/visible rogue, pvp allowed zone, not allowed zone, gm on/off, pets, etc?

NoxMax:
> Although i kinda agree its blizzard like so we should not fix that

It wasn't intended to be a fix as it's disabled by default. But yeah ganking at hotspots is (mostly) an allowed Blizzlike experience. Some can find it annoying so I thought to give the option to adjust it for those who want.

Bluewalling though (which bots do practice near hotspots or anywhere else) was apparently [patched by Blizz in WOD](https://wowpedia.fandom.com/wiki/Bluewall). I remember the first time I tried it many years, I was 65 sneaking up on a PVP flagged 72. My treachery was not fruitful 😂

NoxMax:
Back implemented another part from #1818: Made sure the check if either the bot or their target IsPvpProhibited.

Also, there's a confusing naming mismatch that can be fixed in a second PR. AttackersValue uses `attacker` to refer to a bot's target (or in another words, the bot's attacker). Meanwhile AttackAction uses `target` to refer to the same thing.

Suggestion: Everything should be unified to use the term `target`, and maybe AttackersValue should be renamed to AggroTargetsValue to align it with PossibleTargetsValue

hermensbas:
> Back implemented another part from #1818: Made sure the check if either the bot or their target IsPvpProhibited.
> 
> Also, there's a confusing naming mismatch that can be fixed in a second PR. AttackersValue uses `attacker` to refer to a bot's target (or in another words, the bot's attacker). Meanwhile AttackAction uses `target` to refer to the same thing.
> 
> Suggestion: Everything should be unified to use the term `target`, and maybe AttackersValue should be renamed to AggroTargetsValue to align it with PossibleTargetsValue

Sounds like a good call

hermensbas:
But we need more ppl to test this current fix before we can merge it

src/strategy/actions/PetsAction.cpp
```diff
@@ -134,6 +134,12 @@ bool PetsAction::Execute(Event event)
             botAI->TellError("Target is not a valid attack target for the bot.");
             return false;
         }
+        if (sPlayerbotAIConfig->IsPvpProhibited(bot->GetZoneId(), bot->GetAreaId())
+            && (targetUnit->IsPlayer() || targetUnit->IsPet()))
+        {
+            botAI->TellError("I cannot command my pet to attack players in PvP prohibited areas.");
```
kadeshar: Text should be translatable.

NoxMax: Could you please elaborate?

kadeshar: Text should be taken from database like

obraz
obraz

Example: https://github.com/mod-playerbots/mod-playerbots/pull/1433/files

NoxMax: I see. For this fix, I only followed the pre-existing format used here https://github.com/mod-playerbots/mod-playerbots/pull/1498/files

Now between AttackAction and PetAction there are +20 pre-existing tells. Technically I can go through each one, but I'm not sure that's within the scope of this fix. Also there's the issue that I am not familiar with the localization structure, so I will need help if this done in this PR.

kadeshar: For fix existing texts there is posted separate issue on github. We only dont want introduce more.

NoxMax: Oh okay. So I don't have worry about the other tells in these files? I looked at the translation structure and I think I understand it, but I have a couple of questions:
1) From my observation at https://github.com/mod-playerbots/mod-playerbots/tree/master/data/sql/playerbots/updates, it looks like translations are being done language by language for existing tells, rather than all translations for tells. Right?
2) For the id, I am seeing that the latest one for merged commits is 1716. The  id for the yet to be merged fishing feature uses 1717 for its error. Should I use 1717 and then change it to 1718 if the fishing feature is merged first, or should I use 1718 from the start?

kadeshar: 1. Translate only this language which you know.
2. Use next free id if possible. But good catch. We need one file will all of this translation for easy check which id is next.

NoxMax:
Translated PetAction message outputs. Due to the nature of translation, this was of course done with a chatbot so take it with a grain of salt.
The localizations use message ids 1717-1735. If #1433 is merged first, this PR will be updated to use 1718-1736. If this is merged first, then @Celandriel please make sure to update your sql file to use id 1736 instead of 1717 and rename it to a date after 2025_11_25_00

kadeshar:
Wow, nice work. Thanks for your contribution :)

PR #1837, opened 2025-11-13
Added doc for our module script hooks we use

hermensbas:
These script hooks are prolly most important thing to understand and where to start
when problems occur related to race conditions and where and when executed. How strategies patterns are called in which context.


hermensbas:
Set on draft priority does to core update and thread handling PR

Issue #1835, opened 2025-11-12
Cure strategies need updated


crispinknox:
**Bug Description**
Class spells for cure strategies are mismatched eg:
shaman strat references cure poison and cure disease instead of cure toxins.
Not sure about other classes

Dreathean:
https://github.com/mod-playerbots/mod-playerbots/pull/2234

PR #1844, opened 2025-11-14
Fix Shaman cure logic to use Cure Toxins in WotLK

Wishmaster117:
This PR fixes Shaman cure behavior in WotLK (see #1835):

Route Shaman cure actions to cast “Cure Toxins” instead of the obsolete “Cure Poison” / “Cure Disease” spells.

Keep existing action and trigger names ("cure poison", "cure disease", "... on party") for backwards compatibility.

Restrict Cleanse Spirit usage to curses only, removing its misuse for poison and disease.

This aligns Shaman dispel logic with WotLK mechanics while preserving the current AI configuration and other classes’ cure systems.


Wishmaster117:
@crispinknox can you confim that all it's ok when you have some time? I looked code of other class and its ok, only shamans are bugued because:
Before Wrath of the Lich King (in Classic and The Burning Crusade), shamans had two distinct spells:
Cure Poison -> removed poisons.
Cure Disease -> removed diseases.
Starting with Wrath of the Lich King (patch 3.0), Blizzard simplified and merged these spells:
Both were replaced by Cleanse Spirit, which dispelled curses by default, and with a talent (Improved Cleanse Spirit in the Restoration tree), it could also remove diseases and poisons.
Therefore, in patch 3.3.5 (2009-2010), shamans no longer had Cure Poison or Cure Disease as separate spells. They used Cleanse Spirit to handle these.

brighton-chi:
Is it a lot to change action and trigger names as well? If they aren't changed, it seems to me like there could be confusion in the future.

Wishmaster117:
> Is it a lot to change action and trigger names as well? If they aren't changed, it seems to me like there could be confusion in the future.

I volontary not changed that for naw as after module stability improvement i think we will to consider a code refactor, so let's say that for now is a small patch to make the things work.

hermensbas (dismissed):

hermensbas:
I will approve again when updated with action refactors of @SmashingQuasar  :)

Wishmaster117:
> I will approve again when updated with action refactors of @SmashingQuasar :)

Done

hermensbas (dismissed):

noisiver:
They should use Cleanse Spirit for everything if they have it. It removes everything Cure Toxins does and curses as well. If they don't have the talent they should use Cure Toxins instead.

Celandriel:
Just checking in here if you were going to implement any changes based on revisions comment?

PR #1845, opened 2025-11-15
Core - Fixe raid markers persists after target dead causes issue with bots running off

### LINKED ISSUE #1833: raid markers persists after target dead causes issue with bots running off

problem seems to be raid markers do not get removed on creature death so when they respawn they are still marked and bots can see those markers from huge distances and drop everything and runs off to kill it
this has happened on numerous occasions thought someone would have noticed their weird behaviour by now 

also would be nice if someone could check if raid markers can be seen by bots as far as over continents or zone borders since i suspect this might be the case was testing the loot roll fix this morning and my bots just ran off like they were chasing something but i had not used any raid markers since yesterday so got me a bit baffled to why they just decided to run off i tried resetting AI / summoning them nothing worked they ignored everything and just ran off

why i suspect raid markers is because i have seen it happen before when creatures have respawned they still have skull marked on them and bots drop everything and start chasing it down also not sure if this is core or playerbots related somewhere in the back of my mind i have a memory of raid markers dissapearing on the core version "without bots" but that was a long time ago so i could be wrong

was not sure if i should create the issue as a bug so went with a safe option ;)

a check of some sort on creature death that also removes the raid markers

**Additional context**

Wishmaster117:
This PR fixes #1833  where bots could keep chasing a raid target icon (usually skull) across very large distances or even different maps after the mark was set. It replaces the previous attempt with a simpler design that keeps values generic and moves context logic into triggers/actions.

Changes

- RtiTargetValue now ignores RTI targets that are farther than sPlayerbotAIConfig->sightDistance from the bot or the master (same map only), while still using AttackersValue::IsValidTarget and LOS checks.
- AttackersValue is reverted to its original behavior (no special RTI + IsInCombat logic).
- RTI triggers only react to "rti target" in combat, as suggested in review (“that condition should be check in trigger…”).
- AttackRtiTargetAction keeps a small local fallback: if "rti target" is null, it resolves the raid icon directly from the group so chat commands like “attack rti target” still work, including out of combat.

Behavior

- Bots no longer run across the world to chase a stale skull far away from the group.
- When the group comes back near the marked mob (within sight distance), the skull is again used as a normal focus hint.
- Automatic RTI behavior is limited to combat via triggers, while explicit chat commands still work out of combat.

Testing

- Marked a mob with skull, killed it, moved far away / changed zone, then let it respawn: bots did not chase it from afar.
- In a normal dungeon/raid pack, bots still focus skull in combat.
- Passing near a previously marked mob: bots only react once it is back within sight distance.

### DIFF
--- a/src/strategy/actions/ChooseTargetActions.cpp
+++ b/src/strategy/actions/ChooseTargetActions.cpp
@@ -88,9 +88,7 @@ bool DropTargetAction::Execute(Event event)
     {
         Spell const* spell = bot->GetCurrentSpell(CURRENT_AUTOREPEAT_SPELL); // Get the current spell being cast by the bot
         if (spell && spell->m_spellInfo->Id == 75) //Check spell is not nullptr before accessing m_spellInfo
-        {
             bot->InterruptSpell(CURRENT_AUTOREPEAT_SPELL); // Interrupt Auto Shot
-        }
     }
     bot->AttackStop();
 
@@ -155,9 +153,7 @@ bool AttackRtiTargetAction::Execute(Event event)
             {
                 ObjectGuid const guid = group->GetTargetIcon(index);
                 if (!guid.IsEmpty())
-                {
                     rtiTarget = botAI->GetUnit(guid);
-                }
             }
         }
     }
@@ -173,9 +169,7 @@ bool AttackRtiTargetAction::Execute(Event event)
         }
     }
     else
-    {
         botAI->TellError("I dont see my rti attack target");
-    }
 
     return false;
 }


src/strategy/values/AttackersValue.cpp
```diff
@@ -146,7 +146,9 @@ bool AttackersValue::IsPossibleTarget(Unit* attacker, Player* bot, float range)
 {
     Creature* c = attacker->ToCreature();
     bool rti = false;
-    if (attacker && bot->GetGroup())
+    // Restrict skull (RTI) handling to in-combat situations so that any future
+    // use of 'rti' does not cause bots to chase marks while out of combat.
+    if (attacker && bot->GetGroup() && bot->IsInCombat())
```
kadeshar: For me rti value is not used at all in this method. I dont know why you adding new conditions which change nothing.

src/strategy/values/RtiTargetValue.cpp
```diff
@@ -51,18 +51,33 @@ Unit* RtiTargetValue::Calculate()
...
     //
     //////////////////////////////////////////////////////end: delete below check
 
-    Unit* unit = botAI->GetUnit(guid);
-    if (!unit || unit->isDead() || !bot->IsWithinLOSInMap(unit) || !AttackersValue::IsValidTarget(unit, bot) ||
-        sServerFacade->IsDistanceGreaterThan(sServerFacade->GetDistance2d(bot, unit),
-                                             sPlayerbotAIConfig->sightDistance))
+    // Ignore raid icons as a targeting hint while the bot is not in combat
+    // to prevent running after stale marked units out of combat.
+    if (!bot->IsInCombat())
```
kadeshar: For me its wrong place resulting that you need implement special logic in place where you want proper rti value. That condition should be check in trigger which makes bots attacking rti target when they are not in combat.

Wishmaster117: Code changed

src/strategy/values/AttackersValue.cpp
Wishmaster117: Code changed

brighton-chi:
@Wishmaster117 

When you say "In a normal dungeon/raid pack, bots still focus skull in combat," does that mean it is intended that marking with skull in a raid will initiate combat? That is what is happening for me so I'd like to confirm.

Also, raid markers are not being cleared in raids. I was using skull to mark targets in the gauntlet leading up to Reliquary of Souls, and those mobs respawn every 15 seconds. They would respawn with the skull still on them. Bosses also do not have RTI marks removed after a wipe, but is it intended not to clear boss marks?

Wishmaster117:
> @Wishmaster117
> 
> When you say "In a normal dungeon/raid pack, bots still focus skull in combat," does that mean it is intended that marking with skull in a raid will initiate combat? That is what is happening for me so I'd like to confirm.
> 
> Also, raid markers are not being cleared in raids. I was using skull to mark targets in the gauntlet leading up to Reliquary of Souls, and those mobs respawn every 15 seconds. They would respawn with the skull still on them. Bosses also do not have RTI marks removed after a wipe, but is it intended not to clear boss marks?

This was happening before my PR when u put the skull, this silly tank immediately attack it.

Abour raid markers, No raids markers doesn't automatically desapear after u kill the mob, but this is a core side and blizz mechanics (if i rememberwell, theydiscussedaboutthatin azerothcore repo). So I can do nothing without patching the core. So if you are near a mob you marked with a raid marker, bots always attack it till u remove the mark manually.
My patch only avoid to bots attack raid markers who are faraway. 
About the stuff of combat initiated when u put the skull we can do something.

brighton-chi:
> > @Wishmaster117
> 
> > 
> 
> > When you say "In a normal dungeon/raid pack, bots still focus skull in combat," does that mean it is intended that marking with skull in a raid will initiate combat? That is what is happening for me so I'd like to confirm.
> 
> > 
> 
> > Also, raid markers are not being cleared in raids. I was using skull to mark targets in the gauntlet leading up to Reliquary of Souls, and those mobs respawn every 15 seconds. They would respawn with the skull still on them. Bosses also do not have RTI marks removed after a wipe, but is it intended not to clear boss marks?
> 
> 
> 
> This was happening before my PR when u put the skull, this silly tank immediately attack it.
> 
> 
> 
> Abour raid markers, No raids markers doesn't automatically desapear after u kill the mob, but this is a core side and blizz mechanics (if i rememberwell, theydiscussedaboutthatin azerothcore repo). So I can do nothing without patching the core. So if you are near a mob you marked with a raid marker, bots always attack it till u remove the mark manually.
> 
> My patch only avoid to bots attack raid markers who are faraway. 
> 
> About the stuff of combat initiated when u put the skull we can do something. 

Got it. Really just some questions on my part.

I wonder if it is authentic for the Core that a raid mark would remain on a mob even after it dies and respawns.

Wishmaster117:
> > > @Wishmaster117
> > 
> > 
> > > 
> > 
> > 
> > > When you say "In a normal dungeon/raid pack, bots still focus skull in combat," does that mean it is intended that marking with skull in a raid will initiate combat? That is what is happening for me so I'd like to confirm.
> > 
> > 
> > > 
> > 
> > 
> > > Also, raid markers are not being cleared in raids. I was using skull to mark targets in the gauntlet leading up to Reliquary of Souls, and those mobs respawn every 15 seconds. They would respawn with the skull still on them. Bosses also do not have RTI marks removed after a wipe, but is it intended not to clear boss marks?
> > 
> > 
> > This was happening before my PR when u put the skull, this silly tank immediately attack it.
> > Abour raid markers, No raids markers doesn't automatically desapear after u kill the mob, but this is a core side and blizz mechanics (if i rememberwell, theydiscussedaboutthatin azerothcore repo). So I can do nothing without patching the core. So if you are near a mob you marked with a raid marker, bots always attack it till u remove the mark manually.
> > My patch only avoid to bots attack raid markers who are faraway.
> > About the stuff of combat initiated when u put the skull we can do something.
> 
> Got it. Really just some questions on my part.
> 
> I wonder if it is authentic for the Core that a raid mark would remain on a mob even after it dies and respawns.

I found the topic in azerothcore git:
https://github.com/azerothcore/azerothcore-wotlk/issues/24187
So to be honest i don't understand if it's blizzlike or a bug now lon

PR #1847, opened 2025-11-16
Karazhan Refactor + Nightbane Strategy

brighton-chi:
This PR completely refactors the Karazhan raid strategies to clean up the code and polish up/introduce new strategies, including for Nightbane.

General changes with respect to the code:
- Moved gating checks for action methods to triggers instead of relying on isUseful functions
- Got rid of the nonsensical helper class I made and switched to a helper namespace
- Broke up some longer action classes into separate methods and/or private members
- Deleted/consolidated some excess code
- Made greater use of early returns
- Renamed methods, multipliers, and triggers to make them easier to understand
- Generally made edits to conform with AC code standards

Below are the implemented strategies. I’m including all of them, not just new/modified ones, because I don’t think the first set of strategies was ever tested. So each boss could probably use complete testing.

### **Trash**

I added strategies for a trash mob that I find particularly annoying:
- Mana Warp: These are, IMO, the most annoying trash mobs in Karazhan. They blow up when low on health, and having two blow up in quick succession is enough to take out a decent chunk of your raid. The strategy directs bots to use pretty much every stun in the book on Mana Warps when they’re low on HP, which is the only way to prevent the explosion.

### **Attumen**

- During the first phase, bots will focus on Midnight (taking either Midnight or Attumen to 25% starts the next phase). When Attumen spawns, the assist tank will pick him up and move him away so that bots don’t get cleaved.
- When Attumen mounts Midnight, starting phase 2, threat is wiped, and bots will pause DPS for a few seconds to allow the main tank to get aggro. All bots, other than the main tank and any bot that pulls aggro, will stack behind Attumen (~6 yards for ranged so Hunters can still attack).

### **Moroes**

- As before, bots will mark and prioritize adds and the boss in the recommended kill order: Dorothea, Catriona, Keira, Rafe, Robin, Crispin, and Moroes. In practice, the enemies will probably be stacked up, and the bots will AoE them down in accordance with their typical AoE strategies, but classes without AoE capabilities should still prioritize the skull.
- Based on testing feedback, added a method for the main tank to prioritize Moroes

### **Maiden of Virtue**

I’ve made only minor changes to Revision’s original strategy here. 

- The tank with aggro will position Maiden in the middle of the room and move her to a healer when Repentance is cast so that the Holy Ground will break the healer’s stun.
- Ranged bots have assigned positions between the columns around the middle of the room (to prevent chain damage from Holy Wrath).

### **The Big Bad Wolf**

- The tank with aggro brings the boss to the front left of the stage. 
- If a bot gets turned into Little Red Riding Hood, it will run around the stage in a counter-clockwise rectangle until the transformation fades. I tweaked this strategy a bit; it's still not perfect, but it works better than before.

### **Romulo and Julianne**

There are no substantive changes to this strategy. As before, in phase 3, when both bosses are active, the bots switch back-and-forth between the bosses by alternating the skull icon based on which boss has lower HP (10% differential).

### **The Wizard of Oz**

There are no substantive changes to this strategy. As before, bots mark the bosses with a skull icon in their recommended kill order: Dorothee, Tito (assuming he spawns before you kill Dorothee), Roar, Strawman, Tinhead, and the Crone. Additionally, Mages will spam Scorch on Strawman to daze him.

### **The Curator**

- The tank will drag the boss to a fixed spot down the hallway. Ranged bots will spread out to avoid chain damage from Arcing Sear, and bots will mark Astral Flares with the skull icon to prioritize them down. 
- Those strategies already existed, but now I made the assist tank also focus on the boss to try to stay second in aggro and therefore absorb Hateful Bolts.
- Added a multiplier to save Bloodlust/Heroism until Evocation.

### **Terestian Illhoof**

There are no substantive changes to this strategy. The bots will mark targets with the skull icon in the following order: Demonic Chains, Kil'rek, and Illhoof.

### **Shade of Aran**
I redid the strategies a bit, and I think (hope) they work better.

- Flame Wreath: Bots will stop moving until the aura fades. There is a bug in which Flame Wreath will sometimes persist long beyond its supposed 20-second duration. If Aran casts Arcane Explosion during this time, you will almost certainly wipe, so that’s frustrating. I made it so that bots will stay in place still and eat the Arcane Explosion because it’s the lesser of two evils, and if you are overgeared, you may be able to survive. In the previous strategy, bots would stop actions entirely, but now they should keep attacking/casting without moving.
- Arcane Explosion: No substantive changes here--bots will run out immediately and stay out of range until the cast finishes.
- Conjured Elementals: No substantive changes here--they will be marked one-by-one by the skull icon, except that the marking will skip any elemental that is banished.
- Ranged Positioning: I redid this strategy. Ranged bots will now maintain a distance between 11 and 15 yards from the boss. This keeps them out of the 10-yard radius in which Aran silences while also keeping them from getting too far away and getting stuck in the alcoves.

### **Netherspite**

I significantly refactored the action methods here, but substantively the original strategy remains mostly intact.

- Red (Tank) Beam: One tank will be assigned to block the beam for each Portal Phase. The assigned tank will dance in and out of the beam (5 seconds in, 5 seconds out). Tanks intentionally do not avoid Void Zones (it was the lesser of two evils for them to take that damage vs. trying to dynamically avoid them, moving the boss, and possibly getting everybody out of position.
- Blue (DPS) Beam: DPS other than Rogues and Warriors are eligible to be assigned (one-by-one) to block this beam. When the assigned blocker reaches 25 stacks of the debuff, they will leave the beam, and the next assigned blocker will take their place. If a Void Zone drops under the assigned blocker, the bot will move along the beam to get out of the Void Zone so that they do not stop blocking.
- Green (Healer) Beam: This works the same way as the Blue Beam, except that eligible blockers are healers, Rogues, and DPS Warriors. Healers that are assigned to block will swap in the same way as Blue Beam blockers. If a Rogue or DPS Warrior is the assigned blocker, however, they will stand in the beam for the entire Portal Phase since they do not suffer any adverse effects from the beam. In this PR, I made the strategy prioritize Rogues and DPS Warriors over healers to try to avoid the need for bots to swap (and to avoid the irritating scenario in which a healer would block the beam for the first half of a phase and then tag in a Rogue or DPS Warrior, which would be wasted by blocking only half of a phase).
- Non-Blockers: They will stay at least 5 yards away from each beam until called to be an assigned blocker. They will also avoid Void Zones.
- Banish Phase: The only strategy I implemented was for bots to avoid residual Void Zones from the Portal Phase.
- Phase Transitions: Bots should pause DPS at the beginning of the encounter and whenever Netherspite transitions back from the Banish Phase to the Portal Phase (which is an aggro reset). Note that this doesn't wipe DOTs, and there's not much I can do about that.

### **Prince Malchezaar**

The action methods are significantly refactored, but the strategy substantively is not changed very much.

- Bots will maintain distance from Infernals. The tank has a larger avoidance radius to give DPS a little bit of margin to work with. Depending on Infernal placement, it is possible for bots to get stuck in some bad positions. In that case, the best solution is to put them on “flee” and lead them to a better position.
- Bots that get Enfeebled will run out of Shadow Nova range. They should pick a path that does not cross within any Infernal's Hellfire radius.
- Added a multiplier to save Bloodlust/Heroism until Phase 3.

### **Nightbane**

**Disclaimer**: Bots are terrible at this encounter, in large part because the map is awful (even before the recent Core changes). So the strategies are not ideal because they need to operate within the bots’ limitations. I STRONGLY suggest you clear the entire Livery Stables (including the upper level) because the mobs in them have a high risk of pulling through the floor of the Master’s Terrace. Ideally, you should clear out the Scullery too.

The strategy uses waypoints toward the Northeastern door to the Master’s Terrace. I tried several different locations, and that worked best for me based on where Nightbane lands (note that he has a different landing spot for the encounter start vs. the start subsequent ground phases).

- Ground Phase, main tank: The main tank uses two waypoints after it picks up the boss—the movement pattern should be kind of like a reverse checkmark, where the tank moves back along the inner edge of the terrace, then pivots and moves at a bit of an angle to the outer edge. I did this as a way to get the tank to face Nightbane sideways across the terrace, which is how he’s supposed to be tanked. The main tank will not get out of Charred Earth. This is intended. The tank cannot move dynamically enough to avoid it while also not turning the boss and wiping the raid.
- Ground phase, ranged: Ranged bots rotate between three waypoints. They start stacked at the same position. If Charred Earth is dropped on that position, the bots will rotate to the second position. If Charred Earth is dropped on that position, they will rotate to the third position. The maximum number of Charred Earths that can be active is two, so if one is dropped on the third position, the ranged bots should rotate back to the first position.
- Ground Phase, other melee: Melee bots have no coded Charred Earth avoidance strategy. They do decently enough with the general “avoid aoe” strategy.
- Flight Phase: Bots move to Nightbane’s flight position—Nightbane is bugged and cannot be attacked in the air in AC, but all bots still need to stay near him or he will wipe the raid with Fireball Barrage. Bots stack on the same position; when Rain of Bones is cast (one time, snapshotted on the target’s location), all bots will move away to a second nearby position. They will then kill the Restless Skeletons. The Flight Phase lasts for 45 seconds, but Nightbane emotes after 35 seconds and goes to land—during the final 10-second period, bots are freed from all strategies and will follow the master. You need to lead them back to the Northeastern part of the terrace before Nightbane lands so that bots can get immediately positioned when he lands.
- Phase Changes: Whenever Nightbane lands, bots should pause DPS for the main tank to get aggro. 
- Managing bots/pets: During the Flight Phase, bots and pets have a tendency to chase Nightbane outside of the boundaries of the map and pull mobs from other parts of the instance as well as cause Restless Skeletons to spawn out of bounds (and then further cause bots/pets to go out of bounds to attack the skeletons). The strategy should solve for this, but if there are still issues, it should be noted. My resolution was to implement the following: (1) when Nightbane takes flight, bots stop attacking and mark him with the moon icon, and Hunters and Warlocks put pets on passive and recall them (they will put pets on defensive again when Nightbane lands), and (2) all temporary pets are disabled for the duration of the fight (Water Elementals, Shaman wolves, Druid treants, etc.).

**Known Issues:**

- The approach to getting Nightbane to turn sideways is not always spot on since it’s a workaround to account for what should be dynamic movement. He can end up at a bit of an angle. I’ve tweaked the positions such that if he is at an angle, it should not be a situation in which any ranged position is exposed to Smoldering Breath or Cleave. That does increase the risk that poor positioning may expose a ranged position to Tail Sweep. This is obviously suboptimal, but that attack is much more survivable.
- Ranged bots move between the three waypoints by reading the presence of the Charred Earth aura on themselves. Sometimes, a bot may reach the next waypoint with the Charred Earth aura from the previous waypoint still on them (depending on timing of ticks), in which case the bot will keep moving to the next waypoint. This can cause an issue in which the ranged group gets split. So you could have Charred Earth dropped on position 1, and some ranged bots will go to position 2 but others will go to position 3, and then if the second Charred Earth is dropped on position 3, the bots at position 3 will move back to position 1 (which is still in Charred Earth) before then moving to position 2. Balancing spacing between waypoints, positioning with respect to the boss, line of sight, and maximum distance is very difficult to do. I messed with the positioning a lot, and I am not sure if this possibility cannot be entirely avoided without creating other issues. I have at least reached a result in which I have not seen any bots cycle indefinitely (though if testing observes this, it is obviously a problem).

Ultimately, I wouldn’t say I’m totally satisfied with the strategy. Wipes are still possible on account of bad RNG. But I think it does make the fight a lot more manageable, as it is a fight that is very difficult with IP nerfs in the absence of any strategy. Previously, I needed to bring 3 healers, and they would still all be out of mana by the first Flight Phase and reliant on MP5 for the remainder of the fight, and I would wipe multiple times before a kill.

### DIFF
--- a/src/strategy/raids/karazhan/RaidKarazhanActions.cpp
+++ b/src/strategy/raids/karazhan/RaidKarazhanActions.cpp
@@ -55,23 +55,19 @@ bool AttumenTheHuntsmanMarkTargetAction::Execute(Event event)
             return Attack(attumenMounted);
         }
     }
-    else
+    else if (Unit* midnight = AI_VALUE2(Unit*, "find target", "midnight"))
     {
-        Unit* midnight = AI_VALUE2(Unit*, "find target", "midnight");
-        if (midnight)
+        if (IsMapIDTimerManager(botAI, bot))
+            MarkTargetWithStar(bot, midnight);
+
+        if (!botAI->IsAssistTankOfIndex(bot, 0))
         {
-            if (IsMapIDTimerManager(botAI, bot))
-                MarkTargetWithStar(bot, midnight);
+            SetRtiTarget(botAI, "star", midnight);
 
-            if (!botAI->IsAssistTankOfIndex(bot, 0))
+            if (bot->GetTarget() != midnight->GetGUID())
             {
-                SetRtiTarget(botAI, "star", midnight);
-
-                if (bot->GetTarget() != midnight->GetGUID())
-                {
-                    bot->SetTarget(midnight->GetGUID());
-                    return Attack(midnight);
-                }
+                bot->SetTarget(midnight->GetGUID());
+                return Attack(midnight);
             }
         }
     }


brighton-chi:
Converting to draft because I don't think anybody should waste their time reviewing the code before the strategy goes through testing (unless you feel like reviewing of course) since testing could result in many changes.

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
```diff
@@ -1,316 +1,355 @@
...
 
-bool RaidKarazhanHelpers::IsFlameWreathActive()
-{
-    Unit* boss = AI_VALUE2(Unit*, "find target", "shade of aran");
-    Spell* currentSpell = boss ? boss->GetCurrentSpell(CURRENT_GENERIC_SPELL) : nullptr;
-    if (currentSpell && currentSpell->m_spellInfo && currentSpell->m_spellInfo->Id == SPELL_FLAME_WREATH)
+        return nullptr;
+    }
+
+    Unit* GetNearestPlayerInRadius(Player* bot, float radius)
```
hermensbas, line 148: Shouldnt this be playerbotAI function?

brighton-chi:
I decided to remove the Spectral Retainer marking because it's very easy for players to do, so introducing the strategy unnecessarily takes away player agency.

isitsimon:
Evening, I've tested Kara tonight. 
Thank you Crow for the efforts you've put in to allow me to test. when testing I weirdly found tank agro is be an issue at the start but not so much towards the end of the raid, using a prot warr and bear druid but maybe they were itemised poorly. Auto geared to blues and IP setting to 50% HP healing and damage. 
Note - I did not revert the HP nerf to the bosses introduced with Wrath.

**Huntsmans**

Confirm - Bots focus down Midnight instead of splitting dps
Confirm - Bots stack up in P2

Suggestions - Bots stacking up works, however this stops the Hunter bots from being able to shoot. Shouldn't stop you killing the boss, but it is sub optimal. 
 - the boss insta killed one of my healers on pull, not sure if he ripped agro or something on pull. 

**Morose**

Confirm - Marked order works 

Suggestions - I found when the bots apply skull marks, it was causing tanks to also ignoring other targets which caused boss to run round the dps if their aoe ripped agro on pull. Maybe this was just me, since my Ele and Fire mage seemed to be blasting.
Maybe, in the same way you did with Curator (tank ignores the skull marker), one tank can focus the boss, the other picks up the adds.

**Maiden** 

Confirm - positioning all worked

Suggestions - None


**TBBW**

Confirmed - positioning

Confirmed - Kiting

Suggestions - None

**OZ**

Confirm - Marked as planned, no issues with agro at all

Suggestions - None

**R&J**

Confirmed - balance of HP

**Nightbane**

Weird and kinda buggy fight animations. look all off from the boss as he's approaching the balcony.

Confirmed - Positioning
Confirmed - moving out of AoE
Confirmed - Stacking for adds (small thing, hunters cant shoot adds when stacked like this)

Suggestions - None, you've done amazing to get this to work

**Curator**

Confirmed - Positioning
Confirmed - Mark on adds 
Confirmed - Off tank on boss

Suggestions - None

**Illhoof**

Confirm - Marks work on Prio Targets

Suggestions - None

**Aran**

Confirm - Flame Wreath movement
Confirm - Arcane explosion movements
Confirm - water ele marking

Suggestions - None

**Netherspite**

Confirm - all beams were soaked
Confirm - all void zones were avoided
Confirm - No agro issues between phases

Suggestions - None, again, very impressive what you've pulled off here.

**Chess Event** - Did not test

**Prince**

Confirm - bots will run out with Enfeeble
Confirm - bots will move away from infernals

Had some issues with Ret paladin not attacking the boss. Not sure if related to this strategy or not. Posted the debug chat to the discord. 


Thank you very much Crow for letting me test this PR. What a throw back Kara has been! My suggestions are just my opinions, feel free to ignore :)

brighton-chi:
Thanks a bunch for this.

Re your healer getting killed by Midnight, how did you do the pull? 

I can try a different stack distance behind Attumen for hunters. The current distance is intended to keep melee in range, but you're right that that screws hunters. The charge minimum range is 8 yards so I could put hunters at say 6 and they should still be able to attack and not get charged.

That's a good idea for Moroes. I can try having the MT set their own RTI target on Moroes. I think it's not perfect because Moroes drops targets with Gouge, but it should at least buy some time for bots to get adds down.

isitsimon:
> Thanks a bunch for this.
> 
> Re your healer getting killed by Midnight, how did you do the pull?
> 
> I can try a different stack distance behind Attumen for hunters. The current distance is intended to keep melee in range, but you're right that that screws hunters. The charge minimum range is 8 yards so I could put hunters at say 6 and they should still be able to attack and not get charged.
> 
> That's a good idea for Moroes. I can try having the MT set their own RTI target on Moroes. I think it's not perfect because Moroes drops targets with Gouge, but it should at least buy some time for bots to get adds down.

It was my pleasure 😁. 

I pulled using @ tank attack my target (button on multibot). 

Yes that new ranged distance for hunters should work, good thinking!.

src/strategy/raids/karazhan/RaidKarazhanHelpers.h
```diff
@@ -1,85 +1,132 @@
...
-    std::vector<Unit*> GetSpawnedInfernals() const;
+    extern std::unordered_map<uint32, time_t> netherspiteDpsWaitTimer;
+    extern std::unordered_map<ObjectGuid, time_t> redBeamMoveTimer;
+    extern std::unordered_map<ObjectGuid, bool> lastBeamMoveSideways;
+    // Nightbane
+    extern std::unordered_map<uint32, time_t> nightbaneDpsWaitTimer;
+    extern std::unordered_map<ObjectGuid, uint8> nightbaneTankStep;
+    extern std::unordered_map<ObjectGuid, uint8> nightbaneRangedStep;
+    extern std::unordered_map<uint32, time_t> nightbaneFlightPhaseStartTimer;
+    extern std::unordered_map<ObjectGuid, bool> nightbaneRainOfBonesHit;
```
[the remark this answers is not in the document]
brighton-chi: Unless somebody tells me otherwise, I'm going to ignore this because I already have setting/resetting of global timers limited to a single designated bot.

brighton-chi:
This is fine to be reviewed. I have no plans to make further changes.

If Nightbane's Charred Earth is centered on a dynamic object, now that I know how to get those, refactoring to use a dynamic object lookup _would_ be better than the current approach. But given that the current approach seems to work well enough, I don't think it's worth the effort. So I'm calling this done unless there is further testing that reveals issues.

src/strategy/raids/karazhan/RaidKarazhanHelpers.cpp
brighton-chi: Weird, says this comment was from 3 weeks ago, but I'm just seeing it now. Anyway, I don't think there is another reason this should be a PlayerbotAI function. It doesn't need to access botAI for any reason; there's no sorting of raid members relating to botAI (such as role), and the function is not intended to be limited to bots because it's just for generally checking distance with respect to all raid members, including human players.

Dreathean: Not just you, this also just popped up for me while saying from 3 weeks ago

brighton-chi: I think Bash probably made the comment on his end 3 weeks ago but didn't finalize his comments so it didn't submit. But when it is submitted, it keeps the original timestamp.

isitsimon:
I tested this 2 weeks ago @kadeshar. Seeing there have been some small alterations since then. 

Do you want a full re test?

kadeshar:
I waiting for other mods response :) Both PR and your tests are impressive. Thanks both for your contribution

noisiver (approved):

PR #1852, opened 2025-11-17
Improve Molten Core Strategy

HennyWilly:
This is my first attempt of implementing playerbot strategies.

A team of 40 can steamroll Molten Core relatively easy, even with lower item levels.
Regardless, this PR adds resistance triggers and actions to mitigate some damage.
Additionally, improvements were made for the encounters with Garr, Baron Geddon, Shazzrah and Golemagg.
A short summary per boss is listed below.
All planned features are included, but feedback is of course appreciated.

### Lucifron
- Shadow resistance: mitigate damage from [Impending Doom](https://www.wowhead.com/classic/spell=19702/impending-doom) and [Shadow Shock](https://www.wowhead.com/classic/spell=19460/shadow-shock).

### Magmadar
- Fire resistance: mitigate damage from [Lava Bomb](https://www.wowhead.com/classic/spell=19411/lava-bomb) and [Magma Spit](https://www.wowhead.com/classic/spell=19450/magma-spit).
- Like King Dred and the fraction commander (Nexus), this fight might profit from an anti-fear strategy in order to counter [Panic](https://www.wowhead.com/classic/spell=19408/panic). Not implemented here.

### Gehennas
- Shadow resistance: mitigate damage from [Shadow Bolt](https://www.wowhead.com/classic/spell=19728/shadow-bolt) and increase the chance to resist [Gehennas' Curse](https://www.wowhead.com/classic/spell=19716/gehennas-curse).

### Garr
- Fire resistance: mitigate damage from the Firesworn adds ([Immolate](https://www.wowhead.com/classic/spell=20294/immolate) and [Eruption](https://www.wowhead.com/classic/spell=19497/eruption)).
-  Disabled dps aoe abilities via multiplier. This one is important because multiple exploding adds at once might delete bots rather quick...

### Baron Geddon
- Refactored the existing strategy.
- Fire resistance: mitigate damage from [Ignite Mana](https://www.wowhead.com/classic/spell=19659/ignite-mana), [Inferno](https://www.wowhead.com/classic/spell=19695/inferno) and [Living Bomb](https://www.wowhead.com/classic/spell=20475/living-bomb).
- Better Inferno handling: Before moving away, bots stop attacking and interrupt their spells. Additionally, the new multiplier prevents bots from running back to Geddon while Inferno is still active.

### Shazzrah
- Ranged bots now position themselves in a sweet spot that prevents them from getting hit with [Arcane Explosion](https://www.wowhead.com/classic/spell=19712/arcane-explosion) but still close enough to dps and heal.

### Sulfuron Harbinger
- Fire resistance: mitigate damage from [Hand of Ragnaros](https://www.wowhead.com/classic/spell=19780/hand-of-ragnaros) and [Immolate](https://www.wowhead.com/classic/spell=20294/immolate). To be fair, this one is quite negligible...

### Golemagg
- Fire resistance: mitigate damage from [Magma Splash](https://www.wowhead.com/classic/spell=13880/magma-splash) and [Pyroblast](https://www.wowhead.com/classic/spell=20228/pyroblast).
-  Disabled dps aoe abilities via multiplier. Kind of a preference on my side. Otherwise, the Core Ragers spam emotes about not wanting to die.

### Majordomo Executus
- Shadow resistance: mitigate damage from [Aegis of Ragnaros](https://www.wowhead.com/classic/spell=20620/aegis-of-ragnaros), [Shadow Shock](https://www.wowhead.com/classic/spell=20603/shadow-shock) and [Shadow Bolt](https://www.wowhead.com/classic/spell=21077/shadow-bolt). This one is also negligible, TBF.

### Ragnaros
- Fire resistance: mitigate damage from [Wrath of Ragnaros](https://www.wowhead.com/classic/spell=20566/wrath-of-ragnaros) and [Lava Burst](https://www.wowhead.com/classic/spell=21158/lava-burst).

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -2266,9 +2266,7 @@ bool PlayerbotAI::IsAssistTankOfIndex(Player* player, int index, bool ignoreDead
         }
 
         if (ignoreDeadPlayers && !member->IsAlive())
-        {
             continue;
-        }
 
         if (group->IsAssistant(member->GetGUID()) && IsAssistTank(member))
         {
@@ -2290,9 +2288,7 @@ bool PlayerbotAI::IsAssistTankOfIndex(Player* player, int index, bool ignoreDead
         }
 
         if (ignoreDeadPlayers && !member->IsAlive())
-        {
             continue;
-        }
 
         if (!group->IsAssistant(member->GetGUID()) && IsAssistTank(member))
         {
--- a/src/strategy/raids/moltencore/RaidMcActions.cpp
+++ b/src/strategy/raids/moltencore/RaidMcActions.cpp
@@ -190,10 +190,7 @@ bool McGolemaggAssistTankAttackCoreRagerAction::Execute(Event event)
     }
 
     if (bot->GetVictim() != myCoreRager)
-    {
-        // Step 3.3: Attack our Core Rager in case we previously switched in 3.2.
-        return Attack(myCoreRager);
-    }
+        return Attack(myCoreRager); // Step 3.3: Attack our Core Rager in case we previously switched in 3.2.
 
     // Step 4: Prevent Golemagg's Trust on Core Ragers
     if (myCoreRager->HasAura(SPELL_GOLEMAGGS_TRUST) ||


src/strategy/raids/moltencore/RaidMcMultipliers.h
```diff
@@ -0,0 +1,27 @@
+#ifndef _PLAYERBOT_RAIDGRUULSLAIRMULTIPLIERS_H
```
Celandriel: Change to _PLAYERBOT_RAIDMCMULTIPLIERS_H

src/strategy/raids/moltencore/RaidMcActions.h
```diff
@@ -1,6 +1,7 @@
 #ifndef _PLAYERBOT_RAIDMCACTIONS_H
 #define _PLAYERBOT_RAIDMCACTIONS_H
 
+#include "AttackAction.h"
```
Celandriel, line 4: Is this being used?

src/strategy/raids/moltencore/RaidMcMultipliers.h
```diff
@@ -0,0 +1,27 @@
...
+#define _PLAYERBOT_RAIDGRUULSLAIRMULTIPLIERS_H
```
Celandriel: _PLAYERBOT_RAIDMCMULTIPLIERS_H

src/strategy/raids/moltencore/RaidMcHelpers.h
```diff
@@ -0,0 +1,14 @@
+#ifndef RAID_GRUULSLAIRHELPERS_H
+#define RAID_GRUULSLAIRHELPERS_H
```
Celandriel: RAID_MCHELPERS_H

src/strategy/raids/moltencore/RaidMcActions.h
Celandriel, line 4: Is this needed here? I dont see any Attack Actions referenced.

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -28,16 +32,30 @@ bool McCheckShouldMoveFromGroupAction::Execute(Event event)
...
         }
     }
     return false;
 }
+
+bool McShazzrahMoveAwayAction::Execute(Event event)
+{
+    if (Unit* boss = AI_VALUE2(Unit*, "find target", "shazzrah"))
+    {
+        constexpr float radius = 26.0f; // make sure we don't get hit by Arcane Explosion but still be in casting range
```
Celandriel: Avoid Magic numbers. ARCANE_EXPLOSION_RANGE

src/strategy/raids/moltencore/RaidMcMultipliers.h
HennyWilly: Changed to _PLAYERBOT_RAIDMCMULTIPLIERS_H

src/strategy/raids/moltencore/RaidMcMultipliers.h
HennyWilly: Changed to _PLAYERBOT_RAIDMCMULTIPLIERS_H

src/strategy/raids/moltencore/RaidMcActions.h
HennyWilly: No, removed

src/strategy/raids/moltencore/RaidMcHelpers.h
HennyWilly: Changed to _PLAYERBOT_RAIDMCHELPERS_H for consistent naming.

src/strategy/raids/moltencore/RaidMcActions.h
HennyWilly: No, this is just a remnant of when I tried to implement an AttackAction. Removed.

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: Added static constexpr for INFERNO_DISTANCE (20.0f) and ARCANE_EXPLOSION_DISTANCE (26.0f) at the beginning of RaidMcActions.cpp.
Is this acceptable?

Celandriel: Yea. Perfect!

src/strategy/raids/moltencore/RaidMcActionContext.h
```diff
@@ -10,13 +10,33 @@ class RaidMcActionContext : public NamedObjectContext<Action>
 public:
     RaidMcActionContext()
     {
+        creators["mc lucifron shadow resistance action"] = &RaidMcActionContext::lucifron_shadow_resistance_action;
```
brighton-chi: I think names for actions and triggers should be consistent--either use "action" or "trigger" at the end of the name or not. I don't think it matters either way as long as it's consistent.

src/strategy/raids/moltencore/RaidMcActionContext.h
```diff
@@ -10,13 +10,33 @@ class RaidMcActionContext : public NamedObjectContext<Action>
...
         creators["mc move from baron geddon"] = &RaidMcActionContext::move_from_baron_geddon;
+        creators["mc shazzrah move away"] = &RaidMcActionContext::shazzrah_move_away;
+        creators["mc sulfuron harbinger fire resistance action"] = &RaidMcActionContext::sulfuron_harbinger_fire_resistance_action;
+        creators["mc golemagg fire resistance action"] = &RaidMcActionContext::golemagg_fire_resistance_action;
+        creators["mc majordomo shadow resistance action"] = &RaidMcActionContext::majordomo_shadow_resistance_action;
+        creators["mc ragnaros fire resistance action"] = &RaidMcActionContext::ragnaros_fire_resistance_action;
     }
 
 private:
+    static Action* lucifron_shadow_resistance_action(PlayerbotAI* ai) { return new BossShadowResistanceAction(ai, "lucifron"); }
```
brighton-chi: Sometimes ai is used and other times botAI is used--it should be consistent.

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,12 +1,21 @@
...
+
+using namespace MoltenCoreHelpers;
 
 bool McCheckShouldMoveFromGroupAction::Execute(Event event)
 {
-    if (bot->HasAura(20475))  // barron geddon's living bomb
+    if (bot->HasAura(SPELL_LIVING_BOMB)) // baron geddon's living bomb
     {
-        if (!botAI->HasStrategy("move from group", BotState::BOT_STATE_COMBAT))
+        if (!botAI->HasStrategy("move from group", BOT_STATE_COMBAT))
```
brighton-chi: I know this is not something you added, but I wonder why it is necessary to set a bot strategy for move from group anyway. I think this has always been an issue with the MC strategy. I'm going to post in the PR about this since it's a broader point.

src/strategy/raids/moltencore/RaidMcTriggers.cpp
```diff
@@ -1,22 +1,30 @@
...
+using namespace MoltenCoreHelpers;
 
 bool McLivingBombDebuffTrigger::IsActive()
 {
-    // if bot has barron geddon's living bomb, we need to add strat, otherwise we need to remove
+    // if bot has baron geddon's living bomb, we need to add strat, otherwise we need to remove
     // only do when fighting baron geddon (to avoid modifying strat set by player outside this fight)
     if (Unit* boss = AI_VALUE2(Unit*, "find target", "baron geddon"))
     {
         if (boss->IsInCombat())
```
brighton-chi: I don't think IsInCombat() checks are necessary for boss strategies. FindTargetValue gets targets from the bot's hostile reference list, which is going to require the boss to be in combat with the bot.

src/strategy/raids/moltencore/RaidMcStrategy.h
```diff
@@ -9,9 +9,9 @@ class RaidMcStrategy : public Strategy
 {
 public:
     RaidMcStrategy(PlayerbotAI* ai) : Strategy(ai) {}
-    virtual std::string const getName() override { return "mc"; }
-    virtual void InitTriggers(std::vector<TriggerNode*>& triggers) override;
-    // virtual void InitMultipliers(std::vector<Multiplier*> &multipliers) override;
+    std::string const getName() override { return "mc"; }
```
brighton-chi: This is just my opinion, but I think we should take the opportunity to rename the Molten Core strategy from "mc" to "moltencore" (which would require corresponding changes in PlayerbotAI.cpp and RaidStrategyContext.h). The name "mc" is just too vague imo--I think it's better for co ? or nc ? to show "moltencore" instead because it is much clearer, and such strategy name is not that long anyway.

src/strategy/raids/moltencore/RaidMcStrategy.cpp
```diff
@@ -1,13 +1,72 @@
...
 
+#include "RaidMcMultipliers.h"
 #include "Strategy.h"
 
 void RaidMcStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
+    // Lucifron
+    triggers.push_back(
+        new TriggerNode("mc lucifron shadow resistance trigger",
+                        NextAction::array(0, new NextAction("mc lucifron shadow resistance action", ACTION_RAID), nullptr)));
```
brighton-chi: I don't think we should have strategies to toggle on Paladin' Shadow Resistance Aura. The aura doesn't stack with Priests' Shadow Protection, so it's better to use the buff so that the Paladin can have a different aura. Maybe the strategy can instead have Priests cast Prayer of Shadow Resistance when combat starts if any bot doesn't have the Shadow Resistance aura.

Note I think it could actually be useful to have such a strategy because the Priest default strategy doesn't cast Shadow Resistance. You can toggle on rshadow, but it will cast only the single target version. That probably should be fixed but obviously is outside of the scope of this PR.

src/strategy/raids/moltencore/RaidMcMultipliers.cpp
```diff
@@ -0,0 +1,68 @@
...
+{
+    if (AI_VALUE2(Unit*, "find target", "garr"))
+    {
+        if (IsDpsBotWithAoeAction(bot, action))
+            return 0.0f;
+    }
+    return 1.0f;
+}
+
+float BaronGeddonInfernoMultiplier::GetValue(Action* action)
```
brighton-chi: What about disabling abilities that would take the bot back into the aoe (e.g., Charge)?

brighton-chi:
Thanks for doing this--love to see more work on boss strategies.

I wanted to bring this up (for you and approvers), but I've always thought the MC strategy is suboptimal because it toggles bot strategies. That's something I think generally to be avoided because it can persist, and I have had the issue with Baron Geddon where bots that have Living Bomb when Geddon dies will stay in place after he dies. I don't know exactly what causes it, but my suspicion is the usage of a bot strategy is the issue.

Ultimately, I don't understand why the move from group strategy needs to exist. I would have thought it would be fine to directly call MoveFromGroup (I've never used it though so not sure what it does), or use MoveAway or a custom MoveTo that tries to account for all group members at once. Then you simply have an action that executes at the applicable time in combat and has no effect outside of the active encounter.

src/strategy/raids/moltencore/RaidMcActionContext.h
HennyWilly: Looks like most dungeon/raid strategies don't use the "action" and "trigger" suffix. Removed.

src/strategy/raids/moltencore/RaidMcActionContext.h
HennyWilly: You mean the "ai" parameter name, right? Renamed to "botAI".

src/strategy/raids/moltencore/RaidMcStrategy.h
HennyWilly: Makes sense. Renamed "mc" to "moltencore" in this file, PlayerbotAI.cpp and RaidStrategyContext.h.

src/strategy/raids/moltencore/RaidMcMultipliers.cpp
HennyWilly: `if (dynamic_cast<CastReachTargetSpellAction*>(action))` handles this. `CastChargeAction`, `CastInterceptAction`, `CastFeralChargeCatAction` and `CastFeralChargeBearAction` are sub-classes of this class.

src/strategy/raids/moltencore/RaidMcStrategy.cpp
HennyWilly: I see your point. I always wondered why the priest's shadow resistance isn't always applied together with the stamina and spirit buffs. Or alternatively/additionally, adding the priest shadow buff to the shadow resistance action is also a good idea.

src/strategy/raids/moltencore/RaidMcTriggers.cpp
HennyWilly: Check removed.

HennyWilly:
@brighton-chi Thank you very much for your feedback. And @Celandriel too of course.

I understand your point about adding strategies that last even after the encounter.
It looks like the "move from group" strategy triggers a `MoveFromGroupAction` which in turn calls `MoveFromGroup`.
I did a little testing and calling `MoveFromGroup` directly from a custom `MovementAction` actually works really well.
The only additional modification in that regard was to add the check for Living Bomb into the multiplier. This prevents bots from running back into the group while they have the bomb.
Feel free to test 😄 

When the review is complete, I will rebase my new commits into one commit if required.

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: [See my comment below.](https://github.com/mod-playerbots/mod-playerbots/pull/1852#issuecomment-3549728491)

isitsimon:
Hello all,

I tested this PR in MC this evening. All 39 bots were wearing Autogeared Green armour, and IP 50% reduction in, damage, HP and Healing. I had a lot of fun clearing this and I would like to thank the author of this PR.
My evaluation is:

**Lucifron** - Confirmed aura

Suggestions - None

**Magmadar** - Confirmed aura

Suggestions - Fear wards for tanks would be useful

**Gehennas** - Confirmed Aura (on use of Gehennas Curse)

Suggestions - None

**Garr**   - Confirmed Aura

Suggestions - maybe pull boss tank away from adds prior to death, a tank getting blown away from healers is the only way I see them dying.

**Geddon**   - Confirmed Aura
	              - Confirmed hellfire + living bomb (if geddon dies with living bomb active, bot will run back into raid)
	             
Suggestions  - annoyingly tanks or melee would carry living bomb progressively towards Shaz (and then pull him) not game breaking, happened on a few pulls probably due to slow kill times

**Shaz**     - movement works

Suggestions - None

**Sulfuron** - Confirmed Aura 
	 
Suggestions - None

**Golemagg**  - Confirmed Aura
	          
Suggestions - dps should avoid targeting dogs 
	             - needs to be tanked away from the dogs

**Majordomo** - Confirmed Aura
	
Suggestions - bots should swap target on damage shield but not that important

**Rag** - Confirmed Aura

 Suggestions  - a slight spread to avoid the ranged knockback

 Suggestions are just my opinions :) (except Golemagg, that guy wiped us 3 times and I think splitting the dogs is required)

HennyWilly:
@isitsimon Thanks for the detailed testing and suggestions.

I fixed the issue where bots would run back into the raid if Geddon dies with Living Bomb still active. (see 40208af)

Regarding the Golemagg behaviour: implementing proper dog-splitting and related positioning logic requires a more complex strategy. At the moment I don’t feel confident enough with the codebase to implement this cleanly. If someone with more experience wants to take this part, I would appreciate it. Otherwise I’d prefer to get this PR merged as-is and follow up with a separate improvement once I’m more familiar with the internals.

brighton-chi:
You could try just having the bots mark Golemagg with a skull. I am just guessing here, but the issue might be that if you disable AoE actions, a bunch of bots end up single targeting dogs instead.


I think marking outside of contexts where targets need to be split should generally be a limited strategy because it takes away player agency, but for a situation like this where there is no sane reason to target the dogs, it's probably a good idea.

isitsimon:
> You could try just having the bots mark Golemagg with a skull. I am just guessing here, but the issue might be that if you disable AoE actions, a bunch of bots end up single targeting dogs instead.
> 
> I think marking outside of contexts where targets need to be split should generally be a limited strategy because it takes away player agency, but for a situation like this where there is no sane reason to target the dogs, it's probably a good idea.

While I agree, the dogs need to be tanked. There is also a concern that 2 buffed dogs with 100% increased damage + the boss all agro to one tank. 

That tank will not last long 🤣. 

Ideally an off tank would take the dogs away and MT would sit on boss.

brighton-chi:
> > You could try just having the bots mark Golemagg with a skull. I am just guessing here, but the issue might be that if you disable AoE actions, a bunch of bots end up single targeting dogs instead.
> 
> > 
> 
> > I think marking outside of contexts where targets need to be split should generally be a limited strategy because it takes away player agency, but for a situation like this where there is no sane reason to target the dogs, it's probably a good idea.
> 
> 
> 
> While I agree, the dogs need to be tanked. There is also a concern that 2 buffed dogs with 100% increased damage + the boss all agro to one tank. 
> 
> 
> 
> That tank will not last long 🤣. 
> 
> 
> 
> Ideally an off tank would take the dogs away and MT would sit on boss. 
> 
> 

I see. The simplest solution without needing to split targets might be to have a different mark on Golemagg then and then have DPS prioritize that mark but don't make changes with respect to tank RTI.

Then the player can still control the tanks separately, but DPS will attack only Golemagg.

Celandriel:
@HennyWilly Checking if you are still working on this, or if I should review.

HennyWilly:
@Celandriel I started working on an extended Golemagg strategy yesterday. The initial results look promising, but some parts still need to be implemented and refined. I will push the changes once the behaviour is stable enough for review.

HennyWilly:
Here is the improved Golemagg strategy:

This strategy works best with three tanks.
The main tank attacks the boss and keeps him near his default location. Assist tank 1 and 2 each pick up one Core Rager and pull them away from the boss in order to mitigate [Golemagg's Trust](https://www.wowhead.com/classic/spell=20553/golemaggs-trust). The assist tanks will position the Core Ragers near the stalagmites on the left.
Golemagg is marked with a skull, so DPS will automatically focus him. DPS AoE abilities remain disabled for this encounter.

I might have overengineered the strategy, trying to make it as adaptive as possible:
- If there is only one tank, that tank handles both the boss and the adds.
- With two tanks, the main tank handles the boss and the assist tank handles the two adds.
- If there are more than three tanks, the additional assist tanks will attack the boss.
- If a tank dies at any point, a remaining tank will step in. If there is no additional tank, the remaining tank bots will change their strategy accordingly (as mentioned above).
- If Golemagg is too close to the Core Rager tanking position and affects them with Golemagg's Trust, the assist tanks will try to get the adds away from the boss. This can happen if the main tank dies or a non-bot tank positions the boss at a different location.

This is my first complex strategy, so feedback is welcome.

isitsimon:
> Here is the improved Golemagg strategy:
> 
> This strategy works best with three tanks. The main tank attacks the boss and keeps him near his default location. Assist tank 1 and 2 each pick up one Core Rager and pull them away from the boss in order to mitigate [Golemagg's Trust](https://www.wowhead.com/classic/spell=20553/golemaggs-trust). The assist tanks will position the Core Ragers near the stalagmites on the left. Golemagg is marked with a skull, so DPS will automatically focus him. DPS AoE abilities remain disabled for this encounter.
> 
> I might have overengineered the strategy, trying to make it as adaptive as possible:
> 
> * If there is only one tank, that tank handles both the boss and the adds.
> * With two tanks, the main tank handles the boss and the assist tank handles the two adds.
> * If there are more than three tanks, the additional assist tanks will attack the boss.
> * If a tank dies at any point, a remaining tank will step in. If there is no additional tank, the remaining tank bots will change their strategy accordingly (as mentioned above).
> * If Golemagg is too close to the Core Rager tanking position and affects them with Golemagg's Trust, the assist tanks will try to get the adds away from the boss. This can happen if the main tank dies or a non-bot tank positions the boss at a different location.
> 
> This is my first complex strategy, so feedback is welcome.

That sounds perfect mate, I will give it a try this weekend! 👍

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,220 @@
...
+            float moveX = bot->GetPositionX() + (dX / dist) * maxDistance;
+            float moveY = bot->GetPositionY() + (dY / dist) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, bot->GetPositionZ(), false, false,
+                          false, false, MovementPriority::MOVEMENT_COMBAT, true,
+                          true);
+        }
+
+        float orientation = atan2(target->GetPositionY() - bot->GetPositionY(),
+                                  target->GetPositionX() - bot->GetPositionX());
+        bot->SetFacingTo(orientation);
```
noisiver: Is it really necessary to force their orientation? Wouldn't it make more sense to let them face the boss on their own since it doesn't actually affect the boss in any way?

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,220 @@
...
+    }
+    else if (!bot->IsWithinMeleeRange(target))
```
noisiver: Don't they do this on their own? They always run up the target they're attacking, is there a reason why you're manually handling this on top of that?

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: To be fair, this code was inspired by [HighKingMaulgarMainTankAttackMaulgarAction](https://github.com/mod-playerbots/mod-playerbots/blob/38e2d8584b8772894439e07314facd98345ff823/src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp#L12), [HighKingMaulgarSecondAssistTankAttackBlindeyeAction](https://github.com/mod-playerbots/mod-playerbots/blob/38e2d8584b8772894439e07314facd98345ff823/src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp#L95) and [GruulTheDragonkillerMainTankPositionBossAction](https://github.com/mod-playerbots/mod-playerbots/blob/38e2d8584b8772894439e07314facd98345ff823/src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp#L531).
I thought this was the correct method for positioning enemy NPCs.
Can I assume that this is generally not the case?

Interestingly, setting the orientation doesn't happen for [HighKingMaulgarFirstAssistTankAttackOlmAction](https://github.com/mod-playerbots/mod-playerbots/blob/38e2d8584b8772894439e07314facd98345ff823/src/strategy/raids/gruulslair/RaidGruulsLairActions.cpp#L55). 🤔

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: Same as above

src/strategy/raids/moltencore/RaidMcActions.cpp
brighton-chi: Heh, there's quite a bit that I don't like that I did for Gruul. Anyway, I think setting a fixed orientation is relevant only if there is a reason to have a particular orientation. With Maulgar, I was finding that without doing so, his facing could be inconsistent, and I needed him to face the wall because of his cleave and knockback. It doesn't really matter for any of the other ogres actually.

Olm is weird because he doesn't chase properly, so that tank position has the tank run past where you actually want Olm and then go back once Olm is at the spot you actually want him at. I did that one later so I might have just decided orientation didn't matter at that point and didn't change it.

noisiver: Yes but this is setting the **bot** orientation, not the boss orientation. That's why it confuses me. The boss will always face the player (or bot) so does the bots orientation ever actually matter?

src/strategy/raids/moltencore/RaidMcActions.cpp
brighton-chi: I'm not really settled on what I think is the best way to move a mob to a tank location, despite having tried various things. I think depending on how you set things up, it can be helpful to keep the tank in melee position of the boss as it moves the boss (makes it harder for tank to lose aggro along the way), but the logic has to be set up differently for that, and I don't really think it's overall necessary.

However, even though it seems unnecessary, IIRC, when I removed that block for the HKM tank methods, tanks did not respond as quickly if they lost aggro while going to their tank spot. It may have something to do with the higher priority of the raid action, but I think it's probably worth testing how things look ingame without that block.

src/strategy/raids/moltencore/RaidMcActions.cpp
brighton-chi: In my experience it did, but it's not like I tested this specific logic many times so it could just be up to chance. It is possible that the orientation could influence how the bot chooses to engage the boss, but I can't say anything definitive.

But I will say if there isn't a compelling reason for the boss to face a particular definition, then there's no reason to set orientation for sure.

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,220 @@
...
-    if (bot->HasAura(20475))  // barron geddon's living bomb
+    return MoveFromGroup(LIVING_BOMB_DISTANCE);
+}
+
+bool McMoveFromBaronGeddonAction::Execute(Event event)
+{
+    if (Unit* boss = AI_VALUE2(Unit*, "find target", "baron geddon"))
     {
-        if (!botAI->HasStrategy("move from group", BotState::BOT_STATE_COMBAT))
+        float distToTravel = INFERNO_DISTANCE - bot->GetDistance(boss);
```
brighton-chi: Do you know if the AoE range is measured in three dimensions? I prefer GetDistance2d (if accounting for the target's size) or GetExactDist2d (if measuring from the target's middle). It's not clear to me that AC handles AoE radii correctly as it seems to measure at least some spells from the edge of the caster's hitbox (but I can't get a clear answer on what is right vs. wrong). I don't know about the 2d vs. 3d matter, though.

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,220 @@
...
+    {
+        float distanceToTankPosition = bot->GetExactDist2d(tankPosition.GetPositionX(), tankPosition.GetPositionY());
+        if (distanceToTankPosition > maxDistance)
+        {
+            float dX = tankPosition.GetPositionX() - bot->GetPositionX();
+            float dY = tankPosition.GetPositionY() - bot->GetPositionY();
+            float dist = sqrt(dX * dX + dY * dY);
+            float moveX = bot->GetPositionX() + (dX / dist) * maxDistance;
+            float moveY = bot->GetPositionY() + (dY / dist) * maxDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, bot->GetPositionZ(), false, false,
```
brighton-chi: You can use a constant for the Molten Core map ID and replace all GetMapId() with the constant to avoid the lookup.

src/strategy/raids/moltencore/RaidMcActions.cpp
brighton-chi: I've done some more testing in SSC with that block removed, and I think it can be taken out (unless testing for any particular situation shows otherwise).

HennyWilly:
Here is an updated version. The comments above about `SetFacingTo` and `IsWithinMeleeRange` have been addressed.

I also fixed an issue that prevented assist tank 2 from picking up the free Core Rager if assist tank 0 or 1 died. The problem was that `IsAssistTankOfIndex` doesn't take dead players/bots into account.
For example, if assist tank 1 dies, assist tank 2 would not be "promoted" and `IsAssistTankOfIndex(bot, 1)` would still return false.
To address this, I added an optional parameter `bool ignoreDeadPlayers = false` to `IsAssistTankOfIndex`.
For the above example, with this parameter set to true, `IsAssistTankOfIndex(bot, 1, true)` would return true for assist tank 2 if either assist tank 0 or 1 is dead.
This should not break existing strategies, but I understand that this is a change outside of the Molten Core strategy. I will think of something else if this change is not desired.

Lastly, the bots didn't seem to use their taunt abilities, so I added `botAI->DoSpecificAction("taunt spell", ...)`. With this addition, the bots will use their class specific taunt abilities if their respective targets don't attack them back.
Death Knights learn their taunt ability [Dark Command](https://www.wowhead.com/wotlk/spell=56222) at level 65, so DK tanks below that level may require a little more time to (re-)gain aggro.
If there is a better way to do this, let me know.

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: I tested the encounter quite a few times without setting the orientations and the tanks still work fine. Removed.

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: Same as above. There seems to be no downsides without the `IsWithinMeleeRange` check. Removed.

brighton-chi:
Yeah I believe @nl-saw is going to modify the various assistant roles to make them exclude dead bots but for now it makes sense to manually.

I'll try to take a look at the overall strategy.

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,218 @@
...
-    if (bot->HasAura(20475))  // barron geddon's living bomb
+    return MoveFromGroup(LIVING_BOMB_DISTANCE);
+}
+
+bool McMoveFromBaronGeddonAction::Execute(Event event)
+{
+    if (Unit* boss = AI_VALUE2(Unit*, "find target", "baron geddon"))
     {
-        if (!botAI->HasStrategy("move from group", BotState::BOT_STATE_COMBAT))
+        float distToTravel = INFERNO_DISTANCE - bot->GetDistance(boss);
```
brighton-chi: Unless the radius for this attack should be measured in 3d, I would use a 2d distance measurement (either GetDistance2d if taking the boss's hitbox into account or GetExactDist2d if not).

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,218 @@
...
+    {
+        float distanceToTankPosition = bot->GetExactDist2d(tankPosition.GetPositionX(), tankPosition.GetPositionY());
+        if (distanceToTankPosition > maxDistance)
+        {
+            float dX = tankPosition.GetPositionX() - bot->GetPositionX();
+            float dY = tankPosition.GetPositionY() - bot->GetPositionY();
+            float dist = sqrt(dX * dX + dY * dY);
+            float moveX = bot->GetPositionX() + (dX / dist) * stepDistance;
+            float moveY = bot->GetPositionY() + (dY / dist) * stepDistance;
+            return MoveTo(bot->GetMapId(), moveX, moveY, bot->GetPositionZ(), false, false,
```
brighton-chi, line 88: You can use a constant for the Molten Core map ID and replace all GetMapId() with the constant to avoid the lookup.

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,218 @@
...
+            !PlayerbotAI::IsAssistTankOfIndex(otherCoreRagerPlayerVictim, 0, true) &&
+            !PlayerbotAI::IsAssistTankOfIndex(otherCoreRagerPlayerVictim, 1, true))
         {
-            // float angle = bot->GetAngle(boss) + M_PI;
-            // return Move(angle, distToTravel);
-            return MoveAway(boss, distToTravel);
+            // Assume we are the only assist tank or the other assist tank is dead => pick up other Core Rager!
+            if (bot->GetVictim() != otherCoreRager)
+                return Attack(otherCoreRager);
+            return botAI->DoSpecificAction("taunt spell", event, true);
```
brighton-chi, line 188: I'm not sure I understand this--if the bot's victim is already the othercorerager, it will try to taunt? Or does DoSpecificAction "taunt spell" otherwise have conditions that would make the bot not try to taunt in that situation?

src/strategy/raids/moltencore/RaidMcActions.h
```diff
@@ -21,4 +22,46 @@ class McMoveFromBaronGeddonAction : public MovementAction
...
+    bool Execute(Event event) override;
+};
+
+class McGolemaggTankAction : public AttackAction
+{
+public:
+    McGolemaggTankAction(PlayerbotAI* botAI, std::string const name)
+        : AttackAction(botAI, name) {}
+protected:
+    bool MoveUnitToPosition(Unit* target, const Position& tankPosition, float maxDistance, float stepDistance = 3.f);
```
brighton-chi: Typo for stepDistance?

src/strategy/raids/moltencore/RaidMcMultipliers.cpp
```diff
@@ -0,0 +1,115 @@
...
+                !dynamic_cast<McMoveFromBaronGeddonAction*>(action))
+        return false;
+
+    if (dynamic_cast<CastReachTargetSpellAction*>(action))
+        return false;
+
+    return true;
+}
+
+float BaronGeddonAbilityMultiplier::GetValue(Action* action)
```
brighton-chi, line 59: For convenience, I would have a return 1.0f for WipeAction when the multiplier is restricting bots to a specific action (since otherwise they will not respond to the wipe command).

src/strategy/raids/moltencore/RaidMcStrategy.cpp
```diff
@@ -1,13 +1,81 @@
...
 
 void RaidMcStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
+    // Lucifron
+    triggers.push_back(
+        new TriggerNode("mc lucifron shadow resistance",
+                        NextAction::array(0, new NextAction("mc lucifron shadow resistance", ACTION_RAID), nullptr)));
+
+    // Magmadar
+    // TODO: Fear ward / tremor totem, or general anti-fear strat development. Same as King Dred (Drak'Tharon) and faction commander (Nexus).
```
brighton-chi, line 14: Are you planning this still? FWIW, since tremor totem is a class strategy now, I don't think you really need a raid strategy to use it.

src/PlayerbotAI.cpp
```diff
@@ -2266,6 +2265,11 @@ bool PlayerbotAI::IsAssistTankOfIndex(Player* player, int index)
             continue;
         }
 
+        if (ignoreDeadPlayers && !member->IsAlive())
```
brighton-chi, line 2268: I don't personally care, but the Codestyle should be not to use braces here for single-line statements after if

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: I'm not sure about this one. I don't like the redundancy between that constant and [ApplyInstanceStrategies in PlayerbotAI](https://github.com/HennyWilly/mod-playerbots/blob/def428f6461009eac6c1597c6e81d743c0062317/src/PlayerbotAI.cpp#L1463).
Additionally, `GetMapId` just returns the internal member `m_mapId`. So not much of a performance hit.
If it's okay with you, I would like to keep this one as is.

src/strategy/raids/moltencore/RaidMcMultipliers.cpp
HennyWilly: `WipeAction` is a sub-class of `Action`.
Correct me if I'm wrong but I don't see anything inside the multipliers that would prevent a `WipeAction` from being executed.

src/PlayerbotAI.cpp
HennyWilly: I agree. However, the same style is used in this method (and most other `isX` methods).
So I thought I use the same style here as well.
If you insist, I can remove the curly braces here. But for consistency, every other occurrence in this file should be changed too. 😅 
I would prefer to leave it as it is for now.

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: Changed to `GetDistance2d`. Tested and still works. Same for Shazzrah.

src/strategy/raids/moltencore/RaidMcActions.h
HennyWilly: You mean `3.f`? `3.f` is equivalent to `3.0f`. That’s why I omitted the 0.
In retrospect, however, the "longer" variant is more commonly used in mod-playerbots.
Added 0 for consistency.

src/strategy/raids/moltencore/RaidMcStrategy.cpp
HennyWilly: There were TODOs for the Faction Commander (Nexus) and King Dred (Drak'Tharon), so I just added a TODO for Magmadar as well.
I wasn't planing to implement an anti-fear strategy myself tbh.
This would still be a nice feature. If there is no shaman in the group, there should be a fallback to a priest's Fear Ward.

src/strategy/raids/moltencore/RaidMcActions.cpp
HennyWilly: This was my intention, yes.
All taunt spells are CastSpellActions or CastMeleeSpellActions. And the default `GetTargetName` is "current target".
There are some checks in `isUseful` and `isPossible` but they seem to check if the bot is mounted or in a vehicle.
The warrior taunt action has a custom isUseful implementation `GetTarget() && GetTarget()->GetTarget() != bot->GetGUID()`.
The `execute` method returns the result of `botAI->CastSpell(spell, GetTarget());`.
If I interpret all of the above correctly, there should be no problems that prevent the tank bot from taunting its current target.

src/strategy/raids/moltencore/RaidMcMultipliers.cpp
brighton-chi: My bad, I thought the multiplier was referencing a method, but I see now it's just a helper.

src/PlayerbotAI.cpp
brighton-chi: Heh, for sure there is so much inconsistency with AC's standards throughout the module and throughout the Core's code itself. If the powers that be don't raise a concern this time, it's all the same to me.

src/strategy/raids/moltencore/RaidMcMultipliers.cpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (dynamic_cast<DpsAoeAction*>(action) || dynamic_cast<CastConsecrationAction*>(action) ||
+            dynamic_cast<CastStarfallAction*>(action) || dynamic_cast<CastWhirlwindAction*>(action) ||
+            dynamic_cast<CastMagmaTotemAction*>(action) || dynamic_cast<CastExplosiveTrapAction*>(action) ||
+            dynamic_cast<CastDeathAndDecayAction*>(action))
+            return true;
+
+        if (auto castSpellAction = dynamic_cast<CastSpellAction*>(action))
+            if (castSpellAction->getThreatType() == Action::ActionThreatType::Aoe)
+                return true;
```
kadeshar, line 30: multiline if without {}

HennyWilly: Fixed

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,218 @@
...
+    if (bot->GetVictim() != target)
+        return Attack(target);
+    if (target->GetVictim() == bot)
+    {
+        float distanceToTankPosition = bot->GetExactDist2d(tankPosition.GetPositionX(), tankPosition.GetPositionY());
+        if (distanceToTankPosition > maxDistance)
+        {
+            float dX = tankPosition.GetPositionX() - bot->GetPositionX();
+            float dY = tankPosition.GetPositionY() - bot->GetPositionY();
+            float dist = sqrt(dX * dX + dY * dY);
```
brighton-chi, line 85: Sorry, I thought I made a comment on this one the other day but I obviously didn't. I was going to ask if this calculation is needed, as I think it's actually the same as distanceToTankPosition.

(I wouldn't be surprised if the Gruul strategies did something like this)

HennyWilly: Yes, it looks like this code does exactly the same as the above `GetExactDist2d`.
And yes, the Gruul strategies do the same thing. 

Would you mind keeping this as is for now? 
I already have two approvals, and pushing further changes might reset them.  
The only remaining blocker is the outstanding "requested change", which is already resolved on my side.

noisiver (approved):

isitsimon:
This is awesome HennyWilly. 👌🏻👏🏻

HennyWilly:
Is there anything else I need to take care of?
If there are no further issues, feel free to press the merge button at any time. 😄

isitsimon:
> Is there anything else I need to take care of? If there are no further issues, feel free to press the merge button at any time. 😄

BWL next? 👀😁

src/PlayerbotAI.cpp
noisiver: We aim to match the code style of azerothcore and the enforcement of it is still in progress so it should be single line if = no curly brackets. Because of that any code you add should follow that, no matter what the rest of the code looks like.

src/PlayerbotAI.cpp
```diff
@@ -2285,6 +2289,11 @@ bool PlayerbotAI::IsAssistTankOfIndex(Player* player, int index)
...
+        {
```
noisiver: Don't use curly brackets for single line if

src/strategy/raids/moltencore/RaidMcActions.cpp
```diff
@@ -1,43 +1,218 @@
...
-            return MoveAway(boss, distToTravel);
+            // Assume we are the only assist tank or the other assist tank is dead => pick up other Core Rager!
+            if (bot->GetVictim() != otherCoreRager)
+                return Attack(otherCoreRager);
+            return botAI->DoSpecificAction("taunt spell", event, true);
+        }
+    }
+
+    if (bot->GetVictim() != myCoreRager)
+    {
```
noisiver: Don't use curly brackets for single line if

noisiver:
I'm sorry for blocking this but it really should match what we aim for.

The use of curly brackets in existing files for new code should be avoided.

As I mentioned we do aim for the same standard as azerothcore and the code should at some point be changed to not use curly brackets for single line ifs (and more). Adding new code should do that already so there's less to change further down the road.

HennyWilly:
@noisiver No worries, I understand the concern.  
I’ve updated the code accordingly and removed the curly brackets in the places you pointed out.  
Let me know if anything else needs to be aligned.

PR #1853, opened 2025-11-18
Add locale-aware guild and arena team names for random bots

Wishmaster117:
### Description:

**This PR introduces a simple multi-language system for random bot guild names and arena team names in mod-playerbots, so each server can use names that match its own language.**

### What this changes

**Database**

- playerbots_guild_names

- Added locale-specific columns:

- name_koKR
- name_frFR
- name_deDE
- name_zhCN
- name_zhTW
- name_esES
- name_esMX
- name_ruRU

name remains the base enUS name and is still required and unique.

- playerbots_arena_team_names

- Added the same locale-specific columns:

- name_koKR
- name_frFR
- name_deDE
- name_zhCN
- name_zhTW
- name_esES
- name_esMX
- name_ruRU

Again, name stays as the base enUS value.

If a localized column is NULL or empty for a given row, the code automatically falls back to the base name (enUS), so it’s safe to migrate gradually.

### Configuration
Set your locale in worldserver conf file:

```
#
#    DBC.Locale
#        Description: DBC language settings.
#        Default:     255 - (Auto Detect)
#                     0   - (English)
#                     1   - (Korean)
#                     2   - (French)
#                     3   - (German)
#                     4   - (Chinese)
#                     5   - (Taiwanese)
#                     6   - (Spanish)
#                     7   - (Spanish Mexico)
#                     8   - (Russian)

DBC.Locale = X
```

**Avoid to use 255, if you use it you will have english names.**

This value is read by PlayerbotAIConfig and used in RandomPlayerbotFactory when generating names for:

CreateRandomGuildName()
CreateRandomArenaTeamName()

### Implementation details

Both guild and arena team name generation now select all locale columns from their respective tables.
The factory chooses the correct column based on AiPlayerbot.RandomBotGuildArenaNamesLocale.
If the chosen localized column is empty, it falls back to the base name.
No core changes are required; everything is handled inside mod-playerbots and the existing AzerothCore guild/arena systems are used as-is.

### How to use

Copy files in another place and open them:
2025_11_17_01_french_guilds_name.sql
2025_11_17_03_french_arena_team_names.sql

Rename them as:
2025_11_xx_xx_yourlanguage_guilds_name.sql
2025_11_xx_xx_yourlanguage_arena_team_names.sql

Translate in your language (or rename as you want) and change the name of the column like this:

For exemple if you want to translate de DE:
For guild:
replace name_frFR by name_deDE
like:
UPDATE `playerbots_guild_names` set `name_deDE` = 'Reich der Abgründe' where `name_id` = 1;

For Arena Teams:
Same as guild like:
UPDATE `playerbots_arena_team_names` SET `name_deDE` = 'Neuanfang' WHERE `name_id` = 1;

Put your two files in \mod-playerbots\data\sql\characters\updates.

Share yous translations with community if you want.

**Set your server locale for bot names in playerbots.conf **

# French server
In wolrdserver conf file:
DBC.Locale = 2 # frFR

### Of course, the already existing guilds and arena teams will not be translated, as they are already registered in the AzerothCore database.
 
**If you want to regenerate all random bot guilds in the new language, you can temporarily enable:**

```
AiPlayerbot.DeleteRandomBotGuilds = 1
AiPlayerbot.DeleteRandomBotArenaTeams = 1
```

**then restart, and set it back to 0 afterwards.**

### Why

This allows each server to run mod-playerbots in its own language without changing the core or hardcoding names per locale. Guild names and arena team names can now be localized in the database in the same spirit as other localized data (creatures, gameobjects, etc.), while keeping a simple fallback to enUS.


privatecore:
> AiPlayerbot.RandomBotGuildArenaNamesLocale = -1

Why not use the default server locale for this?
https://github.com/azerothcore/azerothcore-wotlk/blob/539581aa7a1df6683c9f532bd5afa0b84111d256/src/server/game/World/World.h#L161

Wishmaster117:
> > AiPlayerbot.RandomBotGuildArenaNamesLocale = -1
> 
> Why not use the default server locale for this? https://github.com/azerothcore/azerothcore-wotlk/blob/539581aa7a1df6683c9f532bd5afa0b84111d256/src/server/game/World/World.h#L161

I actually tried going this way and ran into a practical problem with playerbots.
mod-playerbots currently only works reliably with enUS DBCs. Even if I set:
DBC.Locale = 2   # frFR
the core still ends up using the enUS DBCs, and sWorld->GetDefaultDbcLocale() resolves to LOCALE_enUS at runtime.

Because of that, “using the default server locale” in this context doesn’t give us a meaningful “server language” for bot names etc... in a supported playerbots setup it will effectively always be enUS. That means guild / arena would always fall back to the name (enUS) column, and the localized columns (name_frFR, name_esES, etc.) would never be used.

The goal of this PR is to allow:
running playerbots with enUS DBCs (required today),
and still using localized names from the database (frFR, deDE, esES, …).
That’s why I kept an explicit setting on the playerbots side: it decouples bot-related names from the DBC locale and lets servers pick the language they want for those names, even when the core is forced to use enUS DBCs.

The only way i see to avoir add another config in playerbots config file is to "read" :
int32 dbcLocale = sConfigMgr->GetOption<int32>("DBC.Locale", 0);
And use it to choose the table field but we have to deal with 255 value if the user sets it to that. Fallback tu enUS?

Wishmaster117:
Ok removed variable from playerbots conf file and decided to use : int32 dbcLocale = sConfigMgr->GetOption("DBC.Locale", 0); instead.

privatecore:
> the core still ends up using the enUS DBCs, and sWorld->GetDefaultDbcLocale() resolves to LOCALE_enUS at runtime.

You mean when using Docker? It's because the AC team creates the image client data with the enUS locale.

The default locale is set based on the DBC files the server is using. If you use enUS data, the server's default locale will also be enUS (even if you set DBC.Locale to smth different). So, if you want a fully localized server setup, you need DBC files in the desired language.

https://github.com/azerothcore/azerothcore-wotlk/blob/539581aa7a1df6683c9f532bd5afa0b84111d256/src/server/game/World/World.cpp#L1045

Wishmaster117:
> The default locale is set based on the DBC files the server is using. If you use enUS data, the server's default locale will also be enUS (even if you set DBC.Locale to smth different). So, if you want a fully localized server setup, you need DBC files in the desired language.

Yes but try to put for exemple french DBC in your server and try to run playerbots, doesn't work to make this work you need DBC's with all alanguages fields filled and i don't know why blizzard doesn't do taht when you extract dbs'c from french client for exemple you will get only french textes in the dbs.

My actual approach works and it more secure.

privatecore:
> Yes but try to put for exemple french DBC in your server and try to run playerbots, doesn't work to make this work you need DBC's with all alanguages fields filled

Do you mean playerbots only work with the enUS locale and cannot function properly with other DBC files?

Wishmaster117:
> > Yes but try to put for exemple french DBC in your server and try to run playerbots, doesn't work to make this work you need DBC's with all alanguages fields filled
> 
> Do you mean playerbots only work with the enUS locale and cannot function properly with other DBC files?

Yes, you got it! :D
I want to try one day to merge all languages to the enUS dbc to test what's happend

Wishmaster117:
Opened a new one

PR #1760, opened 2025-10-22
feat: Improve bot mount behavior to faster close distance between bot and master

icemansparks:
# Fix bot mount behavior when master dismounts

## Summary

Improves bot mount/dismount logic to ensure better coordination with the master player. The bot now remains mounted when closing distance to a recently dismounted master and mounts up to assist the master in combat.

The changes have been tested using the testcases described in the second half of the PR description, which provide some directions on how this PR can be tested and verified.

Closes: #1660

## Changes

- Add masterInCombat variable to track master combat state
- Modify target-based logic to consider master combat state
- Change mount priority when bot is not in combat to favor master-following
- Remove combatReach from distance calculations (duplicate)

## Implementation

Added two methods:
- `StayMountedToCloseDistance()` - prevents premature dismounting when the master dismounts
- `ShouldMountToCloseDistance()` - determines when to mount for master assistance, even if master is not mounted at this time

Modified Execute() method:
- Target-based shouldMount/shouldDismount considers the master's combat state
- Combat assistance logic separated from general following
- Mount when master in combat, but the bot is not

Distance logic:
- Combat: dismount at CalculateDismountDistance() (18-25 yards)
- Non-combat: dismount at 10 yards
- Mount threshold: 21+ yards

## Result
- Bots mount to assist masters in combat
- Bots stay mounted longer during travel
- Different dismount distances based on context
- Existing mount selection logic unchanged

# Mount Behavior Testing

## Prerequisites

1. Add a test bot: `.playerbots bot add <charactername>` or `.playerbots bot addclass <class>`
2. Test in a level-appropriate outdoor area (mobs should not die with one hit, as this makes testing combat assistance impossible)
3. Both master and bot must have mounts available

## Test 1: Combat Assistance Mounting

**Objective**: Verify bot mounts to assist the master in combat

**Detailed Steps**:
1. Position both master and bot dismounted
2. Command bot to stay: `/w <botname> stay`
3. Move master 35+ yards away from bot
4. Target a nearby mob and attack the mob
5. Command bot to assist: `/w <botname> follow`

**Expected Results**:
- Bot immediately mounts up (within 1-2 seconds cast time)
- Bot rides toward master while mounted
- Bot dismounts at ~18-25 yards from master/mob
- Bot engages in combat to assist

**Failure Indicators**:
- Bot runs dismounted (old behavior)
- Bot never mounts despite distance
- Bot mounts but doesn't dismount at proper assist range

## Test 2: Non-Combat Dismount Distance

[...]

PR #1855, opened 2025-11-18
Revert "feat: Improve bot mount behavior to faster close distance between bot and master"

nl-saw:
Reverts mod-playerbots/mod-playerbots#1760

This, as it is causing issues in BG, where bots just don't dismount and just stand there instead.

image


noisiver (approved):

icemansparks:
It appears that they can become trapped in a loop of being in combat while waiting for the target to get close enough for them to dismount, which never happens, as the target is also stuck in a similar loop.

A possible solution could be to add the `inBattleground` check to the if block here: `if (currentTarget && (bot->IsInCombat() || masterInCombat || inBattleground))`

We need to test this more thoroughly, regarding BG behaviour - I'll open a PR again when I have tested it in the upcoming days.

PR #1610, opened 2025-09-02
Smarter Loot Roll Logic for Playerbots

Wishmaster117:
### Summary

This PR improves the loot roll decision-making for Playerbots.

This update introduces spec-aware, stat-aware, and etiquette-friendly rules to better mimic real player behavior.

**Caution, you have to set this config to 2 : AiPlayerbot.LootRollLevel = 2**

### Key Changes
---
**Smart NEED by spec & stats**

- New SmartNeedBySpec system checks whether an item’s stats align with the bot’s main specialization.
- Physical classes (warriors, rogues, hunters, DKs, ferals, ret, enh, prot) never NEED caster gear (SP/INT/SPI/MP5).
- Casters/healers never NEED pure melee DPS gear (STR/AP/ARP/EXP without SP/INT).
- Edge cases (jewelry, trinkets, off-armor) are filtered accordingly.
---
**Cross-armor upgrade logic**

- NEED on off-armor pieces (cloth for paladin, leather for hunter, etc.) only allowed if the new score is significantly higher (CrossArmorExtraMargin).
---
**BoP / BoE / BoU etiquette**

- BoP: allow NEED if the item fills an empty slot and matches spec.
- BoE / BoU: default to GREED, unless explicitly enabled in config.
---
**Token handling (T7–T10)**

- Bots NEED tokens only if they represent a likely upgrade (ilvl + configurable margin).
- Otherwise, they GREED. Unknown-slot tokens default to GREED.
---
**Unique-equip items**

- Bots never NEED duplicates if already equipped or in bags.
---
** Locked chests**
- Rogue lockpicking -> if they have the hability rogues always need in locked chest.
 ---
**Weapon handling**
- So that NEED rolls are more consistent.
---
**System for tank-only loot**
- Prevent DPS from rolling NEED on items with defense, parry, etc.
---
**Disenchant button (Need Before Greed)**

- If enabled, bots choose disenchant on loot if the bot is enchanting
---
**Professions recipes**
- If bot know the profession, have required level and don't know or have the recipe it don NEED otherwise GREED (You can enable or disable rank check and NEED in recipes in the config file)
---
**Special item verification via stat-pattern hook**
This introduces a small, modular hook in the loot roll logic that detects special stat patterns and adjusts NEED/GREED accordingly without disrupting existing rules.

- Pattern registry (extensible): a lightweight matcher/decider path that runs at the start of IsPrimaryForSpec.
- First implemented pattern: INT + Attack Power (incl. Ranged AP).
[...]

PR #1857, opened 2025-11-18
fix: Made bots roll in a more reasonable time on group loots.

SmashingQuasar:
# Description

This PR changes the way loot rolls are being evaluated.

It puts a maximum priority on the loot action so it does not hang for so long.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -1384,8 +1384,6 @@ void PlayerbotAI::DoNextAction(bool min)
     else if (bot->isAFK())
         bot->ToggleAFK();
 
-    PlayerbotAI* masterBotAI = nullptr;
-
     if (master && master->IsInWorld())
     {
         float distance = sServerFacade->GetDistance2d(bot, master);
--- a/src/strategy/generic/WorldPacketHandlerStrategy.cpp
+++ b/src/strategy/generic/WorldPacketHandlerStrategy.cpp
@@ -5,9 +5,6 @@
 
 #include "WorldPacketHandlerStrategy.h"
 
-#include "Playerbots.h"
-#include "Strategy.h"
-
 void WorldPacketHandlerStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
     PassTroughStrategy::InitTriggers(triggers);


src/PlayerbotAI.cpp
```diff
@@ -246,6 +246,21 @@ void PlayerbotAI::UpdateAI(uint32 elapsed, bool minimal)
...
+	if (group && group->isRollLootActive() && engines[BOT_STATE_NON_COMBAT])
+	{
+		ActionResult actionResult = engines[BOT_STATE_NON_COMBAT]->ExecuteAction("loot roll");
+
+		if (actionResult == ACTION_RESULT_OK || actionResult == ACTION_RESULT_UNKNOWN)
+		{
+			return;
+		}
+	}
+
```
SmashingQuasar: Considering the `LootRollAction::IsUseful` already includes this check, we could do without; however, because of the current complexity of the code base, I opted to add it here as well in case someone mistakenly changes `LootRollAction::IsUseful` without understanding the consequences of their action.

src/strategy/actions/LootRollAction.cpp
```diff
@@ -13,6 +13,30 @@
...
+		{
+			isUseful = true;
+
+			break;
+		}
+	}
+
+	return isUseful;
+}
+
```
SmashingQuasar: This had never been implemented. This means this action was constantly being ran and would eat ticks for no reason on many occasions. This code attempts to exit as quickly as possible to avoid unnecessary actions.
Since we will iterate over rolls, it is useless to call `->isLootRollActive()` because in case it is `false`, then `rolls.size() == 0` and the code will return immediately, otherwise we will need to iterate anyway.

SmashingQuasar:
I completely changed my approach when I was told about action priorities.
This is something that really needs to be documented and we shouldn't hard code float but always use the constant instead to avoid creating confusion.

src/PlayerbotAI.cpp
```diff
@@ -1386,7 +1386,6 @@ void PlayerbotAI::DoNextAction(bool min)
     else if (bot->isAFK())
         bot->ToggleAFK();
 
-    Group* group = bot->GetGroup();
     PlayerbotAI* masterBotAI = nullptr;
```
privatecore: Some ninja fixes?) `masterBotAI` is not used either.

src/strategy/generic/WorldPacketHandlerStrategy.cpp
```diff
@@ -6,6 +6,7 @@
 #include "WorldPacketHandlerStrategy.h"
 
 #include "Playerbots.h"
+#include "Strategy.h"
```
privatecore: `Strategy.h` is already included in `PassTroughStrategy.h`.

src/PlayerbotAI.cpp
SmashingQuasar: `group` was never called anywhere so I just removed it. I can open another PR for this if needed and good catch for the masterBotAI. Let me know if we delete them both now.
I actually had done a completely different code originally which used `group` and when I deleted it I forgot that I wasn't the one who created the `group` variable since it's never used.

src/strategy/generic/WorldPacketHandlerStrategy.cpp
SmashingQuasar: Good catch, I'm trying out another fix right now but once I am done testing it, I'll delete this unnecessary include.

src/PlayerbotAI.cpp
privatecore: > `group` was never called anywhere so I just removed it

`masterBotAI` is never called anywhere either ;)
So, if you delete `group`, why not delete `masterBotAI` as well?

kadeshar:
That was made by purpose. To make loots look like made by real persons. Proposed PR can be make by custom strategy if someone dont like current approach. Personally i would say best will be configuration that loot should be instant or rgp-look-like.
Im disagree that we should merge current PR state.

biship:
Bots take way to long to loot by default and this PR doesn't make it instant, but makes it better.

SmashingQuasar:
> That was made by purpose. To make loots look like made by real persons. Proposed PR can be make by custom strategy if someone dont like current approach. Personally i would say best will be configuration that loot should be instant or rgp-look-like. Im disagree that we should merge current PR state.

This PR does not actually make it instant but reasonable instead.
Immersion is fine, until it becomes obnoxious and the way it was before was awful because of the extremely low priority on roll. Messing with world packet triggers is also generally a pretty bad idea so it would be preferable if they all behaved in the same way, even if it's only a question of priority.
This change was also discussed and requested by contributors and admins' on Discord.

SmashingQuasar:
PR was rebased on most recent `master`.

kadeshar:
Never had problem with that. But maybe other reviewers can comment that subject.

hermensbas:
Whats the effective change between current and new?

SmashingQuasar:
> Whats the effective change between current and new?

Previously bots would take often several minutes or time out the roll. Now they roll in about 10-15 seconds.

hermensbas:
I have no real opinion or experience in this matter, so i trust @kadeshar and @noisiver opinion more then my own on things like this. As they both are more advanced and experienced in actual playing with the bots :)

privatecore:
> Never had problem with that. But maybe other reviewers can comment that subject.

Theoretically, a low priority may lead to long wait times if many other higher-priority actions are being processed. I think this was discussed somewhere in Discord(?), although, of course, it would be better to have an issue or feature request.

biship:
Personally, I like the reduced roll times. Especially along with the much improved rolling method here: https://github.com/mod-playerbots/mod-playerbots/pull/1610
Usually within 5-10s I can see if I won the roll, or the stupid mage took my loot.
Waiting a minute or more to see the result really slows down progress through a dungeon.

Dreathean:
Can probably change the title since it is within 10-15 seconds instead of instant like originally setup. Agreed rolling currently takes way too long, it's been a bit for me but I definitely recall it taking upwards of a minute or more for every roll which is a significant QOL issue and also breaks immersion.

SmashingQuasar:
> Can probably change the title since it is within 10-15 seconds instead of instant like originally setup. Agreed rolling currently takes way too long, it's been a bit for me but I definitely recall it taking upwards of a minute or more for every roll which is a significant QOL issue and also breaks immersion.

Good point, I modified the title. I forgot to change it after making the roll rely only on the action priority to respect the original design.

Wishmaster117:
Okay, here's my opinion even though it doesn't count.
Overall, I’m in favor of this PR, especially for a server with many bots/players, but with one nuance: ideally, this behavior should be configurable, not forced as a global default.

### Pros:

- Loot decisions happen in ~10–15 seconds instead of taking up to a minute or even expiring, which massively improves dungeon/raid flow, especially on busy or bot-heavy servers.
- The behavior still feels human-like (not instant ninja-loot), but we’re no longer waiting for bots to “finish reading the item description three times” before clicking.
- Raising the priority of loot-roll handling makes the behavior more consistent under load instead of starving this action behind other AI tasks.

### Possible improvements:

- Ideally this should be configurable (e.g. a loot roll delay / style option) so servers that like very slow, “RPG-style” rolls can keep that flavor.

hermensbas:
> > Never had problem with that. But maybe other reviewers can comment that subject.
> 
> Theoretically, a low priority may lead to long wait times if many other higher-priority actions are being processed. I think this was discussed somewhere in Discord(?), although, of course, it would be better to have an issue or feature request.

Thats actually interesting and sounds plausible, has this issue been verified correctly? As in is the large roll delay always there? Or sometimes? or often? If not in what conditions/situations does it seem to occure? Because if this the result of high priority taking the lead into low priority we need to deal with that instead, or even its working as intended, and not work around by basically adding more back pressure by increasing the amount how often we check or changing the priority. It Might help in certain conditions and for quick results not sure about intended in regard of other actions and priority mechanism. 

Before this is considered we should determine what we are trying to fix here and whether the issue is confirmed. Maybe this simply the result of handling priorities in combination with misconfiguration where you flood your own server due running to many bots. As result the high priority vs low priority has much larger impact due the back pressure of actions that needs to be processed.

From that perspective if there to many actions being process means prolly only the high/med priority will be processed and lower priority discarded for long(er) time.


**I wonder if this delay this still happens when you have 100 bots in your game while ur server runs on decent hardware.**

SmashingQuasar:
> > > Never had problem with that. But maybe other reviewers can comment that subject.
> > 
> > 
> > Theoretically, a low priority may lead to long wait times if many other higher-priority actions are being processed. I think this was discussed somewhere in Discord(?), although, of course, it would be better to have an issue or feature request.
> 
> Thats actually interesting and sounds plausible, has this issue been verified correctly? As in is the large roll delay always there? Or sometimes? or often? If not in what conditions/situations does it seem to occure? Because if this the result of high priority taking the lead into low priority we need to deal with that instead, or even its working as intended, and not work around by basically adding more back pressure by increasing the amount how often we check or changing the priority. It Might help in certain conditions and for quick results not sure about intended in regard of other actions and priority mechanism.
> 
> Before this is considered we should determine what we are trying to fix here and whether the issue is confirmed. Maybe this simply the result of handling priorities in combination with misconfiguration where you flood your own server due running to many bots. As result the high priority vs low priority has much larger impact due the back pressure of actions that needs to be processed.
> 
> From that perspective if there to many actions being process means prolly only the high/med priority will be processed and lower priority discarded for long(er) time.
> 
> **I wonder if this delay this still happens when you have 100 bots in your game while ur server runs on decent hardware.**

This issue has been discussed and reported several times, by @noisiver and @Dreathean if I remember correctly. @noisiver does not use random bots from my understanding so he only has 39 bots running around with him.

Since this action is tied to a world packet, I would assume there is no check every tick, instead it is called as a callback when a certain world packet is received. Could you detail a bit more what you mean regarding the back pressure build up? Maybe I am just confused in how the world packet triggers are being handled.

biship:
Depending on server load (because it's dynamic), I have anywhere from 500 to 1500 bots.  My server averages around 22-28ms. This PR makes dungeon instances much smoother and enjoyable.

privatecore:
> Depending on server load (because it's dynamic), I have anywhere from 500 to 1500 bots. My server averages around 22-28ms. This PR makes dungeon instances much smoother and enjoyable.

I think it would be more informative if you could provide performance monitor data -- this way we can confirm the presence of the problem, try to reproduce it, and localize it if possible.

hermensbas:
> > Depending on server load (because it's dynamic), I have anywhere from 500 to 1500 bots. My server averages around 22-28ms. This PR makes dungeon instances much smoother and enjoyable.
> 
> I think it would be more informative if you could provide performance monitor data -- this way we can confirm the presence of the problem, try to reproduce it, and localize it if possible.

Ye or use some reference configuration, that might help.

Celandriel:
I dont think this is a description of a problem, just a statement enjoying the changes

Celandriel (approved):

Issue #1854, opened 2025-11-18
The entire log is littered with one error


Regrad:
**Bug Description**
The server log is flooded with this message. It outputs 10-20 messages per second. This has never happened before. This behavior began to manifest itself on all three realms at once.

```
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
[...]

PR #1858, opened 2025-11-19
fix: warning updating movement flags while rooted

hermensbas:
fixes https://github.com/mod-playerbots/mod-playerbots/issues/1854
-----
Also includes fixes for:
-----
* Bots swimming with waterWalk kept switching between swimming and walking, as result jittering effect swimming under water when water walking active
* Bots flying close above water they would land on water and start walking, now they stay flying unless on solid ground they will land and start walking by design

-----
Moved all flag setting to updateMovementState:
   * So all movement flag are handled in updateMovementState which also contains the restricted movement logic.
   * Handle restricted movement logic and preventing SendMovementFlagUpdate while being restricted.

-----
Known issue when flying the following bots feel a bit jittering, wont touch for now at least till core movement changes quirks has been dealt with.

The current code is the extended version of what is originally was before core merge with refactored movements. Once the core movement refactors are settled a bit more i would like to revisit this code; as i would expect more imperative code and less manual flag setting e.g. bot->SetWaterWalking, SetGravitiy..SetCanFly etc.

### DIFF
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -1048,7 +1048,6 @@ void MovementAction::UpdateMovementState()
 
         if (movementFlagsUpdated)
             bot->SendMovementFlagUpdate();
-
     }
 
      // Save current state for the next check


noisiver (approved):

SmashingQuasar:
Unfortunately I cannot help test this because I have too many opened PRs that are currently on my plate, also working on inventory management, but I can't wait for this to be merged because it's very difficult to debug anything with this root flag error. Thanks a lot for taking care of it!

Issue #1840, opened 2025-11-14
Recent crashes posted by @Regrad that need fixing


Wishmaster117:
> [ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_2-39-26].txt](https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt) [ae28c3e1dcc4_worldserver.exe_[13-11_4-12-13].txt](https://github.com/user-attachments/files/23529093/ae28c3e1dcc4_worldserver.exe_.13-11_4-12-13.txt)

### Crash 1 Source
**[ae28c3e1dcc4_worldserver.exe_[13-11_17-9-50].txt](https://github.com/user-attachments/files/23529089/ae28c3e1dcc4_worldserver.exe_.13-11_17-9-50.txt)**
Crash : ACCESS_VIOLATION in Item::CheckSoulboundTradeExpired (Item.cpp, AzerothCore Playerbot core).
Called by Player::UpdateSoulboundTradeItems in PlayerStorage.cpp during a player's tick.
Faulty thread : MapUpdater worker, without a direct call to the mod-playerbots module.
Most likely cause: null pointer due to an inconsistent state in the trade data of a soulbound item (BOP tradable) (item deleted or invalid but still present in the trade logic).
The playerbots module does not contain the faulty function, but can more easily trigger this case due to the massive looting/trading by bots.
@blinkysc : Perhaps post this crash report on the Azerothcore GitHub page?

### Crash 2 Source
**https://github.com/user-attachments/files/23529092/ae28c3e1dcc4_worldserver.exe_.13-11_2-39-26.txt:**
**Immediate cause of the crash:**
An RTSC "see spell" event arrives with an empty WorldPacket in SeeSpellAction::Execute.
The code reads from it without any checks → out-of-bounds reading of the ByteBuffer → C++ exception → uncaught → crash in the MapThread.
**Faulty area in the module:**
mod-playerbots/src/strategy/actions/SeeSpellAction.cpp, function SeeSpellAction::Execute(Event event) (line ~40 in your build).
**What needs to be fixed :**
Before running p >> ..., verify that the packet contains enough bytes for castCount + spellId + castFlags.
[...]

PR #1859, opened 2025-11-19
Core - Thread-safety: Secure PacketHandlingHelper in playerbots

Wishmaster117:
This patch makes the PacketHandlingHelper class thread-safe by adding a mutex around the packet stack.
Added std::mutex and std::lock_guard to protect concurrent access to the packet stack.
Packets are copied under lock and processed outside the critical section to minimize blocking and avoid re-entrancy issues.
Only packets with a registered handler are enqueued, reducing unnecessary work.
The AI logic and bot behavior remain unchanged, this patch only secures packet handling.
This change is compatible with the threadSafety PR #1816 and improves robustness of the playerbots module. It should not introduce any side effects on AI or other parts of the system.

This fixes the **source 4** of the crash posted by @Regrad 
https://github.com/mod-playerbots/mod-playerbots/issues/1840

### DIFF
--- a/src/PlayerbotAIConfig.cpp
+++ b/src/PlayerbotAIConfig.cpp
@@ -725,8 +725,8 @@ std::string const PlayerbotAIConfig::GetTimestampStr()
     //       HH     hour (2 digits 00-23)
     //       MM     minutes (2 digits 00-59)
     //       SS     seconds (2 digits 00-59)
-    char buf[20];
-    snprintf(buf, 20, "%04d-%02d-%02d %02d-%02d-%02d", aTm->tm_year + 1900, aTm->tm_mon + 1, aTm->tm_mday, aTm->tm_hour,
+    char buf[32];
+    snprintf(buf, sizeof(buf), "%04d-%02d-%02d %02d-%02d-%02d", aTm->tm_year + 1900, aTm->tm_mon + 1, aTm->tm_mday, aTm->tm_hour,
              aTm->tm_min, aTm->tm_sec);
     return std::string(buf);
 }
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -27,7 +27,9 @@
 #include "PlayerbotAIConfig.h"
 #include "PlayerbotDbStore.h"
 #include "PlayerbotFactory.h"
+#include "PlayerbotOperations.h"
 #include "PlayerbotSecurity.h"
+#include "PlayerbotWorldThreadProcessor.h"
 #include "Playerbots.h"
 #include "RandomPlayerbotMgr.h"
 #include "SharedDefines.h"
@@ -85,7 +87,6 @@ public:
 
 void PlayerbotHolder::AddPlayerBot(ObjectGuid playerGuid, uint32 masterAccountId)
 {
-    // bot is loading
     if (botLoading.find(playerGuid) != botLoading.end())
         return;
 
@@ -195,7 +196,9 @@ void PlayerbotHolder::HandlePlayerBotLoginCallback(PlayerbotLoginQueryHolder con
     }
 
     sRandomPlayerbotMgr->OnPlayerLogin(bot);
-    OnBotLogin(bot);
+
+    auto op = std::make_unique<OnBotLoginOperation>(bot->GetGUID(), this);
+    sPlayerbotWorldProcessor->QueueOperation(std::move(op));
 
     botLoading.erase(holder.GetGuid());
 }
@@ -316,11 +319,9 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
         if (!botAI)
             return;
 
-        Group* group = bot->GetGroup();
-        if (group && !bot->InBattleground() && !bot->InBattlegroundQueue() && botAI->HasActivePlayerMaster())
-        {
-            sPlayerbotDbStore->Save(botAI);
-        }
+        // Queue group cleanup operation for world thread
+        auto cleanupOp = std::make_unique<BotLogoutGroupCleanupOperation>(guid);
+        sPlayerbotWorldProcessor->QueueOperation(std::move(cleanupOp));
 
         LOG_DEBUG("playerbots", "Bot {} logging out", bot->GetName().c_str());
         bot->SaveToDB(false, false);
@@ -549,6 +550,7 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
 
     botAI->TellMaster("Hello!", PLAYERBOT_SECURITY_TALK);
 
+    // Queue group operations for world thread
     if (master && master->GetGroup() && !group)
     {
         Group* mgroup = master->GetGroup();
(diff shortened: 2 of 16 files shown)


hermensbas:
Couldnt we simply use <boost/lockfree/stack.hpp> as thread stack with no hard locking? Looking the locking i assume the object doesnt exist anymore and then gets called again through via UpdateAIInternal. Or whats your take on this?

As in the playerbotAI has been destroyed while its still in the loop (somewhere in a thread) sort of say, then comes back...poooef

So maybe the problem cause starts with the destroying the botAI object while still being in the loop. e.g ' delete botAI'. My best quess would be the one in DisablePlayerBot in the PlayerbotMgr which directly removes the AI object while the AI object should only be removed after the core has verified the player has been destroyed.

```c++
void PlayerbotHolder::DisablePlayerBot(ObjectGuid guid)
{
    if (Player* bot = GetPlayerBot(guid))
    {
        PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
        if (!botAI)
        {
            return;
        }
        botAI->TellMaster("Goodbye!");
        bot->StopMoving();
        bot->GetMotionMaster()->Clear();

        if (group && !bot->InBattleground() && !bot->InBattlegroundQueue() && botAI->HasActivePlayerMaster())
        {
            sPlayerbotDbStore->Save(botAI);
        }

        LOG_DEBUG("playerbots", "Bot {} logged out", bot->GetName().c_str());

        bot->SaveToDB(false, false);

        if (botAI->GetAiObjectContext())  // Maybe some day re-write to delate all pointer values.
        {
            TravelTarget* target = botAI->GetAiObjectContext()->GetValue<TravelTarget*>("travel target")->Get();
            if (target)
                delete target;
        }

        RemoveFromPlayerbotsMap(guid);  // deletes bot player ptr inside this WorldSession PlayerBotMap

        delete botAI;
    }
}

void PlayerbotHolder::RemoveFromPlayerbotsMap(ObjectGuid guid)
{
    playerBots.erase(guid);
}

```

Used OnPlayerbotLogout flow in one of scripts hooks.  while also having  void OnDestructPlayer(Player* player) override
and there few other places in the code where botAI seems to be deleted.

hermensbas:
alot of these problems could have been avoided by simply making the AI as internal property from Player.h instead of using these reverse index maps.

Wishmaster117:
> Couldnt we simply use <boost/lockfree/stack.hpp> as thread stack with no hard locking? Looking the locking i assume the object doesnt exist anymore and then gets called again through via UpdateAIInternal. Or whats your take on this?
> 
> As in the playerbotAI has been destroyed while its still in the loop (somewhere in a thread) sort of say, then comes back...poooef
> 
> So maybe the problem cause starts with the destroying the botAI object while still being in the loop. e.g ' delete botAI'. My best quess would be the one in DisablePlayerBot in the PlayerbotMgr which directly removes the AI object while the AI object should only be removed after the core has verified the player has been destroyed.
> 
> ```c++
> void PlayerbotHolder::DisablePlayerBot(ObjectGuid guid)
> {
>     if (Player* bot = GetPlayerBot(guid))
>     {
>         PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
>         if (!botAI)
>         {
>             return;
>         }
>         botAI->TellMaster("Goodbye!");
>         bot->StopMoving();
>         bot->GetMotionMaster()->Clear();
> 
>         Group* group = bot->GetGroup();
>         if (group && !bot->InBattleground() && !bot->InBattlegroundQueue() && botAI->HasActivePlayerMaster())
>         {
>             sPlayerbotDbStore->Save(botAI);
>         }
> 
>         LOG_DEBUG("playerbots", "Bot {} logged out", bot->GetName().c_str());
> 
>         bot->SaveToDB(false, false);
> 
>         if (botAI->GetAiObjectContext())  // Maybe some day re-write to delate all pointer values.
>         {
>             TravelTarget* target = botAI->GetAiObjectContext()->GetValue<TravelTarget*>("travel target")->Get();
>             if (target)
>                 delete target;
>         }
> 
>         RemoveFromPlayerbotsMap(guid);  // deletes bot player ptr inside this WorldSession PlayerBotMap
> 
>         delete botAI;
>     }
> }
> 
> void PlayerbotHolder::RemoveFromPlayerbotsMap(ObjectGuid guid)
> {
>     playerBots.erase(guid);
> }
> ```
> 
> Used OnPlayerbotLogout flow in one of scripts hooks. while also having void OnDestructPlayer(Player* player) override and there few other places in the code where botAI seems to be deleted.

You’re absolutely right that my patch with the mutex only addresses concurrency on the packet stack itself. It doesn’t solve the deeper issue of the lifecycle of the PlayerbotAI object.

Even if we switch to a lock‑free structure like boost::lockfree::stack, we would still face the same fundamental risk: if the PlayerbotAI instance is destroyed while another thread is inside Handle or UpdateAIInternal, we end up with a dangling pointer and a crash. The locking vs. lock‑free choice doesn’t change that.

The real root cause is that DisablePlayerBot (and similar flows) can call delete botAI while the core may still be executing logic on that object. As you suggest, the safer approach would be to tie the AI’s lifecycle directly to the Player object, or at least defer deletion until the core has confirmed the player is fully destroyed. That way, we avoid races between destruction and usage.

So,
My patch makes PacketHandlingHelper thread‑safe and prevents data corruption.
It does not fix the lifecycle/destruction race.
A more robust long‑term solution would be to integrate PlayerbotAI as an internal property of Player (instead of managing it via reverse index maps), so its lifetime is naturally bound to the player’s.

hermensbas:
Lets do abit more research on this

hermensbas:
Lets keep this in draft for now, to we can explore a bit more. This PR and other thread related does have my attention. If you dont mind @Wishmaster117 ?

Wishmaster117:
> Lets keep this in draft for now, to we can explore a bit more. This PR and other thread related does have my attention. If you dont mind @Wishmaster117 ?

No problemo it's just a piece of tape 😂

Regrad:
image

Wishmaster117:
@Regrad have you now a clear message of intercepted packets ?

Regrad:
> @Regrad have you now a clear message of intercepted packets ?

Need to update the PR to the last version of the main branch. Error when trying to apply it.

Regrad:
Hmm. I don't know why, but there's still a merge error. I apply 1843 first and immediately after it this PR. 

```
CONFLICT (content): Merge conflict in src/PlayerbotAI.cpp
error: Failed to merge in the changes.
```

```
     // Only enqueue packets we actually handle
     if (handlers.find(packet.GetOpcode()) == handlers.end())
     {
+<<<<<<< HEAD
+        LOG_ERROR("playerbots", "Intercepted unhandled packet (opcode: {}), ignored to prevent crash.",
+                  static_cast<uint32>(packet.GetOpcode()));
+=======
         LOG_ERROR("playerbots",
             "Intercepted unhandled packet (opcode: {}), ignored to prevent crash.",
             static_cast<uint32>(packet.GetOpcode()));
+>>>>>>> Correct LOG_ERROR
         return;
     }
```

Wishmaster117:
This will nor result have to found another way

PR #1860, opened 2025-11-19
[DRAFT] fix: Made bots drink responsibly by implementing a real check algorithm on drinking.

SmashingQuasar:
# Description

> [!WARNING]
> This PR is in progress. I am testing those changes to see how the bots behave.
> I tried bringing more structure to the code but the final structure might change depending on the feedback on this PR.

This PR entirely reworks the way bots drink.
Drinking was originally based on a sleep of 12 to 18 seconds depending on the context (battleground, non-battleground).
This led to a softlock of the bot AI during this time. Because of this, bots would stop responding to any command, they wouldn't do anything. They wouldn't react to world packets, move, even fight back.

Unfortunately, we **must** check that the bot is drinking in `PlayerbotAI::DoNextAction()`. This is the only time we can prevent the bot from getting out of drinking whilst still letting it respond to world packet and be able to snap out of the drinking.

The bots will snap out of drinking under any of the following circumstances:
- It is in combat.
- Its total mana is above 95%.
- Its master is farther away than 20 yards.
- It's under water, this way the bots will no longer drown themselves by drinking.
- It does not use mana, in case someone makes a bot drink while not having mana...
- Its master is in combat.

The check is done every second instead of every tick to avoid polluting the main loop with useless calls. I originally lowered it to 500ms and was quite happy with it so we could also opt for this.

It is likely that other conditions will arise from testing this PR.
People are highly encouraged to test this even in draft because it is quite difficult for me to try a lot of different situations.
So far the bots are considerably more responsive with this change and there is no visible performance hit.

> [!NOTE]
> I tried to optimise the checks to go from the cheapest one to the most expansive one to bail out as soon as possible.
> I also separated the `DrinkAction` class from the `NonCombatAction` file because... I'm upset that we cram several symbols in one file.
> I use inline functions to evaluate the drinking conditions in a header-only file to make it as fast as possible. It's also intentional to avoid adding yet another method to `PlayerbotAI` that has more than enough.
> I did not create a mega condition intentionally within `shouldKeepDrinking`. Mega conditions serve no purpose performance-wise and are extremely hard to maintain.

### DIFF
--- a/src/strategy/actions/non-combat/drink/DrinkAction.cpp
+++ b/src/strategy/actions/non-combat/drink/DrinkAction.cpp
@@ -9,11 +9,11 @@
 
 bool DrinkAction::isPossible()
 {
-    bool possible = !bot->IsInCombat() &&
-        !bot->IsMounted() &&
-        !botAI->HasAnyAuraOf(GetTarget(), "dire bear form", "bear form", "cat form", "travel form",
+    bool possible = !this->bot->IsInCombat() &&
+        !this->bot->IsMounted() &&
+        !this->botAI->HasAnyAuraOf(GetTarget(), "dire bear form", "bear form", "cat form", "travel form",
             "aquatic form","flight form", "swift flight form", nullptr) &&
-        (botAI->HasCheat(BotCheatMask::food) || UseItemAction::isPossible());
+        (this->botAI->HasCheat(BotCheatMask::food) || UseItemAction::isPossible());
 
     return possible;
 }
@@ -30,28 +30,28 @@ bool DrinkAction::isUseful()
 
 bool DrinkAction::Execute(Event event)
 {
-    std::string name = bot->GetName().c_str();
+    std::string name = this->bot->GetName().c_str();
 
-    bot->ClearUnitState(UNIT_STATE_CHASE);
-    bot->ClearUnitState(UNIT_STATE_FOLLOW);
+    this->bot->ClearUnitState(UNIT_STATE_CHASE);
+    this->bot->ClearUnitState(UNIT_STATE_FOLLOW);
 
-    botAI->InterruptSpell();
+    this->botAI->InterruptSpell();
 
-    if (bot->isMoving())
-        bot->StopMoving();
+    if (this->bot->isMoving())
+        this->bot->StopMoving();
 
-    bot->SetStandState(UNIT_STAND_STATE_SIT);
+    this->bot->SetStandState(UNIT_STAND_STATE_SIT);
 
-    if (botAI->HasCheat(BotCheatMask::food))
+    if (this->botAI->HasCheat(BotCheatMask::food))
     {
-        bot->AddAura(DRINK_AURA_GRACCU_MINCE_MEAT_FRUITCAKE, bot);
+        this->bot->AddAura(DRINK_AURA_GRACCU_MINCE_MEAT_FRUITCAKE, bot);
 
-        botAI->SetNextCheckDelay(1000);
+        this->botAI->SetNextCheckDelay(1000);
 
         return true;
     }
 
-    botAI->SetNextCheckDelay(1000);
+    this->botAI->SetNextCheckDelay(1000);
 
     return UseItemAction::Execute(event);
 }
--- a/src/domain/core/inspector/abstract/AbstractInspector.h
+++ b/src/domain/core/inspector/abstract/AbstractInspector.h
@@ -1,0 +1,19 @@
+#pragma once
+
+class AbstractInspector
+{
+protected:
+    AbstractInspector() = default;
+    ~AbstractInspector() = default;
+
+    AbstractInspector(AbstractInspector const&) = default;
+    AbstractInspector& operator=(AbstractInspector const&) = default;
+
+    AbstractInspector(AbstractInspector&&) = default;
+    AbstractInspector& operator=(AbstractInspector&&) = default;
+
+	virtual bool isInspectable() const
+	{
+		return true;
+	}
+};
--- a/src/domain/player/inspector/global/GlobalPlayerInspector.h
+++ b/src/domain/player/inspector/global/GlobalPlayerInspector.h
@@ -6,12 +6,11 @@
 #include "ObjectAccessor.h"
 #include "SharedDefines.h"
 #include "Player.h"
+// #include "GridNotifiers.h"
 
 #include "AiFactory.h"
 #include "PlayerbotAI.h"
 #include "AbstractPlayerInspector.h"
-#include "ItemActionStruct.h"
-#include "ItemActionEnum.h"
 
 class GlobalPlayerInspector : public AbstractPlayerInspector
 {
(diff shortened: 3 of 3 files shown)


src/utility/isDrinking.h
```diff
@@ -0,0 +1,33 @@
+#ifndef IS_DRINKING_H
+#define IS_DRINKING_H
+
+#include <vector>
+#include <cstdint>
+
+#include "Unit.h"
+#include "Player.h"
+
+inline bool isDrinking(Player* player)
```
privatecore: Take a look at how eating and drinking is handled in the core:
https://github.com/azerothcore/azerothcore-wotlk/blob/3c4a68c2e283805d3265e93ca2c8c303ec8d8e54/src/server/game/Entities/Player/Player.cpp#L1820

SmashingQuasar: Yes we had a discussion about this but this seems to be quite intensive compared to a simple lookup within a vector.
We had the discussion regarding people possibly adding drinkable things, but my argument was that:
- I have yet to see anyone do this on any core so far.
- It would not break the bot's AI, the bot would just gulp all the drinks back to back until it runs out.
- Vector lookup is extremely fast and this function is called all the time so I would rather not use a heavier code to support hypothetical situations that might never exist.

The core has to support it so they have no other way but I feel like we aren't in the same boat.

Wishmaster117:
Tired of the boats in EoTS that start drinking right after picking up the flag :D

src/utility/isDrinking.h
SmashingQuasar: Though I think that we might want to do it that way to avoid drinking while under innervate. This kind of situation can be quite tricky. Same for not interrupting evocate.

SmashingQuasar:
I haven't handled the situations where the bot is in danger as I am having trouble finding a good code example of how it's being handled.

Wishmaster117:
> I haven't handled the situations where the bot is in danger as I am having trouble finding a good code example of how it's being handled.

Scan the enemies from a certain distance?
When you're a player if you see a danger you don't drink, you run or you attack

SmashingQuasar:
Yes, I don't know how to do that in AC yet.

src/utility/isDrinking.h
privatecore: The problem is not with vector lookup, but with hardcoded aura IDs. Drinking auras can be easily determined, and AC has an example of how to do this:
A spell/aura should have the `SPELL_AURA_MOD_POWER_REGEN` effect + the `AURA_INTERRUPT_FLAG_NOT_SEATED` interrupt flag.

> The core has to support it so they have no other way but I feel like we aren't in the same boat.

What are you talking about?

hermensbas: Isnt the following enough to determine your drinking or not?

```c++
inline bool IsDrinking(Player* bot)
{
    if (!bot || bot->HasAuraType(AURA_INTERRUPT_FLAG_NOT_SEATED))
        return false;

    // Drinking = mana regen aura
    bool drinking =
        bot->HasAuraType(SPELL_AURA_MOD_POWER_REGEN) ||
        bot->HasAuraType(SPELL_AURA_MOD_POWER_REGEN_PERCENT);

    // Eating = health regen aura
    bool eating =
        bot->HasAuraType(SPELL_AURA_MOD_REGEN) ||
        bot->HasAuraType(SPELL_AURA_MOD_REGEN_PERCENT);

    return drinking || eating;
}
```

or maybe the sit state needs a bit more, but check core Unit(line:4600) and Player

SmashingQuasar: > The problem is not with vector lookup, but with hardcoded aura IDs. Drinking auras can be easily determined, and AC has an example of how to do this: A spell/aura should have the `SPELL_AURA_MOD_POWER_REGEN` effect + the `AURA_INTERRUPT_FLAG_NOT_SEATED` interrupt flag.
> 
> > The core has to support it so they have no other way but I feel like we aren't in the same boat.
> 
> What are you talking about?

I mean that because this is a very hot function, I want to keep it optimised. Direct vector lookup (and/or unordered set, depending on what you want to do) is the fastest you can be because a contiguous vector of unsigned integer is the smallest possible unit of rich data that can't fit in a byte mask.
Regarding hard-coding aura ids, I get what you mean, but is there a use case against it? AC has been around for a very long time and I can't find a module that adds any aura regen. I feel like we are discussing a phantom use case.
With that in mind, I still believe you are right due to some other edge cases that might break as I mentioned so I'm going to change it. I'll use @hermensbas suggestion since it uses built-in methods which can eventually be optimised more by the code rather than custom code.

privatecore: > AC has been around for a very long time and I can't find a module that adds any aura regen. I feel like we are discussing a phantom use case.

```cpp
    // Handles the emotes for drinking and eating.
    // According to sniffs there is a background timer going on that repeats independed from the time window where the aura applies.
    // That's why we dont need to reset the timer on apply. In sniffs I have seen that the first call for the spell visual is totally random, then after
    // 5 seconds over and over again which confirms my theory that we have a independed timer.
    if (m_foodEmoteTimerCount >= 5000)
    {
        std::vector<AuraEffect*> auraList;
        AuraEffectList const& ModRegenAuras = GetAuraEffectsByType(SPELL_AURA_MOD_REGEN);
        AuraEffectList const& ModPowerRegenAuras = GetAuraEffectsByType(SPELL_AURA_MOD_POWER_REGEN);

        auraList.reserve(ModRegenAuras.size() + ModPowerRegenAuras.size());
        auraList.insert(auraList.end(), ModRegenAuras.begin(), ModRegenAuras.end());
        auraList.insert(auraList.end(), ModPowerRegenAuras.begin(), ModPowerRegenAuras.end());

        for (auto itr = auraList.begin(); itr != auraList.end(); ++itr)
        {
            // Food emote comes above drinking emote if we have to decide (mage regen food for example)
            if ((*itr)->GetBase()->HasEffectType(SPELL_AURA_MOD_REGEN) && (*itr)->GetSpellInfo()->AuraInterruptFlags & AURA_INTERRUPT_FLAG_NOT_SEATED)
            {
                SendPlaySpellVisual(SPELL_VISUAL_KIT_FOOD);
                break;
            }
            else if ((*itr)->GetBase()->HasEffectType(SPELL_AURA_MOD_POWER_REGEN) && (*itr)->GetSpellInfo()->AuraInterruptFlags & AURA_INTERRUPT_FLAG_NOT_SEATED)
            {
                SendPlaySpellVisual(SPELL_VISUAL_KIT_DRINK);
                break;
            }
        }
        m_foodEmoteTimerCount -= 5000;
    }
```

https://github.com/azerothcore/azerothcore-wotlk/blob/3c4a68c2e283805d3265e93ca2c8c303ec8d8e54/src/server/game/Entities/Player/Player.cpp#L1820

```cpp
                // Food / Drinks (mostly)
                if (AuraInterruptFlags & AURA_INTERRUPT_FLAG_NOT_SEATED)
                {
                    bool food = false;
                    bool drink = false;
                    for (uint8 i = 0; i < MAX_SPELL_EFFECTS; ++i)
                    {
                        if (!Effects[i].IsAura())
                            continue;
                        switch (Effects[i].ApplyAuraName)
                        {
                            // Food
                            case SPELL_AURA_MOD_REGEN:
                            case SPELL_AURA_OBS_MOD_HEALTH:
                                food = true;
                                break;
                            // Drink
                            case SPELL_AURA_MOD_POWER_REGEN:
                            case SPELL_AURA_OBS_MOD_POWER:
                                drink = true;
                                break;
                            default:
                                break;
                        }
                    }

                    if (food && drink)
                        return SPELL_SPECIFIC_FOOD_AND_DRINK;
                    else if (food)
                        return SPELL_SPECIFIC_FOOD;
                    else if (drink)
                        return SPELL_SPECIFIC_DRINK;
                }
```

https://github.com/azerothcore/azerothcore-wotlk/blob/3c4a68c2e283805d3265e93ca2c8c303ec8d8e54/src/server/game/Spells/SpellInfo.cpp#L2160

It’s literally in the AC codebase.

hermensbas:
Haven hadt time to fully look into the code, but if i read the code: the general idea is having a shorter checkDelay so we have the chance to interupt the drinking/eating? Way better then whats its now then blunty waiting and interupt the AI. We should prevent long checkDelays in all formates tbh andel ensure the AI stays responsive.

SmashingQuasar:
> Haven hadt time to fully look into the code, but if i read the code: the general idea is having a shorter checkDelay so we have the chance to interupt the drinking/eating? Way better then whats its now then blunty waiting and interupt the AI. We should prevent long checkDelays in all formates tbh andel ensure the AI stays responsive.

Yes, the problem with the existing code is that it does an 18 seconds sleep which entirely blocks the bot's AI. Because of that the bots are very easy to abuse in PvP, they drown themselves when under water, they become unresponsive in dungeons (even to master's commands), they don't react to world packet (which is a bit dangerous in my opinion).

We need to wait a bit after applying the initial drinking aura. Otherwise it ticks so fast that the aura does not have time to be applied before the next tick and the bot gets confused. There is no reason not to drink for at least one second anyway.
For the regular check, I set it to 1 second because I want to preserve performance as much as possible. I did try with 500ms on my own server and didn't notice any performance degradation whatsoever. In my opinion, 500ms is fine because it makes them even more reactive and does not have that huge of a cost.

I also want to prevent the bots from drinking when there are hostile players in the vicinity. Right now they will mindlessly start drinking if you let them get out of combat, even if you are in melee range, giving you a free critical hit on them.

src/utility/isDrinking.h
SmashingQuasar: I changed it to @hermensbas suggestion to the difference that I use the built-in `IsSitState()` instead of checking for an aura. I think it's preferable to use built-in methods. That way we benefit from possible future optimisations and don't recode what already exists. It's a bit like `AI_VALUE2()` to check if the bot has mana. It already exists within the code through a more explicit method.

SmashingQuasar: @privatecore I have a feeling we are not talking about the same thing and we don't understand each other on that matter.
The core does things in a certain way which are not binding to our own logic. The core needs to cover more situations than us. We should also avoid taking the core code directly and use exposed methods as this is cleaner and stays consistent with what the core exposes instead of re-using code snippets. The second one could benefit from a bit of refactoring to make it more maintainable in particular.

In any case it's been changed to use the code suggested by bash.

privatecore: @SmashingQuasar I have the same feeling as if I'm talking to an AI. I wrote: "Look at how eating and drinking is handled in the core". In response, received a lengthy text stating that: no one has ever done this; it doesn't break the logic; vector lookup is faster...

No one told you to "GRAB CODE FROM AC AND USE IT!" I told you to look at how AC determines when a player/character is drinking/eating.

Dreathean: No reason to get heated from a perfectly reasonable and thorough response. Does not read like AI to me at all.

privatecore: > No reason to get heated from a perfectly reasonable and thorough response. Does not read like AI to me at all.

I see, you also don't get the point. The main problem here is that the solution from https://github.com/mod-playerbots/mod-playerbots/pull/1860#discussion_r2543818703 uses the same logic as the AC codebase, but instead of reading and trying to understand the code, one gets these philosophical reflections:

> The core does things in a certain way which are not binding to our own logic. The core needs to cover more situations than us. We should also avoid taking the core code directly and use exposed methods as this is cleaner and stays consistent with what the core exposes instead of re-using code snippets. The second one could benefit from a bit of refactoring to make it more maintainable in particular.

SmashingQuasar: @privatecore I'm discussing reasonably and respectfully with you and I am trying to improve the code.
If you aren't able to have a reasonable argument I think there is no point going further.
I understand the code from AC very well, thank you. I just care about performance because this function is going to be called on every tick for every single bot. It matters to me that the code stays efficient. Sizing arrays and doing several fetches is less efficient than doing a direct vector lookup, that's just a fact. I also said that I agreed with you, so I am not sure why you keep arguing here and getting upset. In any case, I am resolving this discussion since it has lost all constructiveness.

privatecore: @SmashingQuasar I apologize if my words have offended you in any way, but from your side, it looked like endless arguing against using the existing logic in the core. I also want to note that the original message did not contain a call to use a specific piece of code -- on the contrary, it was a call to read and understand the logic behind it. :peace_symbol:

hermensbas:
Ye i also noticed that during testing they started to drink while being under water and actually drowning themself, haha. I think in general we need to keep prevent long delays when possible, sleeps which kinda breaks the whole idea of AI system. So indeed rather recheck that we need to continue drinking and interrupt if needed.

src/utility/isEatingOrDrinking.h
```diff
@@ -0,0 +1,28 @@
+#ifndef IS_DRINKING_H
+#define IS_DRINKING_H
+
+#include "SpellAuraDefines.h"
+#include "Player.h"
+
+inline bool isEatingOrDrinking(Player* player)
```
hermensbas: Mmm these are basically extension functions of Player.h

Should we (a)buse PlayerAI wrapper object for such extension methods or consider this approach instead? Pro its clean if implemented and used correctly from now on, down-side you need to be aware of their existence since the ide auto-complete wont give you any hints. 

Actually the other doesnt exclude the other. Eitherway maybe we should place them into src/extension/player? or src/function/player?

SmashingQuasar:
PR was rebased on most recent `master`. The code is quite stable on my end by the way, I've been using it for a few days without issue and bots quit drinking at appropriate times. I will do a similar change for eating in another PR.

src/PlayerbotAI.cpp
```diff
@@ -16,7 +16,6 @@
 #include "CharacterPackets.h"
 #include "ChatHelper.h"
 #include "Common.h"
-#include "CreatureAIImpl.h"
```
privatecore, line 19: Why are you making such modifications that are not even mentioned in your PR?

src/PlayerbotAI.cpp
```diff
@@ -1373,6 +1372,15 @@ void PlayerbotAI::DoNextAction(bool min)
 
     bool minimal = !AllowActivity();
 
+	if (!minimal && isEatingOrDrinking(bot) && shouldKeepDrinking(this))
```
privatecore, line 1449: All changes in this file are redundant -- `currentEngine->DoNextAction(nullptr, 0, (minimal || min));` already handles all necessary actions. The whole file should be restored to its original state; nothing related to "drinking" or "eating" is present here.

src/strategy/actions/ActionContext.h
```diff
@@ -43,10 +43,10 @@
 #include "MoveToTravelTargetAction.h"
 #include "MovementActions.h"
 #include "NonCombatActions.h"
+#include "DrinkAction.h"
 #include "OutfitAction.h"
 #include "PositionAction.h"
 #include "DropQuestAction.h"
-#include "RaidNaxxActions.h"
```
privatecore: Why?

src/strategy/actions/non-combat/drink/DrinkAction.cpp
```diff
@@ -0,0 +1,57 @@
...
+        return false;
+
+    bool keepDrinking = shouldKeepDrinking(botAI);
+
+    return keepDrinking;
+}
+
+bool DrinkAction::Execute(Event event)
+{
+    std::string name = bot->GetName().c_str();
```
privatecore: Not used variable.

src/strategy/actions/non-combat/drink/DrinkAction.cpp
```diff
@@ -0,0 +1,57 @@
...
+        bot->AddAura(DRINK_AURA_GRACCU_MINCE_MEAT_FRUITCAKE, bot);
+
+        botAI->SetNextCheckDelay(1000);
+
+        return true;
+    }
+
+    botAI->SetNextCheckDelay(1000);
+
+    return UseItemAction::Execute(event);
```
privatecore, line 56:
```cpp
    botAI->SetNextCheckDelay(1000);
    if (botAI->HasCheat(BotCheatMask::food))
    {
        bot->AddAura(<PROPER_SPELL_ID_WITH_SPELL_AURA_MOD_POWER_REGEN>, bot);
        return true;
    }

    return UseItemAction::Execute(event);
```

`PROPER_SPELL_ID_WITH_SPELL_AURA_MOD_POWER_REGEN` -- 25990 does not have the proper EffectAura, so you need to pick a high-level aura that will be detected as a drinking/eating one.

src/strategy/actions/non-combat/drink/definition/enum/DrinkAuraEnum.h
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU AGPL v3 license, you may redistribute it
+ * and/or modify it under version 3 of the License, or (at your option), any later version.
+ */
+
+#ifndef _PLAYERBOT_DRINK_AURA_ENUM_H
+#define _PLAYERBOT_DRINK_AURA_ENUM_H
+
+enum DrinkAuraEnum {
```
privatecore, line 9: Not used enum. Also, most of the defined auras here are not drinking/eating auras.

src/utility/isEatingOrDrinking.h
```diff
@@ -0,0 +1,28 @@
...
+    bool drinking =
+        player->HasAuraType(SPELL_AURA_MOD_POWER_REGEN) ||
+        player->HasAuraType(SPELL_AURA_MOD_POWER_REGEN_PERCENT);
+
+    // Eating = health regen aura
+    bool eating =
+        player->HasAuraType(SPELL_AURA_MOD_REGEN) ||
+        player->HasAuraType(SPELL_AURA_MOD_HEALTH_REGEN_PERCENT);
+
+    return drinking || eating;
```
privatecore, line 25:
```cpp
return player->HasAuraType(SPELL_AURA_MOD_POWER_REGEN) || player->HasAuraType(SPELL_AURA_MOD_REGEN);
```

Drinking and eating auras only have `SPELL_AURA_MOD_POWER_REGEN` and `SPELL_AURA_MOD_REGEN` EffectAura values. Check the DBC files if in doubt.

privatecore:
Not sure why you are trying to refactor the code and create separate files/paths for each utility method. This question should be addressed to the maintainers: @hermensbas, @kadeshar, @nl-saw. In my opinion, it’s better to add these methods to the drinking action file rather than create separate utility files.

src/PlayerbotAI.cpp
SmashingQuasar: That's entirely wrong.
If you remove this code, the bots will drink for one second and immediately get out of drinking state to initiate another action. This has also been explained in the PR description.

src/PlayerbotAI.cpp
SmashingQuasar: It's a draft PR, which means it's... a draft.¯\_(ツ)_/¯ 
Also not used, so removed by my IDE automatically.

src/strategy/actions/ActionContext.h
SmashingQuasar: It's a draft PR, which means it's... a draft.¯\_(ツ)_/¯ 
Also not used, so removed by my IDE automatically.

src/strategy/actions/non-combat/drink/DrinkAction.cpp
SmashingQuasar: Used for logging on my end. It's a draft.

src/strategy/actions/non-combat/drink/DrinkAction.cpp
SmashingQuasar: 25990 was already being used. I did not change this behaviour whatsoever. It's the [Graccu's mince meat aura](https://www.wowhead.com/spell=25990/graccus-mince-meat-fruitcake) which was picked originally because it restores percentage based health and mana.

hermensbas:
I am waiting to review till the PR is tagged to be ready for review.  

@privatecore are you discord by any chance?

src/strategy/actions/non-combat/drink/definition/enum/DrinkAuraEnum.h
SmashingQuasar: Literally used in the code you commented just before. :)

src/utility/isEatingOrDrinking.h
SmashingQuasar: It's a code suggested by @hermensbas but you make an interesting point so I will have a look. I assume Stoneharry's could give some insight about those auras and probably Keira too.

Wishmaster117:
> Not sure why you are trying to refactor the code and create separate files/paths for each utility method. This question should be addressed to the maintainers: @hermensbas, @kadeshar, @nl-saw. In my opinion, it’s better to add these methods to the drinking action file rather than create separate utility files.

I think the idea (and it's not bad) is to gradually refactor the module.

SmashingQuasar:
> Not sure why you are trying to refactor the code and create separate files/paths for each utility method. This question should be addressed to the maintainers: @hermensbas, @kadeshar, @nl-saw. In my opinion, it’s better to add these methods to the drinking action file rather than create separate utility files.

We had a discussion with @hermensbas about the need for refactoring. We both agreed that there is a need for structure and that it is important to get out of those god-files. The function in question is also inlined and headers only to optimise it for performance. Due to how C++ handles header files and inline functions, it's preferable to not put them in god files to avoid build time bubbling.
It also cannot be put in the `DrinkAction`, because the `DrinkAction` should not be manually invoked within `PlayerbotAI::DoNextAction` where we need to make the check. Originally I did add it to `DrinkAction` but it turned out to be a poor choice as it would require me to manually build an `Event` instance to pass to `DrinkAction`.

noisiver:
I agree with @SmashingQuasar - we are in desperate need of a refactor. The fact that I'm a maintainer doesn't really change anything because we shouldn't *always* have final word on things if it benefits the project and/or is a much needed/requested change. We are here to help maintain it, keep it alive and look after it, but all the changes are done by volunteers and we have to take the opinion of others into mind when making decisions.

There are of course situations where a change may be rejected regardless of the demand. It could be one of these, but it's impossible for me to write every possible reason that could come up so the list would probably be quite long.

- The code is poorly optimized or riddled with bugs
- The code hasn't been tested thoroughly
- The changes affects other areas in unexpected (and usually negative) ways
- A change is introduced to change the experience and is enforced so even users who don't want it are affected. An example is introducing bot chatter without allowing users to disable it.

Wishmaster117:
> I originally lowered it to 500ms and was quite happy with it so we could also opt for this.

You could perhaps put this value in a variable so that everyone can adjust it to their liking and (I admit I haven't looked at the config file because I'm on my phone) add an in-game command to completely disable and re-enable the auto-drink function, so that in case of a raid with a master, the bots will only drink when the master tells them to?

src/PlayerbotAI.cpp
privatecore: > If you remove this code, the bots will drink for one second and immediately get out of drinking state to initiate another action. This has also been explained in the PR description.

Do you know why this happens?

src/PlayerbotAI.cpp
privatecore: You did the same thing in your other PR. So, what's the point of these changes?

src/strategy/actions/non-combat/drink/definition/enum/DrinkAuraEnum.h
privatecore: So, did I get it right -- you created a whole enum with irrelevant values just to use one of them, right?
Also:
> 25990 does not have the proper EffectAura

privatecore:
@noisiver Everyone agrees that the current codebase leaves much to be desired. However, any changes, including refactoring, need to be discussed beforehand. It's essential to create a clear action plan and determine the order of execution. It's important to address aspects such as codestyles, project structure, etc.

I believe that the initiator of such changes and the development of the plan should not be handled by one person alone. The initiative should come from a well-motivated specialist, preferably one of the maintainers of the project. They are, after all, [the original owners of the organization](https://github.com/mod-playerbots/mod-playerbots/discussions/1724).

SmashingQuasar:
> @noisiver Everyone agrees that the current codebase leaves much to be desired. However, any changes, including refactoring, need to be discussed beforehand. It's essential to create a clear action plan and determine the order of execution. It's important to address aspects such as codestyles, project structure, etc.
> 
> I believe that the initiator of such changes and the development of the plan should not be handled by one person alone. The initiative should come from a well-motivated specialist, preferably one of the maintainers of the project. They are, after all, [the original owners of the organization](https://github.com/mod-playerbots/mod-playerbots/discussions/1724).

These changes were discussed heavily on Discord. You did not take part of this discussion since you don't seem to be on Discord. I am quite baffled by this response since you jumped on a **draft** PR you had no information about to make a review on something that had already been discussed with maintainers on Discord repeatedly and got concept approval. I don't jump on changes before discussing them with contributors and administrators on Discord first.

src/PlayerbotAI.cpp
SmashingQuasar: Yes, this happens because actions are called using a priority order and rely on their `isUseful` method to identifiy if they should proceed. So if you want to not put 3 lines of code in `PlayerbotAI::DoNextAction`, you will have to go through every single action that needs not to be called when a bot is drinking to make the check to see if the bot is drinking to avoid using it.

src/PlayerbotAI.cpp
SmashingQuasar: Let me get this straight.
- You complain about unused code.
- You complain when people remove unused code.

Again, it's a draft PR and my IDE is configured to remove unused calls. It's not even something that was done deliberately. Also, including unused headers just bubbles build time for no reason. What's the point of even calling this out?

src/strategy/actions/non-combat/drink/definition/enum/DrinkAuraEnum.h
SmashingQuasar: No, I used those values originally, and did not remove them as I am still satisfied with the structure of this PR. You were still wrong, and you still did not read, tested nor understood this code.
Still, 25990 is the proper effect aura as linked before, it was already using this aura before and this has not been changed. This is a non-behavioural change and I linked you the aura in question.

SmashingQuasar:
> > I originally lowered it to 500ms and was quite happy with it so we could also opt for this.
> 
> You could perhaps put this value in a variable so that everyone can adjust it to their liking and (I admit I haven't looked at the config file because I'm on my phone) add an in-game command to completely disable and re-enable the auto-drink function, so that in case of a raid with a master, the bots will only drink when the master tells them to?

Yes it's a good idea. We can add configuration variables on this.
As I mentioned, I am not super satisfied with the structure of the code. It is out of question to cram more methods in `PlayerbotAI` and as I explained we can't put `isDrinking` in `DrinkAction`. So I just put it outside for now but I would prefer if it was tied to a service at least to avoid having stray functions everywhere.

src/PlayerbotAI.cpp
privatecore: > So if you want to not put 3 lines of code in `PlayerbotAI::DoNextAction`, you will have to go through every single action that needs not to be called when a bot is drinking to make the check to see if the bot is drinking to avoid using it.

Any chance this can be handled in another way? Let's say, by interrupting processing actions if the previous action has a higher priority and an interrupt flag?

src/strategy/actions/non-combat/drink/definition/enum/DrinkAuraEnum.h
privatecore: > No, I used those values originally, and did not remove them as I am still satisfied with the structure of this PR. You were still wrong, and you still did not read, tested nor understood this code.
> Still, 25990 is the proper effect aura as linked before, it was already using this aura before and this has not been changed. This is a non-behavioural change and I linked you the aura in question.

image

SPELL_AURA_OBS_MOD_HEALTH (20)
SPELL_AURA_OBS_MOD_POWER (21)

Your own code checks against different aura types. But, I'm "wrong", okay :smile:

src/PlayerbotAI.cpp
privatecore: > What's the point of even calling this out?

Ultimately, these changes are not related to the pull request, which is why the question was asked -- what is the purpose of these changes being here? If your IDE automatically removes code, is there a chance that you will eventually revert it and handle such changes yourself, documenting the process?

nl-saw:
Just my 2 cents; I think that in general most of the above discussions are best suited for discussion on Discord. This is after all, still a draft. However, it would also be best to discuss scope/approach of the changes, prior of making a PR, so that a debate i.e. regarding helper files can be tackled before even working on it, to both save post-discussions and ultimately everyone's time.

Wishmaster117:
Okay guys, this is going to turn into an endless discussion. @SmashingQuasar has proposed two pull requests (PRs) with a code refactoring.

These PRs are drafts (which also implies proposals).
So, he's telling us not to merge this thing, please. He's saying, "Hey, I have an idea to refactor the module; here's what I propose."
So, for me, it's not a bad thing to show something new.

Afterward, I agree that it need to be discussed on Discord and make a project plan, etc.
So, the best thing is for you to schedule a Discord meeting with the maintainers and @SmashingQuasar to lay everything out on the table.

This will avoid pointless discussions.

This is just my point of view; I haven't followed the thread on Discord, but we all agree that the module needs a major refactoring, which could be done in several phases. For the moment we are "just putting pieces of tape on something that is already fragile".

SmashingQuasar:
> Just my 2 cents; I think that in general most of the above discussions are best suited for discussion on Discord. This is after all, still a draft. However, it would also be best to discuss scope/approach of the changes, prior of making a PR, so that a debate i.e. regarding helper files can be tackled before even working on it, to both save post-discussions and ultimately everyone's time.

It was discussed on discord prior to opening this PR. I even showed code snippets, shared the logic, the pitfalls, I did multiple iterations. It's just that several people do not follow the discord discussions.

I agree with @Wishmaster117 . We can lay out a global plan. This is something that I believe @hermensbas is pretty interested in doing so we can have a clear view on what is the target architecture. As you perfectly said Wishmaster, I am just making a suggestion on a possible structure basis. We can opt for something completely different, it's alright. I just want it to be clear and understandable and not have all the files are `src/` root level with thousands and thousands of lines in them.

I can make this change by sticking strictly to the original design, but it's just going to increase the size of `PlayerbotAI` again.
In any case, this PR is still under development and testing so we can totally flip the architecture.

nl-saw:
> These PRs are drafts (which also implies proposals).

Point being there's potential that hours of work could be in vain - that's all. Up to you how you want to present proposals.

notOrrytrout:
I don’t drink anymore,… Of course I don’t drink any less either

PR #1861, opened 2025-11-19
Create IsMeleeDps Method and Combine Existing Role Checks

brighton-chi:
I don't know why we don't have this already.

### DIFF
--- a/src/strategy/raids/obsidiansanctum/RaidOsTriggers.cpp
+++ b/src/strategy/raids/obsidiansanctum/RaidOsTriggers.cpp
@@ -64,7 +64,7 @@ bool SartharionDpsTrigger::IsActive()
 
 bool SartharionMeleePositioningTrigger::IsActive()
 {
-    if (botAI->IsMeleeDps(bot)) { return false; }
+    if (!botAI->IsMeleeDps(bot)) { return false; }
 
     Unit* boss = AI_VALUE2(Unit*, "find target", "sartharion");
     if (!boss) { return false; }


nl-saw:
We already have
```cpp
bool PlayerbotAI::IsMelee(Player* player, bool bySpec) { return !IsRanged(player, bySpec); }
```
That not sufficient? All other code is using that already, so adding IsMeleeDps seems quite redundant.

brighton-chi:
> We already have
> 
> ```cpp
> 
> bool PlayerbotAI::IsMelee(Player* player, bool bySpec) { return !IsRanged(player, bySpec); }
> 
> ```
> 
> That not sufficient? All other code is using that already, so adding IsMeleeDps seems quite redundant.

That includes tanks.

Edit: Although it's probably worthwhile to change all instances of IsMelee && IsDps to IsMeleeDps. I can do that as part of this.

nl-saw:
> That includes tanks.
> Edit: Although it's probably worthwhile to change all instances of IsMelee && IsDps to IsMeleeDps. I can do that as part of this.

Yep it does include tanks. I'm not sure this change is worth it, unless something related is broken that could only be fixed with this proposal. Reason being that the logic behind the IsMelee and IsMeleeDps are slightly different, so there is not an easy way of saying which ones you can simply replace as all the logic around it is based on how currently IsDps functions. Might be better to focus on other things for now.

brighton-chi:
> > That includes tanks.
> 
> > Edit: Although it's probably worthwhile to change all instances of IsMelee && IsDps to IsMeleeDps. I can do that as part of this.
> 
> 
> 
> Yep it does include tanks. I'm not sure this change is worth it, unless something related is broken that could only be fixed with this proposal. Reason being that the logic behind the IsMelee and IsMeleeDps are slightly different, so there is not an easy way of saying which ones you can simply replace as all the logic around it is based on how currently IsDps functions. Might be better to focus on other things for now.
> 
> 

I'm saying there are usages of "IsMelee && IsDps" in the code that could be replaced. If something currently is just IsMelee alone, we would not replace it. I wouldn't want to make any substantive changes.

Ultimately, it seems silly to me to have IsRangedDps yet not have IsMeleeDps. IsMeleeDps of course would be a very basic function but makes usages in boss strategies more readable as I am frequently needing to apply methods to melee dps only.

nl-saw:
Okay, I misunderstood your IsMelee && IsDps part thinking you wanted to replace both of those. I agree with that it would look nicer to have such alignment, so if there are no functional changes then sure, I suppose.

src/strategy/raids/obsidiansanctum/RaidOsTriggers.cpp
```diff
@@ -64,7 +64,7 @@ bool SartharionDpsTrigger::IsActive()
 
 bool SartharionMeleePositioningTrigger::IsActive()
 {
-    if (!botAI->IsMelee(bot) || !botAI->IsDps(bot)) { return false; }
+    if (botAI->IsMeleeDps(bot)) { return false; }
```
nl-saw: Should be ! botAI->IsMeleeDps(bot)) and also does not take ranged dps, which it did before.

brighton-chi: Thanks, changed to not. I think !botAI->IsMeleeDps(bot) is by itself the same as before, though, and ranged DPS were also previously excluded. THe previous check returned false if not melee OR not DPS; since ranged DPS is not melee, they would have been excluded.

nl-saw: Your IsMeleeDps
```cpp
bool PlayerbotAI::IsMeleeDps(Player* player, bool bySpec) { return IsMelee(player, bySpec) && IsDps(player, bySpec); }
```
has && operator, excluding ranged dps, whereas the original check was either IsMelee OR IsDps, the latter including ranged DPS. Hence your change seems to no longer parse ranged dps.

brighton-chi: I don't think that is right.

New: if (!botAI->IsMeleeDps(bot)) 
New meaning: if (!(botAI->IsMelee(bot) && botAI->IsDps(bot))
Old (same meaning): if (!botAI->IsMelee(bot) || !botAI->IsDps(bot))

nl-saw: original:
    if (!botAI->IsMelee(bot) || !botAI->IsDps(bot)) { return false; }
vs
    if (!botAI->IsMeleeDps(bot)) { return false; }

meaning original: If no melee **or** if no ranged (IsDps) - return false
And new: if no melee - return false

That clearly is a difference.

nl-saw: I'm bored talking about this, I won't approve this. This is what I meant with this change not being worth it.
Maybe someone else will.

kadeshar:
For me method have sense when is implemented itself not using existing IsMelee and IsDps. I know there is IsRangedDps but that also feel not right for me. 2 times is making same operations (getting botAI, checking bySpec, checking strategy or class/spec)

brighton-chi:
> For me method have sense when is implemented itself not using existing IsMelee and IsDps. I know there is IsRangedDps but that also feel not right for me. 2 times is making same operations (getting botAI, checking bySpec, checking strategy or class/spec)

So you want to do that with both IsRangedDps and IsMeleeDps? There is no additional burden by creating the method, there is just no creating a better performance outcome from it either. It's purely visual. 

It makes zero sense to have one but not the other. So either actual work is done to have more efficient versions of both IsMeleeDps and IsRangedDps, or you have them both exist but be only reading convenience, or just neither exists at all. Whatever is done is not my call, but I think preserving the status quo is nonsensical.

kadeshar:
Removing operation duplication dont creating better performance? Explain please.

brighton-chi:
> Removing operation duplication dont creating better performance? Explain please.

Because there is already duplication. Right now you would do IsMelee && IsDps separately in the boss strategy if you wanted to get only Melee Dps.

I'm not saying it wouldn't be better to get rid of duplication. I'm just saying that is already existing.

src/strategy/raids/obsidiansanctum/RaidOsTriggers.cpp
noisiver: I think it makes sense to keep this specific file the same until it can be determined if it's wrong or not. The checks it runs sound off based on the name of the class but that is not what this PR is about. If it turns out to be wrong it can be addressed in another PR but this one should focus on replacing variations of IsMelee and IsDps with IsMeleeDps

brighton-chi:
Ok, I took a look at the code and think it's simple to make IsRangedDps and IsMeleeDps be direct lookups to class/spec. So there should be a refactoring. Hopefully I can get around to that, so I'll close this in favor of a future PR.

PR #1864, opened 2025-11-21
Fix - Druid remove bear form when out of combat

Wishmaster117:
This will remove bear form of tank druids when out of combat, the druid doesn't stay stuck anymore in bear form during all the raid or dungeon.

### DIFF
--- a/src/PlayerbotMgr.cpp
+++ b/src/PlayerbotMgr.cpp
@@ -1602,8 +1602,26 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
 
 void PlayerbotMgr::OnPlayerLogin(Player* player)
 {
+    if (!player)
+        return;
+
+    WorldSession* session = player->GetSession();
+    if (!session)
+    {
+        LOG_WARN("playerbots", "Unable to register locale priority for player {} because the session is missing", player->GetName());
+        return;
+    }
+
+    // DB locale (source of bot text translation)
+    LocaleConstant const databaseLocale = session->GetSessionDbLocaleIndex();
+
+    // For bot texts (DB-driven), prefer the database locale with a safe fallback.
+    LocaleConstant usedLocale = databaseLocale;
+    if (usedLocale >= MAX_LOCALES)
+        usedLocale = LOCALE_enUS; // fallback
+
     // set locale priority for bot texts
-    sPlayerbotTextMgr->AddLocalePriority(player->GetSession()->GetSessionDbcLocale());
+    sPlayerbotTextMgr->AddLocalePriority(usedLocale);
 
     if (sPlayerbotAIConfig->selfBotLevel > 2)
         HandlePlayerbotCommand("self", player);
@@ -1611,7 +1629,7 @@ void PlayerbotMgr::OnPlayerLogin(Player* player)
     if (!sPlayerbotAIConfig->botAutologin)
         return;
 
-    uint32 accountId = player->GetSession()->GetAccountId();
+    uint32 accountId = session->GetAccountId();
     QueryResult results = CharacterDatabase.Query("SELECT name FROM characters WHERE account = {}", accountId);
     if (results)
     {
--- a/src/PlayerbotTextMgr.cpp
+++ b/src/PlayerbotTextMgr.cpp
@@ -190,26 +190,29 @@ bool PlayerbotTextMgr::GetBotText(std::string name, std::string& text, std::map<
 
 void PlayerbotTextMgr::AddLocalePriority(uint32 locale)
 {
-    if (!locale)
+    if (locale >= MAX_LOCALES)
+    {
+        LOG_WARN("playerbots", "Ignoring locale {} for bot texts because it exceeds MAX_LOCALES ({})", locale, MAX_LOCALES - 1);
         return;
+    }
 
     botTextLocalePriority[locale]++;
 }
 
 uint32 PlayerbotTextMgr::GetLocalePriority()
 {
-    uint32 topLocale = 0;
-
     // if no real players online, reset top locale
-    if (!sWorldSessionMgr->GetActiveSessionCount())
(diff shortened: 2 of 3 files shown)


src/strategy/triggers/GenericTriggers.cpp
```diff
@@ -192,6 +192,31 @@ bool NoAttackersTrigger::IsActive()
...
+bool NoMoreAttackersTrigger::IsActive()
+{
+    // Trigger is active only when the bot is fully out of combat,
+    // has no attackers left and no valid hostile target to attack.
+
+    // If the bot (or its nearby group) is still in combat, stay in bear.
+    if (AI_VALUE2(bool, "combat", "self target"))
+        return false;
+
+    // If somebody is still hitting the bot, stay in bear.
```
kadeshar, line 204: Its generic trigger. We shouldnt mention which action its affect.

src/strategy/triggers/GenericTriggers.cpp
```diff
@@ -192,6 +192,31 @@ bool NoAttackersTrigger::IsActive()
     return !AI_VALUE(Unit*, "current target") && AI_VALUE(uint8, "my attacker count") > 0;
 }
 
+bool NoMoreAttackersTrigger::IsActive()
+{
+    // Trigger is active only when the bot is fully out of combat,
+    // has no attackers left and no valid hostile target to attack.
+
+    // If the bot (or its nearby group) is still in combat, stay in bear.
```
kadeshar, line 200: Its generic trigger. We shouldnt mention which action its affected

src/strategy/triggers/GenericTriggers.cpp
```diff
@@ -192,6 +192,31 @@ bool NoAttackersTrigger::IsActive()
...
+    if (AI_VALUE2(bool, "combat", "self target"))
+        return false;
+
+    // If somebody is still hitting the bot, stay in bear.
+    if (AI_VALUE(uint8, "my attacker count") != 0)
+        return false;
+
+    Player* bot = botAI->GetBot();
```
kadeshar, line 208: My algorithms detected code which was commented many times.
Loading program "Skynet"

src/strategy/triggers/GenericTriggers.cpp
```diff
@@ -192,6 +192,31 @@ bool NoAttackersTrigger::IsActive()
     return !AI_VALUE(Unit*, "current target") && AI_VALUE(uint8, "my attacker count") > 0;
 }
 
+bool NoMoreAttackersTrigger::IsActive()
```
kadeshar, line 195: Im feel that this trigger is redundant. Already exists NoAttackersTrigger which is similar.

src/strategy/druid/BearTankDruidStrategy.cpp
```diff
@@ -173,4 +173,8 @@ void BearTankDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
     triggers.push_back(
         new TriggerNode("bash on enemy healer",
                         NextAction::array(0, new NextAction("bash on enemy healer", ACTION_INTERRUPT + 1), nullptr)));
+    // Out of combat: if there are no attackers anymore, try to return to caster form
+    triggers.push_back(
+        new TriggerNode("no more attackers",
+                        NextAction::array(0, new NextAction("caster form", ACTION_HIGH + 8), nullptr)));
```
kadeshar, line 179: I dont know its good idea. Thats mean that tank will pull target without bear form? Exiting this form have some adventages.

src/strategy/druid/GenericDruidNonCombatStrategy.cpp
```diff
@@ -122,6 +122,11 @@ void GenericDruidNonCombatStrategy::InitTriggers(std::vector<TriggerNode*>& trig
 
     triggers.push_back(new TriggerNode("often", NextAction::array(0, new NextAction("apply oil", 1.0f), nullptr)));
 
+    // Out of combat: regularly force a return to caster form if shapeshifted
+    triggers.push_back(
+        new TriggerNode("often",
+                        NextAction::array(0, new NextAction("caster form", ACTION_HIGH + 9), nullptr)));
```
kadeshar, line 128: Every spec of druid should go into caster form?

Wishmaster117:
I'm on my phone i answer you tomorrow Kadesharminator

Vortikai:
I think it'd be cool to add a 1-8 second random wait time after the last attacker is dead so it wouldn't feel like a bot, and emulate more human behavior. That's my suggestion but take from it what you will

noisiver:
I don't think this is a good idea. There are situations where them being in a form is bad, like telling them to use hearthstone stops that from happening because they don't leave their form to do it but they shouldn't get out of bear/cat form whenever they're not in combat. That's not realistic. When I play tank as a druid I am in bear form at all times

wetbrownsauce:
In all my years of raiding I have never seen a bear tank out of bear form unless they're riding a mount or dead.

Also leaving bear form dumps all your pooled rage so IF this is going to go through, they shouldn't do it in dungeon environments. Or pvp. Honestly why would you want this?

Vortikai:
> I don't think this is a good idea. There are situations where them being in a form is bad, like telling them to use hearthstone stops that from happening because they don't leave their form to do it but they shouldn't get out of bear/cat form whenever they're not in combat. That's not realistic. When I play tank as a druid I am in bear form at all times

Same. If we go forward with it I'd like it an optional strategy

Wishmaster117:
Okay man, no problem, the majority wins, we won't do it. And what about the Tree of Life form?

Vortikai:
Tree of Life is 13% of base mana each time one enters and leaves combat going to waste. It's not a huge dent but still cancels out a little bit of its benefits if performed often.

I'm not opposed to developing a strategy for this but the player should be the one to decide if they wanna is all

PR #1865, opened 2025-11-21
Fix - Druid Remove Tree of Life Form when out of combat

Wishmaster117:
This will remove Tree of Life form to druid where he's out of combat.

### DIFF
--- a/src/strategy/druid/GenericDruidNonCombatStrategy.cpp
+++ b/src/strategy/druid/GenericDruidNonCombatStrategy.cpp
@@ -126,7 +126,7 @@ void GenericDruidNonCombatStrategy::InitTriggers(std::vector<TriggerNode*>& trig
     triggers.push_back(
         new TriggerNode("no more attackers",
                         NextAction::array(0, new NextAction("cancel tree form", ACTION_HIGH + 8), nullptr)));
-						
+
     triggers.push_back(
         new TriggerNode("party member critical health",
                         NextAction::array(0,


src/PlayerbotAI.cpp
```diff
@@ -4604,7 +4604,7 @@ void PlayerbotAI::RemoveShapeshift()
     RemoveAura("swift flight form");
     RemoveAura("aquatic form");
     RemoveAura("ghost wolf");
-    // RemoveAura("tree of life");
```
nl-saw, line 4607: This was disabled here for a specific reason: https://github.com/Wishmaster117/mod-playerbots/commit/d45a7dc5b838022d0078008324cfb731f86ece07

Have you checked this won't re-introduce that problem?

Wishmaster117: > This was disabled here for a specific reason: [Wishmaster117@d45a7dc](https://github.com/Wishmaster117/mod-playerbots/commit/d45a7dc5b838022d0078008324cfb731f86ece07)
> 
> Have you checked this won't re-introduce that problem?

If you could help me by telling me why this was removed, I didn't see any abnormal behavior during my tests, but anyway, like any pull request, it needs to be tested on a larger scale than my small local server.

What was the problem it introduced?

Why does the rest of the code that refers to it still exist?

If I can understand the problem it causes, I can either try to fix it or do it another way; right now I've taken the simplest approach.
There's so much commented code in this module that's no longer used that I don't even know what it does anymore.

nl-saw: Unfortunately, all the info I got atm is the commit and just that description: fix party member dispel.
Perhaps @liyunfan1223 can shed some light on this, yet try testing stuff involving dispelling I suppose.

Wishmaster117: Okay, thanks.
I'll look at the code for the occurrences and what they imply.
So far, from what I've tested, I haven't seen anything unusual happen, which is why I don't understand why it was commented out, but I'll investigate further.

Wishmaster117: Okay i take a look at the RemoveAura("tree of life").
Uncommenting it don' break something in terms of crashes, but, we revert the module design, the druid may leave Tree of Life more often than intended, we bypass part of the logic meant to make resto druids behave like real players (stay in Tree of Life while healing).
So is better to let it commented, have to find another way.

Wishmaster117: I'm so stupid, I'd already done 3/4 of the work with the other PR, I just needed to use that work lol

Wishmaster117:
This PR have to be merged after this one:
https://github.com/mod-playerbots/mod-playerbots/pull/1864

src/PlayerbotAI.cpp
liyunfan1223: The difference is that this will cause the restro druid to remove tree of life form before dispel, but this is not necessary (dispel can be cast while in tree form). It's not a big deal, clarifying to make it clearer.

Wishmaster117: @liyunfan1223, it's great to see you again! We miss your expertise and in-depth knowledge of this module :)
Regarding this discussion, I've taken a different approach.

Issue #1866, opened 2025-11-21
Random bots leave RDF group before player loads in


digspp:
**Bug Description**
After queuing into RDF, some random bots (usually all, but not always) drop out of the group before loading in the dungeon.
Tried queuing for different difficulties (normal level 80, heroic level 80, random classic in different level ranges) and the behavior is observed across all scenarios.
No custom modifications were made to the server other than the installed mods.
Behavior was working normally before recent updates (not sure when it started, would guess around mid october, maybe even early november).
Server is running with 3000 random bots.

**Commit Hash**
AzerothCore rev. 3732047b9e25 2025-11-17 21:02:36 +0100 (Playerbot branch) (Unix, RelWithDebInfo, Static)

**How To Reproduce Bug**
Simply queue into RDF solo.

**Expected Behavior**
Expected behavior is what was observed in the past, where all random bots would stay in the group and not leave before player loads in.

**Screenshots**
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/841f8d21-bde5-461a-bcda-ccfd54f86f60" />

**Modules**
mod-TimeIsTime
mod-account-achievements
mod-account-mounts
mod-ah-bot
mod-junk-to-gold
mod-learn-spells
mod-no-hearthstone-cooldown
mod-player-bot-level-brackets
mod-playerbots
mod-pvp-titles
mod-skip-dk-starting-area
mod-transmog

**Playerbot Settings**
* 3000 random bots
* AiPlayerbot.BotActiveAlone = 10
* AiPlayerbot.botActiveAloneSmartScale = 1

**System**
OS: Ubuntu Server 24.04
Hardware: 32GB RAM, 4-core linux VM

**Additional Context**
Screenshot may not be useful, just shows that I see no "has left the party" message, only that bots joined the group, while I join an empty dungeon.
Behavior is consistent across server restarts and database wipe.
Please let me know if I can provide additional info on the issue.

yqsy2025:
The problem lies in the function PlayerbotAI::UpdateAIGroupAndMaster(),this is my code:void PlayerbotAI::UpdateAIGroupAndMaster()
{
    if (!bot)
        return;

    PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
    if (!botAI)
        return;

    Group* group = bot->GetGroup();
    bool IsRandomBot = sRandomPlayerbotMgr->IsRandomBot(bot);

    // 检查是否为随机本队伍
[...]

PR #1876, opened 2025-11-25
Bug: Fix bots leaving LFG groups before master

Celandriel:
I removed bots checking if they should leave group every tick, and will rely on the LeaveGroupFarAway action. I also increased the timer from 5 seconds to 20 seconds. No need to check this that often.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -365,7 +365,7 @@ void PlayerbotAI::UpdateAI(uint32 elapsed, bool minimal)
     }
 
     // Update the bot's group status (moved to helper function)
-    UpdateAIGroupAndMaster();
+    UpdateAIGroupMaster();
 
     // Update internal AI
     UpdateAIInternal(elapsed, minimal);
@@ -373,7 +373,7 @@ void PlayerbotAI::UpdateAI(uint32 elapsed, bool minimal)
 }
 
 // Helper function for UpdateAI to check group membership and handle removal if necessary
-void PlayerbotAI::UpdateAIGroupAndMaster()
+void PlayerbotAI::UpdateAIGroupMaster()
 {
     if (!bot)
         return;
@@ -431,8 +431,6 @@ void PlayerbotAI::UpdateAIGroupAndMaster()
                 botAI->ChangeStrategy("-follow", BOT_STATE_NON_COMBAT);
             }
         }
-        else if (!newMaster && !bot->InBattleground())
-            LeaveOrDisbandGroup();
     }
 }
 
--- a/src/PlayerbotAI.h
+++ b/src/PlayerbotAI.h
@@ -612,7 +612,7 @@ private:
     static void _fillGearScoreData(Player* player, Item* item, std::vector<uint32>* gearScore, uint32& twoHandScore,
                                    bool mixed = false);
     bool IsTellAllowed(PlayerbotSecurityLevel securityLevel = PLAYERBOT_SECURITY_ALLOW_ALL);
-    void UpdateAIGroupAndMaster();
+    void UpdateAIGroupMaster();
     Item* FindItemInInventory(std::function<bool(ItemTemplate const*)> checkItem) const;
     void HandleCommands();
     void HandleCommand(uint32 type, const std::string& text, Player& fromPlayer, const uint32 lang = LANG_UNIVERSAL);
--- a/src/strategy/generic/GroupStrategy.cpp
+++ b/src/strategy/generic/GroupStrategy.cpp
@@ -11,7 +11,6 @@ void GroupStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
     triggers.push_back(new TriggerNode("often", NextAction::array(0, new NextAction("invite nearby", 4.0f), nullptr)));
     triggers.push_back(new TriggerNode("random", NextAction::array(0, new NextAction("invite guild", 4.0f), nullptr)));
-    triggers.push_back(new TriggerNode("often", NextAction::array(0, new NextAction("leave far away", 4.0f), nullptr)));
-    triggers.push_back(
-        new TriggerNode("seldom", NextAction::array(0, new NextAction("reset instances", 1.0f), nullptr)));
+    triggers.push_back(new TriggerNode("random", NextAction::array(0, new NextAction("leave far away", 4.0f), nullptr)));
+    triggers.push_back(new TriggerNode("seldom", NextAction::array(0, new NextAction("reset instances", 1.0f), nullptr)));
 }


Celandriel:
Fixes #1866

yqsy2025:
If modified this way, when a real player leaves the group midway through a random dungeon, the bots will remain in the instance and will not automatically leave the group.

Celandriel:
Wouldnt that also happen with your proposed? If you always exclude LFG groups from leaving you would see the same effect.

Celandriel:
Converted to draft. Needs more work

Celandriel:
I've gone through the code and don't understand why there is a behavior difference between the two after ce51191e8f88a8ddf8e3601427f4bbb72189f62e. 
In the case where the bot has no master, it needs to iterate through players in the group to find a real player. If none is found then it should leave the group. My read of the code is identical in both cases.  (if not help me see why) 
Before a player arrives in to the lfg group the bots are supposed to leave based on code before this PR. 
 The solution is to make it so leaving group is not evaluated every tick, but on a different timer as this PR suggests to do. The only issue is that they aren't leaving on their own. Need to review why the trigger isn't running.

src/PlayerbotAI.cpp
```diff
@@ -431,8 +431,6 @@ void PlayerbotAI::UpdateAIGroupAndMaster()
                 botAI->ChangeStrategy("-follow", BOT_STATE_NON_COMBAT);
             }
         }
-        else if (!newMaster && !bot->InBattleground())
-            LeaveOrDisbandGroup();
```
Wishmaster117, line 435: Hi,
Maybe a middle-ground approach here would be to put back the LeaveOrDisbandGroup() logic for “normal” groups, but explicitly skip it for LFG groups instead of removing it entirely from UpdateAIGroupMaster().
The idea would be keep the existing behavior for non-LFG groups (bots auto-leave when there is no real master anymore), but for LFG groups, do not call LeaveOrDisbandGroup() from the generic group-maintenance code and let LFG / “leave far away” handle cleanup
Something along these lines inside PlayerbotAI::UpdateAIGroupMaster() (right where you currently had the else if (!newMaster && !bot->InBattleground())), somethink like :

```
else if (!newMaster && !bot->InBattleground())
{
    if (Group* group = bot->GetGroup())
    {
        // Do not auto-leave LFG groups here – LFG is supposed to drive their lifecycle.
        if (!group->isLFGGroup())
            LeaveOrDisbandGroup();
    }
    else
    {
        // No group at all, keep the old behavior.
        LeaveOrDisbandGroup();
    }
}
```
I guess this way the original bug bots dropping the LFG group before the master loads in should be avoided, because LFG groups are no longer touched by the generic “no master -> leave” path, and the concern mentioned above (bots staying forever in a “normal” dungeon/quest group if the last player leaves midway) should still be covered, because LeaveOrDisbandGroup() is still called for non-LFG groups.

Celandriel: Yea that's the way that yqsq proposed to solve this issue.
It could work, but right now I want to figure out why they just freeze after the player leaves.

Wishmaster117: Looking at the current UpdateAIGroupMaster() you posted, the “freeze after the player leaves” actually makes sense to me.
After the RDF player leaves the dungeon  group, the bots usually stay in a group together.
In that situation, Group* group = bot->GetGroup(); is still not null (bots are still grouped).
The master pointer on the bot AI still points to the original player, even if that player left the group or went offline.
Now if we walk through UpdateAIGroupMaster() in that state:

This block is not executed, because group is not null

```
if (!group)
{
    if (master && IsRandomBot)
    {
        SetMaster(nullptr);
        Reset(true);
        ResetStrategies();
    }
    return;
}
```
So we never clear the master or reset strategies here.

The battleground-only case is also skipped, because we’re in an RDF dungeon, not a BG

```
if (bot->InBattleground() && IsRandomBot && master && !master->GetGroup())
    SetMaster(nullptr);
```
Then we get to

```
if (master)
    masterBotAI = GET_PLAYERBOT_AI(master);

if (!master || (masterBotAI && !masterBotAI->IsRealPlayer()))
{
    Player* newMaster = FindNewMaster();
    ...
}
```
master is still non-null, it still points to the original player.
For a real player, masterBotAI is either nullptr or IsRealPlayer() == true.

That means the condition (!master || (masterBotAI && !masterBotAI->IsRealPlayer())) is false, so we never go into this block either.
we never call FindNewMaster(),
we never do SetMaster(nullptr),
and we never call ResetStrategies() here.

So after the human leaves, random bots end up in this situation
They are still in a valid Group*.
Their master pointer still points to the old player.
None of the code paths that call SetMaster(nullptr), Reset(true) or ResetStrategies() ever run.

From the AI’s point of view, they are “grouped bots with a master”, but in reality there is no valid human master left and no logic that transitions them back to “random world bot” mode or makes them leave the group. As a result they just stand there in the instance and look frozen.

yqsy2025:
try this :
```

if (group && !bot->InBattleground())
    {
        if (GetMaster() && HasRealPlayerMaster()
            && (bot->IsInSameGroupWith(GetMaster()) || bot->IsInSameRaidWith(GetMaster())))
            return;

        if (GetMaster() && HasRealPlayerMaster() &&
            !bot->IsInSameGroupWith(GetMaster()) && !bot->IsInSameRaidWith(GetMaster()))
        {
            LeaveOrDisbandGroup();
            SetMaster(nullptr);
            Reset(true);
            ResetStrategies();
            return;
        }

        if (!HasRealPlayerMaster())
        {
            Player* newMaster = FindNewMaster();
            if (newMaster)
            {
                SetMaster(newMaster);
                ResetStrategies();
                botAI->ChangeStrategy("+follow", BOT_STATE_NON_COMBAT);

                if (GetMaster() && GetGroupMaster())
                {
                    if (GetMaster() == GetGroupMaster())
                    {
                            botAI->TellMaster("你好, 我会跟着你!");
                    }
                    else
                    {
                            botAI->TellMaster(!urand(0, 2) ? "老铁!" : "大佬!");
                    }
                }
                else
                {
                        botAI->ChangeStrategy("-follow", BOT_STATE_NON_COMBAT);
                }
            }
            else
            {
                LeaveOrDisbandGroup();
                botAI->ChangeStrategy("-follow", BOT_STATE_NON_COMBAT);
            }
        }
    }
```

src/PlayerbotAI.cpp
Celandriel: I guess I was hoping that the LeaveFarAwayAction would clean up the group disbanding it once no real players were left.

Wishmaster117: Yeah, that was my understanding of your intent too, but with the current checks in LeaveFarAwayAction::isUseful() it never runs once there’s no valid master anymore, so it can’t actually do that cleanup in practice.

Celandriel:
> If modified this way, when a real player leaves the group midway through a random dungeon, the bots will remain in the instance and will not automatically leave the group.

When you said this, did you mean alt bots or  random bots? I cannot exactly reproduce this with random bots. ALTHOUGH IT APPEARS THAT THEY ARE STUCK. 

I used two gm accounts. I created an LFG group with two players. 
Then the master of the group left the instance and was teleported out. I turned of GM mode and teleported back into the instance. 
Then I left on the second account from the instance. 
At this point it looked like the three bots were still in the instance, but they were 'Dummies'. If I appeared to their location with the second GM account they were out in the world doing other things and not actually in the instance. If I whispered `nc +debug` to them then the real player outside the instance began to say out loud the debug log. 

I did this again with an altbot with the same effect. 

The reason the code is supposed to work when leaving is that the bots process the server packet where the group master leaves the group by either finding a new master, or leaving the group, and this appears to happen every time. 

If you have a specific scenario that is different that does not work then please let me know.

kadeshar (approved):

PR #1888, opened 2025-11-28
Implement Serpentshrine Cavern Strategies

brighton-chi:
Edited: Below description of methods were brought up to date as of the PR coming off of draft.

### General
I've starting leveraging, to the extent possible, an out-of-combat method to erase map keys. This is mostly useful for timers that need to start upon the pull because I dislike having to rely on a check for a boss to be at 100% HP (or 99.9% or whatever) because it can be unreliable sometimes.

### Trash
Underbog Colossi: Some Colossi leave behind a lake of toxin when they die that quickly kills any player that is standing in it. The pool is a dynamic-object-generated AoE, and bots will not avoid it on their own (I think because the AoE is out of combat, plus the radius is much larger than the default avoidance radius in the config). The method does not require bots to be in combat, and simply gets bots to run out of the toxin. You will probably still get a couple of idiots who drink in the middle of it, but in my experience, the vast majority of the raid gets out, and healers that escape can easily keep up a couple of fools until they've drank to full.

Greyheart Tidecallers: Bots will mark and destroy Water Elemental Totems immediately.

### Hydross the Unstable
The strategy uses 2 tanks, with the main tank assigned to the frost phase and the 1st assistant tank assigned to the nature phase.

- The main tank will tank the frost phase, and the first assistant tank will tank the nature phase. They each have designated spots and will wait at their spots twiddling their thumbs while Hydross is in the other phase.
- Hunters will misdirect to the applicable tank upon the pull and after each phase change.
- The phase change process begins 1 second after Hydross reaches 100% Marks. The current tank will begin moving to the next phase tank's spot for the next tank to take over as soon as Hydross transitions.
- DPS is ordered to stop after Hydross reaches 100% Marks until 5 seconds after he transitions.
- Bots will prioritize the elementals adds after every phase change, unless Hydross is under 10% HP, in which case they should ignore the adds and burn the boss.
- Ranged bots should spread during the frost phase to mitigate the impact of Water Tombs.

### The Lurker Below

- There is a designated spot for the main tank.
- Ranged DPS will fan out over a 120-degree arc that is centered directly across from the tank spot (to try to spread to reduce Geyser damage while also keeping them behind Lurker).
- When Spout begins, all bots will run around behind Lurker. The intent is to keep a distance with a radius of 20 or 21 yards and within 45 degrees (either side) of directly behind him. Movement is specifically tangential along an arc so bots don't run in front of Lurker.
- Spout's duration is tracked by a timer. The mechanics of the spell itself are rather unique and don't involve a continuous cast or aura to track easily so I settled for the timer.
- If you have 3 (or more) tanks, each of the first 3 tanks will be assigned to one of the 3 Coilfang Guardians during the submerge phase.

### Leotheras the Blind
The fight is designed for a Warlock tank. You can choose the Warlock tank by giving a Warlock the Assistant flag. If you don't do that, your highest HP Warlock will be picked. Do NOT  switch the Warlock tank to a co +tank strategy--the designated Warlock is hardcoded to spam Searing Pain on Demon Leo and otherwise will engage in normal DPS strategies. If you don't have a Warlock at all, the strategy has some methods built in to try to make things work as best as possible with a melee tank.

- The Spellbinders get marked with skulls and killed in order.
- There is no designated spot or designated tank for the human phase. Your tanks will fight for aggro. Ranged bots will attempt to keep some distance, and when Whirlwind starts, everybody will run away from Leotheras.
- During the demon phase, your melee tanks should take a backseat to your Warlock tank, who will receive help in the form of Misdirection. Bots will get the hell away from the Warlock tank so the Warlock tank should be taking every Chaos Blast alone.
- During the final phase, your regular tanks will tank Leotheras, and the Warlock tank will tank his Shadow. The melee tanks will attempt to separate Leotheras from his Shadow so bots can focus down Leotheras without getting hit with Chaos Blasts.
- Bots will wait 5 seconds to DPS after every transition into human phase, 12 seconds to DPS after every transition into demon phase, and 8 seconds to DPS after the transition into the final phase. There is no waiting on DPS after Whirlwinds, even though it would be ideal. It's not a big deal to live without, and for various reasons, it would have been a pain in the ass to deal with.
- Bots will save Bloodlust/Heroism until after Spellbinders are down.
- To deal with the Inner Demons, I disabled DPS assist for bots who are targeted and force them to focus only on their Inner Demons. This is sufficient in my experience for all DPS bots and Protection Warriors and Paladins to kill their Inner Demons, even at 50% damage. Feral Tank Druids and Healers still need help, so the strategy hardcodes their actions while fighting Inner Demons. For example, Resto Druids are coded to shift out of Tree Form, cast Barkskin on themselves, and just spam Wrath until the Inner Demon is dead. There are no bot strategy changes used for this method.

### Fathom-Lord Karathress
You will need 4 tanks. Your main tank will tank Karathress, and an assistant tank will tank each Fathom Guard. If you have fewer than 4 tanks, then the priority order for tank assignment will be Karathress, Caribdis, Sharkkis, and then Tidalvess.

- Roughly, the tank spots are (1) for Karathress, near where he starts but closer to the ledge for LoS reasons, (2) for Sharkkis, North from his starting location on the other side of the ramp, (3) for Tidalvess, Northwest from his starting location near the pillar, and (4) for Caribdis, far to the West of her starting position, near the corner.
- Note that the tanks will probably clip through the terrain a bit when going to their positions. This is due to me implementing a forced MoveTo to the tank position coordinates. There is something weird about the maps in Karathress's room, and the tanks will take some really screwed up paths without making them go directly to the exact coordinates. So this looks stupid but is necessary.
- One healer will be assigned to heal the Caribdis tank. Because AC Playerbots does not yet have a focus heal strategy, this just means that such healer has a designated location near the Caribdis tank's location. This healer can be selected with the Assistant flag.
- Hunters will misdirect the Fathom Guards onto their applicable tanks. If you don't have three Hunters, the priority is Caribdis, Tidalvess, then Sharkkis.
- DPS will wait 12 seconds to begin attacking. After that, they will prioritize targets as follows:
- (1): Melee will always prioritize Spitfire Totems as soon as they spawn. This will continue through the duration of the fight.
- (2): All bots will kill Tidalvess first.
- (3): Melee bots will move to Sharkkis, and ranged bots will move to Caribdis. I understand this is not the standard kill order for players, which would have the entire raid kill Sharkkis next. The reasons I have done this differently are because melee DPS is much stronger with 3.3.5 talents vs. in retail TBC, and because bots get really thrown off by Cyclones and therefore they struggle to kill Caribdis quickly. You do not want Karathress below 75% HP before all Fathom-Guards are dead or he gets a huge damage buff.
- (4) If Caribdis dies first, ranged bots will help with Sharkkis.
- (5) Everybody kills Sharkkis's pet.
- (6) Everybody kills Karathress.

### Morogrim Tidewalker

- The main tank will pull the boss to the Northeast pillar, with the tank's back against the pillar.
- A hunter will misdirect the boss onto the main tank upon the pull.
- When the boss gets to 26% HP, the main tank will begin moving the boss to the Northeast corner of the room in preparation for Phase 2 (which begins at 25%). The tank will move in two steps to get around the pillar.
- When the boss gets to 25% HP, ranged will follow the main tank to the corner and stack up right behind the boss. They will also move in two steps.
- There is no method for melee since they will just naturally follow the boss anyway.

### Lady Vashj

**Phase 1**:
- The main tank will tank Vashj in the center of the arena.
- If a Shaman is in the main tank's group, that Shaman will attempt to keep a Grounding Totem down in range of the main tank to absorb Shock Blast. This should continue in Phase 3.
- Ranged bots will spread out in a semicircle around the center of the arena.
- If any bot other than the main tank gets Static Charge, it will run away from other bots. If the main tank gets Static Charge, other bots will run away from the main tank. This method should continue in Phase 3.
- If any bot is Entangled and has Static Charge, the bot will attempt to use Cloak of Shadows if it is a Rogue, and Paladins will attempt to use Hand of Freedom. This method should continue in Phase 3 (with some modifications).
- Bots will not use Bloodlust or Heroism (saved for Phase 3). Bots will not use any other major cooldowns, either, such as Metamorphosis (saved for Phase 2 and 3).

**Phase 2**:
There are two central mechanics to this phase, both of which were challenging to get bots to execute properly. First is the system of prioritizing adds. The large playing field and multiple types of adds coming from random directions make this phase not doable with realistic DPS under the standard Playerbots target selection system. Therefore, I took inspiration from liyunfan's Naxx strategy for Phase 1 of Kel'Thuzad to disable dps assist and create a custom target selection system.

First, a cheat with respect to the Coilfang Striders: 
- Tanks will permanently have the Fear Ward aura applied to them if you have raid cheats enabled. This allows them to tank the Coilfang Striders. The standard strategy was to have an Elemental Shaman kite the Strider around the perimeter of the arena, with ranged players (including healers) spamming DoTs on the Strider. If you can make bots do this, then great, but it's far beyond my capabilities. Therefore, with the cheat, the first assistant tank is responsible for tanking Striders and keeping them away from Core passers (described below) and Vashj. Evidently it was (and is, in TBC Classic) possible to tank (and melee DPS) Striders by wearing a Dire Maul Ogre Suit, which would give you enough reach to stay out of the Strider's fear. I actually tried that, and it does not work, either because AC's radiuses are not the same or just because bots do not maintain the same level of precise positioning. But anyway, the point is that technically the Striders are tankable by real players, so maybe that will make you feel better about using this cheat (it's fine enough rationalizing for me). I found this fight to be unmanageable without this cheat (i.e., using a method that would only have bots try to run away from Striders) because each Strider was guaranteed to wipe out a couple of bots, and you really cannot afford to lose anyone. YMMV though.
- If cheats are enabled for Striders, Hunters will attempt to Misdirect the Striders to the first assist tank.
- If cheats are not enabled, bots will attempt to use slows/roots to stop the Striders. I have some logic for them to use Netherweave Nets, but I suspect it does not actually work so I may remove it instead of trying to get it to function properly.

Target priority is as follows:
- Hunters and Mages: Enchanted Elementals, Coilfang Striders, Coilfang Elites.
- Other Ranged Bots: Elites, Striders, Elementals. 
- Melee DPS: Elementals, Elites.
- Tanks: Elites, Elementals (except if cheats are enabled, the first assistant tank will instead prioritize Striders and then Elementals)
- Everybody else (basically means healers): Elementals, Elites, Striders
- If there is more than one of the same target, bots will prioritize the one that is closer to Vashj.
- In all cases, the valid attack ranged is limited so that bots should not leave the central platform.
- If somehow a bot ends up too far from the center of the room and is not actively attacking anything, there is logic to make them run back.

Handling Tainted Elementals and the Tainted Core: I will make another post about this later. It is easily the most complicated strategy I've ever worked on (far beyond anything on Kael'thas even) so will necessitate a long explanation. The tl;dr is that there is a chain of two-to-four bots that receive/pass the Tainted Core before using it on a Shield Generator, and if you are playing by yourself, you probably need to turn raid cheats on, in which case there will also be a bot that teleports to, kills, and loots the Tainted Elementals (i.e., the bots will then handle the entire sequence of shutting down Shield Generators).

**Phase 3**:
- The main tank will pick up Vashj immediately and try to keep her away from Enchanted Elementals.
- DPS will burn down residual adds from Phase 2 in the order of (1) elementals, (2) strider for ranged only (if you have more than one up, you're dead), and (3) elites (hopefully you have only one up, but two with one almost dead is possible).
- Hunters will kill Toxic Sporebats. This works quite well, but they (and anybody else if ordered to target Sporebats) have a tendency to levitate up into the pipes at the top of the room when killing the Sporebats. To counteract this, a method forcibly teleports bots to the ground if they get more than 2 yards above the ground.
- The Phase 1 Cloak of Shadows/Hand of Freedom method is now expanded to include bots Entangled in the Sporebat poison pools (with Hand of Freedom usage prioritized on the main tank).
- There is a specific method to avoid the Sporebat poison pools. The Vashj tank will move backwards when avoiding poison.

### DIFF
--- a/src/Ai/Raid/SerpentshrineCavern/Multiplier/RaidSSCMultipliers.cpp
+++ b/src/Ai/Raid/SerpentshrineCavern/Multiplier/RaidSSCMultipliers.cpp
@@ -351,7 +351,7 @@ float LeotherasTheBlindWaitForDpsMultiplier::GetValue(Action* action)
 
     constexpr uint8 dpsWaitSecondsPhase2 = 12;
     Unit* leotherasPhase2Demon = GetPhase2LeotherasDemon(botAI);
-    Player* demonFormTank = GetLeotherasDemonFormTank(botAI, bot);
+    Player* demonFormTank = GetLeotherasDemonFormTank(bot);
     if (leotherasPhase2Demon)
     {
         if (demonFormTank && demonFormTank == bot)
@@ -655,13 +655,9 @@ float LadyVashjStaticChargeStayAwayFromGroupMultiplier::GetValue(Action* action)
     return 1.0f;
 }
 
-// If raid cheat (which enables bot looting of the core) is not enabled
-// Bots should not loot the core
+// Bots should not loot the core with normal looting logic
 float LadyVashjDoNotLootTheTaintedCoreMultiplier::GetValue(Action* action)
 {
-    if (!botAI->HasCheat(BotCheatMask::raid))
-        return 1.0f;
-
     if (AI_VALUE2(Unit*, "find target", "lady vashj"))
     {
         if (dynamic_cast<LootAction*>(action))
--- a/src/Ai/Raid/SerpentshrineCavern/Util/RaidSSCHelpers.h
+++ b/src/Ai/Raid/SerpentshrineCavern/Util/RaidSSCHelpers.h
@@ -138,7 +138,7 @@ namespace SerpentShrineCavernHelpers
     Unit* GetPhase2LeotherasDemon(PlayerbotAI* botAI);
     Unit* GetPhase3LeotherasDemon(PlayerbotAI* botAI);
     Unit* GetActiveLeotherasDemon(PlayerbotAI* botAI);
-    Player* GetLeotherasDemonFormTank(PlayerbotAI* botAI, Player* bot);
+    Player* GetLeotherasDemonFormTank(Player* bot);
 
     // Fathom-Lord Karathress
     extern const Position KARATHRESS_TANK_POSITION;
--- a/src/Ai/Raid/SerpentshrineCavern/Action/RaidSSCActions.cpp
+++ b/src/Ai/Raid/SerpentshrineCavern/Action/RaidSSCActions.cpp
@@ -13,7 +13,7 @@ using namespace SerpentShrineCavernHelpers;
 
 // General
 
-bool SerpentShrineCavernEraseTimersAndTrackersAction::Execute(Event event)
+bool SerpentShrineCavernEraseTimersAndTrackersAction::Execute(Event /*event*/)
 {
     const uint32 instanceId = bot->GetMap()->GetInstanceId();
     const ObjectGuid guid = bot->GetGUID();
@@ -63,7 +63,7 @@ bool SerpentShrineCavernEraseTimersAndTrackersAction::Execute(Event event)
 // Trash Mobs
 
 // Move out of toxic pool left behind by some colossi upon death
-bool UnderbogColossusEscapeToxicPoolAction::Execute(Event event)
+bool UnderbogColossusEscapeToxicPoolAction::Execute(Event /*event*/)
 {
     Aura* aura = bot->GetAura(SPELL_TOXIC_POOL);
     if (!aura)
@@ -129,7 +129,7 @@ bool UnderbogColossusEscapeToxicPoolAction::Execute(Event event)
                   true, MovementPriority::MOVEMENT_FORCED, true, false);
 }
 
-bool GreyheartTidecallerMarkWaterElementalTotemAction::Execute(Event event)
+bool GreyheartTidecallerMarkWaterElementalTotemAction::Execute(Event /*event*/)
 {
     if (Unit* totem = GetFirstAliveUnitByEntry(botAI, NPC_WATER_ELEMENTAL_TOTEM))
         MarkTargetWithSkull(bot, totem);
@@ -142,7 +142,7 @@ bool GreyheartTidecallerMarkWaterElementalTotemAction::Execute(Event event)
 // (1) When tanking, move to designated tanking spot on frost side
 // (2) 1 second after 100% Mark of Hydross, move to nature tank's spot to hand off boss
 // (3) When Hydross is in nature form, move back to frost tank spot and wait for transition
-bool HydrossTheUnstablePositionFrostTankAction::Execute(Event event)
+bool HydrossTheUnstablePositionFrostTankAction::Execute(Event /*event*/)
 {
     Unit* hydross = AI_VALUE2(Unit*, "find target", "hydross the unstable");
     if (!hydross)
@@ -224,7 +224,7 @@ bool HydrossTheUnstablePositionFrostTankAction::Execute(Event event)
 // (1) When tanking, move to designated tanking spot on nature side
 // (2) 1 second after 100% Mark of Corruption, move to frost tank's spot to hand off boss
 // (3) When Hydross is in frost form, move back to nature tank spot and wait for transition
-bool HydrossTheUnstablePositionNatureTankAction::Execute(Event event)
+bool HydrossTheUnstablePositionNatureTankAction::Execute(Event /*event*/)
 {
     Unit* hydross = AI_VALUE2(Unit*, "find target", "hydross the unstable");
     if (!hydross)
@@ -303,7 +303,7 @@ bool HydrossTheUnstablePositionNatureTankAction::Execute(Event event)
     return false;
 }
 
-bool HydrossTheUnstablePrioritizeElementalAddsAction::Execute(Event event)
+bool HydrossTheUnstablePrioritizeElementalAddsAction::Execute(Event /*event*/)
 {
     Unit* waterElemental = GetFirstAliveUnitByEntry(botAI, NPC_PURE_SPAWN_OF_HYDROSS);
     if (waterElemental)
@@ -331,7 +331,7 @@ bool HydrossTheUnstablePrioritizeElementalAddsAction::Execute(Event event)
 }
 
 // To mitigate the effect of Water Tomb
-bool HydrossTheUnstableFrostPhaseSpreadOutAction::Execute(Event event)
+bool HydrossTheUnstableFrostPhaseSpreadOutAction::Execute(Event /*event*/)
 {
     if (!AI_VALUE2(Unit*, "find target", "hydross the unstable"))
         return false;
@@ -354,7 +354,7 @@ bool HydrossTheUnstableFrostPhaseSpreadOutAction::Execute(Event event)
     return false;
 }
 
-bool HydrossTheUnstableMisdirectBossToTankAction::Execute(Event event)
+bool HydrossTheUnstableMisdirectBossToTankAction::Execute(Event /*event*/)
 {
     Unit* hydross = AI_VALUE2(Unit*, "find target", "hydross the unstable");
     if (!hydross)
@@ -424,7 +424,7 @@ bool HydrossTheUnstableMisdirectBossToTankAction::TryMisdirectToNatureTank(
     return false;
 }
 
-bool HydrossTheUnstableStopDpsUponPhaseChangeAction::Execute(Event event)
+bool HydrossTheUnstableStopDpsUponPhaseChangeAction::Execute(Event /*event*/)
 {
     Unit* hydross = AI_VALUE2(Unit*, "find target", "hydross the unstable");
     if (!hydross)
@@ -470,7 +470,7 @@ bool HydrossTheUnstableStopDpsUponPhaseChangeAction::Execute(Event event)
     return false;
 }
 
-bool HydrossTheUnstableManageTimersAction::Execute(Event event)
+bool HydrossTheUnstableManageTimersAction::Execute(Event /*event*/)
 {
     Unit* hydross = AI_VALUE2(Unit*, "find target", "hydross the unstable");
     if (!hydross)
@@ -515,7 +515,7 @@ bool HydrossTheUnstableManageTimersAction::Execute(Event event)
 // The Lurker Below
 
 // Run around behind Lurker during Spout
-bool TheLurkerBelowRunAroundBehindBossAction::Execute(Event event)
+bool TheLurkerBelowRunAroundBehindBossAction::Execute(Event /*event*/)
 {
     Unit* lurker = AI_VALUE2(Unit*, "find target", "the lurker below");
     if (!lurker)
@@ -552,7 +552,7 @@ bool TheLurkerBelowRunAroundBehindBossAction::Execute(Event event)
     return false;
 }
 
-bool TheLurkerBelowPositionMainTankAction::Execute(Event event)
+bool TheLurkerBelowPositionMainTankAction::Execute(Event /*event*/)
 {
     Unit* lurker = AI_VALUE2(Unit*, "find target", "the lurker below");
     if (!lurker)
@@ -573,7 +573,7 @@ bool TheLurkerBelowPositionMainTankAction::Execute(Event event)
 }
 
 // Assign ranged positions within a 120-degree arc behind Lurker
-bool TheLurkerBelowSpreadRangedInArcAction::Execute(Event event)
+bool TheLurkerBelowSpreadRangedInArcAction::Execute(Event /*event*/)
 {
     Unit* lurker = AI_VALUE2(Unit*, "find target", "the lurker below");
     if (!lurker)
@@ -636,7 +636,7 @@ bool TheLurkerBelowSpreadRangedInArcAction::Execute(Event event)
 
 // During the submerge phase, if there are >= 3 tanks in the raid,
 // the first 3 will each pick up 1 Guardian
-bool TheLurkerBelowTanksPickUpAddsAction::Execute(Event event)
+bool TheLurkerBelowTanksPickUpAddsAction::Execute(Event /*event*/)
 {
     Player* mainTank = nullptr;
     Player* firstAssistTank = nullptr;
@@ -700,7 +700,7 @@ bool TheLurkerBelowTanksPickUpAddsAction::Execute(Event event)
     return false;
 }
 
-bool TheLurkerBelowManageSpoutTimerAction::Execute(Event event)
+bool TheLurkerBelowManageSpoutTimerAction::Execute(Event /*event*/)
 {
     Unit* lurker = AI_VALUE2(Unit*, "find target", "the lurker below");
     if (!lurker)
@@ -725,7 +725,7 @@ bool TheLurkerBelowManageSpoutTimerAction::Execute(Event event)
 
 // Leotheras the Blind
 
-bool LeotherasTheBlindTargetSpellbindersAction::Execute(Event event)
+bool LeotherasTheBlindTargetSpellbindersAction::Execute(Event /*event*/)
 {
     if (Unit* spellbinder = GetFirstAliveUnitByEntry(botAI, NPC_GREYHEART_SPELLBINDER))
         MarkTargetWithSkull(bot, spellbinder);
@@ -735,7 +735,7 @@ bool LeotherasTheBlindTargetSpellbindersAction::Execute(Event event)
 
 // Warlock tank action--see GetLeotherasDemonFormTank in RaidSSCHelpers.cpp
 // Use tank strategy for Demon Form and DPS strategy for Human Form
-bool LeotherasTheBlindDemonFormTankAttackBossAction::Execute(Event event)
+bool LeotherasTheBlindDemonFormTankAttackBossAction::Execute(Event /*event*/)
 {
     Unit* innerDemon = nullptr;
     auto const& npcs =
@@ -769,7 +769,7 @@ bool LeotherasTheBlindDemonFormTankAttackBossAction::Execute(Event event)
 
 // Stop melee tanks from attacking upon transformation so they don't take aggro
 // Applies only if there is a Warlock tank present
-bool LeotherasTheBlindMeleeTanksDontAttackDemonFormAction::Execute(Event event)
+bool LeotherasTheBlindMeleeTanksDontAttackDemonFormAction::Execute(Event /*event*/)
 {
     bot->AttackStop();
     botAI->Reset();
@@ -778,7 +778,7 @@ bool LeotherasTheBlindMeleeTanksDontAttackDemonFormAction::Execute(Event event)
 
 // Intent is to keep enough distance from Leotheras and spread to prepare for Whirlwind
 // And stay away from the Warlock tank to avoid Chaos Blasts
-bool LeotherasTheBlindPositionRangedAction::Execute(Event event)
+bool LeotherasTheBlindPositionRangedAction::Execute(Event /*event*/)
 {
     constexpr float safeDistFromBoss = 15.0f;
     Unit* leotherasHuman = GetLeotherasHuman(botAI);
@@ -802,7 +802,7 @@ bool LeotherasTheBlindPositionRangedAction::Execute(Event event)
                 continue;
 
             constexpr uint32 minInterval = 0;
-            if (GetLeotherasDemonFormTank(botAI, bot) == member)
+            if (GetLeotherasDemonFormTank(bot) == member)
             {
                 constexpr float safeDistFromTank = 10.0f;
                 if (bot->GetExactDist2d(member) < safeDistFromTank)
@@ -820,7 +820,7 @@ bool LeotherasTheBlindPositionRangedAction::Execute(Event event)
     return false;
 }
 
-bool LeotherasTheBlindRunAwayFromWhirlwindAction::Execute(Event event)
+bool LeotherasTheBlindRunAwayFromWhirlwindAction::Execute(Event /*event*/)
 {
     if (Unit* leotherasHuman = GetLeotherasHuman(botAI))
     {
@@ -838,7 +838,7 @@ bool LeotherasTheBlindRunAwayFromWhirlwindAction::Execute(Event event)
 
 // This method is likely unnecessary unless the player does not use a Warlock tank
 // If a melee tank is used, other melee needs to run away after too many Chaos Blast stacks
-bool LeotherasTheBlindMeleeDpsRunAwayFromBossAction::Execute(Event event)
+bool LeotherasTheBlindMeleeDpsRunAwayFromBossAction::Execute(Event /*event*/)
 {
     if (botAI->CanCastSpell("cloak of shadows", bot))
         return botAI->CastSpell("cloak of shadows", bot);
@@ -864,7 +864,7 @@ bool LeotherasTheBlindMeleeDpsRunAwayFromBossAction::Execute(Event event)
 }
 
 // Hardcoded actions for healers and bear tanks to kill Inner Demons
-bool LeotherasTheBlindDestroyInnerDemonAction::Execute(Event event)
+bool LeotherasTheBlindDestroyInnerDemonAction::Execute(Event /*event*/)
 {
     Unit* innerDemon = nullptr;
     auto const& npcs =
@@ -1033,7 +1033,7 @@ bool LeotherasTheBlindDestroyInnerDemonAction::HandleHealerStrategy(Unit* innerD
 }
 
 // Everybody except the Warlock tank should focus on Leotheras in Phase 3
-bool LeotherasTheBlindFinalPhaseAssignDpsPriorityAction::Execute(Event event)
+bool LeotherasTheBlindFinalPhaseAssignDpsPriorityAction::Execute(Event /*event*/)
 {
     Unit* leotherasHuman = GetLeotherasHuman(botAI);
     if (!leotherasHuman)
@@ -1068,13 +1068,13 @@ bool LeotherasTheBlindFinalPhaseAssignDpsPriorityAction::Execute(Event event)
 }
 
 // Misdirect to Warlock tank or to main tank if there is no Warlock tank
-bool LeotherasTheBlindMisdirectBossToDemonFormTankAction::Execute(Event event)
+bool LeotherasTheBlindMisdirectBossToDemonFormTankAction::Execute(Event /*event*/)
 {
     Unit* leotherasDemon = GetActiveLeotherasDemon(botAI);
     if (!leotherasDemon)
         return false;
 
-    Player* demonFormTank = GetLeotherasDemonFormTank(botAI, bot);
+    Player* demonFormTank = GetLeotherasDemonFormTank(bot);
     Player* targetTank = demonFormTank;
 
     if (!targetTank)
@@ -1106,7 +1106,7 @@ bool LeotherasTheBlindMisdirectBossToDemonFormTankAction::Execute(Event event)
 }
 
 // This does not pause DPS after a Whirlwind, which is also an aggro wipe
-bool LeotherasTheBlindManageDpsWaitTimersAction::Execute(Event event)
+bool LeotherasTheBlindManageDpsWaitTimersAction::Execute(Event /*event*/)
 {
     Unit* leotheras = AI_VALUE2(Unit*, "find target", "leotheras the blind");
     if (!leotheras)
@@ -1164,7 +1164,7 @@ bool LeotherasTheBlindManageDpsWaitTimersAction::Execute(Event event)
 // is crucial to separate Caribdis from the others
 
 // Karathress is tanked near his starting position
-bool FathomLordKarathressMainTankPositionBossAction::Execute(Event event)
+bool FathomLordKarathressMainTankPositionBossAction::Execute(Event /*event*/)
 {
     Unit* karathress = AI_VALUE2(Unit*, "find target", "fathom-lord karathress");
     if (!karathress)
@@ -1200,7 +1200,7 @@ bool FathomLordKarathressMainTankPositionBossAction::Execute(Event event)
 
 // Caribdis is pulled far to the West in the corner
 // Best to use a Warrior or Druid tank for interrupts
-bool FathomLordKarathressFirstAssistTankPositionCaribdisAction::Execute(Event event)
+bool FathomLordKarathressFirstAssistTankPositionCaribdisAction::Execute(Event /*event*/)
 {
     Unit* caribdis = AI_VALUE2(Unit*, "find target", "fathom-guard caribdis");
     if (!caribdis)
@@ -1235,7 +1235,7 @@ bool FathomLordKarathressFirstAssistTankPositionCaribdisAction::Execute(Event ev
 }
 
 // Sharkkis is pulled North to the other side of the ramp
-bool FathomLordKarathressSecondAssistTankPositionSharkkisAction::Execute(Event event)
+bool FathomLordKarathressSecondAssistTankPositionSharkkisAction::Execute(Event /*event*/)
 {
     Unit* sharkkis = AI_VALUE2(Unit*, "find target", "fathom-guard sharkkis");
     if (!sharkkis)
@@ -1270,7 +1270,7 @@ bool FathomLordKarathressSecondAssistTankPositionSharkkisAction::Execute(Event e
 }
 
 // Tidalvess is pulled Northwest near the pillar
-bool FathomLordKarathressThirdAssistTankPositionTidalvessAction::Execute(Event event)
+bool FathomLordKarathressThirdAssistTankPositionTidalvessAction::Execute(Event /*event*/)
 {
     Unit* tidalvess = AI_VALUE2(Unit*, "find target", "fathom-guard tidalvess");
     if (!tidalvess)
@@ -1306,7 +1306,7 @@ bool FathomLordKarathressThirdAssistTankPositionTidalvessAction::Execute(Event e
 
 // Caribdis's tank spot is far away so a dedicated healer is needed
 // Use the assistant flag to select the healer
-bool FathomLordKarathressPositionCaribdisTankHealerAction::Execute(Event event)
+bool FathomLordKarathressPositionCaribdisTankHealerAction::Execute(Event /*event*/)
 {
     Unit* caribdis = AI_VALUE2(Unit*, "find target", "fathom-guard caribdis");
     if (!caribdis)
@@ -1332,7 +1332,7 @@ bool FathomLordKarathressPositionCaribdisTankHealerAction::Execute(Event event)
 }
 
 // Misdirect priority: (1) Caribdis tank, (2) Tidalvess tank, (3) Sharkkis tank
-bool FathomLordKarathressMisdirectBossesToTanksAction::Execute(Event event)
+bool FathomLordKarathressMisdirectBossesToTanksAction::Execute(Event /*event*/)
 {
     Group* group = bot->GetGroup();
     if (!group)
@@ -1418,7 +1418,7 @@ bool FathomLordKarathressMisdirectBossesToTanksAction::Execute(Event event)
 
 // Kill order is non-standard because bots handle Cyclones poorly and need more time
 // to get her down than real players (standard is ranged DPS help with Sharkkis first)
-bool FathomLordKarathressAssignDpsPriorityAction::Execute(Event event)
+bool FathomLordKarathressAssignDpsPriorityAction::Execute(Event /*event*/)
 {
     // Target priority 1: Spitfire Totems for melee dps
     Unit* totem = GetFirstAliveUnitByEntry(botAI, NPC_SPITFIRE_TOTEM);
@@ -1526,7 +1526,7 @@ bool FathomLordKarathressAssignDpsPriorityAction::Execute(Event event)
     return false;
 }
 
-bool FathomLordKarathressManageDpsTimerAction::Execute(Event event)
+bool FathomLordKarathressManageDpsTimerAction::Execute(Event /*event*/)
 {
     Unit* karathress = AI_VALUE2(Unit*, "find target", "fathom-lord karathress");
     if (!karathress)
@@ -1540,7 +1540,7 @@ bool FathomLordKarathressManageDpsTimerAction::Execute(Event event)
 
 // Morogrim Tidewalker
 
-bool MorogrimTidewalkerMisdirectBossToMainTankAction::Execute(Event event)
+bool MorogrimTidewalkerMisdirectBossToMainTankAction::Execute(Event /*event*/)
 {
     Unit* tidewalker = AI_VALUE2(Unit*, "find target", "morogrim tidewalker");
     if (!tidewalker)
@@ -1574,7 +1574,7 @@ bool MorogrimTidewalkerMisdirectBossToMainTankAction::Execute(Event event)
 
 // Separate tanking positions are used for phase 1 and phase 2 to address the
 // Water Globule mechanic in phase 2
-bool MorogrimTidewalkerMoveBossToTankPositionAction::Execute(Event event)
+bool MorogrimTidewalkerMoveBossToTankPositionAction::Execute(Event /*event*/)
 {
     Unit* tidewalker = AI_VALUE2(Unit*, "find target", "morogrim tidewalker");
     if (!tidewalker)
@@ -1666,7 +1666,7 @@ bool MorogrimTidewalkerMoveBossToTankPositionAction::MoveToPhase2TankPosition(Un
 
 // Ranged stack behind the boss in the Northeast corner in phase 2
 // No corresponding method for melee since they will do so anyway
-bool MorogrimTidewalkerPhase2RepositionRangedAction::Execute(Event event)
+bool MorogrimTidewalkerPhase2RepositionRangedAction::Execute(Event /*event*/)
 {
     Unit* tidewalker = AI_VALUE2(Unit*, "find target", "morogrim tidewalker");
     if (!tidewalker)
@@ -1724,7 +1724,7 @@ bool MorogrimTidewalkerPhase2RepositionRangedAction::Execute(Event event)
 
 // Lady Vashj <Coilfang Matron>
 
-bool LadyVashjMainTankPositionBossAction::Execute(Event event)
+bool LadyVashjMainTankPositionBossAction::Execute(Event /*event*/)
 {
     Unit* vashj = AI_VALUE2(Unit*, "find target", "lady vashj");
     if (!vashj)
@@ -1772,7 +1772,7 @@ bool LadyVashjMainTankPositionBossAction::Execute(Event event)
 }
 
 // Semicircle around center of the room (to allow escape paths by Static Charged bots)
-bool LadyVashjPhase1SpreadRangedInArcAction::Execute(Event event)
+bool LadyVashjPhase1SpreadRangedInArcAction::Execute(Event /*event*/)
 {
     std::vector<Player*> spreadMembers;
     if (Group* group = bot->GetGroup())
@@ -1845,7 +1845,7 @@ bool LadyVashjPhase1SpreadRangedInArcAction::Execute(Event event)
 }
 
 // For absorbing Shock Burst
-bool LadyVashjSetGroundingTotemInMainTankGroupAction::Execute(Event event)
+bool LadyVashjSetGroundingTotemInMainTankGroupAction::Execute(Event /*event*/)
 {
     Player* mainTank = nullptr;
     if (Group* group = bot->GetGroup())
@@ -1880,7 +1880,7 @@ bool LadyVashjSetGroundingTotemInMainTankGroupAction::Execute(Event event)
     return false;
 }
 
-bool LadyVashjMisdirectBossToMainTankAction::Execute(Event event)
+bool LadyVashjMisdirectBossToMainTankAction::Execute(Event /*event*/)
 {
     Unit* vashj = AI_VALUE2(Unit*, "find target", "lady vashj");
     if (!vashj)
@@ -1912,7 +1912,7 @@ bool LadyVashjMisdirectBossToMainTankAction::Execute(Event event)
     return false;
 }
 
-bool LadyVashjStaticChargeMoveAwayFromGroupAction::Execute(Event event)
+bool LadyVashjStaticChargeMoveAwayFromGroupAction::Execute(Event /*event*/)
 {
     Group* group = bot->GetGroup();
     if (!group)
@@ -1958,7 +1958,7 @@ bool LadyVashjStaticChargeMoveAwayFromGroupAction::Execute(Event event)
     return false;
 }
 
-bool LadyVashjAssignPhase2AndPhase3DpsPriorityAction::Execute(Event event)
+bool LadyVashjAssignPhase2AndPhase3DpsPriorityAction::Execute(Event /*event*/)
 {
     Unit* vashj = AI_VALUE2(Unit*, "find target", "lady vashj");
     if (!vashj)
@@ -2134,7 +2134,7 @@ bool LadyVashjAssignPhase2AndPhase3DpsPriorityAction::Execute(Event event)
     return false;
 }
 
-bool LadyVashjMisdirectStriderToFirstAssistTankAction::Execute(Event event)
+bool LadyVashjMisdirectStriderToFirstAssistTankAction::Execute(Event /*event*/)
 {
     // Striders are not tankable without a cheat to block Fear so there is
     // no point in misdirecting if raid cheats are not enabled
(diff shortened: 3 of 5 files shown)


brighton-chi:
Edited: Below description of core passing methods were brought up to date as of the PR coming off of draft.

The full process of looting, passing, and using the Tainted Cores involves up to five bots. 

First, I’ll cover looting. If raid cheats are not enabled, players will be responsible for killing Tainted Elementals and looting the Core. I do not think this approach is workable for a single player because where the Elemental spawns is random, and you have 15 seconds to kill and loot it before it despawns. The entire raid is under tremendous pressure during this whole period, so if you are appropriately geared and using IP nerfs, you cannot afford to miss any Elementals. However, the option remains if you want to try it—bots should continue the remainder of the Core passing chain, even if raid cheats are disabled and the player loots the Core.

If raid cheats are enabled, a bot (“designatedLooter”) will be responsible for looting the Core. This bot will be the first returned melee DPS bot. The looter will teleport to the Tainted Elemental when it spawns, set the Tainted Elemental as its RTI target, and kill it based on normal bot attacking behavior. The looter will have the health cheat enabled while it is in close proximity to the Tainted Elemental and will also automatically have the Poison Volley debuff removed. This is necessary for the looter to survive without a healer. I find this cheat approach to be needed because bots cannot effectively use the entire playing area. They already struggle to be efficient when using the entire platform, so trying to extend their activity down the stairs is really messy, even more so due to loss of LoS. I tried assigning bot groups to different areas but found that there wasn’t sufficient flexibility in bot behavior to make the strategy effective. Since the Elemental spawns at the base of the stairs and does not move, teleporting to it was the only viable option I came up with.

Once the Elemental is dead, the looter will call OpenLootAction to loot the corpse (actually a creature object, not a corpse object). After a short delay, the method (LadyVashjLootTaintedCoreAction) sends a packet (CMSG_AUTOSTORE_LOOT_ITEM) to the server to store the Core. I found this delay necessary because otherwise the bot would open the loot window but not actually loot the Core. I cannot explain the mechanics of this method so it will need to be analyzed by somebody with the requisite coding knowledge. It is, however, a genuine looting event and not a workaround--the manual packet that is sent should match the packet that is sent in connection with ordinary looting. As a result, the Elemental is actually looted, with the Core removed from its corpse.

Second, the passing and use of the Core involves a sequence of four “Tainted Core Passer” bots (“firstCorePasser,” “secondCorePasser,” etc.). The first three passers can be assigned by giving the assistant flag to healers. Healers are not very useful during Phase 2, so they are the best bots in my opinion to dedicate to passing. The fourth passer, who is very unlikely to be needed, will be a ranged DPS bot (and can also be assigned by using the assistant flag). Any roles for which an assistant flag is not chosen will be chosen by returning the first member that is not a tank. The roles are mutually exclusive. In most cases, only two passers will be necessary, but it is possible that up to four will be needed; that may happen if you have bad luck and the Elemental spawns on the opposite side of the map from the final Generator.

The trigger for the passing method involves the looter and the four passers by way of a cascading sequence. That is to say, once a bot has fulfilled its role in the chain, the trigger will return false for such bot so that bots will be tied up in the passing logic only as long as is necessary. Setting aside any such early return, the trigger fires when the designated looter teleports to the Elemental and continues as long as any bot has the Core or has had the Core within the past 3 seconds (the delay is so that the methods do not stop running when the Core is in transit between bots). A Fear Ward cheat will apply to bots when they are in the core passing sequence. This is necessary due to the bots' inability to be rooted. To simulate the fight, I force bots to stay in place when they have the Core, as if they were players. However, if they get feared by Striders, they run around because they are not actually rooted. This disrupts the entire sequence since if they end up outside of Generator use or pass range after being feared, they cannot go back to range since their movement is disabled. I did not gate this cheat behind BotCheatMask because I view it as inherently necessary to get core passing to work for bots. If they could ever be properly rooted, this can be changed.

Now for the actual passing method. When an Elemental spawns and the designated looter teleports to it, the closest active Generator (the “closestTrigger”) to the designated looter is determined by a grid search of DB guids of NPC_WORLD_INVISIBLE_TRIGGER (which is the NPC that generates the Generator beam to Vashj and despawns when the corresponding Generator is deactivated). That will be the Generator that the passing sequence should end at. The first passer moves to the closest point to the designated looter on the edge of a circle with a radius of 57.5 yards that is centered on the middle of the arena (basically, near the top of the stairs). When the first passer is in position, the designated looter calls ImbueItem to throw the Core to the first passer. Thank you to @Noscopezz for coming up with this for using items, and @mw108 for using it for bot-to-player passing—I don’t know if I would have ever found the right method otherwise. However, ImbueItem does not work to actually give the Core to the receiving bot; there is an animation of the Core being thrown, and the Core is removed from the passing bot’s inventory, but it is never received. I assume there is some issue here due to ImbueItem not being intended for this type of unique situation, though I don’t understand the mechanics that cause the issue (maybe due to there not being any sort of store item method triggered on account of ImbueItem, I don’t know). Accordingly, the strategy simulates the pass—it calls ImbueItem to show the animation and remove the Core from the passer, and then it manually creates the Core in the inventory of the receiver. There is some built-in delay that is needed to make the animations look less instantaneous and also to keep bots from double-throwing.

After the first passer has the Core, the second passer will move to the closest active Generator if reachable within the throwing range of the first passer (40 yards, but there is some built-in buffer to handle imprecision so distance is not maximized quite as efficiently as real players would). If the closest Generator is not close enough, the second passer will instead move as far as possible to remain in throwing distance on the line between the first passer and the closest Generator. At that point, the third passer will activate and move to the closest Generator (or, if too far from the second passer, the third passer will move as far as possible to remain in throwing distance on the line between the second passer and the closest generator). At that point, the fourth passer will activate and move to the closest Generator (four passers will always be enough).

To use the Generator, a packet (CMSG_USE_ITEM) is sent to the server. This manual packet is intended to simulate UseItemOnGameObject. Even though the client-side interaction is just right-clicking the Generator, the fact that it requires the Tainted Core to be in the inventory seems to cause the Generator usage to require a packet, which is not the case with a simple direct use of a game object (like the Magtheridon cubes, which could be used with a simple “Use” call). The Vashj method does skip the early return for IsNonMeleeSpellCast that is in UseItemOnGameObject. I do not know why, but that check would always cause the Generator use to fail.

Timberpoes:
Raid Composition:
3 Tanks (Player as Prot Paladin/Feral Druid/Prot Warrior)
6 Healers (2x shaman, 2x priest, 1 paladin, 1 druid)
Mix of other DPS, 2x pure DPS with 1x of various hybrid DPS.
Gear standard: Most bots sporting a mix of BiS blues with 50-90% T4 purples.
No IP nerfs to player strength.
Playerbot raid cheats enabled.

**Hydross**:
I tanked frost phase. When I moved boss to nature phase, zero aggro issues. Immediately went to my offtank who it looks like had misdirections because their threat gen was huge. DPS killed adds, skulls went up. Boss did not get another phase.

Strategy a **huge** improvement over no strategy just for bot AI behaviours on changing from frost to nature.

**The Lurker Below**:
I let the bots tank this one. DPS and healers mostly spread around the platform. They got knocked into the water occasionally and swam back onto the platform. One submerge phase, two spout phases.

First spout one death, tree form druid that was behind the pack a bit and eventually the spout caught up to them.

Submerge phase, six deaths from Naga. 5 of those deaths happened within 1s of eachother. One rogue to Arcing Smash from a guardian, then 1s later a mage pulled aggro and died with arcing smash taking out an enh shaman, boomkin and resto druid. 10s later holy priest died to another arcing smash.

Second spout on death, resto shaman.

Spout phase is probably realistic, but submerge phase with Naga felt a bit choatic. Possibly issue with only taking 3 tanks. May benefit from a brief period of no DPS to let the tanks establish aggro.

Strategy a **huge** improvement over no strategy for spout, where I had to try and use flee to get the bots stacked up and following me. Melee DPS automatically positioned behind during spout without strategy. I experienced the same issue on submerge phase with and without strategy - DPS got exploded on guardians and took healers with them.

**Leotheras the Blind**:
Human phase no problem. Raid backed off and spread out during whirlwind and tanks tried to keep aggro, mostly succeeding with some licks of whirlwind against random DPS and/or healers.

Demon phase no problem, although I threw my melee DPS under the bus since I didn't have an appropriately geared warlock bot so I just used my fattest bear tank.

Final split phase huge issues getting DPS to actually attack Leo. Square RTI on demon with Star on Leo, but the bots would not attack and the raid would wipe to DPS attrition.

Without a warlock tank I had much more success `co -ssc` during the final phase and just making my DPS do a desperate last stand nuke down of the human form without the huge waiting time.

Did not test with warlock tank.

Strategy human phase a **huge** improvement over no strategy. Did not do demon or final phases as strategy intended, had to disable to complete fight through a final brute force burn down of human form in final phase. Overall, better than no strategy with huge QoL for whirlwind avoidance behaviours. Will need testing with a warlock tank, may cheat an altbot one in at some point.

**Fathom-Lord Karathress**:
Bit of weirdness with the pull but otherwise it went well.

At the start of the fight the Spitfire Totems were being focused down.

In the final phase, rti markers were no longer being put on Spitfire totems and they were being ignored. I had to manually add triangle RTI to Spitfire Totems in this phase.

Strategy a **huge** improvement in general. Fight was more organised than without. Still got the raid caught in a whirlwind or two and I abused summon to bring them back to the floor, but it is what it is with bots that refuse to fall.

**Morogrim Tidewalker**:
My bot tanked him at the top of the ramp and kept him in the same spot the whole fight.

Healers were scattered around for Watery Grave, was good. At 25% everyone stacked up behind him. Still managed a kill despite a hunter deciding to Disengage into him while doing the Murloc packs around his room.

Murloc adds mid-fight got picked up by my spare tank and myself.

Strategy a **huge** improvement for the entire fight. Just felt more clean, more precise. Very good.

**Lady Vashj**:
Phases 1 and 2 are nice, when they work.

Phase 2 bugged on many of my attempts and it took up to 6-7 tainted elementals to crack. Watching the bots, it seems the logic has chosen healers for the chain throwing and if Vashj's chain lightning interrupts using the core, the bot may then prioiritise healing over using the core and will only re-use the core if they're done healing. With poor luck, this leads to the above mentioned soft lock of them getting feared away.

The melee DPS are the real heroes keeping every elemental at bay single-handedly once naga and striders spawn.

Phase 3 is a total shitshow and it took awhile to get a phase transition without tanks just getting nuked. My tanks bounced between failing to move Vashj away from spore pools at all so melee DPS got dunked, to LoSing down the ramp away from all the healers so they got dunked. Spore pools all over the place and it's only really possible after a perfect 4-tainted phase 2 as my raid was just too drained on resources if it dragged on past that.

However, when the core chain works it **works**. The cheat-enabled bot that teleports to get the core feeds it to the next chain at the top of the ramp that feeds it to its final destination with a quick use. If you can get them to do this 4 times in a row, you're looking at a really good first two phases.

**Conclusion**:
All fights complete with no IP nerfs. My raiding team is **not** T4 optimised. Most of my raiding team have BiS blues with a smattering of 4-6 T4 items and the occasional SSC T5 item.
image
image
image

The strategies largely improve fights over no strategies and Vashj does actually work with some effort and patience.

brighton-chi:
> Raid Composition:
> 
> 3 Tanks (Player as Prot Paladin/Feral Druid/Prot Warrior)
> 
> 6 Healers (2x shaman, 2x priest, 1 paladin, 1 druid)
> 
> Mix of other DPS, 2x pure DPS with 1x of various hybrid DPS.
> 
> Gear standard: Most bots sporting a mix of BiS blues with 50-90% T4 purples.
> 
> No IP nerfs to player strength.
> 
> Playerbot raid cheats enabled.
> 
> 
> 
> **Hydross**:
> 
> I tanked frost phase. When I moved boss to nature phase, zero aggro issues. Immediately went to my offtank who it looks like had misdirections because their threat gen was huge. DPS killed adds, skulls went up. Boss did not get another phase.
> 
> 
> 
> Strategy a **huge** improvement over no strategy just for bot AI behaviours on changing from frost to nature.
> 
> 
> 
> **The Lurker Below**:
> 
> I let the bots tank this one. DPS and healers mostly spread around the platform. They got knocked into the water occasionally and swam back onto the platform. One submerge phase, two spout phases.
> 
> 
> 
> First spout one death, tree form druid that was behind the pack a bit and eventually the spout caught up to them.
> 
> 
> 
> Submerge phase, six deaths from Naga. 5 of those deaths happened within 1s of eachother. One rogue to Arcing Smash from a guardian, then 1s later a mage pulled aggro and died with arcing smash taking out an enh shaman, boomkin and resto druid. 10s later holy priest died to another arcing smash.
> 
> 
> 
> Second spout on death, resto shaman.
> 
> 
> 
> Spout phase is probably realistic, but submerge phase with Naga felt a bit choatic. Possibly issue with only taking 3 tanks. May benefit from a brief period of no DPS to let the tanks establish aggro.
> 
> 
> 
> Strategy a **huge** improvement over no strategy for spout, where I had to try and use flee to get the bots stacked up and following me. Melee DPS automatically positioned behind during spout without strategy. I experienced the same issue on submerge phase with and without strategy - DPS got exploded on guardians and took healers with them.
> 
> 
> 
> **Leotheras the Blind**:
> 
> Human phase no problem. Raid backed off and spread out during whirlwind and tanks tried to keep aggro, mostly succeeding with some licks of whirlwind against random DPS and/or healers.
> 
> 
> 
> Demon phase no problem, although I threw my melee DPS under the bus since I didn't have an appropriately geared warlock bot so I just used my fattest bear tank.
> 
> 
> 
> Final split phase huge issues getting DPS to actually attack Leo. Square RTI on demon with Star on Leo, but the bots would not attack and the raid would wipe to DPS attrition.
> 
> 
> 
> Without a warlock tank I had much more success `co -ssc` during the final phase and just making my DPS do a desperate last stand nuke down of the human form without the huge waiting time.
> 
> 
> 
> Did not test with warlock tank.
> 
> 
> 
> Strategy human phase a **huge** improvement over no strategy. Did not do demon or final phases as strategy intended, had to disable to complete fight through a final brute force burn down of human form in final phase. Overall, better than no strategy with huge QoL for whirlwind avoidance behaviours. Will need testing with a warlock tank, may cheat an altbot one in at some point.
> 
> 
> 
> **Fathom-Lord Karathress**:
> 
> Bit of weirdness with the pull but otherwise it went well.
> 
> 
> 
> At the start of the fight the Spitfire Totems were being focused down.
> 
> 
> 
> In the final phase, rti markers were no longer being put on Spitfire totems and they were being ignored. I had to manually add triangle RTI to Spitfire Totems in this phase.
> 
> 
> 
> Strategy a **huge** improvement in general. Fight was more organised than without. Still got the raid caught in a whirlwind or two and I abused summon to bring them back to the floor, but it is what it is with bots that refuse to fall.
> 
> 
> 
> **Morogrim Tidewalker**:
> 
> My bot tanked him at the top of the ramp and kept him in the same spot the whole fight.
> 
> 
> 
> Healers were scattered around for Watery Grave, was good. At 25% everyone stacked up behind him. Still managed a kill despite a hunter deciding to Disengage into him while doing the Murloc packs around his room.
> 
> 
> 
> Murloc adds mid-fight got picked up by my spare tank and myself.
> 
> 
> 
> Strategy a **huge** improvement for the entire fight. Just felt more clean, more precise. Very good.
> 
> 
> 
> **Lady Vashj**:
> 
> Phases 1 and 2 are nice, when they work.
> 
> 
> 
> Phase 2 bugged on many of my attempts and it took up to 6-7 tainted elementals to crack. Watching the bots, it seems the logic has chosen healers for the chain throwing and if Vashj's chain lightning interrupts using the core, the bot may then prioiritise healing over using the core and will only re-use the core if they're done healing. With poor luck, this leads to the above mentioned soft lock of them getting feared away.
> 
> 
> 
> The melee DPS are the real heroes keeping every elemental at bay single-handedly once naga and striders spawn.
> 
> 
> 
> Phase 3 is a total shitshow and it took awhile to get a phase transition without tanks just getting nuked. My tanks bounced between failing to move Vashj away from spore pools at all so melee DPS got dunked, to LoSing down the ramp away from all the healers so they got dunked. Spore pools all over the place and it's only really possible after a perfect 4-tainted phase 2 as my raid was just too drained on resources if it dragged on past that.
> 
> 
> 
> However, when the core chain works it **works**. The cheat-enabled bot that teleports to get the core feeds it to the next chain at the top of the ramp that feeds it to its final destination with a quick use. If you can get them to do this 4 times in a row, you're looking at a really good first two phases.
> 
> 
> 
> **Conclusion**:
> 
> All fights complete with no IP nerfs. My raiding team is **not** T4 optimised. Most of my raiding team have BiS blues with a smattering of 4-6 T4 items and the occasional SSC T5 item.
> 
> image
> 
> image
> 
> image
> 
> 
> 
> The strategies largely improve fights over no strategies and Vashj does actually work with some effort and patience.

This is super helpful, thank you.

Some questions to help me figure this out:

For Lurker's submerge phase, can you confirm it was the Guardians that were donking everybody? It is very hard to get bots to pick up multiple adds, but if it's just the Guardians that need to be controlled, that's definitely doable.

For Karathress, did you have melee DPS still alive when the bots were all on Karathress?

For Leo's final phase, was it all DPS that would not attack? I had this problem before and thought I fixed it but maybe broke something again. The bots are also supposed to wait for 12 seconds, I think I need to shorten that, but do you know if you waited that long to see if they did anything?

Also for Leo, did you see which classes/specs got demons, and did any fail? I'm also hoping somebody can help me observe the bots that get demons and see what they do, including those that do and do not have the cheat. Fine if you weren't focusing on that, just wondering if you saw anything.

For Vashj, did you use the assistant flag to assign the healers? Did it look like different healers were getting assigned, or was it just the same ones but if they felt like they had to heal they were prioritizing it higher?

Timberpoes:
> For Lurker's submerge phase, can you confirm it was the Guardians that were donking everybody? It is very hard to get bots to pick up multiple adds, but if it's just the Guardians that need to be controlled, that's definitely doable.

I was pulling details from my DPS meter addon as I was writing the post for each boss. For Lurker, all deaths were Guardians with one of the deaths having a bit of multishot damage from one of the non-Guardian adds but that was a negligble sub-2k versus the guardians cleaving for 10k. The inital set of deaths were DPS pulling aggro into cleaves as soon as the adds spawned. The second set I think may have just been bot positioning as a random healer died to a cleave 10s later once everything was stable.

>For Karathress, did you have melee DPS still alive when the bots were all on Karathress?

Pretty much the entire raid was alive. They were just ignoring spitfire totems unless I manually put triangle RTI on them. Karathress, at that point in the raid, had triangle RTI.

>For Leo's final phase, was it all DPS that would not attack? I had this problem before and thought I fixed it but maybe broke something again. The bots are also supposed to wait for 12 seconds, I think I need to shorten that, but do you know if you waited that long to see if they did anything?

> Also for Leo, did you see which classes/specs got demons, and did any fail? I'm also hoping somebody can help me observe the bots that get demons and see what they do, including those that do and do not have the cheat. Fine if you weren't focusing on that, just wondering if you saw anything.

For Leo's final phase, none of the DPS would attack and I noticed that at least one of my bots was glitchy in behaviour. Looking at the demon form then the human form then the demon form back and forth in quick succession as if it was *trying* to attack but couldn't.

I'm not sure what classes and specs got demons. Running raid cheats, I didn't observe any demons spawning at all across 4-5 attempts with complete human-demon-human-both cycles on all fights. At one point I got a demon and I think it just instantly despawned? I never even had to attack or kill it.

> For Vashj, did you use the assistant flag to assign the healers? Did it look like different healers were getting assigned, or was it just the same ones but if they felt like they had to heal they were prioritizing it higher?

For Vashj, I don't know what setting assistant flag means for healers so it was just whatever defaults come out of the box. It was the same set of healers every time. My ret paladin (last to be added to the raid) did the killing, then passed to my tree druid who then passed to a resto shaman. If the resto shaman wasn't in position, my holy paladin moved to become the 4th in the chain. It never needed a 5th. If either the resto shaman or paladin got interrupted by Vashj's lightning attack while using the core, they'd prioritise raid healing until this was no longer possible/a priority, then try using the core again. This led to one instance of fear softlocking, and at least one instance of the bot holding the core until after the the **next** tainted had spawned and been killed because it kept getting interrupted then resorting to healing.

A lot of my raid ran low on mana at the end of phase 2 and it was always the 3rd or 4th core failing - cores 1 and 2 were flawless, so there could possibly be a situation where an OOM healer with the core won't use it because they need to heal but can't as they're OOM and just looping heal logic over use core logic, but that's just pure conjecture with the benefit of hindsight.

brighton-chi:
Substantive updates:

### Trash

Removed Rancid Mushroom method. It was not working, and while I'm pretty sure I know why, it's not important enough to bother fixing.

### Lurker

- Resto Druids shift out of Tree Form during Spout
- During the Submerge phase, if there are at least 3 tanks in the party, the first 3 tanks will each mark one Guardian and pick it up.

### Leotheras

- Removed all logic that is intended as a fallback for if you don't have a Warlock tank. It was not working well so I think it's better to just let the player handle things in that situation.
- Fixed a bug with the Inner Demon cheat. Based on Timberpoes's comment, I realized that bots were targeting each other's Inner Demons (and the player's) with the cheat damage. They now should target only their own, so now players will still have to kill their own Inner Demons, as well bots that are not a healer, tank, or DPS spec that is eligible for the cheat.
- Reduced the DPS wait time in the final phase from 12 to 8 seconds
- Regarding the lock-up of bots in the final phase, I believe it was something that would occur if you did not have a Warlock tank. It should be fixed now.

### Karathress

- Previously, the order of add tanks was Sharkkis, Tidalvess, Caribdis. The new order is Caribdis, Sharkkis, Tidalvess. That way, if you have only 2 or 3 tanks, one will still be dedicated to Caribdis since she is the most important one to pull away from the group. Tidalvess is the least important to have a dedicated tank since he will be killed first (and unlike Sharkkis, he doesn't summon a pet).
- I could not reproduce the issue of bots not marking Spitfire Totems when only Karathress was alive and don't see anything in the code that should be causing that. So this one is subject to more information.

### Vashj
For core passing:
- Core handler bots get a Fear Ward. I don't have a better way to handle this, and honestly I see it as less of a cheat than needing to solve a bug with respect to bots (i.e., Paralyze not rooting them).
- Bots should be using the Generators from a farther distance (I was too strict before vs. the actual interact distance).
- The trigger for the core passing sequence is now based on a bot holding a Core (with a grace period after) instead of the Paralyze aura. This seems to be much more reliable.
- As soon as a Tainted Elemental spawns, core handlers should calculate how many passers will be needed to get to the closest trigger from that elemental, and all needed passers will immediately move into position. This should tighten up the sequence.
- When a core passer is moving to a core passer position or at a core passer position waiting for a pass, it should ignore all movement commands other than to move to the position. When a core passer has the core, it should do nothing other than pass it or use it on the generator (no healing, etc.)

For the final phase:

- Reduced the playable area a bit to keep bots from going down the stairs.
- The tank will move backwards to get out of toxic pools--the tank was moving forwards previously and thus getting dazed by Vashj's melee strikes.
- Grounding Totem usage previously excluded Enhancement Shamans because as a Rogue, I didn't want to give up my precious Windfury. That restriction is gone--any Shaman in the main tank's group will drop Grounding Totems. This is really, really important for Phase 3 because the Shock Burst dazes the main tank for 5 seconds and is the number one reason they do not get out of pools in time. You should probably put two Shamans in the main tank's party.
- There was a targeting problem that resulted from me doing an if, if, if, else where the conditions should have been if, if else, if else, else. What resulted was that ranged bots and tanks would all prioritize enchanted elementals, then elites, then striders, then Vashj in Phase 3 (the general fallback prioritization), instead of having their proper priorities. The biggest problem was this would make the main tank rapidly switch between Vashj and other targets and thus would not properly tank/move Vashj until all adds were dead. This should be fixed.
- I'm hoping all the above will improve things, but Phase 3 will probably still be kind of a shitshow and a gear check. It always was in retail and is what makes the fight notorious. One thing I haven't tried yet but may do so is having different spore trigger avoidance distances for the main tank and other bots. I did something like this for Malchezaar, under the premise that the tank needs to leave more space for melee to operate. Right now, it's set to 7 yards for everybody (the pool's radius is 5 yards). I may try something like 8 or more for the MT and then 6 or 7 for others.

Timberpoes:
For reference, I did another run before the latest changes with a warlock """tank""" on Leo (I just took my highest HP warlock and set them to tank in full DPS gear) and the strategy worked as good as you're going to get, minus re-discovering the problem with Inner Demons in that one spawned on me and got nuked down without my intervention in a few seconds - I shall keep an eye on the changes for this in the latest commit and see how Inner Demons go.

I'm theorising that perhaps the strategy may also work if bots get logic to move away from demon Leo's current target once their debuff stacks pass a threshold. With a warlock tank, that logic would ideally never trigger and could even failsafe against mispositioning for any reason that caused a bot to be placed next to the warlock tank. Without a warlock tank, this would allow melee dps to move in, do some damage and retreat to clear stacks before dying. In theory would that make the strategy both more resilient and more universal? Once Phase 3 hits it's all DPS on the human form anyway so the Demon form offtank just needs to not die.

I'll re-test Karathress 2-bot-tank-1-player-tank too. It was a wipe-fest last night with bots positioning during the no-DPS-on-pull-downtime being based on me as one of the three tanks, which made the raid positioning a total mess.

But my raid comp does include a cat druid I can force into bear if I want a proper 4-tank experience, so I'll see if I can test both ways.

Looking forward to the final phase adjustments for Vashj, too. I'll take note about shamans in the tank group, my last successful kill I stacked 3 of them with the highest threat bot tank I had.

Timberpoes:
So, next big testing clear.

**Hydross**:
Flawless with 2 bot tanks and myself picking up the adds on change between frost/nature. No issues with DPS pulling aggro. Just works.

**Lurker**:
No more spout deaths. Didn't test 3-tank logic. Will make my cat druid a tank on the next kill to pad up to 3.

**Leo**:
In the final phase, tank just dragged Leo away and then the DPS very quickly nuked him down. I'm not even sure anyone was tanking the Demon, but it didn't matter. Zero deaths. 

**Karathress**:
New tank pickup priority order **so** much better. 3 tank strategy - two bots and me. Main bot tank took Karathress. Assist bot tank took Carib and Shark and hid behind the pillar in Carib's spot. I took Tidal and moved to the NW pillar as per tank instructions. No issues with Spitfire Totems. I did have to recover some raid members from the sky with a summon after Carib got taken down, but only one death. DPS still followed me until their timer to attack was done, but because I was on Tidal this worked well.

**Morogrim**:
Tank and spank. One death to the back Murloc pack diving into the healer line but it is what it is. Seems to work with the tank moving around the back of the pillar in the final DPS race to finish.

**Vashj**:
Phase 1 perfect, always perfect, never not perfect across all my 10-15 attempts.

Phase 2 bots always kill the elementals no problem, Vashj never gets any stacks, logic is sound for that. Naga usually don't stack up, Striders usually don't stack up. Priorities seem appropriate, at least for my raid's DPS level.

So the final part that's causing random phase 2 deaths, the Naga.

Bot positioning can be messy for healers and ranged DPS, the Naga tank doesn't respect they can cleave and will happily tank the Naga facing healers or ranged DPS, and will sometimes tank Naga far away from the center. This can lead to slow pickup of future Naga causing the occasional death.

Healers are just scattered in whatever random positions, sorta made worse by a bunch of them running around on core duty. So there's a lot of bot behavioural RNG in phase 2 because ranged and healers are so disorganised with no assigned spots, while tanks just tank the Naga whereever they establish aggro. Which could be on top of a healer or ranged DPS.

image

It's not raid wiping and many attempts go by without any deaths, or at least with few enough that a couple of combat rezzes can't counter, more just an observation of where deaths come from in a way not related to gear.

So, my latest shot under the new commit took 3 attempts to kill - which is a HUGE improvement with one caveat I'll cover at the end.

First attempt lost the Phase 3 DPS race as entered Phase 2 with 2x naga and strider and it just didn't seem to work out for us.

Second attempt lost to Phase 2 core softlock. 3rd shield needed a 4-chain of bots and the 3rd bot in the chain stopped responding, wouldn't acknowledge any chat commands like `inv` or `co ?`. There was no 4th bot stood waiting to use the core on a generator.

Third attempt flawless including a successful 4-chain in phase 2. Phase 3 felt WAY less chaotic, until I looked up at the sky to see if the bots were killing sporebats or something. At least one of my hunters was, but they'd... Teleported up into the sky some 40ft up killing sporebats??? That aside, is a key to this fight better sporebat management? Instead of ending with my raid basically hanging on by a thread and 20 sporebats raining down death, I ended the fight with practically everyone alive. There were 5 deaths in total and I think 3 of them happened after the fight ended to bots AFKing in lingering spore clouds. My raid comp only has 2 hunters.

**Conclusion**:
These strategies now have me safely raiding SSC on farm status. The logic is sound and every boss pre-Vashj I think is in an acceptable standard of strategy.

Hydross is good.

Lurker you really can't do much more with except maybe having bots disperse from the tanks during submerge to protect against errant cleaves, but otherwise it's good and I stopped seeing Spout deaths.

Leo I didn't observe any Inner Demon behaviour on the latest clear, but the logic works with a Warlock tank flawlessly. Maybe could be fine-tuned without one via DPS/healers avoiding the Demon tank when debuff stacks reach a certain level.

Karathress 3-tank is good and the new logic to leave Tidal as the loose 4th tank pickup worked for my use-case (DPS stick by me at the start and as the unassigned "bonus" tank they're right next to me to do that first kill).

Morogrim tank and spank. Very good.

Finally, Vashj feels improved. It felt like tanks were moving better. When cores didn't soft-lock, they were done quickly. But my Phase 3 is a **little** tainted by a hunter bot deciding to attack sporebats from 40ft in the air, so that final phase was a lot less of a DPS burn rush and more managable.

Timberpoes:
Side note on Hydross, your timing is incredibly tight because the phase transition involves the tank very slowly walking backwards.

On my run just now, the tank just got water prison stunned trying to walk back and because it only happens 5 seconds after 100%, it went up to 250% with the 1-shot and usual bouncing-back-and-forth-between-nature-and-not that accompanies a classic Hydross wipe. It's probably better to start the move as soon as the 100% stack happens.

Timberpoes:
Leo plus warlock tank, I got an Inner Demon and had to kill it manually as a prot paladin, which I managed. A 20/0/41 Frost Mage got mind controlled.

Timberpoes:
Karathress with 3 tanks plus me, each bot tank took their individual mob to the appropriate place, Spitfire Totems handled no problem, entire thing worked smoothly.

Timberpoes:
Vashj 1-shot with only 1 dead raid member at the end, the main tank feral druid.
image

No weird shenanigans with teleporting hunters. Multiple 4-chain core passes completed. Phase 3 started with a 100% health Naga. Raid kept spore bats under control without teleporting into the sky for long enough to beat the damage race. New tank logic makes it less chaotic.

Static Charge overrides Toxic Spores movement, so at one point a healer got Static Charge, moved away, got Toxic Spores and just face tanked it until Static Charge expired.

Timberpoes:
So that's two full clears under the updated strategy and a shameful display where the only wipe was Hydross. Me likey. SSC is very, very much doable and I'm sure those raiding under IP damage/healing nerfs will be able to leverage these strategies to great effect.

mw108:
Just wanted to say, amazing work! Thanks guys! ❤ Will try it when I've got time.

I have been working on a Vashj strategy in the past and got it to a point where she was beatable for me. But without all the automation. Still had to go for the elementals myself with the assistance of a temporary bot that I teleported to me and ping-ponged the core back and forth until I could use the shield generator.

brighton-chi:
Good point with Hydross; I hadn't considered the Water Tomb possibility. I originally had the phase change at 100% Marks but then decided to get cute and squeeze a little more damage, but the simpler approach seems better anyway so I'll tweak it.

I can do a method for Leo that triggers if you don't have a Warlock tank and makes bots other than the current tank get away from demon Leo if they have some number of Chaos Blast stacks. That sounds like low-hanging fruit that could benefit some players a lot. But how many stacks is too many? 4? 5? I didn't realize until today that WotLK totally changed the resistance system, and I haven't read into it enough to understand it yet.

I'm trying to think of if there is a good solution for the Naga cleaves. I could try something like have healers move some distance away from Naga or just their front and see how that plays out; I tried something along those lines with Magtheridon and didn't get it to work in a good way so gave up on it. It tends to be a lot harder to get bots _not_ to go somewhere or _not_ to attack something than the opposite, short of just making them do nothing at all. I could try to have tanks drag Naga to some location, but the trickiness with such a dynamic phase is the number of available tanks vs. creatures that need to be tanked can change a lot so I worry about taking away tank flexibility. My deaths during Phase 2 have tended to be because a tank did not pick up a Strider or Naga before it kills some cloth DPS (the fight actually got a lot better for me once I started having my Warlocks Soulstone themselves).

For the soft lock when core passing, do you remember if the third bot was near the core that was supposed to be used or not? If the bot was not responding to anything,

Funny about the midair Hunter. I don't remember how Phase 3 was done back in 2007, but I think killing the Sporebats is probably ideal because they have only 7K HP even after reverting the WotLK HP nerf. It seems in TBC Classic (which I didn't play), nobody killed them, but they were buffed to 42K (!) HP so ignoring them makes sense. I haven't been able to get my Hunters to properly prioritize Sporebats though, maybe because they are just out of range and Hunters are not properly moving to get them in range, perhaps due to the Z-axis factor, but I'm not sure. I see them attack a Sporebat here or there but tend to just go back to shooting Vashj. It sounds like you had Hunters actually focusing the Sporebats consistently, though?

Nice call about the Static Charge overriding Toxic Spores movement. That multiplier needs to be modified not to override--clear mistake and easy fix.

Also, I've had a couple of cases where bots got stuck inside of the generators somehow and had to be summoned out. They were just some random ranged DPS during Phase 2 that shouldn't have been executing any particular method at the time beyond picking targets so I don't think those were strategy-related bugs. Did you ever run into anything like that?

brighton-chi:
Oh, another question about core passing--did you notice any bots ever appear to throw the core twice in quick succession to another bot, or a bot throw a core before it had received a core? I have had a lot of difficulty with synching the visual effects (because the core passing is a combination of passing it and then separately creating the item on the receiver as a workaround because the actual bot-to-bot pass doesn't give the receiver the core).

I think I've fixed it such that bots do not attempt to throw the core multiple times so hopefully you didn't see any of that. But I think what I implemented is not properly introducing delay between steps of the chain (so the pass is not instant from bot-to-bot, which breaks immersion IMO). I don't know why that part is not working so need to figure it out.

Timberpoes:
> I can do a method for Leo that triggers if you don't have a Warlock tank and makes bots other than the current tank get away from demon Leo if they have some number of Chaos Blast stacks. That sounds like low-hanging fruit that could benefit some players a lot. But how many stacks is too many? 4? 5? I didn't realize until today that WotLK totally changed the resistance system, and I haven't read into it enough to understand it yet.

Truthfully not using a ranged tank is a sub-optimal strategy to begin with so it's difficult to complain about losing some melee DPS to inefficiency, but could it be a dynamic trigger based on percent health on getting hit by the splash from the bolt when not tanking? That way you could have the bots back off if a hit takes them to 50% or lower health which should let melee try to pull off as much DPS as possible within the limits of their own gear and the raid's healing/melee setup.

If not, I'd have to go in and test. Obviously some beefy bots like DPS warriors with respoectable health pools survived the entire demon phase soaking every AoE. But it kept my healers very busy. And I've done the fight with my ~10k health warlock with zero resist gear tanking it too without the warlock dying. So the bots clearly aren't reaching being 1-shot status, but with enough of them tanking massive hits it causes attrition. Could also have them back off halfway through the phase to dump stacks as another example of a time-based approach.

> My deaths during Phase 2 have tended to be because a tank did not pick up a Strider or Naga before it kills some cloth DPS (the fight actually got a lot better for me once I started having my Warlocks Soulstone themselves).

I bypassed this problem as I'm one of the tanks, but a low hanging fruit is gonna be tank positioning when not tanking adds. They should return to the center so they're positioned to pick up a Naga or Strider from any side ASAP, currently they just stay in place when the add dies, which causes issues if they pick up an add near the edge of the arena and one spawns the opposite side. It'll have a field day on DPS or healers until the tank hauls ass to get in taunt range.

If there's no simple enough solution it's probably not worth gunning for perfection, though. It's not a deal breaker, just a nice QoL addition.

> For the soft lock when core passing, do you remember if the third bot was near the core that was supposed to be used or not? If the bot was not responding to anything,

The bot was just not quite in range of the generator, I believe. It was not responding to any playerbot commands sent to chat until I GM-deleted the core from their inventory.

> It sounds like you had Hunters actually focusing the Sporebats consistently, though?

I think central sporebats were being killed, the ones orbiting far on the outside weren't. But it definitely felt like there were fewer bats.

> Also, I've had a couple of cases where bots got stuck inside of the generators somehow and had to be summoned out. They were just some random ranged DPS during Phase 2 that shouldn't have been executing any particular method at the time beyond picking targets so I don't think those were strategy-related bugs. Did you ever run into anything like that?

Nothing witnessed here.

> Oh, another question about core passing--did you notice any bots ever appear to throw the core twice in quick succession to another bot, or a bot throw a core before it had received a core?

Same, nothing that seemed untoward. The throw chains seemed to flow well and didn't look unrealistic in that sense.

brighton-chi:
Oh interesting, you've had melee eat the Chaos Blasts for the entire phase without dying? I don't know how AC is calculating resistance, but I can't imagine it's accurate. Even under the WotLK formula (which gives you more value per-point early on I believe), you shouldn't be able to tank Leo with no resistance. I have done it without resistance as well, though I've also had my Warlock die without resistance. My understanding of the (intended) Chaos Blast mechanic is that every stack causes the next hit to do an additional 1675 damage, and the spell is a binary resist so depending on gear you'd be able to take probably 7 or 8 hits as a Warlock before getting one-shot, and he casts the spell close to 20 times during the phase. So theoretically you should need a significant amount of resistance to make it through. The debuff lasts for 45 seconds and the entire phase is 60 seconds so you shouldn't be able to drop stacks. I think I need to pay some attention to trying to understand how the spell is working in AC, though (not a high priority obviously but I'm just curious now).

The tanks _should_ be attacking Elementals when not tanking, but I wonder now if it would be better to have them wait for Naga/Striders rather than falling back to attacking Elementals. Right now, the way the second phase works is melee (including tanks) will initiate combat on anything within 50 yards of the center, and all bots that are not actually doing anything and are more than 35 yards from the center should run back to within 30 yards of the center. So if I were to take Elementals off the valid targets table for tanks, they should come back to within 30 yards if not attacking a Naga or Strider. Theoretically I could make them attack Elementals only if they're within 30 yards of Vashj but then go after Naga or Striders farther than that, though that's a little complicated based on the current structure of the method so maybe not worth it. I'll give this some thought.

For the softlock, that's helpful info. There is probably an edge case in which the bots decide that only 2 or 3 passers are needed but in reality another one is and the last passer falls just short of the generator, and that might be what happened since you didn't see a fourth passer active.

TBH, one of the most irritating things about testing is not having good enough damage meter info. Skada works pretty well but just doesn't provide enough information for the purposes of working on raid strats. Details! does have really comprehensive information, but I find the Felbite version that is supposed to be WotLK-compatible to instead be a broken mess.

Timberpoes:
Huh. I'm using Details! with the WotLK HD client and it's fine for me. No issues and it's what I've been grabbing fight data from.

brighton-chi:
I redid Leotheras so bots can kill their Inner Demons without a cheat. It would be awesome if somebody could test. I also have Hunters killing Sporebats again on Phase 3 of Vashj, and I think I've solved them flying into the air, so testing that would be great too.

For the Inner Demons, the way it works is that (1) for all of phase 2, tanks will swap to DPS strategies (e.g., feral tanks should shift into cat form), and they'll swap back to tank strategies in other phases or when out of combat; (2) when Inner Demons appear, if a healer gets an Inner Demon, it will switch to what I thought was the closest DPS strategy (e.g., Elemental Shaman strategy for Resto Shamans) and attack the demon, and it will switch back again after the demon is killed (or otherwise if the demon is gone or if the bot is out of combat), and (3) all bots with Inner Demons are specifically directed to attack their Inner Demons and no longer have Leotheras as a valid target during that time. I am hoping testing can focus not just on how successful bots are at killing Inner Demons, but even moreso on the strategy swaps and if bots are switching properly, including back to their normal strategies. Also, I'm hoping to have testing of when Leotheras is pushed into Phase 3 while Inner Demons are still active.

brighton-chi:
Moving off of draft. New Leotheras method is still untested except for by me, but I am fine with it as is if nobody ends up testing.

Please note the manual packet usage in LadyVashjLootTaintedCoreAction::Execute() and LadyVashjPassTheTaintedCoreAction::UseCoreOnNearestGenerator(). I tried to provide some context/explanation for those usages in the first two posts, but I freely admit I have no idea what I am doing for those and just copied existing code with some AI assistance to modify. That code will need to be carefully reviewed, and I will not be able to reflect or properly respond to any conceptual comments with respect to that code.

src/Ai/Raid/SerpentshrineCavern/Trigger/RaidSSCTriggers.h
```diff
@@ -0,0 +1,422 @@
...
+
+class LeotherasTheBlindBossIsInactiveTrigger : public Trigger
+{
+public:
+    LeotherasTheBlindBossIsInactiveTrigger(
+        PlayerbotAI* botAI) : Trigger(botAI, "leotheras the blind boss is inactive") {}
+    bool IsActive() override;
+};
+
+class LeotherasTheBlindHumanFormEngagedByMainTankTrigger : public Trigger
```
kadeshar: That class dont have implementation

src/Ai/Raid/SerpentshrineCavern/RaidSSCHelpers.h
```diff
@@ -0,0 +1,201 @@
...
+    void MarkTargetWithSquare(Player* bot, Unit* target);
+    void MarkTargetWithStar(Player* bot, Unit* target);
+    void MarkTargetWithCircle(Player* bot, Unit* target);
+    void MarkTargetWithDiamond(Player* bot, Unit* target);
+    void MarkTargetWithTriangle(Player* bot, Unit* target);
+    void MarkTargetWithCross(Player* bot, Unit* target);
+    void MarkTargetWithMoon(Player* bot, Unit* target);
+    void SetRtiTarget(PlayerbotAI* botAI, const std::string& rtiName, Unit* target);
+    bool IsInstanceTimerManager(PlayerbotAI* botAI, Player* bot);
+    Unit* GetFirstAliveUnitByEntry(PlayerbotAI* botAI, uint32 entry);
```
kadeshar: In new structure i would see RaidBaseHelpers to not require reimplementing this method every raid especial that similar method exists in other helpers

src/Ai/Raid/SerpentshrineCavern/Util/RaidSSCHelpers.h
kadeshar, line 1: File should be in Util folder like for example Karazhan

brighton-chi:
Reverting to draft, pending https://github.com/mod-playerbots/mod-playerbots/pull/2069

brighton-chi:
Note this will not compile on the main branch right now because it is contingent on https://github.com/mod-playerbots/mod-playerbots/pull/2069

Celandriel:
I was hoping changing the base branch would sort that out.
Either way, ready to review?

brighton-chi:
> I was coping changing the base branch would sort that out.
> Either way, ready to review? 

Sorry edited my post to reference the main branch. I'm not sure what is the issue with respect to the test branch. I'll take a look after I put my daughter to bed.

Celandriel:
I don't think the checks are building using  test stage branch but master, so the check won't pass.

brighton-chi:
> I don't think the checks are building using test stage branch but master, so the check won't pass.

Oh ok, then I think it's fine.

src/Ai/Raid/SerpentshrineCavern/Multiplier/RaidSSCMultipliers.cpp
```diff
@@ -0,0 +1,803 @@
...
+        dynamic_cast<FollowAction*>(action) ||
+        dynamic_cast<CastKillingSpreeAction*>(action) ||
+        dynamic_cast<CastReachTargetSpellAction*>(action))
+        return 0.0f;
+
+    return 1.0f;
+}
+
+// If raid cheat (which enables bot looting of the core) is not enabled
+// Bots should not loot the core
```
kadeshar: Confusing comment. Based on implementation if cheat is enabled then bots dont loot the core

brighton-chi: Yeah that doesn't make any sense. I think originally I was thinking that if the player didn't use the cheat, then the player would have to loot the core so you wouldn't want bots to loot it. But really you don't want bots to ever execute LootAction during the Vashj fight. The only time they should loot the core is specifically in accordance with LadyVashjLootTaintedCoreAction.

Going to change this. And push an update to comment out unused parameters.

kadeshar:
Ok, will merge after checks pass :) Great job with this PR

PR #1906, opened 2025-12-08
Fix: Allow bots to duel in PVP prohibited areas

NoxMax:
Noticed that if you ask a bot to duel in a PVP prohibited area, it will accept, and do nothing. I thought about making the bot reject the request, but if you (the real player) want to duel with it, the duel should happen. This is just a minor fix to allow bots to duel if you ask them to in such areas.

Tested with bots in party, random bots of the same faction, and random bots of the opposite faction. All behaved the same before and after fix. An example place to test is Zim'Torga in Zul'Drak which is by default is a PVP prohibited area.

- Before fix, you challenge a bot, they accept and turn red, then they either just stay where they are or wander off.

- After fix, bot attacks you within the PVP prohibited area when the duel starts.

### DIFF
--- a/src/strategy/actions/AttackAction.cpp
+++ b/src/strategy/actions/AttackAction.cpp
@@ -15,7 +15,7 @@
 #include "SharedDefines.h"
 #include "Unit.h"
 
-bool AttackAction::Execute(Event event)
+bool AttackAction::Execute(Event /*event*/)
 {
     Unit* target = GetTarget();
     if (!target)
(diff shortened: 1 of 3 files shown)


hermensbas (dismissed):

Celandriel (approved):

Celandriel (commented):
I tested this and it works just fine as anticipated. 
Im a simple guy, so commenting out the events seems out of style with the rest of the code base...

Issue #1903, opened 2025-12-05
The entire log is littered with one error


Regrad:
**Bug Description**
Created it again. since the last bug report was closed, however, the problem remains.

The server log is flooded with this message. This has never happened before. This behavior began to manifest itself on all three realms at once.

```
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x000000000000741e Type: Player Low: 29726
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
Attempted sending heartbeat with root flag for guid GUID Full: 0x0000000000001ca6 Type: Player Low: 7334
[...]

Issue #1902, opened 2025-12-04
When the robot is frozen in place by a spell, its movement will be affected


gutoukuaiken:
**Bug Description**
When the bots is frozen in place by a spell, and the spell effect disappears, the robot's movement will follow you as if it were teleporting. The more times it freezes, the greater the probability of its occurrence, especially in the "Halls of Reflection" where the probability is extremely high. At the same time, a red warning will appear in the server log: "Attempted to send heartbeat with root flag for guid GUID Full: 0x00000000000005af Type: Player Low: xxxx". This warning was generated from the source code file unit.cpp of azerothcore.

void Unit::SendMovementFlagUpdate(bool self /* = false */)
{
    if (IsRooted())
    {
        // each case where this occurs has to be examined and reported and dealt with.
        LOG_ERROR("Unit", "Attempted sending heartbeat with root flag for guid {}", GetGUID().ToString());
        return;
    }

    WorldPacket data;
    BuildHeartBeatMsg(&data);
    SendMessageToSet(&data, self);
}

**Commit Hash**
All latest

**System**
OS: Ubuntu 24.04

gutoukuaiken:
@privatecore This RP has caused some issues. Here is a temporary fix solution.Status: Open.
#1927

privatecore:
@gutoukuaiken yep, sh#t happens ;) Check this fix: https://github.com/mod-playerbots/mod-playerbots/issues/1927#issuecomment-3677643128

Grimfeather:
now bots no longer teleport to catch up.
they stop moving completely

this happens a lot at the beginning of the military quarter in naxx.

privatecore:
> now bots no longer teleport to catch up.

Test this fix: https://github.com/mod-playerbots/mod-playerbots/pull/1933

Grimfeather:
I've just tested it, it's still happening.
I'll try to isolate the ability that is causing it.

privatecore:
> I've just tested it, it's still happening. I'll try to isolate the ability that is causing it.

Give me coords, let's dig together.

Grimfeather:
it's the first part of the military quarter in naxx
but I can't isolate the issue.

I've been fighting enemies separately and nothing goes wrong.
but if i pack them together in a big pull, I got 10+ bots that refuse to move after the fight ends.
with a worldserver console full of heartbeat root flag errors.

it's the deathknights, deathknight captains and dark touched warriors

privatecore:
[...]

PR #1908, opened 2025-12-09
Rerwite is moving allowed logic +  fix root flag heartbeat spam

privatecore:
Okay, what have been done:

1. Fix heartbeat spam for root flag: check against `MOVEMENTFLAG_ROOT` flag (`IsRooted`) instead of `HasRootAura`.
2. Rewrite `IsMovingAllowed` - place checks from most common to the rarest.
3. Remove unnecessary checks: `HasRootAura`, `HasConfuseAura`, `HasStunAura` - handled by AuraEffects and set unit state flags `UNIT_STATE_ROOT`, `UNIT_STATE_CONFUSED`, `UNIT_STATE_STUNNED` - `UNIT_STATE_LOST_CONTROL` already handles confused and stunned (rooted checked with `IsRooted` method).
4. Combine traveling state checks for taxi flights: `UNIT_STATE_IN_FLIGHT` + MM flag `FLIGHT_MOTION_TYPE`.
5. Simplify check against being in vehicle: use `MOVEMENTFLAG_ONTRANSPORT` as an indicator that the unit is in the vehicle.

Also, update `UpdateMovementState` method with simplified checks and the updated logic (common > rare).

This should fix issues: https://github.com/mod-playerbots/mod-playerbots/issues/1903 and https://github.com/mod-playerbots/mod-playerbots/issues/1902

NOTE: The `PlayerbotAI` class has a method `CanMove` with the same checks, but this method is only used once in the code. We should decide how to properly check if the bot can move or not:

1. Place all logic into `IsMovingAllowed` and drop `CanMove`.
2. Place all logic into `CanMove` and use it inside `IsMovingAllowed`.
3. Use them for different approaches: 
   - `CanMove`: simple checks (unit flags, CC state, death state, travel state, vehicle state);
   - `IsMovingAllowed`: everything from `CanMove` + MM flags checks (not sure about rooted since it still checks for movement flags...).

### DIFF
--- a/src/strategy/actions/MovementActions.cpp
+++ b/src/strategy/actions/MovementActions.cpp
@@ -946,25 +946,37 @@ bool MovementAction::IsWaitingForLastMove(MovementPriority priority)
 
 bool MovementAction::IsMovingAllowed()
 {
-    // do not allow if not vehicle driver
-    if (botAI->IsInVehicle() && !botAI->IsInVehicle(true))
+    // Most common checks: confused, stunned, fleeing, jumping, charging. All these
+    // states are set when handling certain aura effects. We don't check against
+    // UNIT_STATE_ROOT here, because this state is used by vehicles.
+    if (bot->HasUnitState(UNIT_STATE_LOST_CONTROL))
         return false;
 
-    if (bot->isFrozen() || bot->IsPolymorphed() || (bot->isDead() && !bot->HasPlayerFlag(PLAYER_FLAGS_GHOST)) ||
-        bot->IsBeingTeleported() || bot->HasRootAura() || bot->HasSpiritOfRedemptionAura() || bot->HasConfuseAura() ||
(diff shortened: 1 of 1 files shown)


Celandriel:
My thinking is to remove CanMove and replace where its implemented. Otherwise it makes sense to me to replace the independent checks with the core root checking function and align the two.

privatecore:
> My thinking is to remove CanMove and replace where its implemented. Otherwise it makes sense to me to replace the independent checks with the core root checking function and align the two.

I'm also thinking about removing it, but then I saw **hermensbas'** (`UpdateMovementState`) code with just a few checks. That got me to the point: the `CanMove` method can be used for simple checks, while `IsMovingAllowed` can handle more complex situations that involve some expensive calls (MM, auras, etc.)

Celandriel (approved):

PR #1909, opened 2025-12-09
feat. Enable bots to respond to GM messages.

Celandriel:
I found an existing config option that allows bots to speak without a master, and extended its use so that protected commands (nc, co) are taken from a bot by GM players. This is primarily a debugging tool and shouldnt be on generally. 

My testing of this is the following. 
With the option off, no error is seen, but no text output is shown. 

With the option on 
GM accounts (Whether GM is on or off) will respond to nc and co commands by just saying the output directly. Strategies can be changed by GMs for bots whom they are not masters for (Note, im pretty sure that a GM could change the strategies even before this change, you just couldnt see the output of that.)

For non GM characters you get "Invite me to your group first" with the option on/off.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -2771,8 +2771,7 @@ bool PlayerbotAI::TellMaster(std::string const text, PlayerbotSecurityLevel secu
     if (!master)
     {
         if (sPlayerbotAIConfig->randomBotSayWithoutMaster)
-            TellMasterNoFacing(text, securityLevel);
-        return false;
+            return TellMasterNoFacing(text, securityLevel);
     }
     if (!TellMasterNoFacing(text, securityLevel))
         return false;


hermensbas (approved):

src/PlayerbotAI.cpp
```diff
@@ -2769,7 +2768,13 @@ bool PlayerbotAI::TellMaster(std::ostringstream& stream, PlayerbotSecurityLevel
 
 bool PlayerbotAI::TellMaster(std::string const text, PlayerbotSecurityLevel securityLevel)
 {
-    if (!master || !TellMasterNoFacing(text, securityLevel))
+    if (!master)
+    {
+        if (sPlayerbotAIConfig->randomBotSayWithoutMaster)
+            TellMasterNoFacing(text, securityLevel);
```
kadeshar: That shouldnt return true when is executed?

src/PlayerbotAI.cpp
```diff
@@ -910,8 +910,7 @@ void PlayerbotAI::HandleCommand(uint32 type, std::string const text, Player* fro
         fromPlayer->SendDirectMessage(&data);
         return;
     }
-
-    if (!IsAllowedCommand(filtered) &&
+    if (!IsAllowedCommand(filtered) && //!isFromGM &&
```
kadeshar: I think this comment is to delete

src/PlayerbotAI.cpp
Celandriel: It will return true, but I didnt bother to catch and store the value. Maybe its bad style to do that.

kadeshar: I dont understand. How it will return true? In next line is return false

Celandriel: I see your point. I can just return the value it gives.

hermensbas (approved):

Issue #1846, opened 2025-11-16
Disable bot XP gain when master has XP turned off


HennyWilly:
**Is your feature request related to a problem? Please describe.**
When the master disables experience gain using the “Experience Eliminator” NPCs ([Behsten](https://www.wowhead.com/wotlk/npc=35365/behsten) and [Slahtz](https://www.wowhead.com/wotlk/npc=35364/slahtz)), the bots in their group/raid continue receiving XP. This causes bots to out-level the master or gain unintended progression while the master remains locked.

**Describe the solution you'd like**
Bots should automatically inherit the master’s XP-gain state. If the master has XP disabled via an Experience Eliminator NPC, bots should receive no experience from any source until the master re-enables it.

**Describe alternatives you've considered**
I currently use `.levelup [$playername] -1` every time a bot levels up. This is error-prone and time-consuming if multiple bots level up at the same time. It may also reset the bot's talents.

**Additional context**
I know this is a niche feature, but it would significantly improve consistency for players who intentionally keep their level fixed, especially for fixed-level gameplay like raiding in classic or BC.

NoxMax:
There is also `RandomBotFixedLevel`, but of course that is server wide.

HennyWilly:
True, but as you said, `RandomBotFixedLevel` is server wide. That's not really what I had in mind.

I did a little digging and found that both experience eliminator NPCs have a creature script called "npc_experience".
This script is implemented [here](https://github.com/mod-playerbots/azerothcore-wotlk/blob/8f84271af88785badaaafc9d5e58494a3e94fea5/src/server/scripts/World/npcs_special.cpp#L2003).
It looks like players who disable XP are marked with `PLAYER_FLAGS_NO_XP_GAIN`.
I'm wondering if it's enough to set `amount` to 0 in [OnPlayerGiveXP](https://github.com/mod-playerbots/mod-playerbots/blob/bb569b4d3987b7f15e702620eedb22d2ed5cc269/src/Playerbots.cpp#L239) if there is a player master that has the `PLAYER_FLAGS_NO_XP_GAIN` flag.

HennyWilly:
It seems that #1947 reverted the changes of #1910 due to performance issues.
I'm reopening this ticket so the issue doesn't get forgotten.
Maybe someone else can resolve the issue without affecting performance.

NoxMax:
[...]

PR #1910, opened 2025-12-09
Disable bot XP gain when master has XP turned off

HennyWilly:
Addresses #1846

This PR disables XP gain for bots whose non-bot master has disabled XP gain via an "Experience Eliminator" NPC.

If the current master has disabled XP gain, randombots in the group and addclass-bots belonging to the player won't gain XP.
Randombots not grouped with a player will continue to gain XP as normal.

Discussed points:
- `Should this feature only be enabled via a new configuration value?`: Comments currently tend towards no config value.

Open points:
- Should bots be allowed to gain XP until they reach the player's level, even if the player has disabled XP gain?

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -238,12 +238,22 @@ public:
 
     void OnPlayerGiveXP(Player* player, uint32& amount, Unit* /*victim*/, uint8 /*xpSource*/) override
     {
-        // early return
-        if (sPlayerbotAIConfig->randomBotXPRate == 1.0 || !player)
+        // no XP multiplier, when player is no bot.
+        if (!player || !player->GetSession()->IsBot())
             return;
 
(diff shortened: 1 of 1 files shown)


Celandriel:
I wonder if this behavior should be configurable.

kadeshar:
I terms of reason sounds good for me even without config. Then you can twink in group.

HennyWilly:
I agree with the reasoning here and I think this behavior makes sense as a default.
If someone prefers to have this configurable, I can add a config option with the current behavior enabled by default.

One question I had while implementing this: Should bots still be able to gain XP if their level is beneath the master’s level, even if XP is disabled for the master? Just to enable them to "catch up" to the master. Or would that just introduce unnecessary complexity?

Dreathean:
I definitely see how adding that could be useful, but I agree I think that would introduce unnecessary complexity.

PR #1843, opened 2025-11-14
Core - Fix freeze/crash when using "spells" command in MapUpdater worker thread

Wishmaster117:
This will fix Crash 3 Source in issue [#1840](https://github.com/mod-playerbots/mod-playerbots/issues/1840) found in logs posted by @Regrad

## Summary

Fix a freeze/crash caused by the `spells` command when executed in a `MapUpdater` worker thread. The old `CompareSpells` implementation was extremely expensive and could block `Map::Update` for more than 60 seconds, triggering the `FreezeDetector` and `Acore::Abort()`.

## What was wrong

Call path on crash:

- World::Update → MapMgr::Update → MapUpdater::wait()
- MapUpdater::WorkerThread → MapUpdateRequest::call → Map::Update → Player::Update
- ScriptMgr::OnPlayerAfterUpdate → PlayerbotsPlayerScript::OnPlayerAfterUpdate
- botAI->UpdateAI → PlayerbotAI::DoNextAction → Engine::DoNextAction
- Engine::ListenAndExecute → ListSpellsAction::Execute → GetSpellList / CompareSpells

`CompareSpells` rescanned the entire `sSkillLineAbilityStore` DBC table for every comparison inside `std::sort`, which becomes extremely slow on bots with many spells.

## What this PR does

- Replace the heavy `CompareSpells` with a **cheap comparator**:
  - Uses `SpellInfo` only (school + spell name) for sorting.
  - No more `sSkillLineAbilityStore` scan inside the comparator.
- Keep sending the **full spell list** to the master so existing addons that parse chat still work.
- Make filter parsing a bit safer:
  - No out-of-bounds on `spells first` vs `spells first aid`.
  - Support `spells +tailoring`, `spells tailoring+`, etc. for craftable-only filters.
  - Return `"No spells found."` when nothing matches.
  
### Testing

Tested locally on an AzerothCore WotLK 3.3.5a server with mod-playerbots:
**Functional:**
- spells
- spells first aid
- spells tailoring
- spells 20-40
- spells +tailoring
- spells head
- spells qsdqsdqsd (returns No spells found. as expected)

**Scenario with bots having:**
- multiple professions (Tailoring, Blacksmithing, First Aid, etc.),
- many known spells and recipes,
- various reagents in bags and vendor-only reagents.

**Verified that:**
- the full spell list is still sent to the master chat,
[...]

PR #1912, opened 2025-12-12
Followup #1843 Pre-initialize ListSpellsAction caches at world startup

Wishmaster117:
## Summary

This is a small follow-up to the previous `ListSpellsAction` / `CompareSpells` changes.

The goal here is to make the `spells` command more core-friendly by moving the heavy cache initialization (DBC scan + `npc_vendor` query) to world startup, instead of doing it on demand from MapUpdater worker threads.
There is no functional change to how `spells` works in game. It should still show the same spells and the same formatting as before.

## What this PR changes

- Add a static method:
`  static void ListSpellsAction::InitStaticCaches();`

**which:**

fills skillSpells from sSkillLineAbilityStore,
fills vendorItems from:

`SELECT item FROM npc_vendor WHERE maxcount = 0;`

Guard this initialization with std::once_flag so that caches are built only once, even if multiple threads call it.
Call ListSpellsAction::InitStaticCaches() from PlayerbotsWorldScript::OnBeforeWorldInitialized(), so the caches are pre-warmed on the main world thread during startup.
Keep a call to InitStaticCaches() at the beginning of GetSpellList() as a safety net, in case the world startup hook is not executed for some reason. In normal case this is a no-op thanks to std::call_once.

Previously, the first call to GetSpellList() would:
scan the whole SkillLineAbility DBC, run a synchronous WorldDatabase.Query("SELECT item FROM npc_vendor ..."), and this could happen from a MapUpdater worker thread when a bot receives a spells command.
Now, those expensive operations are done once at world startup, on the world thread.

**Testing**

On my local setup I did:

full server restart and watched the logs:
saw the “ListSpellsAction caches initialized at world startup” log once,
no errors during world initialization;

tested in game on several bots:
spells
spells first aid
spells tailoring
spells 20-40
spells +tailoring
spells head
spells qsdqsdqsd (still returns No spells found.);

checked that spell outputs and formatting are the same as before.

I was not able to see any anomaly on my local server, but this is of course very small scale.
**It would be very helpful if somebody could try this on a larger, more busy server (many bots + real players), just to confirm there is no hidden behavior change and that the startup cache works fine under real load.**

### DIFF
--- a/src/strategy/actions/ListSpellsAction.cpp
+++ b/src/strategy/actions/ListSpellsAction.cpp
@@ -93,35 +93,6 @@ void ListSpellsAction::InitStaticCaches()
 
 std::vector<std::pair<uint32, std::string>> ListSpellsAction::GetSpellList(std::string filter)
 {
-/*    if (skillSpells.empty())
-    {
-        for (uint32 j = 0; j < sSkillLineAbilityStore.GetNumRows(); ++j)
-        {
-            if (SkillLineAbilityEntry const* skillLine = sSkillLineAbilityStore.LookupEntry(j))
-                skillSpells[skillLine->Spell] = skillLine;
-        }
-    }
-
-    if (vendorItems.empty())
-    {
-        QueryResult results = WorldDatabase.Query("SELECT item FROM npc_vendor WHERE maxcount = 0");
-        if (results)
-        {
-            do
-            {
-                Field* fields = results->Fetch();
-                int32 entry = fields[0].Get<int32>();
-                if (entry <= 0)
-                    continue;
-
-                vendorItems.insert(entry);
-            } while (results->NextRow());
-        }
-    }*/
-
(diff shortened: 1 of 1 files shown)


Wishmaster117:
@hermensbas can you take a look if this is what you asking for, when you have some time?

src/strategy/actions/ListSpellsAction.cpp
```diff
@@ -8,71 +8,68 @@
 #include "Event.h"
 #include "Playerbots.h"
 
+#include <mutex>
```
Celandriel: Is this being used?

biship:
Well https://github.com/mod-playerbots/mod-playerbots/pull/1928 seems to conflict :/

Celandriel:
Yea, I took this PR and adjusted some things to get it over the fence in #1931

PR #1913, opened 2025-12-12
Refactor guild managment into a singleton

Celandriel:
The idea is to centralize the creation, assignment, and management of bot guilds into a single class that can be referenced. 

The way this is intended to work.

when the manager is created, if the config option to delete guilds is set, then it deletes all bot guilds. 

On startup 
1. Load all guild names from database. Shuffle keys for some randomization. 
2. Load Guilds from database
3. For existing guilds, identify the guild faction, number of members, and assess if the guild is 'full' based on the number of bots set in config. 
4. Determine if the leader of the guild is a real player based on the leader account.
5. Mark any playerbot guild names as not available (false).

The validation process (2-5) is set to run once an hour. 

Guild Creation.
Now guild creation occurs on an as needed bases during the initialization process. Previously, all of the guilds would be created at once, and then randomly assigned. 
When a bot is not in a guild during initialization, it will check if there are any partially filled guilds of that bots faction where the bot can be assigned to. If not, and the cache of bot guilds is less than the set number in config, it will randomly return the available name. This then goes to the CreateGuild function where the core guild manager creates a guild, the guild emblem is set, and the cache updated. 
If a bot is assigned to guild, but fails to join then it throws an error. 

Checking for real player guilds function now lives in the guild manager.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -514,7 +514,6 @@ void AddPlayerbotsScripts()
     new PlayerbotsWorldScript();
     new PlayerbotsScript();
     new PlayerBotsBGScript();
-  
     AddPlayerbotsSecureLoginScripts();
     AddSC_playerbots_commandscript();
     PlayerBotsGuildValidationScript();


src/PlayerbotGuildMgr.cpp
```diff
@@ -0,0 +1,327 @@
...
+#include "DatabaseEnv.h"
+#include "Guild.h"
+#include "GuildMgr.h"
+#include "RandomPlayerbotMgr.h"
+#include "ScriptMgr.h"
+
+PlayerbotGuildMgr::PlayerbotGuildMgr()
+{
+    _randomBotGuildCount = sPlayerbotAIConfig->randomBotGuildCount;
+    _randomBotGuildSizeMax = sPlayerbotAIConfig->randomBotGuildSizeMax;
```
kadeshar: module itself have command to reload config.I think such config caching will break this.

src/PlayerbotGuildMgr.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+void PlayerbotGuildMgr::ResetGuildCache()
+{
+    for (auto it = _guildCache.begin(); it != _guildCache.end();)
+        {
+            GuildCache& cached = it->second;
+            cached.memberCount = 0;
+            cached.faction = 2;
+            cached.status = 0;
+        }
```
kadeshar: This brackets position is strange

src/PlayerbotGuildMgr.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+        void OnUpdate(uint32 diff) override
+        {
+            _validateTimer += diff;
+
+            if (_validateTimer >= _validateInterval) // Validate every hour
+            {
+                _validateTimer = 0;
+                sPlayerbotGuildMgr->ValidateGuildCache();
+                LOG_INFO("playerbots", "Schedueled guild cache validation");
```
kadeshar: Scheduled

src/PlayerbotGuildMgr.cpp
```diff
@@ -0,0 +1,327 @@
...
+            }
+            return;
```
kadeshar: redundant return

src/PlayerbotGuildMgr.cpp
Celandriel: Better to just call direct every time?

Celandriel: Fixed.

Celandriel:
To test this, I would mess with changing the number of guilds  and their sizes. Using the delete guilds option. Look around to see if bots greater than lv 10 have tabards.

src/PlayerbotGuildMgr.cpp
kadeshar: I didnt analized playerbots reload but looks more safetly

kadeshar (approved):

hermensbas:
Fixed merge conflict

TRPB:
This is almost certainly the cause of an issue I've just found: Random bots have suddenly joined my guild.

> Determine if the leader of the guild is a real player based on the leader account.

I assume this should be blocking that. There's no new config option to block this behaviour. Is it a bug? 

If not, how can I disable random bots being put in my guild? The guild was created by my player account and only contained my alts until this update

Celandriel:
It should not have put bots into a real player guild. I'll investigate.

PR #1914, opened 2025-12-13
Fix/Feat: Stop bots in party from PVP when master isn't, and PVP probablity system

NoxMax:
There are two related PVP components in this PR. First is the simple yet fundamental change to bot behaviour when they are in party. Right now bots with a master will go into PVP when there's a nearby PVP target, even if master is not in PVP. This absolutely should not happen. Bots should not consider PVP at all if master is not in PVP. The fix is only 3 lines in EnemyPlayerValue

The second component is introducing PVP probabilities, to make decisions more realistic. Right now even a level 1 bot will 100% go into PVP if it sees a level 80 PVP target. They can't help themselves. So the change here addresses that insanity. Several thresholds (subject to community review) are introduced:

1. Bots will not fight a target 5 or more levels higher than them
2. Bots have a 25% chance starting a fight with a target +/- 4 levels from them.
3. Bots have a 50% chance starting a fight with a target +/- 3 levels from them.
4. Bots have a 75% chance starting a fight with a target +/- 2 levels from them.
5. Bots have a 100% chance starting a fight with a target +/- 1 level from them.
6. Bots have a 25% chance starting a fight with a target 5 or more levels below them (ganking. thought it would be funny, and technically realistic of player behaviour)

Exception of course exist for BG/Arena/Duel, and in capitals where bots will always PVP. Also bots will always defend themselves if attacked.

Few notes: 
1. The if/ else if logic can be further simplified, but only if we use thresholds that are different by one. So current logic allows for flexibility of using values like 10/7/5/3 instead of 5/4/3/2.
2. The caching system is per-bot basis. So for some target X, if some bot decides to attack it, another bot will make its own decision. At first I used a simplified global system (thinking there might be performance concerns) where if one bot decides to attack a target then they all do, but when I switched to the more realistic per-bot basis, I didn't see an effect on performance.
3. Variables are obviously not configurable right now. I'm starting to see Bash's POV that maybe we have too many configs :grimacing: Still, they can be easily exposed in the future, and if someone is reading this then, remember to change constexpr  to const.

### DIFF
--- a/src/strategy/values/PossibleTargetsValue.cpp
+++ b/src/strategy/values/PossibleTargetsValue.cpp
@@ -25,8 +25,9 @@ constexpr int32 HIGH_LEVEL_DIFF = 4;     // 25% chance at +/- this difference
 constexpr int32 MID_LEVEL_DIFF = 3;      // 50% chance at +/- this difference
 constexpr int32 LOW_LEVEL_DIFF = 2;      // 75% chance at +/- this difference
 
-// Cache duration before reconsidering attack decision
+// Cache duration before reconsidering attack decision, and old cache cleanup interval
 constexpr uint32 ATTACK_DECISION_CACHE_DURATION = 2 * MINUTE;
+constexpr uint32 ATTACK_DECISION_CACHE_CLEANUP_INTERVAL = 10 * MINUTE;
 
 // Custom hash function for (botGUID, targetGUID) pairs
 struct PairGuidHash
@@ -49,8 +50,29 @@ void PossibleTargetsValue::FindUnits(std::list<Unit*>& targets)
     Cell::VisitObjects(bot, searcher, range);
 }
 
+static void CleanupAttackDecisionCache()
+{
+    time_t currentTime = time(nullptr);
+    for (auto it = attackDecisionCache.begin(); it != attackDecisionCache.end();)
+    {
+        if (currentTime - it->second.second >= ATTACK_DECISION_CACHE_DURATION)
+            it = attackDecisionCache.erase(it);
+        else
+            ++it;
+    }
+}
+
 bool PossibleTargetsValue::AcceptUnit(Unit* unit)
 {
+    // attackDecisionCache cleanup
+    static time_t lastCleanup = 0;
+    time_t currentTime = time(nullptr);
+    if (currentTime - lastCleanup > ATTACK_DECISION_CACHE_CLEANUP_INTERVAL)
+    {
+        CleanupAttackDecisionCache();
+        lastCleanup = currentTime;
+    }
+
     if (!AttackersValue::IsPossibleTarget(unit, bot, range))
         return false;
 
@@ -123,7 +145,6 @@ bool PossibleTargetsValue::AcceptUnit(Unit* unit)
         if (attackChance < 100)
         {
             std::pair<ObjectGuid, ObjectGuid> cacheKey = std::make_pair(bot->GetGUID(), unit->GetGUID());
-            time_t currentTime = time(nullptr);
 
             auto it = attackDecisionCache.find(cacheKey);
             if (it != attackDecisionCache.end())


src/strategy/values/PossibleTargetsValue.cpp
```diff
@@ -24,7 +48,97 @@ void PossibleTargetsValue::FindUnits(std::list<Unit*>& targets)
...
+        if (bot->InBattleground() || bot->InArena())
+            return true;
+
+        // Skip restrictions if in duel with this player
+        if (bot->duel && bot->duel->Opponent == unit)
+            return true;
+
+        // Capital cities - no restrictions
+        uint32 zoneId = bot->GetZoneId();
+        bool inCapitalCity = (zoneId == 1657 ||  // Darnassus
```
Celandriel: See
https://github.com/mod-playerbots/azerothcore-wotlk/blob/b9345774f94b3af2b22f74baf1b985e4f0fe0719/src/server/game/Maps/AreaDefines.h#L85 for area definitions to change from magic nums

src/strategy/values/PossibleTargetsValue.cpp
```diff
@@ -16,6 +16,30 @@
 #include "SpellAuraEffects.h"
 #include "SpellMgr.h"
 #include "Unit.h"
+#include <unordered_map>
```
Celandriel, line 20: Is this required to add?

src/strategy/values/PossibleTargetsValue.cpp
```diff
@@ -24,7 +48,97 @@ void PossibleTargetsValue::FindUnits(std::list<Unit*>& targets)
...
+            time_t currentTime = time(nullptr);
+
+            auto it = attackDecisionCache.find(cacheKey);
+            if (it != attackDecisionCache.end())
+            {
+                if (currentTime - it->second.second < ATTACK_DECISION_CACHE_DURATION)
+                    return it->second.first;
+            }
+
+            bool shouldAttack = (rand_norm() < attackChance);
```
Celandriel: Im not sure of this, but would moving to urand be better? Make all float probabilities in integers may be smaller and more performing?

src/strategy/values/PossibleTargetsValue.cpp
NoxMax: Thanks for letting me know. Will implement it today.

src/strategy/values/PossibleTargetsValue.cpp
NoxMax: Yes. It's for the dictionary:
```
// Cache for probability-based attack decisions (Per-bot: non-global)
// Map: (botGUID, targetGUID) -> (should attack decision, timestamp)
static std::unordered_map<std::pair<ObjectGuid, ObjectGuid>, std::pair<bool, time_t>, PairGuidHash> attackDecisionCache;
```

src/strategy/values/PossibleTargetsValue.cpp
NoxMax: yeah I often default to float with percentages, but you are right there is no need for it. Int values will actually look a bit cleaner. Will amend.

NoxMax:
Made a tiny PR to AC to add Silvermoon definition. It's been merged https://github.com/azerothcore/azerothcore-wotlk/pull/24199
Will update the definition to this PR when changes are brought downstream.

Celandriel (dismissed):

Celandriel (approved):

src/strategy/values/PossibleTargetsValue.cpp
```diff
@@ -24,7 +49,97 @@ void PossibleTargetsValue::FindUnits(std::list<Unit*>& targets)
...
+
+            auto it = attackDecisionCache.find(cacheKey);
+            if (it != attackDecisionCache.end())
+            {
+                if (currentTime - it->second.second < ATTACK_DECISION_CACHE_DURATION)
+                    return it->second.first;
+            }
+
+            bool shouldAttack = (urand(1, 100) <= attackChance);
+            attackDecisionCache[cacheKey] = std::make_pair(shouldAttack, currentTime);
```
kadeshar, line 157: You never clear this cache?
That not will be problematic for long running servers?

NoxMax: Not the I have noticed with my +7K server that runs 24/7, even before I upgraded it. However, it is good defence. I will look into cleaning up caches that are older than ATTACK_DECISION_CACHE_DURATION every... 15 minutes. I don't expect there would be any measurable overhead if I cleaned every 5, likewise it wouldn't matter even to wait 30 minutes. So 15 sounds good?

kadeshar: Any cleanup will be ok

NoxMax: went with a flat 10 minutes. nullptr location changed because of check on top.

kadeshar (approved):

noisiver (approved):

hermensbas:
[worldserver_gdb_20260111_000520.log](https://github.com/user-attachments/files/24547232/worldserver_gdb_20260111_000520.log)

Crashes the server, concurrent issue on cache erase.

PR #1922, opened 2025-12-16
Remove "potion" strategy from arenas + code cleanups

brighton-chi:
Lines 481 through 484 are the only substantive changes. Bots were having the potion strategy added in arenas, and since potions are not allowed in arenas, this caused them to lock up and attempt to repeatedly drink potions when under the applicable health or mana trigger thresholds. Now they won't have the potion strategy in arenas.

Otherwise, I noticed a bunch of magic numbers for spec tabs and so went ahead and did some refactoring:
1. All references to spec tab numbers now use the appropriate constant name.
2. A bunch of extra braces were deleted.
3. DEATHKNIGHT_TAB was changed to DEATH_KNIGHT_TAB, and HUNTER_TAB_BEASTMASTERY was changed to HUNTER_TAB_BEAST_MASTERY, to reflect the correct names of the class and spec.
4. Deleted some comments that were clearly unneeded. There's much more that can be cleaned up, and probably the entire logic sequence for adding/removing strategies in AiFactory.cpp can be redone, but that's above my pay grade.


brighton-chi:
I realized that a few classes have potion strategies hardcoded into their class strategies, so removing the potion strategy won't stop them from trying to drink potions.

Preferable to pause on this PR and try to address those situations as well? Or deal with that in another PR? I think it's better to address in another PR but will obviously defer to the maintainers.

hermensbas:
> I realized that a few classes have potion strategies hardcoded into their class strategies, so removing the potion strategy won't stop them from trying to drink potions.
> 
> Preferable to pause on this PR and try to address those situations as well? Or deal with that in another PR? I think it's better to address in another PR but will obviously defer to the maintainers.

Prolly best to combine them, so we can revert as whole if needed for whatever reason?

brighton-chi:
> > I realized that a few classes have potion strategies hardcoded into their class strategies, so removing the potion strategy won't stop them from trying to drink potions.
> 
> > 
> 
> > Preferable to pause on this PR and try to address those situations as well? Or deal with that in another PR? I think it's better to address in another PR but will obviously defer to the maintainers.
> 
> 
> 
> Prolly best to combine them, so we can revert as whole if needed for whatever reason?

I think it would be harder to identify issues. Because this one is just a change to the engine for initializing strategies so potions are not initialized in arenas. Whereas to stop classes like Shamans from using potions is instead modifying their class strategies and maybe the general potion usage strategy.

PR #1923, opened 2025-12-17
NextAction refactoring to eliminate sentinel arrays and pointers

SmashingQuasar:
# Description

This PR aims to refactor the NextAction declaration to achieve two goals:

## Eliminate C-style sentinel arrays

Currently, a double pointer (`NextAction**`) approach is being used. This an old pre-C++11 (< 2011) trick before `std::vector<>` became a thing.
This approach is painful for developers because they constantly need to declare their `NextAction` arrays as:
```cpp
NextAction::array(0, new NextAction("foo", 1.0f), nullptr)
```
Instead of:
```cpp
{ new NextAction("foo", 1.0f) }
```
The first argument of `NextAction::array` is actually a hack. It is used to have a named argument so `va_args` can find the remaining arguments. It is set to 0 everywhere but in fact does nothing. This is very confusing to people unfamiliar with this antiquated syntax.
The last argument `nullptr` is what we call a sentinel. It's a `nullptr` because `va_args` is looking for a `nullptr` to stop iterating. It's also a hack and also leads to confusion.

## Eliminate unnecessary pointers for `NextAction`

Pointers can be used for several reasons, to cite a few:
- Indicate strong, absolute identity.
- Provide strong but transferable ownership (unlike references).
- When a null value is acceptable (`nullptr`).
- When copy is expensive.

`NextAction` meets none of these criteria:
- It has no identity because it is purely behavioural.
- It is never owned by anything as it is passed around and never fetched from a registry.
- The only situations where it can be `nullptr` are errors that should in fact throw an `std::invalid_argument` instead.
- They are extremely small objects that embark a single `std::string` and a single `float`.

Pointers should be avoided when not strictly necessary because they can quickly lead to undefined behaviour due to unhandled `nullptr` situations. They also make the syntax heavier due to the necessity to constantly check for `nullptr`. Finally, they aren't even good for performance in that situation because shifting a pointer so many times is likely more expensive than copying such a trivial object.

# End goal

The end goal is to declare `NextAction` arrays this way:
```cpp
{ NextAction("foo", 1.0f) }
```

> [!NOTE]
> Additional note: `NextAction` is nothing but a hacky proxy to an `Action` constructor. This should eventually be reworked to use handles instead of strings. This would make copying `NextAction` even cheaper and remove the need for the extremely heavy stringly typed current approach. Stringly typed entities are a known anti-pattern so we need to move on from those.

### DIFF
--- a/src/strategy/Action.h
+++ b/src/strategy/Action.h
@@ -23,19 +23,7 @@ public:
     std::string const getName() { return name; }
     float getRelevance() { return relevance; }
 
-    static std::vector<NextAction> clone(std::vector<NextAction> actions)
-    {
-        std::vector<NextAction> clone = {};
-
-        for (NextAction const& action : actions)
-        {
-            clone.push_back(action);
-        }
-
-        return clone;
-    };
-
-    static std::vector<NextAction> merge(std::vector<NextAction> what, std::vector<NextAction> with)
+    static std::vector<NextAction> merge(std::vector<NextAction> const& what, std::vector<NextAction> const& with)
     {
         std::vector<NextAction> result = {};
 
@@ -116,15 +104,15 @@ public:
 
     std::vector<NextAction> getContinuers()
     {
-        return NextAction::merge(NextAction::clone(continuers), action->getContinuers());
+        return NextAction::merge(this->continuers, action->getContinuers());
     }
     std::vector<NextAction> getAlternatives()
     {
-        return NextAction::merge(NextAction::clone(alternatives), action->getAlternatives());
+        return NextAction::merge(this->alternatives, action->getAlternatives());
     }
     std::vector<NextAction> getPrerequisites()
     {
-        return NextAction::merge(NextAction::clone(prerequisites), action->getPrerequisites());
+        return NextAction::merge(this->prerequisites, action->getPrerequisites());
     }
 
 private:


yqsy2025:
It is indeed necessary to optimize the performance in this way. I fully support you—keep up the good work!

SmashingQuasar:
I am done with this refactoring. The PR is now ready for review. I've tested it locally with bots running around and doing their thing. They behaved normally and the server did not crash. Anyone will to test this is more than welcome.

kadeshar:
PR dont passed codestyle check. Please fix that.

notOrrytrout:
This looks cool. I think the easier it is for developers to understand the better. What needs to be looked at in testing? That the bots just have the next action working correctly?

SmashingQuasar:
> This looks cool. I think the easier it is for developers to understand the better. What needs to be looked at in testing? That the bots just have the next action working correctly?

Testing is a bit tricky as we need to check that bots have no visible change in behaviour. I tested it myself and couldn't really see any change. Maybe I am missing something since I do not have as much experience with bots behaviour as some people here do.

hermensbas:
Ye its tricky one to test for sure, dont feel comfortable enough whether there behavior changes or not. Maybe someone else can test that who is using the bots more from a player perspective?

Or we simply do general code review, best effort in testing, merge and keep monitoring? I am gonna hold back by that approach since there are issues with warrior AI / shaman AI scout/totem spamming, movement etc. Not sure entirely sure what would be wise.

src/strategy/Action.h
```diff
@@ -24,15 +23,38 @@ class NextAction
     std::string const getName() { return name; }
     float getRelevance() { return relevance; }
 
-    static uint32 size(NextAction** actions);
-    static NextAction** clone(NextAction** actions);
-    static NextAction** merge(NextAction** what, NextAction** with);
-    static NextAction** array(uint32 nil, ...);
-    static void destroy(NextAction** actions);
+    static std::vector<NextAction> clone(std::vector<NextAction> actions)
```
hermensbas: shouldnt actions be const and as reference?

const std::vector<NextAction>& actions

src/strategy/Action.h
```diff
@@ -24,15 +23,38 @@ class NextAction
...
+
+        for (NextAction const& action : actions)
+        {
+            clone.push_back(action);
+        }
+
+        return clone;
+    };
+
+    static std::vector<NextAction> merge(std::vector<NextAction> what, std::vector<NextAction> with)
```
hermensbas: Same?

static std::vector<NextAction> merge(
    const std::vector<NextAction>& what,
    const std::vector<NextAction>& with
)

src/strategy/Action.h
```diff
@@ -73,39 +95,44 @@ class Action : public AiNamedObject
...
-    }
+    ActionNode(
+        std::string name,
+        std::vector<NextAction> prerequisites = {},
+        std::vector<NextAction> alternatives = {},
+        std::vector<NextAction> continuers = {}
+    ) :
+    name(std::move(name)),
+    action(nullptr),
+    continuers(continuers),
```
hermensbas, line 94: Accept vectors by value and move them into members? e.g:

  continuers(std::move(continuers)),
  alternatives(std::move(alternatives)),
  prerequisites(std::move(prerequisites))

SmashingQuasar:
I rebased the PR on the latest `master` to avoid accumulating conflicts.

Celandriel:
I ran this for about 40 min without noticing any difference.

hermensbas:
privatecore
yep + heap alloc and overall performance (a bit faster)

```
------------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations   AllocAvg   AllocMax     Allocs      Bytes
------------------------------------------------------------------------------------------------------------------------------
BM_OLD_CreateSingle/iterations:100            59.0 ns         47.4 ns          100         28         40          2         56
BM_NEW_CreateSingle/iterations:100            29.8 ns         28.9 ns          100         40         40          1         40
BM_OLD_Create6Actions/iterations:100           103 ns          102 ns          100    42.2857         56          7        296
BM_NEW_Create6Actions/iterations:100           153 ns          152 ns          100        150        320          4        600
BM_OLD_Iterate6Actions/iterations:10000       22.1 ns         22.1 ns        10000          0          0          0          0
BM_NEW_Iterate6Actions/iterations:10000       22.8 ns         22.8 ns        10000          0          0          0          0
BM_OLD_Clone3Actions/iterations:100            120 ns          120 ns          100         38         40          8        304
BM_NEW_Clone3Actions/iterations:100            101 ns          101 ns          100        100        160          4        400
BM_OLD_Merge3Plus3/iterations:100              344 ns          343 ns          100    39.3043         56         23        904
BM_NEW_Merge3Plus3/iterations:100              275 ns          275 ns          100    115.556        240          9      1.04k
BM_OLD_RealisticWorkflow/iterations:100        172 ns          172 ns          100    34.6667         40          6        208
BM_NEW_RealisticWorkflow/iterations:100        155 ns          154 ns          100    66.6667         80          3        200
------------------------------------------------------------------------------------------------------------------------------
```

this is with fix pass by ref 😉
overall, if he fixes those issues with passing by value to achieve better performance and reduce memory allocations, the result will be more than profitable

By value it a bit slower, tested by privatecore. Currently is by value.

@SmashingQuasar could you have look at this? So we can merge?

brighton-chi:
Awesome. I have no knowledge of these matters to comment technically, but I can say that the current way of doing it looks like absolute shit so I'll be happy to make the corresponding changes in the open raid strategy PRs once this is merged.

SmashingQuasar:
> privatecore
> yep + heap alloc and overall performance (a bit faster)
> 
> ```
> ------------------------------------------------------------------------------------------------------------------------------
> Benchmark                                        Time             CPU   Iterations   AllocAvg   AllocMax     Allocs      Bytes
> ------------------------------------------------------------------------------------------------------------------------------
> BM_OLD_CreateSingle/iterations:100            59.0 ns         47.4 ns          100         28         40          2         56
> BM_NEW_CreateSingle/iterations:100            29.8 ns         28.9 ns          100         40         40          1         40
> BM_OLD_Create6Actions/iterations:100           103 ns          102 ns          100    42.2857         56          7        296
> BM_NEW_Create6Actions/iterations:100           153 ns          152 ns          100        150        320          4        600
> BM_OLD_Iterate6Actions/iterations:10000       22.1 ns         22.1 ns        10000          0          0          0          0
> BM_NEW_Iterate6Actions/iterations:10000       22.8 ns         22.8 ns        10000          0          0          0          0
> BM_OLD_Clone3Actions/iterations:100            120 ns          120 ns          100         38         40          8        304
> BM_NEW_Clone3Actions/iterations:100            101 ns          101 ns          100        100        160          4        400
> BM_OLD_Merge3Plus3/iterations:100              344 ns          343 ns          100    39.3043         56         23        904
> BM_NEW_Merge3Plus3/iterations:100              275 ns          275 ns          100    115.556        240          9      1.04k
> BM_OLD_RealisticWorkflow/iterations:100        172 ns          172 ns          100    34.6667         40          6        208
> BM_NEW_RealisticWorkflow/iterations:100        155 ns          154 ns          100    66.6667         80          3        200
> ------------------------------------------------------------------------------------------------------------------------------
> ```
> 
> this is with fix pass by ref 😉
> overall, if he fixes those issues with passing by value to achieve better performance and reduce memory allocations, the result will be more than profitable
> 
> By value it a bit slower, tested by privatecore. Currently is by value.
> 
> @SmashingQuasar could you have look at this? So we can merge?

Thanks for running those benchmarks, they are very interesting.
I can change the code for sure; however, I am not certain what you are passing by reference and where. Could you show me what you did so I can understand better what you mean? From my perspective using references here would be unsafe as it can end up in dangling references. I might be confused about the code you did.

kadeshar (approved):

Celandriel:
@privatecore can you share more about how you did your benchmarks?
I'm interested in learning and understanding the potential for performance improvement.

privatecore:
> @privatecore can you share more about how you did your benchmarks? I'm interested in learning and understanding the potential for performance improvement.

@Celandriel Yep, ofc, you can try it by yourself: https://godbolt.org/z/zxovdha7T ;)
I just copy-pasted the old and new impl of `NextAction` and tested the most common cases using google's benchmark. Since it's still synthetic tests, you can try to extend them with real-world examples from the current code.

@SmashingQuasar check at least the `clone` and `merge` methods -- you passed parameters by value. I haven't checked the whole impl, so you can review the entire solution and look for further optimizations.

yqsy2025:
Why haven't you merged yet ?

noisiver:
{36526D69-D4EB-414D-9491-E3DDBB536D57}

On top of this people have lives. This is all voluntary and there have been holidays and things like that that keep people occupied.

SmashingQuasar:
> > @privatecore can you share more about how you did your benchmarks? I'm interested in learning and understanding the potential for performance improvement.
> 
> @Celandriel Yep, ofc, you can try it by yourself: https://godbolt.org/z/zxovdha7T ;) I just copy-pasted the old and new impl of `NextAction` and tested the most common cases using google's benchmark. Since it's still synthetic tests, you can try to extend them with real-world examples from the current code.
> 
> @SmashingQuasar check at least the `clone` and `merge` methods -- you passed parameters by value. I haven't checked the whole impl, so you can review the entire solution and look for further optimizations.

Thanks for pointing that out, excellent observation. After looking at it a bit more, it was just silly of me. I removed the clone method because it's completely pointless and passed both merge parameters by const reference. Now we should be good to go.
It would be nice to see your original benchmark but if I understand how things went, you might not have it around anymore. In any case I thought you want to pass the NextAction instances as references when creating the TriggerNode which would have resulted in UB everywhere. Thanks for clarifying what you originally meant and again, good call on those two methods since they are very hot.

I also rebased on `master` and resolved all conflicts.

@yqsy2025 It's a very large PR so it can't be merged without real testing and clean review. It's a bit annoying to rebase it regularly when changes are being made but I'm fine with it since this will help people in the long run.

hermensbas:
Ye sorry guys, but had my fam holidays and whatnot, and this week new real-life (payed) projects. Doing my best, like everyone else. I have some time tonight today to go some open issues. For those who do not understand the testing/verifying takes a bit more time and just to accept every incoming PR which seems to be alright, which is very depending on the context of functionality throughout. 

We we rather take it a bit more slow, also consider when we merge its not fire and forget. So its not as simple just accept and lets take the next one.  

Either way @privatecore has been running to benchmarks and some talks about here and there, @SmashingQuasar has been working on it etc. Other then that we are a very small group of maintainers with limited experience. Fam/kids/friends and obviously day jobs.

privatecore:
> It would be nice to see your original benchmark but if I understand how things went, you might not have it around anymore. In any case I thought you want to pass the NextAction instances as references when creating the TriggerNode which would have resulted in UB everywhere. Thanks for clarifying what you originally meant and again, good call on those two methods since they are very hot.

Just noticed that it’s actually mixed with some optimizations made while playing around: `for (const New::NextAction& action : handlers)`

Will make a quick review when I have some time beside my IRL stuff... But you can check the current impl yourself for any potential issues/optimizations: pass-by-value, move semantics (constructor optimizations, [rule of 3/5/0](https://en.cppreference.com/w/cpp/language/rule_of_three.html)), vector.reserve, vector.insert vs push_back, etc.

Keep in mind, @hermensbas will be happy with just pass-by-value fixes, and so will I :laughing:

privatecore:
Hmm, it seems like godbolt is broken on mobile browsers -- attached benchmarks below.

[test-pr-1923.cpp](https://github.com/user-attachments/files/24442243/test-pr-1923.cpp)

SmashingQuasar:
> Hmm, it seems like godbolt is broken on mobile browsers -- attached benchmarks below.
> 
> [test-pr-1923.cpp](https://github.com/user-attachments/files/24442243/test-pr-1923.cpp)
> 

Thanks, indeed I figured out that it is because I was on my phone that I couldn't see the benchmarks. I was able to see them on my desktop but it seems godbolt won't run them nor quick bench. I'll have to set this up locally myself but in any case I integrated your optimisations to the PR already.

Celandriel:
I tested this again and didn't notice unusual behavior after a few hours.
I was still getting ganked as normal, so all is well.

hermensbas:
> > Hmm, it seems like godbolt is broken on mobile browsers -- attached benchmarks below.
> > [test-pr-1923.cpp](https://github.com/user-attachments/files/24442243/test-pr-1923.cpp)
> 
> Thanks, indeed I figured out that it is because I was on my phone that I couldn't see the benchmarks. I was able to see them on my desktop but it seems godbolt won't run them nor quick bench. I'll have to set this up locally myself but in any case I integrated your optimisations to the PR already.

You want take another look at those benchmarks before merging? or optimize after if needed?

hermensbas (approved):

PR #1928, opened 2025-12-19
[Stability] Various crash fixes based on Regrad fixes and crashlogs.

hermensbas:
These contains various fixes, fixes that have history worked one in past more then once as person as group, aswell @Wishmaster117. But due various reasons we had to drop them due priority or simply timewise. These fixes have recollected again by @Regrad based on his crash logs.

Most crash logs we have, i am talking 30+ of them, to many to post here. @Regrad running a larger server 100+ real players with bots, which means he will walk into issues that most of us wont or are extremely difficult to reproduce.

@Regrad used LLM to solve them based on crash log and mentioned his server crashes almost disappeared, instead of redoing every single PR and pull them apart. I tried to keep his bunch of changes together as whole, reviewed them, some redone, verified again etc etc. This is not how would normally do this. But since i want @Regrad being able to confirm, we need this in a package as a whole. Pulling them apart in the current situation is simply to much, to complicated in the verification process.

So this PR is open and in my opinion has priority above others, but @Regrad is only person who can give the green light for the mod-playerbot changes for now.  

I, we spend huge amount of time into these issues over last couple of months. I will put other PR's on hold for abit.

PR #1929, opened 2025-12-19
server_crash_fix: OnPlayerGiveXP

hermensbas:
[fef0ed4072c8_worldserver.exe_.18-12_23-30-44.txt](https://github.com/user-attachments/files/24258638/fef0ed4072c8_worldserver.exe_.18-12_23-30-44.txt)

- Just added common logic and defense coding
- Optimized isRandombot while at it

Just for clarification i am aware the trigger of bug lies with the progression mod. Regardless it should not happen.

### DIFF
--- a/src/RandomPlayerbotMgr.cpp
+++ b/src/RandomPlayerbotMgr.cpp
@@ -2600,7 +2600,7 @@ bool RandomPlayerbotMgr::IsRandomBot(Player* bot)
     PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
     if (!botAI || botAI->IsRealPlayer())
         return false;
-        
+
     return IsRandomBot(bot->GetGUID().GetCounter());
 }
 


HennyWilly:
@hermensbas Thanks for fixing this.

Unfortunately there are two checks that prevent my feature (#1910) from working properly. Those checks should come after the `const Player* master = botAI->GetMaster()` block:
- `sPlayerbotAIConfig->randomBotXPRate == 1.0f`: The master's `PLAYER_FLAGS_NO_XP_GAIN` check is independent  from the `randomBotXPRate`. If `randomBotXPRate` is 1 (default), bots keep gaining XP even though the master has XP gain disabled.
- `!sRandomPlayerbotMgr->IsRandomBot(player)`: This check filters addclass-bots.

Mind if I open a new PR? Or would you rather add this into your #1928 PR?

hermensbas:
create a new PR i will resync/rebase 1918 based on ur PR

hermensbas:
```c++
void OnPlayerGiveXP(Player* player, uint32& amount, Unit* /*victim*/, uint8 /*xpSource*/) override
{
    if (!player || !player->IsInWorld())
        return;

    PlayerbotAI* botAI = GET_PLAYERBOT_AI(player);
    if (!botAI)
        return;

    // no XP gain if master is a real player with XP gain disabled
    if (const Player* master = botAI->GetMaster())
    {
        if (WorldSession* masterSession = master->GetSession();
            masterSession && !masterSession->IsBot() &&
            master->HasPlayerFlag(PLAYER_FLAGS_NO_XP_GAIN))
        {
            amount = 0;
            return;
        }
    }

    // from here on we only care about random bots
    if (!sRandomPlayerbotMgr->IsRandomBot(player))
        return; 

    if (sPlayerbotAIConfig->randomBotXPRate == 1.0f)
        return; 

    // no XP multiplier if bot is in a group with at least one real player
    if (Group* group = player->GetGroup())
    {
        for (GroupReference* gref = group->GetFirstMember(); gref; gref = gref->next())
        {
            if (Player* member = gref->GetSource())
            {
                if (WorldSession* session = member->GetSession();session && !session->IsBot())
                    return;
            }
        }
    }

    // otherwise apply XP multiplier
    amount = static_cast<uint32>(std::round(amount * sPlayerbotAIConfig->randomBotXPRate));
}
```

like this i assume?

HennyWilly:
I would add 
```c++
if (WorldSession* session = player->GetSession(); !session || !session->IsBot())
    return;
```

between
```c++
if (!player || !player->IsInWorld())
    return;
```
and
```c++
PlayerbotAI* botAI = GET_PLAYERBOT_AI(player);
if (!botAI)
    return;
```
to make sure that the player is a bot. The rest looks good to me.

HennyWilly:
In hindsight, the `PlayerbotAI* botAI = GET_PLAYERBOT_AI(player)` check might be sufficient for filtering non-bots.
I don't know, if there is any edge case that requires the `session->IsBot()` check.
Alternatively, there is `botAI->IsRealPlayer()`...

hermensbas:
> In hindsight, the `PlayerbotAI* botAI = GET_PLAYERBOT_AI(player)` check might be sufficient for filtering non-bots. I don't know, if there is any edge case that requires the `session->IsBot()` check. Alternatively, there is `botAI->IsRealPlayer()`...

I suppose especially in script hooks the isBot should be safer since its isBot is directly bound to the core player object, meanwhile the other relies on the internal maps administration of the manager.

HennyWilly:
@hermensbas Here is my new PR: #1935
FYI, I re-added the  `static_cast<float>(amount)` at the bottom to fix a warning about a narrowing conversion.

PR #1930, opened 2025-12-19
Fix Assistant Assignment Functions

brighton-chi:
IsHealAssistantOfIndex() and IsRangedDpsAssistantOfIndex() are supposed to iterate through the group and first return members with the applicable role that have the assistant flag, and then iterate through non-assistants only if there are not enough assistants for the designated index. They are not written properly and actually completely ignore the assistant flag.

I rely on these functions for significant roles in SSC and TK (which I have decided I'll PR in the same way as SSC, as a long-term draft). I have them fixed on my own fork, but it is problematic for testers if these functions do not work.

So I've done three things here:
1. Fixed the functions to prefer members with the assistant flag.
2. Added a third parameter for ignoreDeadPlayers, like IsAssistTankOfIndex() has. Note that the parameter is by default false for IsAssistTankOfIndex(), meaning dead players are _not_ ignored. This is not my preferred design choice--I think the default should be to ignore dead players, but I have not changed the default and have made the default the same for IsAssistHealOfIndex() and IsAssistRangedDpsOfIndex(), since I don't know the intent of the pre-existing boss strats that use the functions.
3. Changed the names to IsAssistHealOfIndex() and IsAssistRangedDpsOfIndex() so they parallel IsAssistTankOfIndex(), and made corresponding changes in the few boss strats that use the functions.

Also, note that the functions _do _not_ exclude real players. I think there are arguments for and against excluding real players. A fourth parameter for this could be useful, but I've not made any change in that regard.


brighton-chi:
Resolved merge conflicts so this should be good to go. Sucks that doing that dismisses the reviews.

notOrrytrout:
Am I understanding this correctly?  in IsAssistHealOfIndex, the counter is incremented inside a branch that handles both assistant and non-assistant healers, but the return condition only ever matches assistants on the first pass? 

If so, does that mean non-assistant healers still advance the index even though they can never be returned, effectively changing the ordering compared to the old logic?

Would a two-pass, assistant-first stable selection func work?

brighton-chi:
> Am I understanding this correctly? in IsAssistHealOfIndex, the counter is incremented inside a branch that handles both assistant and non-assistant healers, but the return condition only ever matches assistants on the first pass?
> 
> If so, does that mean non-assistant healers still advance the index even though they can never be returned, effectively changing the ordering compared to the old logic?
> 
> Would a two-pass, assistant-first stable selection func work?

Sorry, I'm confused. Are you perhaps reading the changes backwards? Existing IsHealAssistantOfIndex() increments the counter for every healer, both assistants and non-assistants, and mixes them together for the return, so it is useless. 

This: 
```
    if ((isAssistant && index == counter) || (!isAssistant && index == counter))
    {
        return player == member;
    }
```

Is just saying this:
```
    if (index == counter)
    {
        return player == member;
    }
```


I changed to have a first pass for only assistants and then a second for non-assistants (and renamed it IsAssistHealOfIndex() to match IsAssistTankOfIndex()).

notOrrytrout:
Maybe I am. I will have to look at it again. Thanks for the response. 
We can agree the old code was useless the way it was implemented.

kadeshar (approved):

notOrrytrout:
Testing

notOrrytrout:
I added logging locally. It looks like when the healer is demoted it does not remove selection. I disbanded the raid then remade it and used add class to put healers in well of course it used some of the same healers as before.  One of them was assistant.  He was no longer assistant when he joined. The log still said that it was selecting him.

notOrrytrout:
So I couldn’t get it to change who is selected so I kicked him and it changed to a random other healer. I even had a assistant healer that it could have selected, but it did not.

I kicked the old healer out of the group, logged him out, and it changed to selecting someone else. I re-logged the old healer and it started selecting him right away.  I will provide my modified playerbot ai to ensure I did not do anything except add logs

notOrrytrout:
[PlayerbotAI.cpp](https://github.com/user-attachments/files/24657802/PlayerbotAI.cpp)

brighton-chi:
What did you do to call the function? Did you engage one of the bosses that uses it?

notOrrytrout:
> What did you do to call the function? Did you engage one of the bosses that uses it?

I did not engage a boss. I exercised the logic by forming a raid group and assigning and reassigning raid assistants while remaining out of combat. The assist selection logic appears to run continuously based on group state alone.

On initial raid creation, no assist was selected. After modifying assistant assignments, selection began updating. This was sufficient to reproduce the behavior I described, without requiring combat. Combat could be next if needed. 

I'm new to testing and debugging.

notOrrytrout:
Smells like the code is selecting based on stale in-memory state, not current raid role truth. Idk for sure..

I don’t want to be a pain I just want to get this tested as it is the last step. My logs I was using to test could be wrong too. 😐

brighton-chi:
Sorry, but what is g_assistLog? Where do you get the output from logging in that way?

notOrrytrout:
> way?

It logs to a file, but you have to add a setting to let it know to log it. I forgot about that.

Appender.PlayerbotsAssistFile=3,5,logs/playerbots-assist.log,a
Logger.playerbots.assistdebug=5,Console PlayerbotsAssistFile

I added these to my world conf 
Then the file is in the folder with the world server bin or exe 

there are definitely better ways of logging but I’m new to it. 😆

brighton-chi:
I'm just baffled as to why you got anything to output. I think right now in the main branch, the healer assignment function is never used. I think it was used in Naxxramas, which is now removed. I use it in SSC and TK, but those are not merged. I actually don't think this should even be testable with the main branch (nor would the ranged dps version be, as I think it also was only used in Naxxramas before I started using it in stuff that isn't merged).

The way I'd log something would be like "LOG_DEBUG("playerbots", "[IsAssistHealOfIndex] SELECTED (assistant): {}", member->GetName());" and "LOG_DEBUG("playerbots", "[IsAssistHealOfIndex] SELECTED (nonassistant): {}", member->GetName());"

Then it should output to the playerbots log for every tick that the function is actually called in some boss strategy. Otherwise, I'd think you would get nothing. There shouldn't be anything cached. You could also use LOG_INFO to output to the WorldServer console, but for something like this you would get insane spam.

brighton-chi:
I suppose you could try logging for IsAssistTankOfIndex, which is used in a lot of stuff, because the logic is identical except for tanks.

notOrrytrout:
image
[assist_debug.log.txt](https://github.com/user-attachments/files/24658788/assist_debug.log.txt)

The crazy thing is the other healer is selected in the image. It is in group 2. That one it says was selected even when not in the party.. I'll look into logging like you said.

notOrrytrout:
> I'm just baffled as to why you got anything to output. I think right now in the main branch, the healer assignment function is never used. I think it was used in Naxxramas, which is now removed. I use it in SSC and TK, but those are not merged. I actually don't think this should even be testable with the main branch (nor would the ranged dps version be, as I think it also was only used in Naxxramas before I started using it in stuff that isn't merged).
> 
> The way I'd log something would be like "LOG_DEBUG("playerbots", "[IsAssistHealOfIndex] SELECTED (assistant): {}", member->GetName());" and "LOG_DEBUG("playerbots", "[IsAssistHealOfIndex] SELECTED (nonassistant): {}", member->GetName());"
> 
> Then it should output to the playerbots log for every tick that the function is actually called in some boss strategy. Otherwise, I'd think you would get nothing. There shouldn't be anything cached. You could also use LOG_INFO to output to the WorldServer console, but for something like this you would get insane spam.

I did not test main branch and you use IsAssistHealOfIndex() in the Pr correct?

notOrrytrout:
Sorry never mind I'm going to move to the other PR testing and leave it here. We have brighter minds than me in coding here..

brighton-chi:
Are we needing testing for this one? It would probably be a giant pain in the ass for anybody else because the functions are not correctly used. I can say that I use them on my own fork, and they work as far as I can tell, though I've not pressure tested with disbanding and dropping members and switching configurations and so on. They're the exact same as IsAssistTankOfIndex(), in any case.

notOrrytrout (approved):
I think I was reading it backwards or something. I don’t know what was wrong with my testing, but looking directly at the code. The logic makes sense.  Sorry about that.

notOrrytrout:
It was 100% that my logs were in the wrong place.  Two other people and the author approved this code, so I could not give up on it I had to see what I was doing wrong. 

I fixed the placement and the logs produced the correct bot to be selected.. The strategies seem to trigger when in the area. So when the strategy is running it calls the who is assistant code over and over. If  I reset the bot ai the logs quit till they pick the raid strat. Basically after fixing the log placement I confirmed in my logs that this PR is correct, as we already new by looking at the code. My fault guys, I feel we should merge it.

brighton-chi:
It's cool--if you're going to be this thorough all the time, I sure hope you keep testing PRs.

I was just thinking about whether the loops should also exclude bots that are not in the map. In case you enter a raid with 25 but actually have more in your raid, just not in the instance. Or if a bot releases. Maybe it's not much of a concern.

notOrrytrout:
> It's cool--if you're going to be this thorough all the time, I sure hope you keep testing PRs.
> 
> I was just thinking about whether the loops should also exclude bots that are not in the map. In case you enter a raid with 25 but actually have more in your raid, just not in the instance. Or if a bot releases. Maybe it's not much of a concern.

I was wondering if it would be possible to do only a single call of the function only when the any assistant changes. That way it’s not calling all the time repeatedly.  But then it would need to “remember” and have a cache or something. It could be tied to the master and the master’s strategy would keep track of each assistant. But not for this PR.

brighton-chi:
I think that would be more ideal, though I think it's sort of a high-level concept that would need to be introduced with respect to the module to have a cache-type system with respect to function calls. I certainly wouldn't know how to pull that off.

notOrrytrout:
> > It's cool--if you're going to be this thorough all the time, I sure hope you keep testing PRs.
> > I was just thinking about whether the loops should also exclude bots that are not in the map. In case you enter a raid with 25 but actually have more in your raid, just not in the instance. Or if a bot releases. Maybe it's not much of a concern.
> 
> I was wondering if it would be possible to do only a single call of the function only when the any assistant changes. That way it’s not calling all the time repeatedly. But then it would need to “remember” and have a cache or something. It could be tied to the master and the master’s strategy would keep track of each assistant. But not for this PR.

And then each bot could run a did the assistant change loop if not do nothing keep current. Heal or dps the same.  It might already cheap how it is it’s just about what it is doing consistently.

PR #1935, opened 2025-12-21
Re-enable no-xp-feature with better checks

HennyWilly:
The feature `Disable bot XP gain when master has XP turned off` (#1910) was unintentionally disabled in #1929.
This PR re-enables it with better checks in order to prevent crashes when playing together with the individual progression module.


hermensbas:
looks good, thx

PR #1937, opened 2025-12-22
Quick fix for CMSG_FORCE_MOVE_ROOT_ACK and CMSG_FORCE_MOVE_UNROOT_ACK

privatecore:
**Original issue:** https://github.com/mod-playerbots/mod-playerbots/issues/1902
**Original cause:** charmed bot (with lost client control) got rooted at the same time.

**How to reproduce:**
1. Spawn creatures 11350 (x3) and 11830 (x3) using the command: `.npc add <entry>` in a quiet place.
2. Create any party with random bots or alt bots (should be 60-65 levels), make sure there is at least one healer.
3. Set the healer's mana to a high value, like 100M, using command: `.mod mana <value>`.
4. Start the fight while continuously respawning creatures with: `.resp all`.
5. When console starts to spam 'heartbeat' messages, check your party members' movement flags to identify which one has `MOVEMENTFLAG_ROOT` 0x00000800 (2048) using the command: `.debug move`.

This PR will not fix ALL 'heartbeat' issues, as `ServerFacade::SetFacingTo` still sends `SendMovementFlagUpdate` while bots can have the `MOVEMENTFLAG_ROOT` flag.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -1350,10 +1350,6 @@ void PlayerbotAI::DoNextAction(bool min)
     bool isBotAlive = bot->IsAlive();
     if (currentEngine != engines[BOT_STATE_DEAD] && !isBotAlive)
     {
-        bot->StopMoving();
-        bot->GetMotionMaster()->Clear();
-        bot->GetMotionMaster()->MoveIdle();
-
         // Death Count to prevent skeleton piles
         // Player* master = GetMaster();  // warning here - whipowill
         if (!HasActivePlayerMaster() && !bot->InBattleground())
@@ -1374,6 +1370,8 @@ void PlayerbotAI::DoNextAction(bool min)
     // Change engine if just ressed
     if (currentEngine == engines[BOT_STATE_DEAD] && isBotAlive)
     {
+        bot->SendMovementFlagUpdate();
+
         ChangeEngine(BOT_STATE_NON_COMBAT);
         return;
     }


privatecore:
Keep in mind that this impl is not 'correct' because `CMSG_FORCE_MOVE_ROOT_ACK` and `CMSG_FORCE_MOVE_UNROOT_ACK` should be sent with the proper movement information (not properly set for bots).

I will try to rewrite the movement in a more 'proper' way, which should fix most of the issues with bots' movements and, may be, introduce new ones ;)

privatecore:
@Grimfeather waiting for your tests results...

Grimfeather:
Just in case someone else reads this. I did report back to privatecore.
there are still some issues.

privatecore:
@Grimfeather actually, resurrection is not related to the current issues, since the death state broke MM after all. For ref, check `Unit::setDeathState`. Ready to test... again =)

privatecore:
For those who want to help with testing:
1. Test the case from the PR's original message (see "How to reproduce").
2. Test bots' death and resurrection (from player, from bot, from spirit healer).

Grimfeather:
I've tested the spots where there were always issues before.
At the bug trio in aq40 where they fear and stun, no longer are bots teleport following, or refusing to move. no heartbeat messages in the console.

the other spot in naxx40 with the fear and stuns from whirlwind, also everything is fine.
no heartbeat messages, nobody gets stuck, everyone is moving normally.

same for after resurrecting. they no longer start mass teleport following. they now move normally.
I've let bots get resurrected normally by other bots and by using the release option.

seems all good now.

hermensbas:
such a nice job fixing this issue.

PR #1940, opened 2025-12-23
playerbots_rb_team_size_fix

Regrad:
Fixes an issue where when choosing a "Random battlefield", a team of 10 bots was always formed, even if the battlefield implies a larger team size (15, 40)


kadeshar:
Codestyle fixes required
obraz

Regrad:
I've included this PR in my big PVP PR.

PR #1943, opened 2025-12-23
Implement Tempest Keep: The Eye Strategies

brighton-chi:
Edit: Descriptions of methods are out of date right now. To be updated.

This one comes with the same caveats as SSC about requiring ownership from somebody with C++ knowledge, except I think the matter is even more acute here because these strategies incorporate a novel approach proposed by Timberpoes. By redeclaring the entire bossai class for Kael’thas, it was possible to add new member functions to the class in order to access its private member variables. This allows bots to have visibility into boss mechanics beyond what they could do with ordinary techniques and is similar in approach to what was done by the Naxx strategies, except that this approach does not require any modifications to the core. I used it for only one mechanic, which was to detect Kael’thas’s phase. That was very helpful because the fight is divided into 5 phases, and distinguishing between them with traditional techniques requires lookups of a dozen NPCs and comparisons of their various unit states, react states, and auras; by accessing his bossai, this can all be avoided. However, there is far more potential beyond this if the approach is an acceptable one.

On with the (shit)show.

### Trash
In a perfect world, there would be many strategies for TK trash, which is easily more difficult than two of the bosses. It’s a real pain to do though because to solve the biggest issues properly, each pack would have to be handled a little differently. So the only thing I’ve included is for Mages to cast polymorph on the Crimson Hand Centurions when they are channeling Arcane Flurry. The purpose is not to actually keep them CC’d but to interrupt their channel.

### Al’ar
This fight sucked so much to write a strategy for. The only silver lining is that being the post-nerf version, the boss moves between only 4 platform locations (instead of 6), and movement between them is on a fixed rotation (interrupted by Flame Quills) instead of being random. Thus, a strategy can be consistently replicated, and the fight can be done with only 3 tanks (2 on the platforms for the boss and 1 below for adds).

**Phase 1:**
I’m going to call the platform that Al’ar lands at after the pull “platform 0” because that reflects the indices in the code. In a clockwise direction, the remaining platforms will be referred to as platforms 1, 2, and 3, respectively.
The best way to pull is to first put all ranged, as well as tanks other than your main tank and first assistant tank, on nc +stay below platform 0. Then, go up the ramp to platform 0 with your main tank, first assistant tank, and melee dps following you, then hit Al’ar with any ranged attack or spell to start the fight.

- Your main tank will start at platform 0, and your first assistant tank will immediately move to platform 1. When Al’ar moves to platform 1, your main tank will move to platform 2. When Al’ar moves to platform 2, your first assistant tank will move to platform 3. When Al’ar moves to platform 3, your main tank will move back to platform 0. This assures a tank is available to receive Al’ar after every platform movement (every 30 seconds).
- Melee DPS will follow Al’ar as it moves between platforms.
- Each platform is mapped to a corresponding ground location below it. Ranged DPS and healers will follow Al’ar by moving to the corresponding ground location as it flies between platforms.
- After each platform move, an Ember of Al’ar will spawn. Your second assistant tank will pick up the Ember and move it to the point that is 25 yards away from the ground position corresponding to Al’ar’s platform (on an invisible line between such ground position and the middle of the room). Ranged DPS will then focus down the Ember before switching back to Al’ar (this positioning is so that ranged are not hit by the Ember Blast explosion that happens whenever an Ember dies).
- Each time Al’ar leaves a platform, it has a chance to instead fly up high in the middle of the room to perform Flame Quills, which will one-shot anybody on the upper level or ramps. When Al’ar begins the Flame Quills sequence, all bots on the top level will jump off. FYI, Al’ar’s usage of Flame Quills is not entirely random: there is a 20% chance for it to do so after the first platform move, and the chance increases by another 20% after each subsequent platform move that does not trigger Flame Quills (reset after each Flame Quills sequence).
- After Flame Quills, Al’ar will randomly land at either platform 0 or 3. To prepare for this, bots will move to assigned positions during the Flame Quills sequence:
  - Ranged and the second assistant tank will wait in the middle of the room.
  - Melee DPS will wait at a point that is between the base of each ramp.
  - The main tank will wait at the base of the ramp to platform 0.
  - The first assistant tank will wait at the base of the ramp to platform 3.
- Once Al’ar lands, the regular Phase 1 strategies resume.
- When Al’ar “dies,” it disappears and moves to the center of the room, where it casts Rebirth and returns to full HP. Bots will wait outside of the radius of the Rebirth explosion for Phase 2 to start.

Phase 2:

- Your main tank will tank Al’ar initially. When Al’ar casts Melt Armor, your first assistant tank will taunt Al’ar and take over. The tank swaps will continue back and forth every time Melt Armor is cast.
- Bots will avoid Flame Patches. FWIW, the standard co +avoid aoe strategy does work for Flame Patches, but avoid aoe provides no buffer distance so as you’ve probably noticed, it doesn’t provide for preemptive avoidance. Also, avoid aoe does not consider multiple hazards together so it can be an issue when movement needs to take into account more than one hazard, plus when a strategy requires particular bot movement, it’s better to account for the hazards within that movement strategy instead of relying on separate methods that can create conflicts.
- When Al’ar takes to the sky to perform Dive Bomb, bots will spread out (and continue to avoid Flame Patches). After the Dive Bomb, Al’ar does another Rebirth explosion. I have tried a million different things to properly detect this full sequence (even accessing the bossAI like I did with Kael’thas) and cannot get it to work properly. Ultimately, all I’ve been able to get to work at all with respect to the final explosion is for bots to detect the 2-second cast of the Rebirth and run out. It is not enough time for bots that are too close when the cast happens so some bots may get hit, but if you have adequate gear, they should survive.
- After each Dive Bomb, 2 Embers will spawn. Your second assistant tank will tank one Ember, and either the main tank or first assistant tank, whichever one is not tanking Al’ar at the time, will tank the other Ember. They will both move the Embers away from bots, and ranged DPS will focus both Embers down before switching back to Al’ar.
- Because the room is so large, it is possible for bots to get too far away from active combat (particularly if they are thrown across the room by Ember Blast) so there is also a method for them to run back toward the center if they get too far away.

### Void Reaver
Ironically, what was often considered the easiest boss in 25-player content in TBC is the only boss with an ability (Arcane Orb) that I do not believe can be avoided by bots, even with access to Void Reaver’s boss script. Therefore, every single Arcane Orb is going to hit its target, so the strategy can only try to limit the damage by spreading ranged bots in two rings around Void Reaver (one for healers and one for ranged DPS, to try to ensure sufficient distribution of healers). The tanks will all fight for aggro (necessary due to Knock Away) and try to keep Void Reaver in the middle of the room. Bots that can wipe aggro or otherwise gain invulnerability are directed to use the applicable abilities as soon as they pick up aggro (e.g., Soulshatter). He’s still easy, but if you have IP nerfs, it’s a little bit of a gear check.

### High Astromancer Solarian
No boss was hit harder by nerfs in TBC than Solarian, whose encounter went from a totally unique fight that required arcane resistance to a fight that is kind of just an easier Baron Geddon. IMO, she is the easiest boss in TBC 25-player raids.

- Ranged bots stack up at a distance from Solarian; this leaves all bots with plenty of space to run away from other bots when they get Wrath of the Astromancer.
- When Solarian vanishes, all bots will stack to AoE down the Solarium Agents that spawn.
- When Solarian returns with two Solarium Priests, melee will divide into two groups, with one focused on each Solarium Priest. I think this method is not working correctly right now because when one Priest dies, the bots still on the second Priest are leaving it. I’ll need to decide whether I want to figure it out or just get rid of it because this fight is so easy regardless.
- Priest bots will cast Fear Ward on the main tank to block the Psychic Scream during the final tank-and-spank Voidwalker phase, and the main tank will pick up Voidwalker Solarian as soon as she transforms.

Note that the bots will not be knocked into the air by Wrath of the Astromancer. The issue is due to the presence of a check for knockbacks in Playerbots that causes bots to ignore knockbacks that would launch them at a velocity beyond a hardcoded value. I’ve increased that velocity limit on my own fork, and it does allow Wrath of the Astromancer (and other knockbacks that otherwise don’t work) to work on bots. But that’s obviously a broader issue and not addressed in this PR, and bots don’t take fall damage in any case.

### Kael’thas Sunstrider

So this strategy has 23(!) action methods. But like in retail, this is actually an easy fight once it is learned because it is highly scripted. 

Unlike in other strategies I’ve done, the bots probably cannot do this fight by themselves unless they are way overgeared. This is because there are a few windows during which bots need to position themselves properly based on dynamic factors. But no RTSC is needed—you just need to have bots follow you to the right locations. Also note that the gear check for this strategy is higher than in retail because you have to get all of the legendary weapons down and looted before the advisors aggro in Phase 3, or it’s going to be an absolute shitshow (with human players, you can deal with there still being a couple of weapons up). For a point of reference, when I was first working on this strategy with damage reduced to 50% and bots pretty close to T4 BiS, I had almost no margin of error (I would usually get the weapons down with barely a second to spare).

You will need at least 2 tanks, but 3 is better. Your main tank will need to be able to equip the legendary shield so you must use a Warrior or Paladin. However, it is ideal for the first assistant tank to be a Druid because they can equip the legendary staff.

**Phase 1:**
Fun fact—when you “kill” the advisors in this phase, they don’t actually die but get an aura applied called “Permanent Feign Death” (nice oxymoron).

- _Thaladred_: You’re supposed to kite him, and bots can’t really kite, so the method is a poor man’s method of having the bot move away from him in a straight line when fixated. You want him to die in the far Southern part of the room. If he dies in a bad location, you may as well call a wipe and restart. What will work best for you will depend on your DPS since you don’t want to kill him before he gets to the location you want but also don’t want bots to be trapped up against a wall since they can’t properly kite him. The way that works best for me is to have bots stay back while I aggro the boss, and wait until right before Thaladred switches to his second fixate target before attacking. Note that if you do put bots on stay, when you put them back on follow, the bot that is then being fixated will remain on stay (because they need to disregard movement orders other than running away from Thaladred). So after Thaladred dies, make sure to manually type /follow or the bot that was fixated when you took the bots off of stay will not rejoin the fight.
- _Sanguinar_: He will be tanked by your main tank, who will be targeted by your Priests for Fear Ward. Bots will wait to engage him; I made it a very generous time (12 seconds) because there is absolutely no rush in Phase 1. There’s no sense in being aggressive. During that time, the main tank will drag Sanguinar to the West wall.
- _Capernian_: This is the first make-or-break part of the fight. Phase 1 Capernian was the most frequent cause for wipes for me. 
  - She should be tanked by a Warlock. If you want to pick your Warlock tank, you can do so by the assistant flag, but if you don’t, the strategy will just pick your highest HP Warlock. If you raid without a Warlock, then you’re insane, but at least there’s a guard so your server won’t crash?
  - You do not need to add the tank strategy to your Warlock. There is a method that will automatically switch your selected tank Warlock between DPS and tank strategies at appropriate times because you need to squeeze out every drop of DPS you can get, particularly for Phase 2, where you’ll need your Warlock to be blowing up weapons with Seed of Corruption instead of spamming Searing Pain. You’ll want your Warlock to start with a DPS strategy as usual (since they should be DPSing Thaladred).
  - To engage Capernian, start running East right before Sanguinar dies. She will activate quickly, and you want to try to get in front of her (but not too close) before she aggros.
  - When Capernian aggros, your Warlock tank will immediately switch to the tank strategy and attack. Your main tank will run toward Capernian but not actually attack; their purpose will be to bait her Conflagration to reduce the chance that it hits your Warlock tank. Other melee will not engage Capernian. Ranged DPS will be idle for 12 seconds; during this time, you should run South to make sure they are not in range of Capernian. After 12 seconds, your ranged DPS will activate, move into range and spread out, and attack (it doesn’t seem possible to outrange Conflagrate, so if bots don’t spread, she will annihilate the entire ranged group with a single cast). Ideally, you kill her not too far from her starting position. If she ends up in the middle of the room, you should probably wipe and start over.
- _Telonicus_: He is very easy in retail but actually is a big risk for wipes with respect to bots because his bombs will one-shot any non-tank, and bots will stupidly stand in front of him without a proper strategy. You should keep some distance from him before he aggros. Your first assistant tank will pick him up and move him to the West wall near Sanguinar. Again, there is a 12-second delay before DPS starts. Your melee DPS are coded to stay directly behind him and not get too close so they don’t get hit by bombs. 

**Phase 2:**
Kael’thas will summon all weapons immediately after Telonicus is down. Just before Telonicus is down, you should move to the platform where the advisors originally were—you’ll be in better position for the raid to AoE down the weapons.

- Your main tank will pick up the axe and move it away from the group. The axe is the biggest threat during this phase and can easily one-shot casters if not pulled away.
- One of your Hunters will attempt to get aggro on the bow and move away from the group (as a hacky way of trying to turn the bow away from the group because you can’t really get a bot to do that directly). This method is hit or miss, but it shouldn’t be that big of a deal if your Hunter doesn’t pull it off properly. 
- Everybody else will prioritize weapons in the following order (but most damage will come from AoE, which is what you want or you will not beat the timer): staff, mace, sword, dagger, axe (ranged only), bow, and shield.
- As weapons are defeated, bots will loot and equip them. If you have not disabled bot announcements in your config, you get to see your entire raid go nuts because they looted legendary items.
- Here is what weapons bots will loot and equip. I don't know anything about DKs, having never played WotLK, so tell me if anything is wrong for them.
  - _Healers:_ Mace (if a healer normally uses a staff, it's best if they keep an OH in their bags for this fight)
  - _Tanks:_ Shield and sword for Paladins and DK, shield and dagger for warriors, staff for Druid
  - _Offensive_ casters: Staff
  - _Rogues:_ Sword and dagger if Combat or Subtlety, dagger only if Assassination
  - _DPS Death Knights, Retribution Paladins, Arms Warriors_: Axe
  - _Fury Warriors_: Dagger. I understand that due to Titan Grip, they should also have the Axe for best DPS; however, Fury Warriors have awful DPS (we’re talking barely above Prot-level) at this stage. Thus, my view is it is better to give them only the dagger so they will MH it and help break MC in Phase 4, since they will contribute hardly any DPS regardless.
  - _Cat Druids_: Staff
  - _Enhancement Shamans_: Dagger
  - _Hunters:_ Bow and dagger. Note that I do NOT have them loot the sword because they need the dagger in their mainhand to use to break MC in Phase 4; whatever marginal benefit they get from the sword as a stat stick is not worth losing this capability. If your Hunter uses a 2H, it is best to have them carry a 1H in their inventory so they can put something in the OH after they equip the dagger.
- After looting weapons, bots with the staff will use it (once) to activate the Mental Protection Field. Hunters will use the bow to generate the legendary arrows and equip those (and will continue to do so during the fight if they use up the arrows).
- If you wipe from this point forward, everybody will lose their legendary weapons, and by default, most bots will not automatically reequip their own weapons until a loot event occurs. This was extremely annoying, and therefore there is a noncombat method implemented that causes everybody to equip upgrades when they get within 150 yards of Kael’thas. I considered applying this to the whole instance, but I’m not sure if some people would not like that so I decided to limit things to the Kael’thas encounter.

**Phase 3:**
I highly recommend you have your Shamans drop Tremor Totems (co +tremor) during this phase. Doing so is not coded because I wanted to leave flexibility, but I think it is very helpful for Sanguinar. After the weapons die, you want to move your bots to a central location between the advisors. If Thaladred died closer to the middle of the room, ideally you position to the side of Thaladred so when he fixates he will not chase bots North into the other advisors.

- Shamans will immediately use Heroism/Bloodlust.
- Your melee tanks will bring Sanguinar and Telonicus to their tanking positions (same as Phase 1). If your first assistant tank is a Druid, they will be immune to Telonicus’s Remote Toy due to having the legendary staff’s aura activated and will also make your main tank immune.
- One healer will stay by the Sanguinar and Telonicus tanking positions to heal the tanks. Once IsHealAssistantOfIndex() is fixed, you will be able to select this healer with the assistant flag. Right now, this will just be the last healer that joined your raid (per standard AC logic).
- DPS priority will be Thaladred, Capernian (ranged only), Sanguinar, Telonicus. As with retail, the most chaotic period will be before Thaladred is killed, particularly if he chases bots into other advisors. I don’t have a great solution for this, but Capernian is significantly less dangerous during this phase thanks to the legendary staff. This is the last true breakpoint—if you get Thaladred down with your raid mostly intact, you are very likely to get the kill.

**Phase 4:**
Kael’thas will aggro immediately after all advisors are dead.

- Your main tank will position Kael’thas at his original position.
- Bots will move out of Flame Strikes.
- Assist tanks will pick up Phoenixes. Since they die over time anyway, bots will not waste time attacking them. When Phoenixes die, they turn into an Egg—at that point, bots will switch to the Egg to destroy it before the Phoenix is reborn.
- When Kael’thas puts up Shock Barrier and starts casting Pyroblast on your main tank (a one-shot), all bots will focus DPS on him (even if there is an egg up). You have 4 seconds to break the barrier (80K HP) and interrupt his Pyroblast. It is likely that you will not be able to if you are playing with IP nerfs and are in T4 gear. However, the main tank will use the legendary shield’s ability, which will allow them to absorb one cast, giving you 8 seconds to break the barrier and interrupt Pyroblast. Bots will put top priority on interrupting Pyroblast as soon as the barrier is down.
- If a bot (or player) is mind controlled, bots with the legendary dagger (other than tanks) will move to MC’d players and use the following attacks to break MC: Shiv (Rogues), Hamstring (Warriors), Wing Clip (Hunters), and Stormstrike (Shamans).

**Phase 5:**
At 50% HP, Kael’thas enters a long RP sequence. This is a good time to kill any remaining Phoenixes and/or Eggs.
- Kael’thas stops casting Pyroblast and Mind Control.
- His main new ability is Gravity Lapse, and it doesn’t work properly on bots... He sucks in the entire raid then knocks everybody back in a different direction. What is supposed to happen is that players will end up floating in midair in different directions and at different heights. However, bots will immediately fall to the ground after getting knocked back. They will not actually hit the ground though and instead remain in a flying state right above the floor. 
  - If you could move in 3D space, Netherbeam would be very easy to deal with. However, because that is not available to bots, they can spread only in 2D space and thus need to move farther to get properly spread, and they waste the first moments falling straight down. As a result, the damage from Netherbeam can be quite high, and the beginning of Gravity Lapse requires a lot of healing. I don’t really have a better way of dealing with this.
  - FWIW, I don’t think there is any existing method to make bots disperse in 3D anyway.
- Kael’thas is supposed to use Nether Void when players are in midair, which creates clouds that reduce your max HP and thus make it more challenging to maneuver, but AC is bugged and he doesn’t use the ability at all (there’s been an open issue about this forever).

For fuck's sake, that's all.


isitsimon:
Christmas has arrived early

brighton-chi:
Alright the tanks will now go sit in front of Sanguinar and Telonicus during the 6-second interlude at the beginning of Phase 3 after Kael revives them but before they aggro (same spot where their dedicated healer remains for the whole phase). So hopefully that makes the first moments of Phase 3 less chaotic.

brighton-chi:
I'd like to get feedback on a matter. The strategy contains a method to force bots to reequip when within 150 yards of Kael'thas. This is due to the fact that after a wipe, players lose their legendary weapons and must reequip their own weapons, but bots do not do this without triggering a loot event. I understand now that a simple chat command can resolve this. Accordingly, should the method to reequip bots be removed? FWIW, it does not appear to be particularly resource intensive, but also it's possible that somebody might not want bots to equip what the bot thinks is the best upgrade so it does take away some flexibility from players.

Dreathean:
I would imagine that most players might not even realize those particular bots haven't re-equipped their weapons, so I would think that keeping the method vs relying on players to realize the issue, know the command, and use it is more practical.

brighton-chi:
Note this will not compile on the main branch right now because it is contingent on https://github.com/mod-playerbots/mod-playerbots/pull/2069

kadeshar:
Conflicts to resolve

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+        return false;
+
+    if (centurion->GetHealth() == centurion->GetMaxHealth())
+    {
+        if (!botAI->HasAura("polymorph", centurion))
+        {
+            if (botAI->CanCastSpell("polymorph", centurion))
+                return botAI->CastSpell("polymorph", centurion);
+        }
+    }
```
kadeshar: You can merge that into 1 if without nesting

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+{
+    Unit* capernian = AI_VALUE2(Unit*, "find target", "grand astromancer capernian");
+    if (!capernian)
+        return false;
+
+    // Main tank purposely stays in range to bait Conflagration in Phase 1
+    if (botAI->IsMainTank(bot))
+    {
+        constexpr float desiredDist = 15.0f;
+        if (fabs(bot->GetExactDist2d(capernian) - desiredDist))
```
kadeshar: I dont know what is fabs but return float not bool. That shouldnt be compared with something?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+        if (initialVoidReaverPositions.erase(guid) > 0)
+            erased = true;
+        if (hasReachedInitialVoidReaverPosition.erase(guid) > 0)
+            erased = true;
+    }
+    if (!AI_VALUE2(Unit*, "find target", "kael'thas sunstrider"))
+    {
+        if (advisorDpsWaitTimer.erase(instanceId) > 0)
+            erased = true;
+    }
```
kadeshar: You can merge that to one if

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    if (!centurion)
+        return false;
+
+    if (centurion->GetHealth() == centurion->GetMaxHealth())
+    {
+        if (!botAI->HasAura("polymorph", centurion))
+        {
+            if (botAI->CanCastSpell("polymorph", centurion))
+                return botAI->CastSpell("polymorph", centurion);
+        }
```
kadeshar: You can merge that into one if. Please review whole PR for this comment

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+                tankTarget = member;
+                break;
+            }
+        }
+    }
+    else if (hunterIndex == 2)
+    {
+        advisorTarget = AI_VALUE2(Unit*, "find target", "grand astromancer capernian");
+        tankTarget = GetCapernianTank(bot);
+    }
```
kadeshar: hunterIndex = 0 and 2 are duplicated

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    Unit* targetPriest = AssignSolariumPriestsToBots(priestsPair, meleeMembers);
+    if (!targetPriest)
+        return false;
+
+    auto it = std::find(meleeMembers.begin(), meleeMembers.end(), bot);
+    if (it == meleeMembers.end())
+        return false;
+
+    size_t botIndex = std::distance(meleeMembers.begin(), it);
+    size_t totalMelee = meleeMembers.size();
```
kadeshar: Usused variables

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    float angleOffset = (offsetArc / radius);
+    uint8 ringIndex = botIndex / botsPerRing;
+    uint8 posInRing = botIndex % botsPerRing;
+    float baseAngle = 2 * M_PI * posInRing / botsPerRing;
+    float angle = baseAngle + (ringIndex == 1 ? angleOffset : 0);
+
+    if (ringIndex > 1)
+    {
+        angle = 2 * M_PI * (botIndex % botsPerRing) / botsPerRing;
+        ringIndex = 0;
```
kadeshar: This value is never read

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+        bool opened = open.Execute(Event());
+
+        if (!opened)
+            return opened;
+
+        const ObjectGuid botGuid = bot->GetGUID();
+        const ObjectGuid corpseGuid = guid;
+        constexpr uint8 weaponIndex = 0;
+
+        botAI->AddTimedEvent([this, botGuid, corpseGuid, weaponIndex, itemId]()
```
kadeshar: this is required?

src/Ai/Raid/TempestKeep/Util/RaidTempestKeepHelpers.cpp
```diff
@@ -0,0 +1,497 @@
...
+            ALAR_PLATFORM_1,
+            ALAR_PLATFORM_2,
+            ALAR_PLATFORM_3,
+            ALAR_POINT_QUILL_OR_DIVE,
+            ALAR_POINT_MIDDLE,
+        };
+
+        float minDist = std::numeric_limits<float>::max();
+        int8 locationIndex = LOCATION_NONE;
+        for (int8 i = 0; i < 7; ++i)
```
kadeshar: That 7 is correct? Even that you should read locations size not hardcoded

src/Ai/Raid/TempestKeep/Util/RaidTempestKeepHelpers.cpp
```diff
@@ -0,0 +1,497 @@
...
+            Unit* unit = botAI->GetUnit(npcGuid);
+            if (unit && unit->GetEntry() == npcEntry && bot->GetExactDist2d(unit) < maxSearchRadius)
+                hazardTriggers.push_back(unit);
+        }
+
+        return hazardTriggers;
+    }
+
+    Position FindSafestNearbyPosition(Player* bot, const std::vector<Unit*>& hazards,
+        float maxRadius, float hazardRadius, const Position* center)
```
kadeshar: maxRadius is not used

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    Unit* alar = AI_VALUE2(Unit*, "find target", "al'ar");
+    if (!alar)
+        return false;
+
+    if (!isAlarInPhase2[alar->GetMap()->GetInstanceId()])
+        return HandlePhase1Embers(alar);
+    else
+        return HandlePhase2Embers(alar);
+
+    return false;
```
kadeshar: Unreachable

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    // But this applies to assist tanks only after they get aggro on the mace, dagger, or sword
+    Unit* axe = AI_VALUE2(Unit*, "find target", "devastation");
+    Unit* mace = AI_VALUE2(Unit*, "find target", "cosmic infuser");
+    Unit* dagger = AI_VALUE2(Unit*, "find target", "infinity blades");
+    Unit* sword = AI_VALUE2(Unit*, "find target", "warp slicer");
+    if (axe)
+    {
+        if (botAI->IsDps(bot) || botAI->IsHeal(bot) ||
+            (botAI->IsAssistTank(bot) && (mace && mace->GetVictim() == bot ||
+             dagger && dagger->GetVictim() == bot || sword && sword->GetVictim() == bot)))
```
kadeshar: Uffff hard to read that. Missing parentheses?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    if (!capernian)
+        return false;
+
+    MarkTargetWithCircle(bot, capernian);
+    SetRtiTarget(botAI, "circle", capernian);
+
+    if (bot->GetTarget() != capernian->GetGUID())
+    {
+        if (botAI->CanCastSpell("searing pain", capernian))
+            botAI->CastSpell("searing pain", capernian);
```
kadeshar: That if shouldnt return true?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    if (bot->GetExactDist2d(destX, destY) > 1.0f)
+    {
+        return MoveTo(TEMPEST_KEEP_MAP_ID, destX, destY, targetPosition.GetPositionZ(), false,
+                      false, false, false, MovementPriority::MOVEMENT_FORCED, true, false);
+    }
+
+    return false;
+}
+
+Position VoidReaverSpreadRangedAction::GetRangedBotPosition(const Position& center,
```
kadeshar: That method is used?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+    Unit* astromancer = AI_VALUE2(Unit*, "find target", "high astromancer solarian");
+    if (!astromancer)
+        return false;
+
+    if (bot->GetTarget() != astromancer->GetGUID())
+        return Attack(astromancer);
+
+    if (astromancer->GetVictim() != bot)
+    {
+        const char* taunts[] = { "taunt", "growl", "hand of reckoning", "dark command" };
```
kadeshar: Duplicated code

Image

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,2142 @@
...
+        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+        {
+            Player* member = ref->GetSource();
+            if (member && botAI->IsMainTank(member))
+            {
+                mainTank = member;
+                break;
+            }
+        }
+    }
```
kadeshar: That looks like something to move to helper.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Yeah, I do nesting for single statements sometimes when I feel like the concepts are distinct and when for me it's easier to read them separately. I think combining this one would make sense, but sometimes I think combining makes things meaningfully harder to read. You want me to always combine when possible?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: I'll refactor this. I think I was using fabs originally to compare a +/- range for distance. But then I changed it to send the bot to a fixed distance so I think fabs doesn't really make sense anymore. The tank shouldn't be doing anything here anyway so the method can just calculate a spot 15 yards from Capernian and call MoveTo to that spot, way simpler.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Haha, not missing, but maybe I can do the line breaks in a better way, and maybe its better to combine DPS and Heal checks into a not Tank check. The concept is the following bots should move away from the axe: (1) DPS, (2) Healers, and (3) assist tanks, but only if they have aggro from the mace, dagger, and/or sword.

I did it this way because the main tank needs to move the axe away from everybody, but sometimes he gets other weapons stuck to him too that the assist tanks need to get off him and then move away so melee can attack weapons other than the axe.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Maybe not. I had a delay for Vashj core actions, and if I didn't delay looting or passing the core for Vashj, bots would double up on actions, maybe because it took a bit for the packet to register. I didn't try without it here though.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Thanks. Got into a habit of starting every method by just dropping a return false at the end.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: I don't think so. I messed with this method a ton and haven't looked at it in a long time so I just need to clean this up.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Thanks. I usually start with a giant ugly execute function, and then when I can get the method actually working in game, I might break it out into multiple functions. I think I just forgot to pull this part out.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Actually maybe this can just be DoSpecificAction() for "taunt spell"

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Hm you mean the general raid boss helper? Something like Player* GetMainTank()?

src/Ai/Raid/TempestKeep/Util/RaidTempestKeepHelpers.cpp
brighton-chi: Thanks. Leftover from Vashj, when I needed to limit the search to the platform.

src/Ai/Raid/TempestKeep/Util/RaidTempestKeepHelpers.cpp
brighton-chi: Oops, should be 6. I can define 6 as a constant.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
kadeshar: Yea, we should avoid nesting

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
kadeshar: Maybe GetGroupMainTank

brighton-chi:
Ok, should have all comments addressed now. I was also able to simplify quite a bit of the code, and also I totally redid Void Reaver's ranged spread and it works much, much better I think.

I also added getters for the main tank and first three assist tanks to the general boss helper. This will be useful going forward at least for misdirection methods.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
kadeshar: I mean parameter this. Is usused. Its required?
EDIT:
didnt notice is commented code. Is required for something?

src/Ai/Raid/TempestKeep/Util/RaidTempestKeepHelpers.cpp
kadeshar: Not fixed

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -994,33 +878,18 @@ bool HighAstromancerSolarianTankVoidwalkerAction::Execute(Event /*event*/)
...
     {
-        const char* taunts[] = { "taunt", "growl", "hand of reckoning", "dark command" };
-        for (const char* spellName : taunts)
-        {
-            if (botAI->CanCastSpell(spellName, astromancer))
-                return botAI->CastSpell(spellName, astromancer);
-        }
+        bool taunted = botAI->DoSpecificAction("taunt spell", Event(), true);
+        LOG_DEBUG("playerbots", "Bot {}: taunt on High Astromancer Solarian {}", bot->GetName(),
+                    taunted ? "succeeded" : "failed");
```
kadeshar: That require logs?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,1989 @@
...
+        else if (botAI->IsAssistTankOfIndex(bot, 1, true))
+        {
+            if (!AI_VALUE2(Unit*, "find target", "ember of al'ar") && bot->GetExactDist2d(
+                groundTarget.GetPositionX(), groundTarget.GetPositionY()) > 20.0f)
+            {
+                return MoveInside(TEMPEST_KEEP_MAP_ID, groundTarget.GetPositionX(),
+                                  groundTarget.GetPositionY(), groundTarget.GetPositionZ(),
+                                  20.0f, MovementPriority::MOVEMENT_COMBAT);
+            }
+        }
```
kadeshar: Merge if's

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: Oh, haha. Well actually the entire delay logic is not needed. I tested already, just forgot to delete the commented out stuff. I'll update it.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,1989 @@
...
+    else
+    {
+        MarkTargetWithStar(bot, targetPriest);
+        SetRtiTarget(botAI, "star", targetPriest);
+    }
+
+    if (bot->GetTarget() != targetPriest->GetGUID())
+        return Attack(targetPriest);
+
+    return false;
```
kadeshar, line 893: This false is real? Action not executed?

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,1989 @@
...
+
+    if (movePosition)
+    {
+        if (bot->GetExactDist2d(movePosition->GetPositionX(), movePosition->GetPositionY()) > 2.0f)
+        {
+            return MoveTo(TEMPEST_KEEP_MAP_ID, movePosition->GetPositionX(), movePosition->GetPositionY(),
+                          movePosition->GetPositionZ(), false, false, false, false,
+                          MovementPriority::MOVEMENT_FORCED, true, false);
+        }
+    }
```
kadeshar: Merge if's

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,1989 @@
...
+        {
+            if (Unit* phoenixEgg = GetFirstAliveUnitByEntry(botAI, NPC_PHOENIX_EGG))
+            {
+                MarkTargetWithDiamond(bot, phoenixEgg);
+                SetRtiTarget(botAI, "diamond", phoenixEgg);
+
+                if (bot->GetTarget() != phoenixEgg->GetGUID())
+                    return Attack (phoenixEgg);
+            }
+        }
```
kadeshar: Merge if's

src/Ai/Raid/TempestKeep/Multiplier/RaidTempestKeepMultipliers.cpp
```diff
@@ -0,0 +1,424 @@
...
+        return 1.0f;
+
+    Creature* alarCreature = alar->ToCreature();
+    if (alarCreature && alarCreature->GetReactState() == REACT_PASSIVE)
+    {
+        if (dynamic_cast<MovementAction*>(action) &&
+            !dynamic_cast<AlarMoveAwayFromRebirthAction*>(action) &&
+            !dynamic_cast<AlarAvoidFlamePatchesAndDiveBombsAction*>(action))
+            return 0.0f;
+    }
```
kadeshar: MErge if's

src/Ai/Raid/TempestKeep/Multiplier/RaidTempestKeepMultipliers.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+    Unit* alar = AI_VALUE2(Unit*, "find target", "al'ar");
+    if (alar && bot->GetTarget() == alar->GetGUID())
+    {
+        if (dynamic_cast<CastTauntAction*>(action) ||
+            dynamic_cast<CastGrowlAction*>(action) ||
+            dynamic_cast<CastHandOfReckoningAction*>(action) ||
+            dynamic_cast<CastDarkCommandAction*>(action))
+            return 0.0f;
+    }
```
kadeshar: Merge if's

src/Ai/Raid/TempestKeep/Multiplier/RaidTempestKeepMultipliers.cpp
```diff
@@ -0,0 +1,424 @@
...
+float VoidReaverMaintainPositionsMultiplier::GetValue(Action* action)
+{
+    if (!AI_VALUE2(Unit*, "find target", "void reaver"))
+        return 1.0f;
+
+    if (botAI->IsTank(bot))
+    {
+        if (dynamic_cast<CombatFormationMoveAction*>(action))
+            return 0.0f;
+    }
```
kadeshar: Merge if's

src/Ai/Raid/TempestKeep/Multiplier/RaidTempestKeepMultipliers.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+    if (botAI->IsRanged(bot))
+    {
+        if (dynamic_cast<CombatFormationMoveAction*>(action) ||
+            dynamic_cast<FleeAction*>(action) ||
+            dynamic_cast<CastBlinkBackAction*>(action) ||
+            dynamic_cast<CastDisengageAction*>(action) ||
+            dynamic_cast<ReachTargetAction*>(action))
+            return 0.0f;
+    }
```
kadeshar: Merge all if's in all such situations

brighton-chi:
Actually I didn't push the latest updates, oops. I had stashed some changes then made more updates on another branch but didn't bring them over so I need to resolve that.

So don't review further right now, sorry.

src/Ai/Raid/TempestKeep/Multiplier/RaidTempestKeepMultipliers.cpp
brighton-chi: You want to do that for multipliers too? I think it is confusing to have the blocked actions combined with all the conditions.

kadeshar: Everywhere where nesting is no need

brighton-chi: Ok. It's just harder to read and maintain because even if syntactically it is possible to combine, the two levels cover very different concepts that logically flow from each other: (1) outer level says what is the situation or role for which we need to block actions, and (2) inner level says here are the actions that need to be blocked for that situation or role.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,1989 @@
...
+
+    MarkTargetWithStar(bot, alar);
+    SetRtiTarget(botAI, "star", alar);
+
+    int8 locationIndex = GetAlarCurrentLocationIndex(alar);
+    if (locationIndex == LOCATION_NONE)
+    {
+        Position dest;
+        locationIndex = GetAlarDestinationLocationIndex(alar, dest);
+    }
```
trolloks, line 72: This is duplicated in a few places, wouldn't it make sense to move this to a helper function?

src/Ai/Raid/TempestKeep/Multiplier/RaidTempestKeepMultipliers.cpp
kadeshar: Then use early exit in method. But good practise is merge if's

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -0,0 +1,1989 @@
...
+            groundTarget.GetPositionX(), groundTarget.GetPositionY()) > 8.0f)
+        {
+            return MoveInside(TEMPEST_KEEP_MAP_ID, groundTarget.GetPositionX(),
+                              groundTarget.GetPositionY(), groundTarget.GetPositionZ(),
+                              8.0f, MovementPriority::MOVEMENT_COMBAT);
+        }
+        else if (botAI->IsAssistTankOfIndex(bot, 1, true))
+        {
+            if (!AI_VALUE2(Unit*, "find target", "ember of al'ar") && bot->GetExactDist2d(
+                groundTarget.GetPositionX(), groundTarget.GetPositionY()) > 20.0f)
```
trolloks: Would it make sense to make these numbers constants for better readability?

trolloks (commented):
Still very new with the codebase so just looking at some general coding things.

trolloks:
> Actually I didn't push the latest updates, oops. I had stashed some changes then made more updates on another branch but didn't bring them over so I need to resolve that.
> 
> So don't review further right now, sorry.

Missed this comment, will check back in later

brighton-chi:
Went ahead and also revised the multipliers along with the changes I missed pushing.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
kadeshar: That is not changed and i didnt got answer.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
kadeshar: Still not merged

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: I'm not sure I understand your question. Are you asking if the false is needed? It is for if the bot is already targeting the priest.

kadeshar: Method return false even after succesful marking target and set rti target. I wouldnt say that didnt executed.

brighton-chi: Yeah, that's how it's been in all the boss strats up to this point since they're void functions. I guess they could be changed to boolean functions to return true without making bots get stuck in the function. Marking/setting RTI target shouldn't ever early return for a success though, but I assume using a boolean variable and then returning at the end of the function would get around that issue.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: I can't merge them because phoenixEgg is defined in the inner if, and if I define it in the outer if instead, it has to come before the IsDps() check so then adds performance cost because healers have to do the grid search. Would you prefer the below way instead? It just adds an extra line so it doesn't seem better, but I don't care either way.

```
    if (botAI->IsDps(bot) && !kaelthas->HasAura(SPELL_SHOCK_BARRIER))
    {
        Unit* phoenixEgg = GetFirstAliveUnitByEntry(botAI, NPC_PHOENIX_EGG);
        if (phoenixEgg)
        {
            MarkTargetWithDiamond(bot, phoenixEgg);
            SetRtiTarget(botAI, "diamond", phoenixEgg);

            if (bot->GetTarget() != phoenixEgg->GetGUID())
                return Attack(phoenixEgg);
        }
    }
    else if (botAI->IsDps(bot))
        return false;
```

kadeshar: Aaaaaa ok, then leave :)

kadeshar (dismissed):

brighton-chi:
Apologies, reverting to draft for now while I investigate grid searches further, as I think there are some huge performance changes that potentially could be made, but I need to test what works to find what needs to be found, so to speak.

See: https://github.com/mod-playerbots/mod-playerbots/discussions/2157

brighton-chi:
Ok, sorry about that. Performance on the modified searches is much better though, and the weapon logic was busted as all fuck even though it technically worked.

brighton-chi:
Can't believe I didn't think about this before. Performance should be better with redundant checks that funnel. So instead of directly checking group role iterations like IsAssistTankOfIndex(), first check IsTank() so only tanks run that loop.

kadeshar:
> Can't believe I didn't think about this before. Performance should be better with redundant checks that funnel. So instead of directly checking group role iterations like IsAssistTankOfIndex(), first check IsTank() so only tanks run that loop.

I thinking about that and see inconsistent in naming/implementation.

IsTank -> only if have tank strategy or his spec point that he is tank
IsMainTank -> have group flag main tank or its first IsTank

Not better to IsMainTank method add parameter flag ignore groupFlag check and right before group iteration check that bot isTank? That make IsMainTank method more dynamic.

obraz

kadeshar:
Also please make PR open when you dont plan add more fixes/changes. Adding changes to open PR every few hours make review harder.

brighton-chi:
> Also please make PR open when you dont plan add more fixes/changes. Adding changes to open PR every few hours make review harder.

Yeah, sorry, problem is I always think I'm done then think of more things...

kadeshar:
Thinking more im not sure that you should ignore MEMBER_FLAG_MAINTANK.
Real player dont have tank strategy and can be main tank

brighton-chi:
> > Can't believe I didn't think about this before. Performance should be better with redundant checks that funnel. So instead of directly checking group role iterations like IsAssistTankOfIndex(), first check IsTank() so only tanks run that loop.
> 
> I thinking about that and see inconsistent in naming/implementation.
> 
> IsTank -> only if have tank strategy or his spec point that he is tank IsMainTank -> have group flag main tank or its first IsTank
> 
> Not better to IsMainTank method add parameter flag ignore groupFlag check and right before group iteration check that bot isTank? That make IsMainTank method more dynamic.
> 
> obraz

Trying to understand--the way you have it now, by default, the main tank flag is being ignored entirely, so unless the second parameter is true, then IsMainTank() just returns the first tank? And if the second parameter is false, then you don't check whether the main tank is a tank at all, just iterate the whole group for the flag?

kadeshar:
If flag is false its work like now except adding early check that player istank and isalive before searching first tank alive in group. Its your idea but in other place :P

brighton-chi:
> Thinking more im not sure that you should ignore MEMBER_FLAG_MAINTANK. Real player dont have tank strategy and can be main tank

Real players are tanks though if they have a tank spec because botAi will be nullptr so the function will just evaluate by spec.

```
bool PlayerbotAI::IsTank(Player* player, bool bySpec)
{
    PlayerbotAI* botAi = GET_PLAYERBOT_AI(player);
    if (!bySpec && botAi)
        return botAi->ContainsStrategy(STRATEGY_TYPE_TANK);

    int tab = AiFactory::GetPlayerSpecTab(player);
    switch (player->getClass())
    {
        case CLASS_DEATH_KNIGHT:
            if (tab == DEATH_KNIGHT_TAB_BLOOD)
            {
                return true;
            }
            break;
        case CLASS_PALADIN:
            if (tab == PALADIN_TAB_PROTECTION)
            {
                return true;
            }
            break;
        case CLASS_WARRIOR:
            if (tab == WARRIOR_TAB_PROTECTION)
            {
                return true;
            }
            break;
        case CLASS_DRUID:
            if (tab == DRUID_TAB_FERAL && (player->GetShapeshiftForm() == FORM_BEAR ||
                                           player->GetShapeshiftForm() == FORM_DIREBEAR || player->HasAura(16931)))
            {
                return true;
            }
            break;
    }
    return false;
}
```

kadeshar:
You have right about that but still:
- you ignore MEMBER_FLAG_MAINTANK for bots. Maybe its good but its hidden in code
- changing IsMainTank will optimize performance globally, not only in your raid strategy

brighton-chi:
> If flag is false its work like now except adding early check that player istank and isalive before searching first tank alive in group. Its your idea but in other place :P

Oh, I think it is missing ! in the first check you added then. But otherwise it makes sense to me. What are you thinking for process? Are you going to do a PR for that?

kadeshar:
> > If flag is false its work like now except adding early check that player istank and isalive before searching first tank alive in group. Its your idea but in other place :P
> 
> Oh, I think it is missing ! in the first check you added then. But otherwise it makes sense to me. What are you thinking for process? Are you going to do a PR for that?

Yeap, missing ! :D About where to place changes its your choice. You can add here or in separate PR.

brighton-chi:
> > > If flag is false its work like now except adding early check that player istank and isalive before searching first tank alive in group. Its your idea but in other place :P
> > 
> > 
> > Oh, I think it is missing ! in the first check you added then. But otherwise it makes sense to me. What are you thinking for process? Are you going to do a PR for that?
> 
> Yeap, missing ! :D About where to place changes its your choice. You can add here or in separate PR.

There is a performance consequence though if you let any raid member be the main tank because then every bot is going to iterate the raid to find the flag. If you do botAI->IsTank(bot) && botAI0->IsMainTank(bot), then every bot will check if it is a tank, and only tanks will iterate the group to find the flag, but of course it doesn't work if the flag is put on a non-tank.

I guess it depends how important you think it is for strategies to support real players who want to tank without a tank spec. I tend to think that if you can beat the encounter with a main tank who is not a tank spec, maybe you don't need a strategy at all...

kadeshar:
Yeap, that is something setup by player. If he want try, why not.

brighton-chi:
> Yeap, that is something setup by player. If he want try, why not.

Just performance is a little worse for everybody to allow that. A judgment call.

I think for sure the assistant functions in playerbotai need to be refactored to do the role check before iterating. There is no reason for them not to do that. So if we do that and change main tank then we don't need to do the doubled up checks like I have right now.

So maybe I will refactor those functions also in this PR along with main tank and then in the tempest keep boss strategies remove doubled up checks?

brighton-chi:
Also I need to fix the GetGroup tank functions. Dunno why I did them so stupidly. Loop calling a loop. They can just replicate the assistant functions and return Player instead of bool. Lmao.

Can you put this back on draft, and I'll address tomorrow? I don't think I can put it on draft via the app and I'm not at a computer.

kadeshar:
Ok, if possible when you change all can you send me PM to review that again?

brighton-chi:
@kadeshar 

I updated boolean main tank and assistant tank, healer, and DPS functions to add early returns and the main tank parameter you proposed. Also the assistant functions now do only one loop to check for both assistants and assistants. I also fixed the main and assist tank Player functions in RaidBossHelpers.cpp. And then I modified Void Reaver's strategy to include Arcane Orb avoidance based on the spell script made by @trolloks , which means I also had to register the script--I added it to the bottom of Playerbots.cpp but am not sure if it should be done in some other way.

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
```diff
@@ -721,6 +735,69 @@ int VoidReaverSpreadRangedAction::GetRangedDpsIndex(Group* group, int& rangedDps
...
+    return false;
+}
+
+bool VoidReaverEraseTrackersAction::Execute(Event /*event*/)
+{
+    Unit* voidReaver = AI_VALUE2(Unit*, "find target", "void reaver");
+    if (voidReaver)
+        return false;
+
+    bool erased = false;
```
kadeshar, line 790: This variable is redundant

src/Ai/Raid/TempestKeep/Trigger/RaidTempestKeepTriggers.cpp
```diff
@@ -146,6 +146,41 @@ bool VoidReaverBossLaunchesArcaneOrbsTrigger::IsActive()
...
+    constexpr float safeDistance = 22.0f;
+
+    for (const auto& orb : it->second)
+    {
+        if (getMSTimeDiff(orb.castTime, currentTime) <= orbDuration)
+        {
+            if (bot->GetExactDist2d(orb.destination.GetPositionX(),
+                                    orb.destination.GetPositionY()) < safeDistance)
+                return true;
+        }
```
kadeshar: Merge if's

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
brighton-chi: What do you mean? I want this action to be able to return true if either map is successfully erased or both are successfully erased, but I don't want it to have to run the function twice to erase each map.

kadeshar: I would say something like this will replace this variable

```
return voidReaverArcaneOrbs.erase(bot->GetMap()->GetInstanceId()) |
	hasReachedVoidReaverPosition.erase(bot->GetGUID());
```

brighton-chi: Won't the function return true after the first map is erased then? So it would take two ticks to erase both maps?

kadeshar (approved):

src/Ai/Raid/TempestKeep/Action/RaidTempestKeepActions.cpp
kadeshar: Its will work but with parentheses. But for readability maybe better leave that variable

hermensbas (approved):

PR #1944, opened 2025-12-24
Add pvp gear to bots in bg and arena

Regrad:
If the bot engages in battle in BG or Arena, it disguises itself in PVP gear. When exiting BG or Arena, the bot changes back into its PVE. A different rule applies to characters from the player's account. When entering the BG or Arena, the Bot-alt searches for PVP things in bags and puts them on, and when leaving the BG or arena, he puts back on only from what is in his bags.

In the arena, the level of the bot's items (except for the alt bot) depends on the rating of the arena team, where 1000 ratings correspond to 200 level items, and 2400 ratings correspond to 300 level items (with a margin).

### DIFF
--- a/src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
+++ b/src/strategy/raids/magtheridon/RaidMagtheridonActions.cpp
@@ -401,14 +401,20 @@ std::unordered_map<ObjectGuid, bool> MagtheridonSpreadRangedAction::hasReachedIn
 
 bool MagtheridonSpreadRangedAction::Execute(Event event)
 {
+    Unit* magtheridon = AI_VALUE2(Unit*, "find target", "magtheridon");
+    if (!magtheridon)
+        return false;
+
     Group* group = bot->GetGroup();
     if (!group)
         return false;
 
+    const uint32 instanceId = magtheridon->GetMap()->GetInstanceId();
+
     // Wait for 6 seconds after Magtheridon activates to spread
     const uint8 spreadWaitSeconds = 6;
-    auto it = magtheridonSpreadWaitTimer.find(bot->GetMapId());
-    if (it == magtheridonSpreadWaitTimer.end() ||
+    auto it = spreadWaitTimer.find(instanceId);
+    if (it == spreadWaitTimer.end() ||
         (time(nullptr) - it->second) < spreadWaitSeconds)
         return false;
 
@@ -416,8 +422,8 @@ bool MagtheridonSpreadRangedAction::Execute(Event event)
     if (cubeIt != botToCubeAssignment.end())
     {
         time_t now = time(nullptr);
-        auto timerIt = magtheridonBlastNovaTimer.find(bot->GetMapId());
-        if (timerIt != magtheridonBlastNovaTimer.end())
+        auto timerIt = blastNovaTimer.find(instanceId);
+        if (timerIt != blastNovaTimer.end())
         {
             time_t lastBlastNova = timerIt->second;
             if (now - lastBlastNova >= 49)
@@ -559,8 +565,8 @@ bool MagtheridonUseManticronCubeAction::HandleCubeRelease(Unit* magtheridon, Gam
 
 bool MagtheridonUseManticronCubeAction::ShouldActivateCubeLogic(Unit* magtheridon)
 {
-    auto timerIt = magtheridonBlastNovaTimer.find(bot->GetMapId());
-    if (timerIt == magtheridonBlastNovaTimer.end())
+    auto timerIt = blastNovaTimer.find(magtheridon->GetMap()->GetInstanceId());
+    if (timerIt == blastNovaTimer.end())
         return false;
 
     time_t now = time(nullptr);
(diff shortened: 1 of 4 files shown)


brighton-chi:
How does the gearing work if you are playing arenas at 70?

src/factory/StatsWeightCalculator.cpp
```diff
@@ -91,6 +91,17 @@ float StatsWeightCalculator::CalculateItem(uint32 itemId, int32 randomPropertyId
...
     CalculateSocketBonus(player_, proto);
 
+// PvP gear preference:
+// In BG/Arena prefer items with Resilience; in world prefer non-Resilience.
+bool inPvpContext = player_ && (player_->InBattleground() || player_->InArena());
+bool hasResilience = (collector_->stats[STATS_TYPE_RESILIENCE] > 0.0f);
+if (hasResilience && !inPvpContext)
+{
+    // Penalize resilience gear in PvE so bots swap back to PvE sets.
+    weight_ *= 0.001f;
```
noisiver: Curly brackets with single line if. Remove them

src/PlayerbotAI.cpp
```diff
@@ -434,6 +441,126 @@ void PlayerbotAI::UpdateAIGroupMaster()
...
+
+        if (Battleground* bg = bot->GetBattleground())
+        {
+            if (bg->isArena())
+            {
+                uint8 arenaType = bg->GetArenaType(); // 2,3,5
+                uint8 slot = ArenaTeam::GetSlotByType(arenaType);
+                uint32 teamId = bot->GetArenaTeamId(slot);
+                if (teamId)
+                {
```
noisiver: Curly brackets with single line if. Remove them

Regrad:
> How does the gearing work if you are playing arenas at 70?

Just like the usual pve items. They are selected from a items pool, but resistance begins to be considered the main stat. Wasn't it at level 70?

src/factory/PlayerbotFactory.cpp
```diff
@@ -4099,10 +4102,19 @@ void PlayerbotFactory::InitImmersive()
...
+    // Extra safety: even if InitArenaTeam() is called from somewhere else,
+    // never touch arena teams below the required level.
+    if (bot->GetLevel() < requiredLevel)
+        return;
 
+    // Arena teams are only for random-bot accounts, never for real player alts.
     if (!sPlayerbotAIConfig->IsInRandomAccountList(bot->GetSession()->GetAccountId()))
         return;
 
+
```
noisiver: Blank line for no reason

brighton-chi:
> > How does the gearing work if you are playing arenas at 70?
> 
> 
> 
> Just like the usual pve items. They are selected from a items pool, but resistance begins to be considered the main stat. Wasn't it at level 70?

It was, yeah. I was wondering because you mentioned that item levels increase as rating increases, but of course level 70 would have different item levels. It sounds like it would work the same as 80 though, just with level 70 gear.

I like this idea a lot btw. I don't want all bots running around the world in epics, but then they are never a challenge in arenas.

src/PlayerbotAI.cpp
```diff
@@ -434,6 +441,126 @@ void PlayerbotAI::UpdateAIGroupMaster()
...
+        pvpGearSwapCooldown = 5000;
+        return;
+    }
+
+    // Important: "alt" bots may be temporarily controlled by someone else.
+    // We must not decide this based on master's account.
+    // The module already uses RandomAccountList to distinguish real alts from bot accounts
+    // (see AutoGearAction / PlayerbotAIConfig::IsInRandomAccountList).
+    // - accounts in RandomAccountList => bot accounts (allowed to generate gear)
+    // - all other accounts            => real player alts (bag-only)
```
noisiver: What exactly does that mean? Altbots should be completely unaffected in all situations, including whatever "bag-only" means. Just don't mess with altbots, change this PR to completely ignore altbots.

src/PlayerbotAI.cpp
```diff
@@ -434,6 +441,126 @@ void PlayerbotAI::UpdateAIGroupMaster()
...
+    bool isBotAccount = sRandomPlayerbotMgr->IsRandomBot(bot) ||
+                        sPlayerbotAIConfig->IsInRandomAccountList(bot->GetSession()->GetAccountId());
+    bool bagOnly = !isBotAccount;
```
noisiver: Why does this exist? bagOnly is only called right under this so that if might as well
```cpp
if (!isBotAccount)
```
rather than this definition followed by
```cpp
if (bagOnly)
```

noisiver (changes requested):

src/strategy/actions/BattleGroundJoinAction.cpp
```diff
@@ -8,13 +8,56 @@
...
 #include "BattlegroundMgr.h"
+#include "SharedDefines.h"
 #include "Event.h"
 #include "GroupMgr.h"
 #include "PlayerbotAI.h"
 #include "Playerbots.h"
 #include "PositionValue.h"
 #include "UpdateTime.h"
 
+
```
noisiver: Blank line for no reason

src/PlayerbotAI.cpp
```diff
@@ -434,6 +441,126 @@ void PlayerbotAI::UpdateAIGroupMaster()
...
+    }
+
+
+    PlayerbotFactory factory(bot, bot->GetLevel(), ITEM_QUALITY_LEGENDARY, gsLimit);
+    factory.InitEquipment(false, true);
+
+    pvpGearSwapPending = false;
+}
+
+
```
noisiver: Blank line for no reason

src/PlayerbotAI.h
```diff
@@ -643,6 +645,11 @@ class PlayerbotAI : public PlayerbotAIBase
     BotCheatMask cheatMask = BotCheatMask::none;
     Position jumpDestination = Position();
     uint32 nextTransportCheck = 0;
+    // PvP/BG/Arena gear swap tracking
+    bool lastPvpContext = false;
+    bool pvpGearSwapPending = false;
+    uint32 pvpGearSwapCooldown = 0;
+
```
noisiver: Blank line for no reason

noisiver (changes requested):

Regrad:
Wait! I'm not done yet!)) That's not even all I'm working on!) I've started a complete PVP redesign here. Reasons:
1) The current team creation system does not do everything very efficiently. With 2,000 bots, of which 700 are level 80, only a few teams are actually engaged in the arena.
2) Bots go to PVP in PVE stuff
3) A random BG allows only 10 bots to enter, even if the BG is designed for 40 people.
4) On any rating, the bots are dressed the same
5) There is a constant selection of teams with a rating that does not match the player (due to the small number of teams).
6) There are often situations that prevent bots from playing in some formations at all, for example 5x5.

I have taken on a very difficult task and I hope to be able to complete it. And I hope that my vision will be approved.

noisiver:
It doesn't matter which account an altbot belongs to since they can never be a rndbot, they must not be affected by this. You can't even touch their bags or anything in them. Just exclude them from this.

brighton-chi:
There's certain plenty to improve with PvP. A few things that come to mind:

Bots with potion strategies built into their class strategies need to have those migrated to the overall potions strategy so they will be disabled for arenas.

Rogues still use MH instant OH deadly for PvP. I think poison usage needs to be tied to strategies for MH and OH poisons. Then defaults need to change. In BGs and Arena, they should MH wound and OH crippling by default. Outside of BGs, Assassination spec (after level 50) should MH deadly and OH instant. This is something that is on my backburner to do someday, though of course if somebody else wanted to do it, that would be preferable. 🙃 

Their logic for usage of Cloak of Shadows is also terrible (and tbh for cooldowns in general).

Bots do not have code to use Will of the Forsaken or Every Man for Himself. Those are huge PvP abilities.

brighton-chi:
Actually, for temporary equipment, I wonder if bots should equip the PvP trinket. They would also need to have logic built in to use it properly, though, so maybe it's a bit out of scope.

Regrad:
> It doesn't matter which account an altbot belongs to since they can never be a rndbot, they must not be affected by this. You can't even touch their bags or anything in them. Just exclude them from this.

I don't understand.

PR #1945, opened 2025-12-24
Minor fixes to .dist descriptions

brighton-chi:
Clarified that MinEnchantingBotLevel also determines whether maintenance will socket gems, in addition to applying enchants.

Fixed max iLevel for ZA gear in RandomGearScoreLimit description (iLevel 138 is the max that drops from the final boss, but the 3rd timed chest gives iLevel 141 equipment).


brighton-chi:
@Celandriel 

A few questions re: the fishing strategy (I want to clarify the config since this PR is editing the config anyway, and just personally I want to try out the feature so want to know how it works):
1. Which bots can have the "master fishing" strategy? Is it altbots and rndbots? It sounds like they don't have to be in any master's party, so maybe the better question is what bots _cannot_ fish?
2. FishingDistance is the distance _from a bot_ that such bot will search for water, provided it doesn't have a master?
3. I'm not clear on what FishingDistanceFromMaster means. Does that mean that if the master moves more than the configured yards away from the bot, the bot will stop fishing?
4. Where did you end up with bots reequipping gear after stopping fishing? Will they do so right away? You know where I'm going with this, I'm sure...

Celandriel:
Master fishing is a response to a bot seeing their master fish. This can be any bot that has fishing.

Fishing distance is how far a bot will search for water if it does not have a master.
If a bot has a master, it restricts it to the second distance because it overrides the follow distance which was causing issues with bots ping ponging back and forth. 

You would have to force masterless bots  to have the strategy, but I decided to do it this way for future proofing for when bots do have professions. 

And no, I didn't solve the combat trigger issue. My best bet would be to try and trigger an equip upgrades event. 

Maybe bots need a combat strategy that gets added when certain other activities take place, so that the first combat action is a equip upgrades.

brighton-chi:
> Master fishing is a response to a bot seeing their master fish. This can be any bot that has fishing.
> 
> Fishing distance is how far a bot will search for water if it does not have a master. If a bot has a master, it restricts it to the second distance because it overrides the follow distance which was causing issues with bots ping ponging back and forth.
> 
> You would have to force masterless bots to have the strategy, but I decided to do it this way for future proofing for when bots do have professions.
> 
> And no, I didn't solve the combat trigger issue. My best bet would be to try and trigger an equip upgrades event.
> 
> Maybe bots need a combat strategy that gets added when certain other activities take place, so that the first combat action is a equip upgrades.

Hm, so essentially fishing distance is not a config option that has any practical impact in game right now?

I'm sure I could make the Lurker strat force equip upgrades at the beginning with botAI->DoSpecificAction("equip upgrades"). I don't know if that would annoy people, though. I do that for Kael'thas, but that one is a non-combat version that forces reequip when you're within 150 yards of him to address the issue of losing the temporary legendary weapons after a wipe and bots not reequipping because it is ridiculously annoying without a method like that.

It would be nice to have a chat command to equip upgrades. Right now if I swap specs on a bot, to get them to swap equipment I trade them something like a piece of cloth--it's kind of stupid.

Celandriel:
I think adding the chat command is easiest, the action is already there, all that is missing is the trigger.

Is there time between lurker spawning and combat such that you can detect the spawn event and then trigger equip?

brighton-chi:
> I think adding the chat command is easiest, the action is already there, all that is missing is the trigger.
> 
> Is there time between lurker spawning and combat such that you can detect the spawn event and then trigger equip?

Yeah, well another one for the TBD pile for whoever wants to and can do it (maybe I can, I'd have to take a look).

Lurker aggros right away. I don't think he hits all that hard so if bots have to expend a tick to reequip, that's probably not a problem. He doesn't move at all so there's nothing vital about picking him up ASAP, but you do need somebody in melee range of him at all times. My main concern with forcing equipment upgrades is that somebody might have the config for auto equipping upgrades intentionally disabled because they want to manage their bots' gear depending on situation. Like maybe they carry a mix of threat and mitigation pieces on their tank and swap manually. But anyway a chat command to equip upgrades I think would be broadly useful.

When you have a chance, please let me know if changes to the fishing comments in the config are accurate/make sense.

noisiver (approved):

brighton-chi:
Updated fishing description per Keleborn's feedback. Also broke up a few long lines and tried to clarify some stuff (e.g., what "SpecialSpells" are).

Celandriel (approved):

PR #1949, opened 2025-12-25
Add secure login handling for playerbots

BeardBear33:
Introduces PlayerbotsSecureLoginServerScript to prevent multiple logins to the same bot character by disconnecting existing bot sessions. Registers the new script in Playerbots.cpp to enhance account security for playerbots.

### DIFF
--- a/src/Playerbots.cpp
+++ b/src/Playerbots.cpp
@@ -513,7 +513,7 @@ void AddPlayerbotsScripts()
     new PlayerbotsWorldScript();
     new PlayerbotsScript();
     new PlayerBotsBGScript();
-	AddPlayerbotsSecureLoginScripts();
+    AddPlayerbotsSecureLoginScripts();
 
     AddSC_playerbots_commandscript();
 }


src/PlayerbotsSecureLogin.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+        LOG_DEBUG("playerbots",
+            "Playerbots: Secure login: Account {} (ID: {}) is trying to log into '{}' which is already online. Disconnecting the character.",
+            username, acc, existing->GetName());
+
+        bool ok = ForceLogoutViaPlayerbotHolder(existing);
+        if (!ok)
+        {
+            if (existing->GetSession())
+                existing->GetSession()->KickPlayer();
```
privatecore, line 120: This will not work with bots ;)

BeardBear33: Yes, it does work. I’ve been using it on my server for a year now, and so far nobody has managed to log in in a way where the altbot would still remain in-game. It always kicks the altbot before the real player (client) actually attaches to that character.

I’d appreciate it if you could test it before claiming it doesn’t work. :)

src/PlayerbotsSecureLogin.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include "Player.h"
+#include "AccountMgr.h"
+
+#include "Playerbots.h"
+
+namespace
+{
+    static Player* FindOnlinePlayerByGuid(ObjectGuid guid)
+    {
+        auto const& players = HashMapHolder<Player>::GetContainer();
```
privatecore, line 14: Instead of iterating through all players, you can use `PlayerBotMap playerBots` to search, get, or check bot using its guid -- method `GetPlayerBot`.

src/PlayerbotsSecureLogin.cpp
```diff
@@ -0,0 +1,130 @@
...
+            return true;
+
+        PlayerbotAI* existingAI = GET_PLAYERBOT_AI(existing);
+        bool isAltbot = (existingAI && !existingAI->IsRealPlayer());
+        if (!isAltbot)
+            return true;
+
+        uint32 acc = session->GetAccountId();
+        std::string username = "unknown";
+        AccountMgr::GetName(acc, username);
```
privatecore, line 110: Execute DB query just to show the original username in the log?

src/PlayerbotsSecureLogin.cpp
BeardBear33: Look, if you have a better way to do it, nothing is stopping you from trying. I was simply asked to submit my fix as a PR instead of keeping it as a separate module that I’ve been using so far.

I did my part and implemented a working kick on the altbot to prevent a server crash from a duplicate login.

So if you have an idea to improve it — go for it. :) But as it is right now, it works reliably. Since opening the PR, I’ve done over 50 test logins to an altbot and every single one worked, so that’s where I’m coming from.

src/PlayerbotsSecureLogin.cpp
privatecore: `existing->GetSession()->KickPlayer();` -> will not work with bots; not ALL your changes, just this call is redundant

privatecore:
@BeardBear33 I don't really understand how I upset you. I'm also confused about what your message means -- do you want me to make PR for your changes, or don't you want to consider any improvements to your impl because you think it's "final"?

Dreathean:
Hey @BeardBear33, I don't see how anything Privatecore said could be interpreted as negative? They seemed like fairly standard reviews with offering suggestions for improvement? Every PR undergoes some degree of review, that is totally normal. I appreciate you opening this PR, but I simply don't understand what caused you to get offended.

Issue #1934, opened 2025-12-21
The null pointer issue in the PlayerbotAI::HasAuraToDispel function caused the server to crash.


gutoukuaiken:
**Bug Description**
Server crash

[gdb.txt](https://github.com/user-attachments/files/24276901/gdb.txt)

**Commit Hash**
Latest

**System**
OS: Ubuntu 24.04

Problem analysis:
1. Crashing location
On line 4022, an attempt was made to access an AuraApplication* object through the iterator itr and call its GetBase() method.
However, at this point, itr->second (that is, AuraApplication*) might be a null pointer (nullptr) or a released dangling pointer, resulting in accessing illegal memory → SIGSEGV.
2. Context clues
The call stack shows that this occurred when searching for the teammates to dispel (PartyMemberToDispel).
When dispelType = 2, it corresponds to DISPEL_MAGIC (based on the DispelType enumeration in WoW).
During the traversal of the visible auras on the target, an AuraApplication object was invalid.
3. Possible Causes
Multithreaded Competition: AzerothCore uses multithreaded MapUpdater for map updates, while playerbots access the Aura that is being removed or destructed in another thread.
Missing Null Pointer Check: The code assumes that itr->second is always valid, but it may actually have been removed (for example, the aura has just been dispelled, expired, or the unit has died).
Iterator Failure: When iterating over the container returned by target->GetVisibleAuras(), this container may be modified during the iteration (such as when other threads remove the aura).

Code modification suggestions:
```--- a/modules/mod-playerbots/src/PlayerbotAI.cpp
+++ b/modules/mod-playerbots/src/PlayerbotAI.cpp
@@ -4015,23 +4015,36 @@ bool PlayerbotAI::IsRanged(Unit* target)
 
 bool PlayerbotAI::HasAuraToDispel(Unit* target, uint32 dispelType)
 {
-    if (!target->IsInWorld())
+    if (!target || !target->IsInWorld())
     {
         return false;
     }
+
+    if (!bot || !bot->IsInWorld())
+        return false;
+
     bool isFriend = bot->IsFriendlyTo(target);
     Unit::VisibleAuraMap const* visibleAuras = target->GetVisibleAuras();
+    if (!visibleAuras)
+        return false;
+
     for (Unit::VisibleAuraMap::const_iterator itr = visibleAuras->begin(); itr != visibleAuras->end(); ++itr)
     {
-        Aura* aura = itr->second->GetBase();
[...]

PR #1951, opened 2025-12-25
[CRASH FIX] Various crash fixes in regard to defensive checks around unit/targets and finer grained checks on tele ack

hermensbas:
Some basic defense checks on mainly unit/targets used in public functions, and some minor tweaks with teleport ack.
https://github.com/mod-playerbots/mod-playerbots/issues/1934
https://github.com/mod-playerbots/mod-playerbots/issues/1957

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -2861,7 +2861,7 @@ bool PlayerbotAI::TellMaster(std::string const text, PlayerbotSecurityLevel secu
 
 bool IsRealAura(Player* bot, AuraEffect const* aurEff, Unit const* unit)
 {
-    if (!unit && !unit->IsInWorld() && unit->IsDuringRemoveFromWorld())
+    if (!unit || !unit->IsInWorld() || unit->IsDuringRemoveFromWorld())
         return false;
 
     if (!aurEff)
@@ -3864,7 +3864,7 @@ bool PlayerbotAI::CastVehicleSpell(uint32 spellId, Unit* target)
     if (!spellId)
         return false;
 
-     if (!IsValidUnit(target))
+    if (!IsValidUnit(target))
         return false;
 
     Vehicle* vehicle = bot->GetVehicle();


hermensbas:
The added "aura->IsRemoved()" suggest there might be a race condition, if so its prolly best to make a snapshot of the VisibleAuraMap before iterating over it. The IsRemoved() is a really cheap check adding copy not so much, lets only add that if we find proof of a race condition exists.

Reporter1337:
Hello why is this not pushed yet? Can I add this or does it needs to be tested first?

Dreathean:
It's been less than 24 hours. As with any PR it requires review from another maintainer and testing is preferable. You are certainly free to add it to test it yourself.

hermensbas:
and holiday/xmass times my friend :)

hermensbas:
There is another closely related bug which i am gonna include aswell. Hence draft

hermensbas:
Alright this should do it, in essence we still have design issues regarding threading, packets handlings, out of sync manager mappings with player objects. But lets get stabile and work for there. Meanwhile survive the refactors of the core. Still have logout on my list, but first the reported crash logs and monitoring the fixes when merged.

movement is still issue, but we getting closer and closer in our understanding. Havent used the valid checks everywhere, lets try to add those based on crash logs much as possible.

hermensbas:
rebased with master

src/PlayerbotAI.cpp
```diff
@@ -2822,13 +2861,18 @@ bool PlayerbotAI::TellMaster(std::string const text, PlayerbotSecurityLevel secu
 
 bool IsRealAura(Player* bot, AuraEffect const* aurEff, Unit const* unit)
 {
+    if (!unit && !unit->IsInWorld() && unit->IsDuringRemoveFromWorld())
```
kadeshar: Should be || not &&

src/PlayerbotAI.cpp
```diff
@@ -3804,6 +3864,9 @@ bool PlayerbotAI::CastVehicleSpell(uint32 spellId, Unit* target)
     if (!spellId)
         return false;
 
+     if (!IsValidUnit(target))
```
kadeshar: Extra space before if

src/PlayerbotAI.cpp
hermensbas: aaah lol dumbass mistake

noisiver (approved):

PR #1952, opened 2025-12-25
Exclude additional test enchants from bots

brighton-chi:
Added two test enchants to exclude from the list of enchants that bots can apply: 19927 and 44119. I also reordered the list to be in alphanumeric order.

The important exclusion to add is 44119: https://www.wowhead.com/wotlk/spell=44119/enchant-bracer-template

It is a test enchant that gives 34 AP on bracers; this is stronger than any other bracer enchant for level 70 physical attackers so bots are applying it.

### DIFF
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -121,8 +121,11 @@ void PlayerbotFactory::Init()
     uint32 maxStoreSize = sSpellMgr->GetSpellInfoStoreSize();
     for (uint32 id = 1; id < maxStoreSize; ++id)
     {
-        if (id == 7218 || id == 19927 || id == 35791 || id == 44119 || id == 47147 || id == 47181 ||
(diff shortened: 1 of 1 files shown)


brighton-chi:
Added 35791. Took me forever to find this one. It's 43 spellpower, but the spell is for 81 healing (https://www.wowhead.com/wotlk/spell=35791/qaenchant-weapon-81-healing). When healing and spell damage were merged in 3.0.2, weapons that had the healing enchant at the time were converted to have 43 spellpower, but the enchant was changed going forward to give 40 spellpower (same as the old spell damage enchant). So the 43 spellpower is a grandfathered enchant from pre-3.0.2 and clearly should not be available for bots to apply to weapons (it is however the standard caster weapon enchant being applied at 70 currently).

Added 39405 also. Same deal as above, grandfathered 30 healing bracer enchant that was converted to 16 spellpower (going forward, the equivalent spellpower bracer enchant was 15).

noisiver (approved):

PR #1955, opened 2025-12-28
Align the playerbots database structure with the azeroth core database structure 

Wishmaster117:
This PR Align the Playerbots database structure with the AzerothCore database structure.
```
\MODULES\MOD-PLAYERBOTS\DATA
\---sql
    +---archive
    |   \---db_playerbots
    +---base
    |   \---db_playerbots
    +---characters
    |   +---base
    |   \---updates
    +---create
    |   \---db_playerbots
    +---custom
    |   \---db_playerbots
    +---old
    |   \---db_playerbots
    +---updates
    |   \---db_playerbots
    \---world
        +---base
        \---updates
```

**\azerothcore-wotlk\src\server\database\Updater\DBUpdater.cpp Need to be updated core side.**


Wishmaster117:
After further research, we’re already core‑like for both world and characters. Here is an excerpt from the AzerothCore code.
```
        // data/sql
        for (auto const& itr : moduleList)
        {
            std::string path = _sourceDirectory->generic_string() + "/modules/" + itr + "/data/sql/" + _dbModuleName; // modules/mod-name/sql/world

            Path const p(path);
            if (!is_directory(p))
                continue;
```

NoxMax:
notwithstanding a review by someone more intimately knowledgeable of the DB, make sure formatting follows AC's standard https://github.com/azerothcore/wiki/blob/master/docs/cpp-code-standards.md

privatecore:
One of the problems with the current `mod-playerbots` impl is the external database, which is not handled properly with `dbimport`. Because of this, every SQL file inside the data folder is treated as an update. To properly recreate the database structure after a certain stage, the entire create table process should be called.

Below is one of my stashes that forces `dbimport` to work with the `playerbots` external database in the same way as others, so all databases are managed the same way.

Im not sure if I finished these changes, but you can give it a try yourself ;)
[test-stash-5.patch](https://github.com/user-attachments/files/24405248/test-stash-5.patch)

Wishmaster117:
> One of the problems with the current `mod-playerbots` impl is the external database, which is not handled properly with `dbimport`. Because of this, every SQL file inside the data folder is treated as an update. To properly recreate the database structure after a certain stage, the entire create table process should be called.
> 
> Below is one of my stashes that forces `dbimport` to work with the `playerbots` external database in the same way as others, so all databases are managed the same way.
> 
> Im not sure if I finished these changes, but you can give it a try yourself ;) [test-stash-5.patch](https://github.com/user-attachments/files/24405248/test-stash-5.patch)

Okay, I’m actually far from what was expected. I’m switching it to draft and I’ll take another look at it later.
Thanks for the stash!!

Wishmaster117:
Made a new clean and complete one

Issue #1905, opened 2025-12-07
Warriors Repeatedly Spam Battle Shout


brighton-chi:
**Bug Description**
Since today's Core merge, Warriors of all specs spam Battle Shout when in a raid with a Paladin that has applied Blessing of Might. I suspect this is because Battle Shout will not override Blessing of Might, assuming Blessing of Might is stronger or equally strong (which would be the case if BoM is talented, or not talented but BS also not talented).

**Commit Hash**
Playerbots: 52c3e966416ffe881acd5ff24a375debbbeddde9
However, I think the AC hash is the relevant one: 8f84271af88785badaaafc9d5e58494a3e94fea5

**How To Reproduce Bug**
Start a raid. Include a Paladin that uses improved BoM and a Warrior. Have the Paladin apply BoM. Enter combat and observe the Warrior repeatedly attempt to Battle Shout.

**Expected Behavior**
I think this is not so much a bug in the core but an issue that needs to be addressed by the module. My suspicion is that the following commit to the core has caused this, as previously, Battle Shout would be applied even if BoM was active (it would not override BoM, they would both be applied, though I'm not sure if functionally their effects stacked):

https://github.com/azerothcore/azerothcore-wotlk/commit/2f7f9bd72f188298d8805e1faaa4019c1e03016c

So I think probably there is a need to modify Warrior strategies so that they do not attempt to Battle Shout repeatedly if a player already has Blessing of Might.

@noisiver Has also experienced this issue.

brighton-chi:
This is actually brutal right now and way past warriors. Shamans are pretty much useless too because they keep trying to drop totems over and over when their buffs don't apply. I suspect this impacts any sort of non-stacking buff situation.

It is bad enough that I think it is worth considering reverting the merge until this is fixed because group play is totally busted.

Wishmaster117:
I don't think we should revert the core update; the issue is that the coexistence of BoM and BS is simply not handled right now. We just need to patch the Paladin buff strategies and the Warrior buff logic so this condition is managed.  
It would be easier to do this once the entire Paladin buff side is merged, because the following PR is still missing: https://github.com/mod-playerbots/mod-playerbots/pull/1785

noisiver:
Reverting the master merge is not a solution as mentioned. It's simply easier to address the issue in the warrior AI

brighton-chi:
[...]

PR #1956, opened 2025-12-29
Refactor Paladin blessing coordination to reduce redundant group scans

Wishmaster117:
This PR refactors the Paladin blessing system to reduce redundant group scans and centralize decision logic without changing gameplay behavior.
The coordination logic (bstats/bmana/bdps) now uses a cached state value, and blessing resolution is moved to dedicated helpers. 

**Why this is better / less resource‑intensive**

- Fewer repeated group scans: the previous logic iterated the group multiple times per action and per tick. The new PaladinBlessingState caches the paladin count and role election with a short interval, avoiding repeated loops.
- Centralized blessing decisions: Might/Wisdom/Sanctuary resolution is now in one place, which reduces duplicated checks and avoids diverging logic paths.
- Deterministic role coordination: role election is computed once and reused, preventing multiple paladins from overwriting each other’s buffs while keeping the original behavior.

**Behavior parity**
Same triggers and priorities (Sanctuary > Kings in bstats).
Same buff eligibility rules and targeting logic.

**Notes**
This will also solve issues:
https://github.com/mod-playerbots/mod-playerbots/issues/1905
https://github.com/mod-playerbots/mod-playerbots/issues/1677


brighton-chi:
So this may be addressed by this PR but just wanted to offer an observation if not, which is I've noticed that after Paladins have stopped buffing, if I relog the group (with Multibot), they start buffing again.

Also, another observation--I run a setup with two holy, one prot, and one ret. So by default, two have strategy set to wisdom, one to might, and one to sanctuary. I have seen that sometimes (uncommon and I am not sure about a specific trend), nobody will buff kings. But most of the time, they do, even without any of them set to buff kings.

Wishmaster117:
> So this may be addressed by this PR but just wanted to offer an observation if not, which is I've noticed that after Paladins have stopped buffing, if I relog the group (with Multibot), they start buffing again.
> 
> Also, another observation--I run a setup with two holy, one prot, and one ret. So by default, two have strategy set to wisdom, one to might, and one to sanctuary. I have seen that sometimes (uncommon and I am not sure about a specific trend), nobody will buff kings. But most of the time, they do, even without any of them set to buff kings.

This may have been caused by AI collisions; try with this refactor to see if it still happens.

hermensbas (dismissed):

Wishmaster117:
With the latest pull:
Amortized state calculation: the blessing state (roles + counters) is cached every 2 seconds, which limits group loops per tick and keeps the server load light.
Simple and targeted loops: retargeting only iterates through the group when necessary, with no global scans or large‑scale resets.

Wishmaster117:
Done! Ready for final review!

hermensbas:
Alrighty i will make some time this week.

brighton-chi:
I have some consistent issues with the Paladin greater buffs strategies; does this PR address any of these matters? These are the results from running with 4 Paladins in a 25-player raid. Switching their buff strategies does not seem to make any difference.

- Enhancement Shaman receives Kings and Wisdom, but no Might
- Hunter receives Kings and Might, but no Wisdom
- Paladin receives Kings and Wisdom, but sometimes no Might
- Druid receives Kings and Wisdom, but sometimes no Might
- Sometimes, caster classes receive Might and Kings, but no Wisdom. This seems most common with Mages, but maybe that's just where I notice it most. This doesn't always happen, though.
- As noted, the Paladins appear not to care about what buff strategies they have. For example, if I have 2 Holy Paladins, 1 Retribution Paladin, and 1 Protection Paladin, and I set the Retribution Paladin to buff Might, the Protection Paladin to buff Kings, and the Holy Paladins to buff Wisdom:
   - I (playing as a Rogue) will usually receive Might from a Holy Paladin and not the Retribution Paladin. This is not good since the Retribution Paladin has Improved Blessing of Might so I would think they should be prioritized for casting that blessing.
   - My Protection Paladin will usually buff nobody except itself with Sanctuary, despite being the only bot with Kings set as a strategy

Wishmaster117:
Yes this PR  solve this issues but conflict need to be solved.

PR #1961, opened 2026-01-01
Removed strategies for Naxxramas until they can be rewritten without core changes

noisiver:
Merging this isn't required but rather to provide an option to core changes.

The scripts can be written from scratch when someone feels like doing it without touching the core.

This shouldn't be merged unless https://github.com/mod-playerbots/azerothcore-wotlk/pull/130 is merged too.


Wishmaster117:
Great idea, the fewer changes we make to AzerothCore, the easier long‑term maintenance becomes.
Excellent work so far, is there any reason not to merge this?
I’m not sure I have the skills required to write full strategies, but a few ideas come to mind to avoid relying on headers:
– Strategies based on observable signals (spells, auras, UnitState)
– An entry/spell‑ID driven approach using stable DBC/DB data
– Using InstanceScript + BossState without AI headers
I’m not sure how feasible these ideas are, but it should definitely be possible to build a script using these mechanics.

kadeshar (approved):

noisiver:
> Great idea, the fewer changes we make to AzerothCore, the easier long‑term maintenance becomes. Excellent work so far, is there any reason not to merge this? I’m not sure I have the skills required to write full strategies, but a few ideas come to mind to avoid relying on headers: – Strategies based on observable signals (spells, auras, UnitState) – An entry/spell‑ID driven approach using stable DBC/DB data – Using InstanceScript + BossState without AI headers I’m not sure how feasible these ideas are, but it should definitely be possible to build a script using these mechanics.

The only reason not to merge it would've been if people thought the sacrifice wasn't worth the gain. But just as you say, long-term maintainability and less core changes is a big benefit.

We've had discussions about these strategies in the past and no one can tell why Naxxramas would need core headers to function, and they could be done without them.

Better to get it over with instead of delaying it further, adding more effort to maintaining core merges because of Naxxramas now that CC moved into WotLK.

PR #1966, opened 2026-01-03
Add a Combat Delay so bots and their pets dont run in and pull everything.

jads147:
Add a configurable Combat Delay so Playerbots only attack after a certain amount of time.

Video of change in Action:
https://youtu.be/5rylZkHyKVc

### DIFF
--- a/src/Config/PlayerbotAIConfig.cpp
+++ b/src/Config/PlayerbotAIConfig.cpp
@@ -82,6 +82,7 @@ bool PlayerbotAIConfig::Initialize()
     sitDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.SitDelay", 20000);
     returnDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.ReturnDelay", 2000);
     lootDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.LootDelay", 1000);
+    combatDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.CombatDelay", 0);
     minBotsForGreaterBuff = sConfigMgr->GetOption<int32>("AiPlayerbot.MinBotsForGreaterBuff", 3);
     rpWarningCooldown     = sConfigMgr->GetOption<int32>("AiPlayerbot.RPWarningCooldown", 30);
     disabledWithoutRealPlayerLoginDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.DisabledWithoutRealPlayerLoginDelay", 30);
--- a/src/Config/PlayerbotAIConfig.cpp
+++ b/src/Config/PlayerbotAIConfig.cpp
@@ -82,7 +82,7 @@ bool PlayerbotAIConfig::Initialize()
     sitDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.SitDelay", 20000);
     returnDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.ReturnDelay", 2000);
     lootDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.LootDelay", 1000);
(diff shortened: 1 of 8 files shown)


brighton-chi:
This is cool, although I think the best way to approach this issue is to make this a strategy instead of a config option so it can be toggled on/off in game and also applied only to specific bots.

isitsimon:
Cmangos has a 
"Wait attack time" strategy which is configured in game to delay by a specified number of seconds.
Could probably implement something similar?

src/Config/PlayerbotAIConfig.cpp
```diff
@@ -82,6 +82,7 @@ bool PlayerbotAIConfig::Initialize()
     sitDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.SitDelay", 20000);
     returnDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.ReturnDelay", 2000);
     lootDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.LootDelay", 1000);
+    combatDelay = sConfigMgr->GetOption<int32>("AiPlayerbot.CombatDelay", 0);
```
kadeshar: Default value in config file is 3000

src/strategy/actions/AttackAction.cpp
```diff
@@ -59,6 +62,31 @@ bool AttackAction::Attack(Unit* target, bool /*with_pet*/ /*true*/)
...
+        if (targetFirstSeenTime.find(targetGuid) == targetFirstSeenTime.end())
+        {
+            targetFirstSeenTime[targetGuid] = currentTime;
+            return false; 
+        }
+        uint32 timeSinceFirstSeen = currentTime - targetFirstSeenTime[targetGuid];
+        if (timeSinceFirstSeen < sPlayerbotAIConfig->combatDelay)
+        {
+            return false; 
+        }
```
kadeshar: without {}

src/Core/Ai/Generic/Actions/AttackAction.cpp
```diff
@@ -59,6 +62,31 @@ bool AttackAction::Attack(Unit* target, bool /*with_pet*/ /*true*/)
...
+        if (timeSinceFirstSeen < sPlayerbotAIConfig->combatDelay)
+        {
+            return false; 
+        }
+
+        // Clean up old entries to prevent memory leak
+        if (targetFirstSeenTime.size() > 100)
+        {
+            targetFirstSeenTime.clear();
+        }
```
kadeshar, line 91: Without {}

This cleanup should check time instead of size?
If uint32 time + combatDelay  + some threshold > current time then delete from list

kadeshar:
obraz

src/Config/PlayerbotAIConfig.cpp
kadeshar: Still not fixed

src/Core/Ai/Generic/Actions/AttackAction.cpp
```diff
@@ -62,28 +62,28 @@ bool AttackAction::Attack(Unit* target, bool /*with_pet*/ /*true*/)
...
-        if (targetFirstSeenTime.size() > 100)
+        // Remove entries older than combatDelay + threshold
+        for (auto it = targetFirstSeenTime.begin(); it != targetFirstSeenTime.end();)
         {
-            targetFirstSeenTime.clear();
+            if (currentTime - it->second > sPlayerbotAIConfig->combatDelay + 10000)
+                it = targetFirstSeenTime.erase(it);
+            else
+                ++it;
         }
```
kadeshar, line 91: That cleanup loop is made in every attack which can be performance issue now. I think better approach is check that in some interval like once a minute or something

kadeshar:
Could you solve conflicts?

PR #1968, opened 2026-01-03
Relaxed queue/melee follow: Move if further than x meter but less then y meter away from following target.

jads147:
Works for queue/melee Formation, makes queue much more nice and (most times) actually like a queue. 

Demonstration with
follow distance: 6
relaxed follow distance: 3

https://youtu.be/JDjhsL8_k9c


Wishmaster117:
Hi,  if merged add Multibot need update tag please.

kadeshar:
Could you resolve conflicts?

src/Ai/Base/Actions/FollowActions.cpp
```diff
@@ -23,7 +23,38 @@ bool FollowAction::Execute(Event event)
     bool moved = false;
     if (!target.empty())
     {
-        moved = Follow(AI_VALUE(Unit*, target));
+        // Relaxed follow: stay in radius, don't adjust angle/formation
+        if (botAI->HasStrategy("relaxed follow", BOT_STATE_NON_COMBAT))
```
kadeshar, line 27: For movement in group behavior we using formation not strategy

Issue #1964, opened 2026-01-02
Altbots Not Moving Vehicles in Ulduar 25m


Xyletic:
**Bug Description**
Altbots will go and find a seat within the vehicle units in Ulduar 25 man, but then do not move. I have attempted to remove and add follow from NC, logging out the bots and back in, re-entering the raid, reforming the raid, logging in bots one at a time, etc. Once they enter their vehicle there is no trigger I can find that will get them to follow me. I have also waited for an extended period of time and gone further into Ulduar to try and see if there are any triggers to get them to follow me.

I have not tried on 10 man, but would like to know if there is something I am potentially missing here.

**Commit Hash**
b16789f

**How To Reproduce Bug**
Enter the raid, talk to Bran and enter one of the vehicles.

**Expected Behavior**
I've seen a YT video where the bots follow the player's siege device. So I would expect that behavior so I can clear the first boss.

**Modules**
AH Bot, Transmog, Ollama (disabled), Playerbots

**System**
OS: Ubuntu Server 24.04

gutoukuaiken:
This issue is currently being fixed.
#1902

gutoukuaiken:
@privatecore After updating the new code, this problem occurred again.
azcore-tree commit:  8f55eb0ca9bf7be17d221db8f4a6ddd2335651d4
mod-Playerbots commit:  c9cc4324d369475562f7fbe5bcf0fb591bab5860

privatecore:
> After updating the new code, this problem occurred again.

PR is not merged yet, so you need to apply the changes manually.

Xyletic:
Confirmed this is fixed in the latest merge.

PR #1969, opened 2026-01-04
[BUG FIX] the issue where bots on vehicles cant move

privatecore:
Fix the [issue #1964](https://github.com/mod-playerbots/mod-playerbots/issues/1964) where bots on vehicles cant move. Transfer the allowed movement logic to `PlayerbotAI::CanMove` -- Ulduar raid strategy also used this method for the Malady of the Mind trigger.

The only remaining `IsRooted` check is in `MovementAction::UpdateMovementState`. I'm not sure if this also needs to be updated for the vehicles case, but on paper, everything should work as intended. A more complex solution to fix this issue can be found in the [comment](https://github.com/mod-playerbots/mod-playerbots/issues/1902#issuecomment-3703795779), but for now, these changes are safer.

### DIFF
--- a/src/PlayerbotAI.cpp
+++ b/src/PlayerbotAI.cpp
@@ -5853,16 +5853,38 @@ int32 PlayerbotAI::GetNearGroupMemberCount(float dis)
 
 bool PlayerbotAI::CanMove()
 {
-    // do not allow if not vehicle driver
-    if (IsInVehicle() && !IsInVehicle(true))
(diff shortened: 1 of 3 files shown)


privatecore:
@gutoukuaiken @Xyletic guys, need some testing over here ;)

Xyletic:
> @gutoukuaiken @Xyletic guys, need some testing over here ;)

I am out of town for a few days but will happily test when I get back! I should be back Thursday.

hermensbas:
Works on my machine ;)

PR #1973, opened 2026-01-05
Fix deprecated vsprintf usage in Engine logging

notOrrytrout:
Replace deprecated vsprintf with vsnprintf to eliminate compiler warning and prevent potential buffer overflow.


notOrrytrout:
Tested in game and it seemed to log actions just fine.
 I just basically added a buffer size by using the current vsnprintf lib instead.

notOrrytrout:
This is no longer needed on the latest commit.

PR #1991, opened 2026-01-08
[BUG FIX] Movement flag update called for bots while rooted

hermensbas:
Fix for the error "Attempted sending heartbeat with root flag for guid"

The core does not allow movement flag updates when unit/player has the MOVEMENTFLAG_ROOT flag. Change scope bots alone.


hermensbas:
Ive tested ths together with https://github.com/mod-playerbots/azerothcore-wotlk/pull/138

Havents single a single error, well obviously since the code checks but still. I havent touched the raid scripts code in regard of movment update flags, shouldnt be needed.

Celandriel (approved):

hermensbas:
cheats enabled

PR #1992, opened 2026-01-08
Fix fishing AI position matching precision

OlegKungurov:
**Problem**: FishingAction::isUseful() used strict `pos == bot->GetPosition()` comparison causing "Rog A:go fishing - USELESS" when bot was within 0.5-1m of fishing spot (normal positioning tolerance).

**Solution**: Replace exact coordinate match with 1.5m tolerance using `fabs(posX - botX) < 1.5f` for all axes. Added `#include "Config.h"` for AI_VALUE stability.

**Result**: Master fishing now reliably triggers and completes cycle:
- move near water → go fishing → use bobber ✓

Closes #fishing-useless-bug

### DIFF
--- a/src/strategy/actions/FishingAction.cpp
+++ b/src/strategy/actions/FishingAction.cpp
@@ -18,7 +18,6 @@
 #include "PlayerbotTextMgr.h"
 #include "Playerbots.h"
 #include "Position.h"
-#include "Config.h"
 
 uint32 const FISHING_SPELL = 7620;
 uint32 const FISHING_POLE = 6256;
@@ -262,7 +261,9 @@ bool MoveNearWaterAction::isUseful()
         return false;
     FishingSpotValue* fishingSpotValueObject = (FishingSpotValue*)context->GetValue<WorldPosition>("fishing spot");
     WorldPosition pos = fishingSpotValueObject->Get();
-    return !pos.IsValid() || fishingSpotValueObject->IsStale(FISHING_LOCATION_TIMEOUT) || pos != bot->GetPosition();
+    return !pos.IsValid() || fishingSpotValueObject->IsStale(FISHING_LOCATION_TIMEOUT) ||
+           bot->GetExactDist(&pos) < 0.1f;
+
 }
 
 bool MoveNearWaterAction::isPossible()
@@ -299,6 +300,17 @@ bool MoveNearWaterAction::isPossible()
                 return true;
             }
         }
+    }
+    // Can the bot fish from current position?
+    WorldPosition waterAtCurrentPos =
+        FindWaterRadial(bot, bot->GetPositionX(), bot->GetPositionY(), bot->GetPositionZ(), bot->GetMap(),
+                        bot->GetPhaseMask(), MIN_DISTANCE_TO_WATER, MAX_DISTANCE_TO_WATER, SEARCH_INCREMENT, true);
+    if (waterAtCurrentPos.IsValid())
+    {
+        SET_AI_VALUE(WorldPosition, "fishing spot",
+                     WorldPosition(WorldPosition(bot->GetMapId(), bot->GetPositionX(), bot->GetPositionY(),
+                                                 bot->GetPositionZ())));
+        return false;
     }
     // Lets find some water where we can fish.
     WorldPosition water = FindWaterRadial(
@@ -442,13 +454,7 @@ bool FishingAction::isUseful()
     if (!pos.IsValid() || fishingSpotValueObject->IsStale(FISHING_LOCATION_TIMEOUT))
         return false;
 
-    float posX, posY, posZ;
-    pos.GetPosition(posX, posY, posZ);
-    float botX, botY, botZ;
-    bot->GetPosition(botX, botY, botZ);
-
-    float dist2d = sqrt(powf(posX - botX, 2) + powf(posY - botY, 2));
-    return dist2d < 2.0f;
+    return bot->GetExactDist(&pos) < 0.1f;
 }
 
 bool UseBobberAction::isUseful()


src/strategy/actions/FishingAction.cpp
Celandriel, line 315: My preferred fix to this issue is actually setting a new position similar to https://github.com/OlegKungurov/mod-playerbots/blob/25307564b72006b6d4d82c7e37c3359f2cf0f0fb/src/strategy/actions/FishingAction.cpp#L280C1-L290C10
 where if the bot current position can fish from their current location, it sets a new value confirming it has LOS to it. There are many places where being off by 0.5 can cause significant LOS issues preventing bots from doing this successfully.

OlegKungurov: Bots approach the water's edge, but they don't even equip a fishing rod.
I'll go collect logs to describe the problem more accurately.
image

Celandriel:
Are they trying to move to the target position? Can you provide coordinates so I can verify?

OlegKungurov:
> Are they trying to move to the target position? Can you provide coordinates so I can verify?

No, they don't move. They just stand there.
`.go xyz -9691.926 836.73065 21.656132 0`
image

src/strategy/actions/FishingAction.cpp
OlegKungurov: I collected logs with the old and new code.
old code - https://github.com/mod-playerbots/mod-playerbots/pull/1992/commits/aae200b6b4499f17cb3253afa317de0cd1b9df4a
```
Рог PUSH:go fishing - 10.000000 (trigger)
[%s] Fishing spot dist: %.2fm (need 0.0m)
[%s] Fishing USELESS: pos!=bot (dist=%.2fm)
Рог A:go fishing - USELESS
```
new code - https://github.com/mod-playerbots/mod-playerbots/pull/1992/commits/9a95a98ca78c2147f5dd0d97eed8e12e24d5fa70
```
Рог PUSH:go fishing - 10.000000 (trigger)
[%s] FishingAction::isUseful() START
  Can fish: %s
  Spot valid: %s, stale: %s
  STRICT MATCH (pos == bot): %s
  2m TOLERANCE (dist=%.2fm): %s
  Bot: (%.1f,%.1f) Spot: (%.1f,%.1f)
[%s] FishingAction USEFUL! (2m fix)
[%s] FishingAction END: %s
Рог A:go fishing - PREREQ
Рог S:+use bobber
Рог A:go fishing - OK
```

brighton-chi:
> > Are they trying to move to the target position? Can you provide coordinates so I can verify?
> 
> 
> 
> No, they don't move. They just stand there.
> 
> `.go xyz -9691.926 836.73065 21.656132 0`
> 
> image
> 
> 

Is this independent then of any issue where bots move around looking for a position to fish and never are able to find the correct position?

OlegKungurov:
> > > Are they trying to move to the target position? Can you provide coordinates so I can verify?
> > 
> > 
> > No, they don't move. They just stand there.
> > `.go xyz -9691.926 836.73065 21.656132 0`
> > image
> 
> Is this independent then of any issue where bots move around looking for a position to fish and never are able to find the correct position?

My еnglish is bad. Maybe I misunderstood you.
But the bots just approach the water (perhaps trying to find a fishing spot) and don't do anything.
From the logs, I see that the condition that the bot is exactly at the fishing spot simply isn't met - https://github.com/mod-playerbots/mod-playerbots/pull/1992#discussion_r2673525435

hermensbas:
Sounds like a improvement to me

hermensbas:
Can someone involved with fishing feature review/test. Otherwise we better off removing the entire fearure of it aint working?

Celandriel:
I'll take a look Monday when I'm at a pc. Either way some changes need to be made to bring this in line.

src/strategy/actions/FishingAction.cpp
```diff
@@ -434,10 +435,20 @@ bool FishingAction::isUseful()
...
         return false;
+
     FishingSpotValue* fishingSpotValueObject = (FishingSpotValue*)context->GetValue<WorldPosition>("fishing spot");
     WorldPosition pos = fishingSpotValueObject->Get();
 
-    return pos.IsValid() && !fishingSpotValueObject->IsStale(FISHING_LOCATION_TIMEOUT) && pos == bot->GetPosition();
+    if (!pos.IsValid() || fishingSpotValueObject->IsStale(FISHING_LOCATION_TIMEOUT))
+        return false;
+
+    float posX, posY, posZ;
```
Celandriel: Replace all below with 
```
    return bot->GetExactDist(&pos) < 0.1f;
```
Only a small tolerance is needed. Anything greater needs a new position.

src/strategy/actions/FishingAction.cpp
Celandriel: Change to look like this
```
    // Can the bot fish from current position?
    WorldPosition waterAtCurrentPos = FindWaterRadial(
        bot, bot->GetPositionX(), bot->GetPositionY(), bot->GetPositionZ(),
        bot->GetMap(), bot->GetPhaseMask(),
        MIN_DISTANCE_TO_WATER,
        MAX_DISTANCE_TO_WATER,
        SEARCH_INCREMENT, true);\
    if (waterAtCurrentPos.IsValid())
    {
        SET_AI_VALUE(WorldPosition, "fishing spot", WorldPosition(WorldPosition(bot->GetMapId(), bot->GetPositionX(), bot->GetPositionY(), bot->GetPositionZ())));
        return false;
    }
    // Lets find some water where we can fish.
```
We will set a new position if current doesnt match.

Celandriel: replace 
```
!pos.IsValid() || fishingSpotValueObject->IsStale(FISHING_LOCATION_TIMEOUT) || bot->GetExactDist(&pos) < 0.1f;
```

src/strategy/actions/FishingAction.cpp
```diff
@@ -18,6 +18,7 @@
 #include "PlayerbotTextMgr.h"
 #include "Playerbots.h"
 #include "Position.h"
+#include "Config.h"
```
Celandriel: Not needed

Celandriel:
The proposed changes are designed to correct only for only precision based errors (I actually think 0.1 is too generous). 
The reason to be strict is that in many places there can be signficant line of sight issues when being 1 or 2f off from position. If that is the case then the position needs to be reevaluated so that bots can be more successful.

The issue was likely introduced with map and movement changes because this worked previously in many places. The way I correct for positioning errors larger than 0.1f is to check in isPossible if the current position is good for fishing, if so, reset the position and do not move. 

I tested the above in the location you proposed, and in booty bay as well.

brighton-chi:
> The proposed changes are designed to correct only for only precision based errors (I actually think 0.1 is too generous). 
> 
> The reason to be strict is that in many places there can be signficant line of sight issues when being 1 or 2f off from position. If that is the case then the position needs to be reevaluated so that bots can be more successful.
> 
> 
> 
> The issue was likely introduced with map and movement changes because this worked previously in many places. The way I correct for positioning errors larger than 0.1f is to check in isPossible if the current position is good for fishing, if so, reset the position and do not move. 
> 
> 
> 
> I tested the above in the location you proposed, and in booty bay as well. 

Did you ever get a chance to check out Lurker's isle?

src/strategy/actions/FishingAction.cpp
OlegKungurov: Done

src/strategy/actions/FishingAction.cpp
OlegKungurov: Done

src/strategy/actions/FishingAction.cpp
OlegKungurov: Done

OlegKungurov:
@Celandriel
I made the changes you mentioned.
I tested it myself – not all bots start fishing right away; some take 2-3 minutes to find a position. But in the end, everything was fine.
image

Celandriel:
Lurkers island works, however I noted that they will not equip upgrades when commanded by chat, so we need a combat strategy to trigger an equip upgrades action.
This will have to be a different pr.

Celandriel:
> @Celandriel
> I made the changes you mentioned.
> I tested it myself – not all bots start fishing right away; some take 2-3 minutes to find a position. But in the end, everything was fine.
> image
> 

I hope this works! When fishing in large groups it can be difficult because the equip actions is triggered by many items received (fish was already excluded) so you can end up in a cycle where they have to equip the fishing pole repeatedly. There is another PR that looks to disable the equip upgrades action unless an equipable item is looted that needs testing (#1994). 

Also the action priority is not very high right now. It's supposed to be more casual. 

Thank you for investigating this!

Issue #1987, opened 2026-01-07
Significant adverse performance impact from EquipUpgradesAction()


brighton-chi:
This is based on a discussion with @Celandriel and @noisiver on Discord.

I have been experiencing extreme latency spikes since yesterday whenever a loot event is triggered. Pmon shows that it is due to "equip upgrades" sometimes resulting in a delay of as high as over 1s _per bot_. After discussion and looking at the code, the likely culprit seems to be one that may have existed all along but was masked due to what was required for it to be noticeably problematic.

What I mean is that EquipUpgradeAction() checks ItemUsageValue and from there, through IsItemUsefulForSkill(), ultimately ends up at RandomItemMgr::ContainsInternal(). ContainsInternal() checks, for any profession, all profession trainers, spells, and recipes in the DB (for me, that is almost 80,000 entries). What this appears to mean is that when equip upgrades is called for a bot (which happens upon every loot event, including trading an item to/from a bot), to determine if a piece of equipment is an upgrade, a bot checks every item in its inventory and, before checking if the item is equippable, checks if the item is useful for a profession (thus, for every profession that the bot has, it checks almost 80,000 DB entries _per item_). I confirmed this via logs to EquipAction.cpp, ItemUsageValue.cpp, and RandomItemMgr.cpp.

The reason the delay became noticeable for me yesterday appears to be due to the other issue I posted about today, re: maintenance giving bots professions. Since I have max professions at 11 per character in WorldServer, this means each altbot in my 25-player raid ended up with 14 professions yesterday when I called maintenance (before, only a handful had professions, and not more than 1 or 2 each). Thus, every bot passed every botAI->HasSkill() check, and equip upgrades was doing over a million checks for each item in each bot's inventory when a loot event was triggered. Every per-item per-profession check was taking 2-4 ms, meaning upwards of 50 ms per item, which added up quickly in this context.

As far as I can tell, this is not a new issue with equip upgrades. But probably people were not running around with a bunch of bots with 14 professions, so the additional delay was not as noticeable (though it certainly could be at scale, so maybe this is contributing to performance problems on big servers).

[...]

PR #1994, opened 2026-01-08
Refactor  of EquipActions

Celandriel:
#PR Description 

The root cause of issue #1987 was the AI Value item usage becoming a very expensive call when bots gained professions accidentally. 

My original approach was to eliminate it entirely, but after inputs and testing I decided to introduce a more focused Ai value "Item upgrade" that only checks equipment and ammo inheriting directly from item usage, so the logic is unified between them.

Upgrades are now only assessed when receiving an item that can be equipped. 

Additionally, I noticed that winning loot rolls did not trigger the upgrade action, so I added a new package handler for that. 

Performance needs to be re-evaluated, but I expect a reduction in calls and in the cost of each call. 

I tested with bots and selfbot in deadmines and ahadowfang keep.

### DIFF
--- a/src/Ai/Base/Value/ItemUsageValue.cpp
+++ b/src/Ai/Base/Value/ItemUsageValue.cpp
@@ -19,7 +19,9 @@
 
 ItemUsage ItemUsageValue::Calculate()
 {
-    GetItemIdFromQualifier();
+    ParsedItemUsage const parsed = GetItemIdFromQualifier();
+    uint32 itemId = parsed.itemId;
+    uint32 randomPropertyId = parsed.randomPropertyId;
     if (!itemId)
         return ITEM_USAGE_NONE;
 
@@ -461,16 +463,20 @@ ItemUsage ItemUsageValue::QueryItemUsageForAmmo(ItemTemplate const* proto)
     return ITEM_USAGE_NONE;
 }
 
-void ItemUsageValue::GetItemIdFromQualifier()
+ParsedItemUsage ItemUsageValue::GetItemIdFromQualifier()
 {
-    size_t pos = qualifier.find(",");
+    ParsedItemUsage parsed;
+
+    size_t const pos = qualifier.find(",");
     if (pos != std::string::npos)
     {
-        this->itemId = atoi(qualifier.substr(0, pos).c_str());
-        this->randomPropertyId = atoi(qualifier.substr(pos + 1).c_str());
+        parsed.itemId = atoi(qualifier.substr(0, pos).c_str());
+        parsed.randomPropertyId = atoi(qualifier.substr(pos + 1).c_str());
+        return parsed;
     }
     else
-        this->itemId = atoi(qualifier.c_str());
+        parsed.itemId = atoi(qualifier.c_str());
+    return parsed;
 }
 // Return smaltest bag size equipped
 uint32 ItemUsageValue::GetSmallestBagSize()
@@ -908,7 +914,9 @@ std::string const ItemUsageValue::GetConsumableType(ItemTemplate const* proto, b
 
 ItemUsage ItemUpgradeValue::Calculate()
 {
-    GetItemIdFromQualifier();
+    ParsedItemUsage parsed = GetItemIdFromQualifier();
+    uint32 itemId = parsed.itemId;
+    uint32 randomPropertyId = parsed.randomPropertyId;
     if (!itemId)
         return ITEM_USAGE_NONE;
 
--- a/src/Ai/Base/Value/ItemUsageValue.h
+++ b/src/Ai/Base/Value/ItemUsageValue.h
@@ -14,7 +14,11 @@ class Player;
 class PlayerbotAI;
 
 struct ItemTemplate;
-
+struct ParsedItemUsage
+{
+    uint32 itemId = 0;
+    int32 randomPropertyId = 0;
+};
 enum ItemUsage : uint32
 {
     ITEM_USAGE_NONE = 0,
@@ -45,10 +49,7 @@ public:
 protected:
     ItemUsage QueryItemUsageForEquip(ItemTemplate const* proto, int32 randomPropertyId = 0);
     ItemUsage QueryItemUsageForAmmo(ItemTemplate const* proto);
-    void GetItemIdFromQualifier();
-    uint32 itemId = 0;
-    uint32 randomPropertyId = 0;
-
+    ParsedItemUsage GetItemIdFromQualifier();
 
 private:
     uint32 GetSmallestBagSize();


brighton-chi:
I know we talked about some other changes yesterday. Should I wait for further updates before testing?

Celandriel:
It'll take me about an hour or so to push a new update.

Celandriel:
I think a good way to test this would be a dungeon run or similar. 
Im seeing a reasonable drop in Equip action triggers based on pmon,

hermensbas:
Could someone else test this pls, so i can focus on some other core issues?

Wishmaster117:
> I think a good way to test this would be a dungeon run or similar. Im seeing a reasonable drop in Equip action triggers based on pmon,

Hi sorry if i missed something in the logic, but have few questions,
You removed the ItemUsage-based filtering (equip/replace/bad_equip) and now equip any inventory item that isn’t INVTYPE_NON_EQUIP; how will you prevent downgrades or non-upgrades from being equipped now?
You no longer query ItemUsageValue::QueryItemUsageForEquip; how will you handle unique-equip rules that were previously enforced there (e.g., already-equipped uniques)?
With the new cheap selection, how will you preserve the class/level checks for armor and weapon usability that were done via CanEquipWeapon/CanEquipArmor?
How will you keep the equipUpgradeThreshold logic that compares item scores and avoids minor/negative upgrades?
The previous logic handled Titan Grip / dual-wield slot expansion, how will the new approach avoid equipping invalid 2H weapons or mis-handling offhand slots?
The old path handled broken-vs-nonbroken item replacement; how will you preserve that behavior after removing the ItemUsage evaluation?
You replaced the item push result filtering from "meat trade goods" to "non-equip inventory type" how will you avoid triggering on other non-upgrade items that were previously excluded by ItemUsage (e.g., vendor/quest items...)?

Celandriel:
So if you look at the original code, the information from item usage is not used in determining is an upgrade or not. It only determines if the equipitems function is called.

That's it.

All those things you are worried about is not changed.

The logic is now very simple. Did I receive something that I can equip? If so check and equip any upgrades.

notOrrytrout:
I took a group into the Stormwind Stockaid the team got a bunch of new items and in the chat, they would say that they are equipping them. When I did an inspection, they did not, however, equip them. I got them to equip the items once I gave them an item. The test was I gave my pally some low level shoes and it put the shoes and some pants it had on.

notOrrytrout:
> This removes ItemUsage-based filtering and it treats any InventoryType != NON_EQUIP item as a candidate. How are the existing ItemUsage semantics (role weighting, upgrade thresholds, weapon/slot rules) preserved under this change?
> 
> If the main cost here is how often ItemUsage gets recomputed, would it make sense to keep the existing ItemUsage semantics but cache or gate the evaluation so it only reruns when the inventory actually changes (e.g., on loot/equip events), rather than broadening the equip filter? If I miss something oops.

I just saw you said it is not used.🧐

 the information from item usage is not used in determining is an upgrade or not. It only determines if the equipitems function is called.

notOrrytrout:
Just to make sure I’m following correctly: is it guaranteed that EquipItems() will reject all non-upgrades even when the candidate set is expanded to “anything equipable”?

My concern isn’t that ItemUsage scores upgrades, but that it previously constrained the candidate set. 

If EquipItems() already enforces all role, threshold, and slot constraints internally, then the simplification is equivalent, but if any of those constraints live in ItemUsage, this would widen the comparison.

Celandriel:
I'll finish this today

Celandriel:
@notOrrytrout I cant reproduce the issue, from what I can tell the filter is correctly triggering on equipable items, and adding them to the list to then evaluate in the upgrades section.

notOrrytrout:
Interesting. Maybe a third party tester is needed or I could test again. If it works it works. 😁

Wishmaster117:
I’m sorry, but there are some things I can’t quite explain to myself. Could you clarify these points for me?

1.
Could you clarify how you ensure that item selection remains equivalent to the old ItemUsage logic (EQUIP / REPLACE / BAD_EQUIP)?
The current implementation uses ItemUsageValue::BuildItemUsageParam + AI_VALUE2("item usage", ...) to filter real upgrades.

2.
In the previous flow, the randomProperty was taken into account through BuildItemUsageParam.
Does the new selection based solely on InventoryType correctly handle these cases?

3.
The previous logic relied on item usage to avoid equipping irrelevant items.
Can you confirm that the new filter (InventoryType != INVTYPE_NON_EQUIP) does not broaden the selection beyond the expected upgrades?

4.
Some reports mention that bots no longer equip items correctly since the change.
Have you been able to validate this behavior on a sample of bots (various classes/levels)?
If so, do you have logs or comparison tests with the old item‑usage flow?

5.
Do you have measured data (CPU time, scan frequency, number of evaluations avoided) to quantify the gain?
The change removes the item‑usage filter, so performance improves but at the cost of different behavior.
Numbers would help justify this trade‑off.

6.
Without the item‑usage filter, potentially more items are passed to EquipItems.
Have you observed an increase in useless equipment attempts or irrelevant equipment messages?

Celandriel:
> Could you clarify how you ensure that item selection remains equivalent to the old ItemUsage logic (EQUIP / REPLACE / BAD_EQUIP)? The current implementation uses ItemUsageValue::BuildItemUsageParam + AI_VALUE2("item usage", ...) to filter real upgrades.

So I was a little over-enthusiastic in my reduction. I saw a lot of logic duplication of evaluating whether an item should be upgraded in EquipItems that I did away with it entirely. I created a more focused AI_Value call to restore original logic.  

> In the previous flow, the randomProperty was taken into account through BuildItemUsageParam. Does the new selection based solely on InventoryType correctly handle these cases?

It probably did not. 

> The previous logic relied on item usage to avoid equipping irrelevant items. Can you confirm that the new filter (InventoryType != INVTYPE_NON_EQUIP) does not broaden the selection beyond the expected upgrades?

No I cant confirm that. Everything was passed whether upgrade or not. 

> Some reports mention that bots no longer equip items correctly since the change. Have you been able to validate this behavior on a sample of bots (various classes/levels)? If so, do you have logs or comparison tests with the old item‑usage flow?

I have not investigated this, and wont as a part of this. 

> Do you have measured data (CPU time, scan frequency, number of evaluations avoided) to quantify the gain? The change removes the item‑usage filter, so performance improves but at the cost of different behavior. Numbers would help justify this trade‑off.


I have results from the internal pmon tool clearly quantifying both a reduction of calls, and a faster processing time, reducing the overall cost by about 50%, and the number of action calls by about 30%. 
 
> Without the item‑usage filter, potentially more items are passed to EquipItems. Have you observed an increase in useless equipment attempts or irrelevant equipment messages?

I didnt specifically see the messages of failing to equip items. 

Thanks for the assessment and having me dig into it more.  Cant always get it right, and its always good to learn.

Either way, now instead of trying to identify every possible item usage, now its focused on checking for upgrades. So we reduced the number of calls, so bots only trigger on receiving something that could be an upgrade. Then we do some pre-filtering, so we dont waste time trying to evaluate non-equippable items, and the new ItemUpgradeValue focuses solely on assessing if it is good to equip, bypassing any logic that assessed trade skills or other issues.

Ill have to re-evaluate the performance now.

Celandriel:
So I just noticed that since winning items in rolling is a different packet than receiving loot the item isnt getting triggered. Ill have to extent the packets that use this.

Wishmaster117:
> So I just noticed that since winning items in rolling is a different packet than receiving loot the item isnt getting triggered. Ill have to extent the packets that use this.

Hi can you consider take a look at this:

https://github.com/Wishmaster117/mod-playerbots/tree/Improve-EquipAction-performances

Performance impact with 1400 bots, 100% active:
Assumptions : ~60 items in bags, ~35 unique itemId+randomProperty pairs, ~8 item-push events per hour per bot.
- Item-push path: ~59 item-usage evaluations saved per event.
- Per bot per hour: ~8 × 59 = ~472 evaluations saved.
- 1400 bots: ~660,800 evaluations saved per hour.

CPU savings (order of magnitude):
- 50 µs per evaluation → ~33s CPU/hour
- 100 µs per evaluation → ~66s CPU/hour
- 200 µs per evaluation → ~132s CPU/hour

These gains scale with item duplication and item-push frequency.

Celandriel:
The issue is "Item Usage" it too generic a check, particularly in its trade-skills check. 
My updated pr focuses the analysis to only if the bot can equip it. 

And now I need to find the other packets so that winning rolls also triggers the event as the current code doesnt do that.

Celandriel:
Updated for loot roll winner.

Celandriel:
> https://github.com/Wishmaster117/mod-playerbots/tree/Improve-EquipAction-performances

Am I correct that the logic you are trying to implement is: If the bot received an item, then equip that item, if a bot observed the looting action, then iterate all items as normal?

Wishmaster117:
> Am I correct that the logic you are trying to implement is: If the bot received an item, then equip that item, if a bot observed the looting action, then iterate all items as normal?

Not exactly. The intent is:
If we can resolve the received item from SMSG_ITEM_PUSH_RESULT, we only evaluate that item using the existing ItemUsage logic (EQUIP / REPLACE / BAD_EQUIP) and equip it if it’s a real upgrade.
If the item cannot be resolved by slot, we fall back to the full inventory scan (same as before).
So it’s not “always equip the received item”; it’s “evaluate only the received item first, using the same item-usage logic, and only then equip if appropriate.” The full scan remains as a safe fallback to preserve behavior.

I see you now moves back toward a real upgrade filter (item upgrade) instead of the earlier InventoryType-only approach, which is a step in the right direction. The ammo handling refactor also looks cleaner.

In my point of view, the good points are:
Reintroduces a proper upgrade decision (closer to the original item usage logic).
Extracts ammo logic into a dedicated helper.

But i'm a litle Concerned by :
The scope is very large (config + EquipAction + ItemUsageValue + ValueContext + triggers). That increases review complexity and regression risk.
A new item upgrade value duplicates part of item usage; this risks divergence over time if one side is updated and the other isn’t.
Multiple functional changes are bundled together without clear validation, which makes it harder to accept upstream.

So this is more ambitious but also riskier. A smaller, localized change that keeps the existing item usage logic intact (and only optimizes redundant work) would be easier to review and safer to merge.

hermensbas:
Have you guys considered looking at: https://github.com/cmangos/playerbots/blob/master/playerbot/strategy/actions/EquipAction.cpp?

src/Ai/Base/Actions/EquipAction.cpp
```diff
@@ -328,7 +328,44 @@ void EquipAction::EquipItem(Item* item)
...
+        //Pre-filter non-equipable items
+        if (itemTemplate->InventoryType == INVTYPE_NON_EQUIP)
+            continue;
+
+        int32 randomProperty = item->GetItemRandomPropertyId();
+        uint32 itemId = item->GetTemplate()->ItemId;
+        std::string itemUsageParam;
+        if (randomProperty != 0)
+            itemUsageParam = std::to_string(itemId) + "," + std::to_string(randomProperty);
+
```
kadeshar: Enter to delete

conf/playerbots.conf.dist
kadeshar, line 1: Conflict to resolve

src/Ai/Base/Actions/EquipAction.cpp
```diff
@@ -328,7 +328,44 @@ void EquipAction::EquipItem(Item* item)
...
+        if (usage == ITEM_USAGE_EQUIP || usage == ITEM_USAGE_REPLACE || usage == ITEM_USAGE_BAD_EQUIP)
+            items.insert(itemId);
+    }
+    return items;
+}
+
+bool EquipUpgradesTriggeredAction::Execute(Event event)
 {
     if (!sPlayerbotAIConfig->autoEquipUpgradeLoot && !sRandomPlayerbotMgr->IsRandomBot(bot))
         return false;
```
kadeshar: This condition fit to change in playerbots.conf.dist?

yqsy2025:
This push will equip the bot with gear that is worse than the current.

Celandriel:
> This push will equip the bot with gear that is worse than the current.

Are you talking about the the first version where I bypassed item usage  or the current one? The logic for determining of its an upgrade is practically unchanged.

Celandriel:
> Have you guys considered looking at: https://github.com/cmangos/playerbots/blob/master/playerbot/strategy/actions/EquipAction.cpp?

I didnt find anything that I would want to implement, sorting by ilvl could be, but im actually not sure what it achieves. 

Ive also considered limiting it just to the received item instead of a full inventory. I think that poses a greater risk than what I have done of items in inventory not being equip.

limetw:
I just recompiled and tested this PR code using “gh pr checkout 1994”.

The latency has significantly decreased, and I completed a full 30-minute dungeon run.

There wasn't a single instance of lag throughout, and the “playerbot pmon” performance report is normal.

Though I don't understand C++ and can't follow your discussion. True, English isn't my native language either—I rely on DeepL for translations.

But I believe this PR offers more practical improvements than the official project updates.

I've learned about “EquipUpgrades” and think updating it could be beneficial.

However, most “playerbot” users run AC on single-player or very small-scale VPS environments.

If everything could be solved with “high-performance” CPUs, “high-performance” IO SSDs, and “high-performance” RAM, it wouldn't be a good project. If every user needed a “high-performance” VPS—say, 12 cores and 64GB RAM—to run AC, why wouldn't they just host servers with real players?

Thanks to the author @Celandriel of this PR for making my bots run smoothly.

As regular players, this is what we truly need. Of course, a smarter, more user-friendly playerbot would be an excellent idea!

Moving forward, I'll actively report bugs. I love Azeroth and am building my own World of Warcraft...

kadeshar (dismissed):

hermensbas:
@Wishmaster117 could you have peek at this?

src/Ai/Base/Actions/EquipAction.cpp
```diff
@@ -328,7 +328,43 @@ void EquipAction::EquipItem(Item* item)
...
+        std::string itemUsageParam;
+        if (randomProperty != 0)
+            itemUsageParam = std::to_string(itemId) + "," + std::to_string(randomProperty);
+        else
+            itemUsageParam = std::to_string(itemId);
+
+        ItemUsage usage = AI_VALUE2(ItemUsage, "item upgrade", itemUsageParam);
+        if (usage == ITEM_USAGE_EQUIP || usage == ITEM_USAGE_REPLACE || usage == ITEM_USAGE_BAD_EQUIP)
+            items.insert(itemId);
+    }
```
Wishmaster117, line 363: Does moving the inventory scan into SelectInventoryItemsToEquip() reduce duplication safely without changing behavior, given that the old logic was duplicated in both EquipUpgradesAction::Execute and EquipUpgradeAction::Execute?
Could this introduce any subtle behavioral change if the new helper is now shared across both manual and triggered paths?

src/Ai/Base/Actions/EquipAction.cpp
```diff
@@ -361,72 +397,18 @@ bool EquipUpgradesAction::Execute(Event event)
         p >> itemId;
 
         ItemTemplate const* item = sObjectMgr->GetItemTemplate(itemId);
-        if (item->Class == ITEM_CLASS_TRADE_GOODS && item->SubClass == ITEM_SUBCLASS_MEAT)
+        if (item->InventoryType == INVTYPE_NON_EQUIP)
             return false;
```
Wishmaster117, line 401: Previously, only meat (trade goods + meat) was ignored; now all INVTYPE_NON_EQUIP items are ignored. Could this prevent the upgrade scan from running when “item push result” is a non‑equipable item, even if there are valid upgrades already in bags, potentially delaying or skipping auto‑equip depending on other triggers?

src/Ai/Base/Actions/EquipAction.cpp
```diff
@@ -328,7 +328,43 @@ void EquipAction::EquipItem(Item* item)
...
+
+        int32 randomProperty = item->GetItemRandomPropertyId();
+        uint32 itemId = item->GetTemplate()->ItemId;
+        std::string itemUsageParam;
+        if (randomProperty != 0)
+            itemUsageParam = std::to_string(itemId) + "," + std::to_string(randomProperty);
+        else
+            itemUsageParam = std::to_string(itemId);
+
+        ItemUsage usage = AI_VALUE2(ItemUsage, "item upgrade", itemUsageParam);
```
Wishmaster117, line 360: By switching from "item usage" to a new "item upgrade" value, are we intentionally narrowing the decision logic to equip/ammo only, and could this diverge from any subtle behaviors previously handled by ItemUsageValue::Calculate() (quest/skill/other usage rules)?

src/Ai/Base/Strategy/WorldPacketHandlerStrategy.cpp
```diff
@@ -42,6 +42,7 @@ void WorldPacketHandlerStrategy::InitTriggers(std::vector<TriggerNode*>& trigger
                                                                                 NextAction("query item usage", relevance),
                                                                                 NextAction("equip upgrades", relevance) }));
     triggers.push_back(new TriggerNode("item push result", { NextAction("quest item push result", relevance) }));
+    triggers.push_back(new TriggerNode("loot roll won", { NextAction("equip upgrades", relevance) }));
```
Wishmaster117, line 45: Since loot roll won now triggers equip upgrades without checking if the roll was won by this bot, could this cause unnecessary full inventory scans in group/raid situations when other players win rolls?

src/Ai/Base/ChatActionContext.h
```diff
@@ -256,7 +256,7 @@ class ChatActionContext : public NamedObjectContext<Action>
     static Action* talents(PlayerbotAI* botAI) { return new ChangeTalentsAction(botAI); }
 
     static Action* equip(PlayerbotAI* botAI) { return new EquipAction(botAI); }
-    static Action* equip_upgrades(PlayerbotAI* botAI) { return new EquipUpgradesAction(botAI); }
+    static Action* equip_upgrades(PlayerbotAI* botAI) { return new EquipUpgradesTriggeredAction(botAI); }
```
Wishmaster117: Because the chat action equip_upgrades is now bound to EquipUpgradesTriggeredAction (which checks AutoEquipUpgradeLoot), could manual use of the “equip upgrades” command be blocked when AutoEquipUpgradeLoot = 0, whereas previously manual invocation would still run?

conf/playerbots.conf.dist
```diff
@@ -269,7 +270,7 @@ AiPlayerbot.UseFastFlyMountAtMinLevel = 70
 AiPlayerbot.RandomBotShowHelmet = 1
 AiPlayerbot.RandomBotShowCloak = 1
 
-# Randombots and altbots automatically equip any items in their inventory that are sufficient upgrades
+# Toggles whether altbots will automatically equip items in their inventory that are sufficient upgrades
```
Wishmaster117, line 273: The comment now mentions altbots only, but the code still allows randombots even when AutoEquipUpgradeLoot = 0. Should the comment be aligned with the real behavior (or should behavior be adjusted to match the comment)?

src/Ai/Base/Actions/EquipAction.cpp
```diff
@@ -328,7 +328,43 @@ void EquipAction::EquipItem(Item* item)
...
+        if (usage == ITEM_USAGE_EQUIP || usage == ITEM_USAGE_REPLACE || usage == ITEM_USAGE_BAD_EQUIP)
+            items.insert(itemId);
```
Wishmaster117, line 362: Since only itemId is stored (not randomProperty), could we equip a worse instance when multiple items share the same itemId but differ by random properties? Should we consider storing GUIDs or a composite key to avoid ambiguity?

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -480,6 +409,65 @@ ItemUsage ItemUsageValue::QueryItemUsageForEquip(ItemTemplate const* itemProto,
     return ITEM_USAGE_NONE;
 }
 
+ItemUsage ItemUsageValue::QueryItemUsageForAmmo(ItemTemplate const* proto)
```
Wishmaster117, line 407: Is this extraction strictly behavior‑preserving compared to the previous inline ammo logic in ItemUsageValue::Calculate()?
Could any subtle difference arise from calling the helper in different contexts (e.g., ItemUpgradeValue vs ItemUsageValue)?

Celandriel: So I did catch a small logical error previously that may have been incorrect for ItemUsageValue. 
With this fix I dont anticipate a change.

src/Ai/Base/ChatActionContext.h
Celandriel: What happened here is that the command already pointed to EquipUpgradesTriggeredAction (I just changed the name not what it pointed to) and the original code was incorrect. I think that both  "equip upgrade" and "equip upgrades"should point to the same method

src/Ai/Base/Actions/EquipAction.cpp
Celandriel: So this is the risk that this change takes, that this action will get triggered less often. Its by design to make the call both lighter and less frequent, but the converse side it may result in bots not automatically upgrading loot when received through some other event. THis is actually why I added in the loot roll win, so that bots will auto apply that. If we find new packets that should trigger this action that are not then we can easily add them.

src/Ai/Base/Strategy/WorldPacketHandlerStrategy.cpp
Celandriel: So I just checked, and this message is received by all bots in the party. This would be consistent with current logic where all party members receive the loot-roll package and then all bots perform the action. 

Both actions could be filtered by bot as well, but it again increases the risk of an item not being upgraded if received through a different package than trade, loot roll, or item push. 
Thoughts?

src/Ai/Base/Actions/EquipAction.cpp
Celandriel: I think that for the exact purposes of this action, if you cant equip it then we dont really care. The other uses of the items just make this mroe expenseive. I suggest keeping as is.

src/Ai/Base/Actions/EquipAction.cpp
Celandriel: Possibly, but this maintains the original logic.

conf/playerbots.conf.dist
Celandriel: I took the 'align the comment with the code' approach. Also, because it think that preventing random bots from upgrading loot is a bad idea.

src/Ai/Base/Actions/EquipAction.cpp
Celandriel: I dont expect it at this stage.

Celandriel:
Thanks for the review @Wishmaster117.
I think there are two questions that need to be resolved.

1. Are we ok with reducing the number of calls to this function be only triggering on equipable items even if it may increase the risk that bots receive an item (I dont know through what packet) and not equip it?
2. Should the equip upgrade action only be performed by the bot receiving the item? Also see above. Its better for performance, but at added risk.

kadeshar (approved):

PR #1997, opened 2026-01-09
Extend SummonWhenGroup to auto-added bots

Wishmaster117:
### Summary
Extend AiPlayerbot.SummonWhenGroup to apply when bots are auto-added to a group (e.g., addclass bots or raidus style auto invites).

### Motivation
Bots added automatically to a group never accept a normal invite, so they do not trigger the summon-on-accept path. When SummonWhenGroup is enabled, these bots should also be teleported next to the master to match expected behavior.

### Implementation details
Hook the summon behavior right after automatic group addition.


Wishmaster117:
I make another one

PR #1998, opened 2026-01-09
Fix teleport to invalid map (xxx) or invalid coordinates (x xxxxxx, y yyyyyy, z  zzzzzzz, o oooooo) given when teleporting player

Wishmaster117:
**This patch fix the bad bot teleports that was spamming invalid coords.**
I add a small helper in WorldPosition to normalize the teleport destination only once, using core helpers.
First we try CanReachPositionAndGetValidCoords (path/nav data), and if it fail we fallback to a safe Z from GetHeight / GetWaterLevel.
So we keep it core‑friendly and no duplicated logic.

Then I use this helper only in the two places where we do “hard” teleports (New RPG stuck teleport and corpse teleport).
If the destination is still invalid we just skip and log a debug line, so no more invalid TeleportTo.

The change is very local and should not touch PlayerbotAI or global movement.

### DIFF
--- a/src/TravelMgr.cpp
+++ b/src/TravelMgr.cpp
@@ -231,6 +231,8 @@ bool WorldPosition::NormalizePositionForTeleport(Player* bot)
 
     Map* map = getMap();
     if (!map)
+        map = sMapMgr->FindBaseMap(getMapId());
+    if (!map)
         return false;
 
     float x = getX();
@@ -242,6 +244,7 @@ bool WorldPosition::NormalizePositionForTeleport(Player* bot)
         setX(x);
         setY(y);
         setZ(z);
+
         return true;
     }
 
@@ -250,15 +253,15 @@ bool WorldPosition::NormalizePositionForTeleport(Player* bot)
     float safeZ = std::max(groundZ, waterZ);
     if (safeZ <= INVALID_HEIGHT || safeZ == VMAP_INVALID_HEIGHT_VALUE)
     {
-        LOG_DEBUG("playerbots",
-                  "[Teleport] Invalid destination after height check ({},{},{},{}) groundZ:{} waterZ:{}",
-                  x, y, z, getMapId(), groundZ, waterZ);
+    LOG_ERROR("playerbots", "[MISSED Teleport by Travel Manager] Normalized destination ({},{},{},{})",
+             x, y, safeZ, getMapId());
         return false;
     }
 
     setX(x);
     setY(y);
     setZ(safeZ);
+
     return true;
 }
 
--- a/src/TravelMgr.cpp
+++ b/src/TravelMgr.cpp
@@ -18,6 +18,7 @@
 #include "TransportMgr.h"
 #include "VMapFactory.h"
 #include "VMapMgr2.h"
+#include "SharedDefines.h"
 #include "Corpse.h"
 
 WorldPosition::WorldPosition(std::string const str)
@@ -221,6 +222,44 @@ bool WorldPosition::isUnderWater()
 bool WorldPosition::IsValid()
 {
     return !(GetMapId() == MAPID_INVALID && GetPositionX() == 0 && GetPositionY() == 0 && GetPositionZ() == 0);
+}
+
+bool WorldPosition::NormalizePositionForTeleport(Player* bot)
+{
+    if (!bot)
+        return false;
+
+    Map* map = getMap();
+    if (!map)
+        return false;
+
+    float x = getX();
+    float y = getY();
+    float z = getZ();
+
+    if (map->CanReachPositionAndGetValidCoords(bot, x, y, z))
+    {
+        setX(x);
+        setY(y);
+        setZ(z);
+        return true;
+    }
+
+    float groundZ = map->GetHeight(x, y, MAX_HEIGHT);
+    float waterZ = map->GetWaterLevel(x, y);
+    float safeZ = std::max(groundZ, waterZ);
+    if (safeZ <= INVALID_HEIGHT || safeZ == VMAP_INVALID_HEIGHT_VALUE)
+    {
+        LOG_DEBUG("playerbots",
+                  "[Teleport] Invalid destination after height check ({},{},{},{}) groundZ:{} waterZ:{}",
+                  x, y, z, getMapId(), groundZ, waterZ);
+        return false;
+    }
+
+    setX(x);
+    setY(y);
+    setZ(safeZ);
+    return true;
 }
 
 WorldPosition WorldPosition::relPoint(WorldPosition* center)
--- a/src/TravelMgr.h
+++ b/src/TravelMgr.h
@@ -142,6 +142,7 @@ public:
     bool isInWater();
     bool isUnderWater();
     bool IsValid();
+    bool NormalizePositionForTeleport(Player* bot);
 
     WorldPosition relPoint(WorldPosition* center);
     WorldPosition offset(WorldPosition* center);
--- a/src/strategy/actions/ReleaseSpiritAction.cpp
+++ b/src/strategy/actions/ReleaseSpiritAction.cpp
@@ -14,6 +14,7 @@
 #include "ServerFacade.h"
 #include "Corpse.h"
 #include "Log.h"
+#include "TravelMgr.h"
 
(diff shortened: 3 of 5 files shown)


hermensbas:
Do we have analyses why we get those errors, and why those coordinates are wrong? Because the current coordinates normalization basically assumes its a Z axes problem if not mistaken?

- The state of player/unit?
- Only happens to bots at randomTeleport? So thereby maybe retrieves corrupts coordinates from DB
- Some the valid are valid but not always for x,y,z reason or something?

hermensbas:
Talked to wishmaster:
errors:

```c++
TeleportTo: invalid map (1) or invalid coordinates (X: 2463.8394, Y: 364.44727, Z: -200000, O: 3.1415927) given when teleporting player (GUID Full: 0x0000000000000922 Type: Player Low: 2338, name: Linklu, map: 1, X: 2465.2395, Y: 364.4464, Z: 96.60649, O: 3.1415927).
```

now the Z axes makes more sense

hermensbas:
Teeeesting as i type.

hermensbas (dismissed):

Celandriel:
Why are bots getting invalid heights? Is it trying to teleport to a coordinate and while trying to get the Z coordinate is starting at a Z below the map?

I'm supportive of a guard, but also prefer root causing the origin.

hermensbas:
i am still seeing the error btw, have to double confirm but my server crashing due another bug which i just reverted

Wishmaster117:
> Why are bots getting invalid heights? Is it trying to teleport to a coordinate and while trying to get the Z coordinate is starting at a Z below the map?
> 
> I'm supportive of a guard, but also prefer root causing the origin.

The most likely root cause is the failure of VMAP/MMAP height queries, not a “Z starting below the map”. The logs we had were of the form Z = -200000, which corresponds to the VMAP sentinel “invalid height”. We can already see this handled elsewhere in the module (e.g. MovementActions), so this isn’t a “real” Z value  it’s a “no height found”.
```
bool MovementAction::Move(float angle, float distance)
{
    float x = bot->GetPositionX() + cos(angle) * distance;
    float y = bot->GetPositionY() + sin(angle) * distance;

    // TODO do we need GetMapWaterOrGroundLevel() if we're using CheckCollisionAndGetValidCoords() ?
    float z = bot->GetMapWaterOrGroundLevel(x, y, bot->GetPositionZ());
    if (z == -100000.0f || z == -200000.0f)
        z = bot->GetPositionZ();
    if (!bot->GetMap()->CheckCollisionAndGetValidCoords(bot, bot->GetPositionX(), bot->GetPositionY(),
```
In New RPG, the fallback teleport used the Z coordinate as-is (coming from the location caches) without recalculating the ground, which is why the -200000 appeared when the source was invalid or when VMAP/MMAP were missing. The fix normalizes this destination before teleporting.
```
bool NewRpgBaseAction::MoveFarTo(WorldPosition dest)
{
    if (dest == WorldPosition())
        return false;

    if (dest != botAI->rpgInfo.moveFarPos)
    {
        // clear stuck information if it's a new dest
        botAI->rpgInfo.SetMoveFarTo(dest);
    }

    // performance optimization
    if (IsWaitingForLastMove(MovementPriority::MOVEMENT_NORMAL))
    {
        return false;
    }

    // stuck check
    float disToDest = bot->GetDistance(dest);
    if (disToDest + 1.0f < botAI->rpgInfo.nearestMoveFarDis)
    {
        botAI->rpgInfo.nearestMoveFarDis = disToDest;
        botAI->rpgInfo.stuckTs = getMSTime();
```

```
bool NewRpgBaseAction::MoveFarTo(WorldPosition dest)
{
    if (dest == WorldPosition())
        return false;

    if (dest != botAI->rpgInfo.moveFarPos)
    {
        // clear stuck information if it's a new dest
        botAI->rpgInfo.SetMoveFarTo(dest);
    }

    // performance optimization
    if (IsWaitingForLastMove(MovementPriority::MOVEMENT_NORMAL))
    {
        return false;
    }

    // stuck check
    float disToDest = bot->GetDistance(dest);
    if (disToDest + 1.0f < botAI->rpgInfo.nearestMoveFarDis)
    {
        botAI->rpgInfo.nearestMoveFarDis = disToDest;
        botAI->rpgInfo.stuckTs = getMSTime();
```

For location generation, random teleport already recalculates the ground via GetHeight and rejects the result if invalid, so it’s less suspicious. The invalid Z values mostly came from the “hard teleport” paths that didn’t validate the destination (New RPG / revive).

```
float ground = map->GetHeight(bot->GetPhaseMask(), x, y, z + 0.5f);
if (ground <= INVALID_HEIGHT)
    continue;

z = 0.05f + ground;

if (!botAI->CheckLocationDistanceByLevel(bot, loc, true))
    continue;

const LocaleConstant& locale = sWorld->GetDefaultDbcLocale();
LOG_DEBUG("playerbots",
          "Random teleporting bot {} (level {}) to Map: {} ({}) Zone: {} ({}) Area: {} ({}) ZoneLevel: {} "
          "AreaLevel: {} {},{},{} ({}/{} "
          "locations)",
          bot->GetName().c_str(), bot->GetLevel(), map->GetId(), map->GetMapName(), zone->ID,
          zone->area_name[locale], area->ID, area->area_name[locale], zone->area_level, area->area_level, x, y,
          z, i + 1, tlocs.size());

if (hearth)
{
    bot->SetHomebind(loc, zone->ID);
}
```
```
if (deadTime > delay)
{
    bot->GetMotionMaster()->Clear();
    bot->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_TELEPORTED | AURA_INTERRUPT_FLAG_CHANGE_MAP);
    WorldPosition safePos = moveToPos;
    if (safePos.NormalizePositionForTeleport(bot))
    {
        LOG_INFO("playerbots", "Teleporting {} corpse to ({},{},{},{})",
                 bot->GetName(), safePos.getX(), safePos.getY(), safePos.getZ(), safePos.getMapId());
        bot->TeleportTo(safePos.getMapId(), safePos.getX(), safePos.getY(), safePos.getZ(), 0);
    }
    else
    {
        LOG_DEBUG("playerbots", "Skip corpse teleport for {} due to invalid destination ({},{},{},{})",
                  bot->GetName(), moveToPos.getX(), moveToPos.getY(), moveToPos.getZ(), moveToPos.getMapId());
```

So the root cause is the destination was sometimes invalid because the height returned by VMAP/MMAP was invalid, and those teleports had no safeguard. The patch adds that safeguard and a terrain fallback to guarantee a consistent coordinate.

Wishmaster117:
Set to draft because missing some teleports who need to be patched with safe helper

liyunfan1223:
I’ve also looked into this issue before, some info for reference:

The spamming message is caused by mage bots casting blink at invalid locations. It was introduced in [this commit](https://github.com/mod-playerbots/mod-playerbots/pull/980/commits/41eaa6515bbb0dd17a505203e9b82027fc7187c4), which enabled blink for them.

This seems to be a core issue rather than bot-specific, as it's reproducible without bots:

1. .go xyz 2463.83 364.44 95 1
2. cast blink

Wishmaster117:
> I’ve also looked into this issue before, some info for reference:
> 
> The spamming message is caused by mage bots casting blink at invalid locations. It was introduced in [this commit](https://github.com/mod-playerbots/mod-playerbots/pull/980/commits/41eaa6515bbb0dd17a505203e9b82027fc7187c4), which enabled blink for them.
> 
> This seems to be a core issue rather than bot-specific, as it's reproducible without bots:
> 
> 1. .go xyz 2463.83 364.44 95 1
> 2. cast blink

Hi thanks for be information , but it's not the only cause, i'm trying to find the others.

notOrrytrout (commented):
Could there still be there be some execution paths where TeleportTo is invoked before, or independently of, normalization (for example, corpse/spirit healer handling or the NewRPG stuck teleport)? In those cases, the guard may not be fully effective.

notOrrytrout:
we should ensure no TeleportTo is executed unless NormalizePositionForTeleport() succeeds

Wishmaster117:
Hi it's a draft for now, so it may be not the final solution. 
Trying to figure out why sometimes core send wrong coords

PR #2003, opened 2026-01-11
[HOTFIX] Revert "Fix/Feat: Stop bots in party from PVP when master isn't, and …

hermensbas:
…PVP probablity system (#1914)"

This reverts commit 02e8465a3b91a12ca2ce56a1a704dfd80368c121.

[worldserver_gdb_20260111_000520.log](https://github.com/user-attachments/files/24547263/worldserver_gdb_20260111_000520.log)

Noticed the server kept crashing all the sudden, first thought it was local issue but see dump.

### DIFF
--- a/src/strategy/values/EnemyPlayerValue.cpp
+++ b/src/strategy/values/EnemyPlayerValue.cpp
@@ -11,10 +11,6 @@
 
 bool NearestEnemyPlayersValue::AcceptUnit(Unit* unit)
 {
-    // Apply parent's filtering first (includes level difference checks)
-    if (!PossibleTargetsValue::AcceptUnit(unit))
-        return false;
-
     bool inCannon = botAI->IsInVehicle(false, true);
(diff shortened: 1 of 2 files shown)


avirar:
I can confirm this crash - encountered the same SIGSEGV at PossibleTargetsValue.cpp:59 in CleanupAttackDecisionCache().

The root cause is a race condition: the static attackDecisionCache lacks thread protection, causing hash table corruption when multiple map worker threads access it simultaneously during bot AI updates.

I've prepared a fix that adds mutex protection following the same pattern already used in WarlockActions.cpp (soulstoneReservations) and RaidIccMultipliers.cpp (plague mechanics). This would restore the PVP probability feature with proper thread safety.

Branch: https://github.com/avirar/mod-playerbots/tree/fix/attack-decision-cache-thread-safety

Happy to submit a PR if there's interest in restoring this functionality.

PR #2008, opened 2026-01-11
Fix/Feat: PVP with master and PVP probablity system (thread-safe remake)

NoxMax:
This is a remake of #1914 that had to be reverted. Original PR had a thread-safe issue where a crash happens if multiple threads access the cache at the same time. Unfortunately this problem was not caught in earlier testing. I don't know if because I was testing on a month old branch, if my settings had only ~2000, or if I needed test runs longer than an hour to find out.

Regardless, this has all been addressed. Test have been run on the latest commits from today (2026/1/11), with all 7500 of my bots active, with a test run that lasted 15 hours. All stable and bots are following the probability system without issue.

~~The new edit uses mutex locking, preventing simultaneous access of the cache by multiple threads.~~
The new edit uses deterministic hashing, thereby not having issues with cache thread safety to begin with. Thank you @hermensbas for catching and reverting the original problem PR. Apologies for not catching the issue myself.

---
Original PR description:

There are two related PVP components in this PR. First is the simple yet fundamental change to bot behaviour when they are in party. Right now bots with a master will go into PVP when there's a nearby PVP target, even if master is not in PVP. This absolutely should not happen. Bots should not consider PVP at all if master is not in PVP. The fix is only 3 lines in EnemyPlayerValue

The second component is introducing PVP probabilities, to make decisions more realistic. Right now even a level 1 bot will 100% go into PVP if it sees a level 80 PVP target. They can't help themselves. So the change here addresses that insanity. Several thresholds (subject to community review) are introduced:

1. Bots will not fight a target 5 or more levels higher than them
2. Bots have a 25% chance starting a fight with a target +/- 4 levels from them.
3. Bots have a 50% chance starting a fight with a target +/- 3 levels from them.
4. Bots have a 75% chance starting a fight with a target +/- 2 levels from them.
5. Bots have a 100% chance starting a fight with a target +/- 1 level from them.
6. Bots have a 25% chance starting a fight with a target 5 or more levels below them (ganking. thought it would be funny, and technically realistic of player behaviour)

Exception of course exist for BG/Arena/Duel, and in capitals where bots will always PVP. Also bots will always defend themselves if attacked.

Few notes: 
1. The if/ else if logic can be further simplified, but only if we use thresholds that are different by one. So current logic allows for flexibility of using values like 10/7/5/3 instead of 5/4/3/2.
2. The caching system is per-bot basis. So for some target X, if some bot decides to attack it, another bot will make its own decision. At first I used a simplified global system (thinking there might be performance concerns) where if one bot decides to attack a target then they all do, but when I switched to the more realistic per-bot basis, I didn't see an effect on performance.
3. Variables are obviously not configurable right now. I'm starting to see Bash's POV that maybe we have too many configs :grimacing: Still, they can be easily exposed in the future, and if someone is reading this then, remember to change constexpr  to const.

### DIFF
--- a/src/Ai/Base/Value/PossibleTargetsValue.cpp
+++ b/src/Ai/Base/Value/PossibleTargetsValue.cpp
@@ -17,8 +17,6 @@
 #include "SpellMgr.h"
 #include "Unit.h"
 #include "AreaDefines.h"
-#include <unordered_map>
-#include <mutex>
 
 // Level difference thresholds for attack probability
 constexpr int32 EXTREME_LEVEL_DIFF = 5;  // Don't attack if enemy is this much higher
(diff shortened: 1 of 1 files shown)


hermensbas:
No worries, there are so many variables when it comes to testing. Basically turns into best effort.

In general i use the amount of bots my system can carry with decent latency (100, 150, 250), for my system thats about 4k bots. And put then on 100% active. 

Then run on AV with self bot and sometimes some dungeons/raid for like 10-20m

i will try to have look at when i can. Not sure we need mutex for this though.

src/Ai/Base/Value/PossibleTargetsValue.cpp
```diff
@@ -24,7 +52,121 @@ void PossibleTargetsValue::FindUnits(std::list<Unit*>& targets)
...
+{
+    if (!AttackersValue::IsPossibleTarget(unit, bot, range))
+        return false;
+
+    // Level-based PvP restrictions
+    if (unit->IsPlayer())
+    {
+        // Self-defense - always allow fighting back
+        if (bot->IsInCombat() && bot->GetVictim() == unit)
+            return true;  // Already fighting
```
kadeshar, line 51: What when bot is in group and other member of his group is attacked by this unit? I can see that is problem in open world pvp

NoxMax: You mean whether or not they would come to the defence of another member of their party, when that member is being attacked by some high level attacker? They would go into PVP and defend their party member, even though the attacker is not a target they would start a fight with on their own.

hermensbas:
This mutex and statefull actions feels wrong, need to dig into that.

NoxMax:
per bot cache might be an alternative, though it requires some more refactoring and additions. not sure there is actual benefit to that approach though.

hermensbas:
How about a more stateless "deterministic" approach with time window? that doesnt require caches and threadsafe by design. These actions are called a chained context per bot per thread etc. Actions it not really the place to have cache in them they rather should be stateless if possible. 

e.g:

https://onlinegdb.com/nukjdI13B

NoxMax:
It would create some wave like behavioral decision making at these time window boundaries. I can look at it and see what it acts like and how it performs.

NoxMax:
Followed Bash's advice and dropped caching altogether. Implemented hashing with FNV. Code is even simpler. Comments about FNV is slightly verbose because this is the first use of it in PB, and it might be helpful for future maintainers.

I won't merge it to this PR just yet. I want someone to take a quick look at it first for any major issues. Compiled, did some light testing and it works so far. Will do longer testing after at least one preliminary review.

https://github.com/mod-playerbots/mod-playerbots/compare/master...NoxMax:mod-playerbots:pvp-master-prob-FNV

NoxMax:
Switched the PR to the deterministic hash approach after several days of running it on my server without issue.

PR #2014, opened 2026-01-14
Fix the duplicate spells issue during randomization

privatecore:
**Original issue:**
Bots/Characters received duplicate spells during randomization process.

**Root cause:**
When `PlayerbotFactory::Randomize` is processed, `InitSkills` is called AFTER `InitAvailableSpells`, which causes the duplicate spells issue because the skillline ability spell is already learned when processing spells from trainers (`InitAvailableSpells`). 

We simply need to change the order of the randomization process: skills should be handled first, then spells. An alternative approach would be to adjust the skillline abilities and check each spell for every skill, but that seems redundant since we already have checks for the trainer's spells.

`InitSkills` -> `SetRandomSkill` -> `SetSkill` -> `learnSkillRewardedSpells` -> `learnSpell` -> duplicate error...

**Steps to reproduce:**
1. create common character and login with it
2. set level for this character eq. 80 (`.set level 79`)
3. create and run macro:
```
/g .playerbots bot initself
/g .saveall
```
4. logout -> login and run macro again

**Note:**
Also added checks for the trainer's spells since `GetSpell` can return nullptr. Updated `LearnQuestSpells` after recent changes and used the same logic and implementation from `Player::learnQuestRewardedSpells`.

Yes, we need to cast spells, or we should handle all spell effects that quests/trainers have (for ex.: `SPELL_EFFECT_SKILL_STEP`, `SPELL_EFFECT_UNLEARN_SPECIALIZATION`, `SPELL_EFFECT_BIND`).

### DIFF
--- a/src/factory/PlayerbotFactory.cpp
+++ b/src/factory/PlayerbotFactory.cpp
@@ -288,17 +288,17 @@ void PlayerbotFactory::Randomize(bool incremental)
             pmo->finish();
     }
 
-    pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Spells1");
-    LOG_DEBUG("playerbots", "Initializing spells (step 1)...");
+    LOG_DEBUG("playerbots", "Initializing skills (step 1)...");
+    pmo = sPerformanceMonitor->start(PERF_MON_RNDBOT, "PlayerbotFactory_Skills1");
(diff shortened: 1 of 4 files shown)


hermensbas:
Tested against all known scenario's and did regression tests aside from reading the code. Thanks alot mr privatecore

PR #2022, opened 2026-01-16
Smart loot system refactored 2026 StatsWeightCalculator ReBased

Wishmaster117:
# Summary

This PR refines and hardens loot roll decision‑making for Playerbots while staying core‑friendly and aligned with AzerothCore’s DB semantics.

The stat‑weight upgrade evaluation is preserved, but wearability rules are centralized, item stat profiles are unified, and loot operations are kept clean. The logic now relies more directly on `StatsWeightCalculator` for spec‑aware stat relevance, which reduces hard‑coded exceptions and keeps behavior consistent with core data.

**Important:** to enable the full behavior you must set:  
`AiPlayerbot.LootRollLevel = 2`

* `0` → bots always PASS  
* `1` → bots never NEED, GREED is conservative  
* `2` → full smart loot logic (recommended)

**New:** group‑wide checks are now gated by config:  
`AiPlayerbot.Roll.UseGroupUsageChecks = 0` (default)

---

# Key Changes

## 1) Weight‑driven smart stat decisions
* A spec‑aware **SmartStat mask** is derived from `StatsWeightCalculator` and reused by higher‑level rules.
* Stat relevance is no longer duplicated across heuristics; it is centralized and consistent with the weight model.

**Result:** loot decisions follow the same stat priorities used for item scoring.

---

## 2) Smart NEED by spec & stats (refined)
* A centralized `ItemStatProfile` aggregates stats and is reused by higher‑level rules.
* Physical specs never NEED caster gear (SP/INT/SPI/MP5), and never NEED weapons/armor they shouldn’t use.
* Casters/healers never NEED pure melee DPS gear (STR/AP/ARP/EXP without SP/INT).
* Healers treat caster Hit gear as off‑spec (GREED), while DPS casters can still NEED.
* **Shield tanks (Prot Paladin / Prot Warrior)** never treat 2H weapons as primary.

**Human‑style:** “Main spec > off spec. If nobody truly needs it, I might NEED for offspec – but not on obviously wrong stats.”

---

## 3) Cross‑armor upgrade logic
* Off‑armor NEED (cloth for paladin, leather for hunter, etc.) is allowed only if:
  * It’s lower‑tier body armor for that class.
  * No primary‑armor bot sees it as an upgrade.
  * The new item is a clear stat‑weight upgrade (`newScore >= bestOld * AiPlayerbot.Roll.CrossArmorExtraMargin`).

* If `AiPlayerbot.Roll.CrossArmorGreedIsPass = 1`, bots PASS instead of GREED on cross‑armor.

---

## 4) BoP / BoE / BoU etiquette
* **BoP:** bots can NEED if the item matches spec and fills an empty slot.  
* **BoE / BoU:** bots GREED unless explicitly allowed by config.

---

## 5) Grey‑gear replacement guard
* Bots only auto‑equip replacements for **poor‑quality gear** if the new item is **green+**.
* Prevents equip/unequip loops on grey items.

---

## 6) Unique‑equip items
* Bots never NEED duplicates of unique‑equip items already owned.  
* Exception: **Book of Glyph Mastery** can be NEEDed multiple times.

---

## 7) Lockboxes & rogues
* Rogues with Lockpicking are allowed to NEED lockboxes.

---

## 8) Targeted debug logs (low‑noise)
* Added focused logs for:
  * `poor-slot upgrade`
  * `cross‑armor` decisions (blocked, score=0, margin fail, equip)
  * `equip check` (score below upgrade threshold)

---

# Improvements
* **Centralized stat relevance** using `StatsWeightCalculator` → fewer hard‑coded rules.
* **Cleaner off‑spec handling**, fewer “ninja NEED” edge cases.
* **Configurable thresholds** for strict‑raid vs casual tuning.
* **Minimal, targeted debug logs** for diagnosing loot decisions.

---

# New Config
AiPlayerbot.Roll.UseGroupUsageChecks = 0  
Default is **disabled** (off by default for performance/fairness concerns).

---

## How the refactor maps to your structure

**StatsWeightCalculator**  
- Now strictly used to compute item value and **derive the SmartStat mask**.

**RandomItemMgr**  
- Continues to be the scoring entrypoint (`CalculateItemWeight`).

**LootValue / ItemUsage**  
- Usage rules, offspec rules, cross‑armor, and stat‑profile logic are centralized here.
- Loot usage queries are consistent and reused by LootRollAction.

**LootRollAction**  
- Reduced to “perform roll + apply etiquette + ask LootUsage.”  
- No business rules embedded in action layer.

**GroupHas* concern**  
- Group‑wide checks are **fully gated by config**.  
- Default is **off**, so bots do not evaluate other bots unless explicitly enabled.


hermensbas:
i will test this in the weekend.

Wishmaster117:
Guys, if you see any strange rolls, please report them to me.

Ben19882:
Hi @Wishmaster117 !

Cool that you’re further developing this awesome rework! I have a question about this PR: is it going to replace the other PR, or is it an additional PR on top of the existing one?
If this is going to replace i know what pr to Merge :-)

Thanks!

Wishmaster117:
> Hi @Wishmaster117 !
> 
> Cool that you’re further developing this awesome rework! I have a question about this PR: is it going to replace the other PR, or is it an additional PR on top of the existing one? If this is going to replace i know what pr to Merge :-)
> 
> Thanks!

This one replace the other one i closed, but wait a little i'me finishing some adjustments about items that have spell power

src/ScriptHook/Playerbots.h
```diff
@@ -49,4 +49,33 @@ int strcmpi(char const* s1, char const* s2);
 #define GAI_VALUE(type, name) sSharedValueContext->getGlobalValue<type>(name)->Get()
 #define GAI_VALUE2(type, name, param) sSharedValueContext->getGlobalValue<type>(name, param)->Get()
 
+// Iterate over all bot members in the same group and invoke a functor.
+// Returns true as soon as the functor signals success by returning true.
+template <typename Func>
```
hermensbas: Prolly best to extract and place is Helper under /Util

hermensbas:
Still under review

Diabloxx:
Testing this aswell, tired or a shaman using a tank ring.

Diabloxx:
Just did ignis in Ulduar, Bots didnt need any gear, just greeded.

Also noticed casters greeding on a tanking trinket that had stamina on it, they are not reading the use/equip stat?

Wishmaster117:
> Just did ignis in Ulduar, Bots didnt need any gear, just greeded.
> 
> Also noticed casters greeding on a tanking trinket that had stamina on it, they are not reading the use/equip stat?

This not happening before the rebase, i tested this un dungeons and ICC

Diabloxx:
> > Just did ignis in Ulduar, Bots didnt need any gear, just greeded.
> > Also noticed casters greeding on a tanking trinket that had stamina on it, they are not reading the use/equip stat?
> 
> This not happening before the rebase, i tested this un dungeons and ICC

I'll do another boss, I notice feral druids needing on strenght ring aswel, I'll do more bosses and post screenshots

Wishmaster117:
> > > Just did ignis in Ulduar, Bots didnt need any gear, just greeded.
> > > Also noticed casters greeding on a tanking trinket that had stamina on it, they are not reading the use/equip stat?
> > 
> > 
> > This not happening before the rebase, i tested this un dungeons and ICC
> 
> I'll do another boss, I notice feral druids needing on strenght ring aswel, I'll do more bosses and post screenshots

I'll take a look, i go to ulduar now with the version before rebase and i tests after with the rebased version, maybe some guards are missing after rebase and i suspect we removed some declarations that have to stay

Wishmaster117:
Just made a ulduar with the version before rebase and loots are accurate, the only thing is that DK don't need in sigils but it's normal i forget to handle them, and bots greed on ulduar tokens, also normal i'm not handling them like in iCC

Diabloxx:
> Just made a ulduar with the version before rebase and loots are accurate, the only thing is that DK don't need in sigils but it's normal i forget to handle them, and bots greed on ulduar tokens, also normal i'm not handling them like in iCC

image
image

Wishmaster117:
> > Just made a ulduar with the version before rebase and loots are accurate, the only thing is that DK don't need in sigils but it's normal i forget to handle them, and bots greed on ulduar tokens, also normal i'm not handling them like in iCC
> 
> image image

I Said:
###  I made a ulduar with the version before rebase, not with this one here.
With the version before rebase all it's ok, now i try to figure out why it not works properly after rebase.

Wishmaster117:
Please try now

Celandriel:
@RovxBot it would be helpful to turn on performance monitoring when that happens.
`playerbots pmon toggle`
And then after a while type
`playerbots pmon stack` to get th output.
Copy to a text editor and share

brighton-chi:
Looking at your log, it's equip upgrades. There is a PR in progress to fix it. Do you by chance have bots with many professions learned in your group?

brighton-chi:
In the meantime, if you want a bandaid, you can add the following after line 376 in EquipAction.cpp:

```

const ItemTemplate* itemProto = item->GetTemplate();
if (!itemProto || itemProto->InventoryType == INVTYPE_NON_EQUIP)
    continue;

```

Wishmaster117:
> > @RovxBot it would be helpful to turn on performance monitoring when that happens. `playerbots pmon toggle` And then after a while type `playerbots pmon stack` to get th output. Copy to a text editor and share
> 
> [Playerbots.log](https://github.com/user-attachments/files/24706976/Playerbots.log) OK. Ran the above and did a full dungeon run, and got some good lag spikes. I wont roll back to the main branch just yet, so let me know if want logs from a longer session.

If we analyze the **TOTAL BOT** section, which shows the distribution of CPU time by category, we can draw the following conclusions:

### **1) Actions → 47.6% of total time**  
This is the heaviest category: **47.627% — 44.848s — Action: Total**.  
It is therefore the largest source of CPU time and, consequently, the main contributor to overall latency.

Within these actions, the most expensive ones by global percentage are:  
- **equip upgrades:** 5.044% (4.750s)  
- **revenge:** 4.497% (4.234s)  
- **new RPG status update:** 4.087% (3.849s)  
- **sinister strike:** 3.933% (3.704s)  
- **execute:** 3.518% (3.313s)  
- **eviscerate:** 3.148% (2.964s)  
- **earth shock:** 2.394% (2.254s)  
- **new RPG do quest:** 1.950% (1.837s)

**Conclusion:** latency comes mainly from combat/RPG bot Actions, with a heavy weight from combat actions and RPG updates.

---

### **2) UpdateAIInternal ≈ 100% of total time, with internal split**  
The Total section shows that most of the time is consumed by **PlayerbotAI::UpdateAIInternal**:

- **100.000% — 94.165s — Total: Total**  
- UpdateAIInternal 1: 43.696%  
- UpdateAIInternal 0: 27.181%  
- UpdateAIInternal I: 17.924%  
- UpdateAIInternal 530: 11.199%

This indicates that the core AI loop is responsible for nearly all the load.  
However, the detailed breakdown shows that **Actions** are what weigh the most inside this loop.

---

### **3) Triggers → 16.5% of total time**  
Triggers are significant but secondary: **16.515% — 15.551s — Trigger: Total**.  
The heaviest triggers remain individually small (<1%), so they are not the main source of latency.

---

### **4) Random Bot Factory → 20.9%**  
The RndBot part is non‑negligible: **20.905% — 19.685s — RndBot: Total**.  
The biggest contributors are:  
- **Refresh:** 10.633%  
- **PlayerbotFactory_Reset:** 7.466%

This overhead is linked to bot regeneration/refresh.

---

### **Summary — the main sources of latency are:**  
- **Bot Actions (47.6%)** — mainly combat/RPG.  
- **Bot refresh/reset (RndBot Refresh/Reset ≈ 18%).**  
- **Triggers (16.5%)**, but clearly secondary.

So the latency is **not exclusively related** to the intelligent loot system PR.  
However, since bots roll for loot and equip themselves automatically, it does add some overhead — but it is **not** the primary cause of latency.

However, it would be interesting if you could provide us with the specifications of your hardware.

biship:
@Wishmaster117 don't you also need to add AiPlayerbot.Roll.SanctificationTokenMode & UseGroupUsageChecks to playerbots.conf.dist ?

Wishmaster117:
> @Wishmaster117 don't you also need to add AiPlayerbot.Roll.SanctificationTokenMode & UseGroupUsageChecks to playerbots.conf.dist ?

Yes sorry i forget to upload the .conf file, done now

Diabloxx:
I can also say, that i've noticed a few lagspikes aswell after adding the PR. 

System my server is running on is

Ryzen 7 Pro 8700GE
64GB DDR5 RAM (5200MHZ)
Windows Server 2022
image
image

Wishmaster117:
> I can also say, that i've noticed a few lagspikes aswell after adding the PR.
> 
> System my server is running on is
> 
> Ryzen 7 Pro 8700GE 64GB DDR5 RAM (5200MHZ) Windows Server 2022 image image

Yes, I see, but the problem isn’t entirely related to this PR. What’s happening is that the bots are "intelligent": on every loot they compare items, decide whether to take the loot or not, and decide whether to equip it. From my point of view, the issue comes from EquipAction, which iterates over items in a flawed way.

A PR proposing an improvement was submitted here:
https://github.com/mod-playerbots/mod-playerbots/pull/1994
but I’m not convinced by the approach, or maybe I didn’t understand it correctly. I’m waiting for the author to clarify it for me.

Diabloxx:
> > I can also say, that i've noticed a few lagspikes aswell after adding the PR.
> > System my server is running on is
> > Ryzen 7 Pro 8700GE 64GB DDR5 RAM (5200MHZ) Windows Server 2022 image image
> 
> Yes, I see, but the problem isn’t entirely related to this PR. What’s happening is that the bots are "intelligent": on every loot they compare items, decide whether to take the loot or not, and decide whether to equip it. From my point of view, the issue comes from EquipAction, which iterates over items in a flawed way.
> 
> A PR proposing an improvement was submitted here: #1994 but I’m not convinced by the approach, or maybe I didn’t understand it correctly. I’m waiting for the author to clarify it for me.

I agree on this, i just read through the comments, and yeah clarity is needed on that pr.

Wishmaster117:
The smart-stat logic now relies fully on StatsWeightCalculator, ensuring the mask is derived from the same weighting pipeline used for item scoring.

hermensbas (dismissed):

src/Ai/Base/Actions/LootRollAction.cpp
```diff
@@ -5,151 +5,84 @@
...
+static inline int32 EncodeRandomEnchantParam(uint32 randomPropertyId, uint32 randomSuffix)
+{
+    if (randomPropertyId)
+        return static_cast<int32>(randomPropertyId);
+
+    if (randomSuffix)
+        return -static_cast<int32>(randomSuffix);
+
+    return 0;
+}
```
SmashingQuasar: Could you clarify the intention here, please?
I am unsure why we are inverting the sign on `randomSuffix`. I looked up the usage afterwards but it is quite unclear to me. Maybe we could add some documentation to this function to make it a bit more understandable.

src/Ai/Base/Actions/LootRollAction.cpp
```diff
@@ -5,151 +5,84 @@
...
+                  guid.ToString(),
+                  itemId);
+
+        group->CountRollVote(bot->GetGUID(), guid, sent);
+        // One item at a time
+        return true;
     }
 
-    return StoreLootAction::IsLootAllowed(proto->ItemId, GET_PLAYERBOT_AI(bot)) ? needVote : PASS;
+    return false;
```
SmashingQuasar, line 73: Is it intended to return `false` if no roll have been processed?
From my understanding on how actions work, this will make the bot trigger this action again, possibly indefinitely.

src/Ai/Base/Actions/LootRollAction.cpp
```diff
@@ -253,10 +171,8 @@ bool RollAction::Execute(Event event)
...
-        if (usage == ITEM_USAGE_EQUIP || usage == ITEM_USAGE_REPLACE || usage == ITEM_USAGE_BAD_EQUIP)
-        {
-            bot->DoRandomRoll(0,100);
-        }
+            if (usage == ITEM_USAGE_EQUIP || usage == ITEM_USAGE_REPLACE)
+                bot->DoRandomRoll(0, 100);
     }
     return true;
-}
+}
```
SmashingQuasar, line 156: You are missing an empty line at the end of the file.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -5,33 +5,130 @@
...
+        parsed.randomPropertyId = atoi(qualifier.substr(pos + 1).c_str());
     }
     else
     {
-        itemId = atoi(qualifier.c_str());
+        parsed.itemId = atoi(qualifier.c_str());
+    }
+
+    return parsed;
+}
```
SmashingQuasar, line 51: You can simplify this method by using an early return:

```cpp
ParsedItemUsage ParseItemUsageQualifier(std::string const& qualifier)
{
    ParsedItemUsage parsed;
    size_t pos = qualifier.find(",");

    if (pos != std::string::npos)
    {
        ParsedItemUsage parsed.itemId = atoi(qualifier.substr(0, pos).c_str());
        parsed.randomPropertyId = atoi(qualifier.substr(pos + 1).c_str());

        return parsed;
    }

    parsed.itemId = atoi(qualifier.c_str());

    return parsed;
}
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+    if (cls == CLASS_DEATH_KNIGHT)
+        return ITEM_SUBCLASS_ARMOR_PLATE;
 
-    bool shouldEquip = false;
-    // uint32 statWeight = sRandomItemMgr->GetLiveStatWeight(bot, itemProto->ItemId);
-    StatsWeightCalculator calculator(bot);
-    calculator.SetItemSetBonus(false);
-    calculator.SetOverflowPenalty(false);
+    return ITEM_SUBCLASS_ARMOR_CLOTH;
+}
```
SmashingQuasar, line 390: Hmm... I understand the intention here, but healers (like holy paladins) don't really limit themselves to plate, especially before WotLK. It's fine to keep it that way here, but it will most likely create bizarre gear being prioritised while leveling.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
             }
-        }
 
-        bool existingShouldEquip = true;
-        if (oldItemProto->Class == ITEM_CLASS_WEAPON && !sRandomItemMgr->CanEquipWeapon(bot->getClass(), oldItemProto))
-            existingShouldEquip = false;
+            bool hasVeryBadItem = !bestProto || bestProto->Quality <= ITEM_QUALITY_NORMAL;
+            return hasVeryBadItem;
+        });
+}
```
SmashingQuasar, line 505: It's really, really hard to read a massive return callable like that. Maybe you could refactor this a bit to make it easier to maintain.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
 
-        // Item is not better but current item is broken and new one is not.
-        if (oldItemIsBroken && !itemIsBroken)
-            return ITEM_USAGE_EQUIP;
+        if (s.hasSP)
+            break;
+    }
+
+    return s;
+}
```
SmashingQuasar: That's a super long method, very hard to follow properly. You you split it at least in two, one for each loop and use a method that calls both.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+static bool IsPrimaryForSpec(Player* bot, ItemTemplate const* proto)
+{
+    if (!bot || !proto)
+        return false;
+
+    const SpecTraits traits = GetSpecTraits(bot);
+    ItemStatProfile const stats = BuildItemStatProfile(proto);
+    const bool hasCaster = stats.hasINT || stats.hasSP || stats.hasMP5;
+    const bool hasPhysical = stats.hasSTR || stats.hasAGI || stats.hasAP || stats.hasARP;
+    const bool hasCasterOffense = stats.hasHIT || stats.hasCRIT || stats.hasHASTE;
```
SmashingQuasar: `stats.hasCRIT` and `stats.hasHASTE` should count for `hasCaster` as well. Holy paladins need a lot of spell critical.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+    if (proto->Class == ITEM_CLASS_WEAPON)
+    {
+        if ((traits.isHunter || traits.isRogue || traits.isEnhSham) && stats.hasSTR)
+            return false;
+
+        if ((traits.isProtPal || traits.isWarProt) && proto->InventoryType == INVTYPE_2HWEAPON)
+            return false;
+
+        if (traits.isTank && hasCaster)
+            return false;
```
SmashingQuasar, line 877: That will not work for protection paladin. They need spell power.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+                proto->SubClass == ITEM_SUBCLASS_WEAPON_SWORD || proto->SubClass == ITEM_SUBCLASS_WEAPON_MACE ||
+                proto->SubClass == ITEM_SUBCLASS_WEAPON_WAND;
+             return isCasterWeapon && (hasCaster || (hasCasterOffense && !hasPhysical));
+        }
+
+        return hasPhysical;
+    }
+
+    if (proto->Class != ITEM_CLASS_ARMOR)
+        return true;
```
SmashingQuasar, line 898: Won't this return make it so that all off hands will always be considered as needed by bots?

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+
+    if (IsJewelryOrCloak(proto))
+    {
+        if (traits.isCaster != hasCaster)
+            return false;
+        return IsFallbackNeedReasonableForSpec(bot, proto);
+    }
+
+    if (IsLowerTierArmorForBot(bot, proto))
+        return false;
```
SmashingQuasar, line 917: This is what I was talking about, I don't believe you should restrict bots to need strictly on their highest armor tier. This will lead to bizarre choices.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+ItemUsage ItemUsageValue::QueryItemUsageForEquip(ItemTemplate const* itemProto, int32 randomPropertyId)
+{
+    if (!sRandomItemMgr->CanEquipForBot(bot, itemProto))
+        return ITEM_USAGE_NONE;
+
+    Item* pItem = Item::CreateItem(itemProto->ItemId, 1, bot, false, 0, true);
+    if (!pItem)
+        return ITEM_USAGE_NONE;
+
+    uint16 dest;
```
SmashingQuasar: You should really initialise this variable. If `CanEquipItem` changes in anyway, you might get an extremely difficult to understand undefined behaviour crash.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+
+    return usage;
+}
+
+ItemUsage ItemUsageValue::QueryItemUsageForEquip(ItemTemplate const* itemProto, int32 randomPropertyId)
+{
+    if (!sRandomItemMgr->CanEquipForBot(bot, itemProto))
+        return ITEM_USAGE_NONE;
+
+    Item* pItem = Item::CreateItem(itemProto->ItemId, 1, bot, false, 0, true);
```
SmashingQuasar: I don't think you need to pass the bot as parameter, `Item::CreateItem` supports having a `nullptr` as player. This will eliminate the need to remove it from the update queue later.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+    // Check is unique items are equipped or not
+    bool needToCheckUnique = false;
+    if (result == EQUIP_ERR_CANT_CARRY_MORE_OF_THIS)
+    {
+        needToCheckUnique = true;
+    }
+    else if (itemProto->HasFlag(ITEM_FLAG_UNIQUE_EQUIPPABLE))
+    {
+        needToCheckUnique = true;
+    }
```
SmashingQuasar: This could easily be refactored in another method.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+
+        // Determine if the unique item is already equipped
+        bool isEquipped = (totalItemCount > bagItemCount);
+
+        if (isEquipped)
+        {
+            return ITEM_USAGE_NONE;  // Item is already equipped
+        }
+        // If not equipped, continue processing
+    }
```
SmashingQuasar: Same here, this is an easy refactor in another method to improve the code.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+            itemProto->Quality >= ITEM_QUALITY_UNCOMMON &&
+            IsFallbackNeedReasonableForSpec(bot, itemProto))
+        {
+            LOG_DEBUG("playerbots",
+                      "[LootRollDBG] poor-slot upgrade: bot={} slot={} oldItemId={} newItemId={} newQ={}",
+                      bot->GetName(), dest + i, oldItemProto->ItemId, itemProto->ItemId, itemProto->Quality);
+            return ITEM_USAGE_EQUIP;
+        }
+        float oldScore = calculator.CalculateItem(oldItemProto->ItemId, oldItem->GetInt32Value(ITEM_FIELD_RANDOM_PROPERTIES_ID));
+        if (oldItem)
```
SmashingQuasar: This condition is unnecessary because you already handled pointer nullity in all situations before.

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+        // No item equipped
+        if (!oldItem)
+        {
+        if (shouldEquipInSlot || IsFallbackNeedReasonableForSpec(bot, itemProto))
+                return ITEM_USAGE_EQUIP;
+            else
+            {
+                return ITEM_USAGE_BAD_EQUIP;
+            }
+        }
```
SmashingQuasar, line 1198: This can be simplified by using an early return:

```cpp

        if (!oldItem)
        {
            if (shouldEquipInSlot || IsFallbackNeedReasonableForSpec(bot, itemProto))
                return ITEM_USAGE_EQUIP;
          
            return ITEM_USAGE_BAD_EQUIP;
        }
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+        {
+            if (!oldItem || oldItemProto->ContainerSlots < itemProto->ContainerSlots)
+            {
+                return ITEM_USAGE_EQUIP;
+            }
+            else
+            {
+                return ITEM_USAGE_NONE;
+            }
+        }
```
SmashingQuasar, line 1219: This can be simplified by using an early return:
```cpp
        if (itemProto->Class == ITEM_CLASS_QUIVER)
        {
            if (!oldItem || oldItemProto->ContainerSlots < itemProto->ContainerSlots)
                return ITEM_USAGE_EQUIP;

            return ITEM_USAGE_NONE;
        }
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+        if (!sRandomItemMgr->CanEquipForBot(bot, oldItemProto))
+            existingShouldEquip = false;
+
+        // uint32 oldItemPower = sRandomItemMgr->GetLiveStatWeight(bot, oldItemProto->ItemId);
+        // uint32 newItemPower = sRandomItemMgr->GetLiveStatWeight(bot, itemProto->ItemId);
+
+        // Compare items based on item level, quality or itemId.
+        bool isBetter = false;
+        if (itemScore > oldScore)
+            isBetter = true;
```
SmashingQuasar: This can be simplified as:
```cpp
const bool isBetter = itemScore > oldScore;
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+                        // return vals will result in an attempted equip action so it wouldn't have much effect currently
+                        if (itemIsBroken && !oldItemIsBroken)
+                            return ITEM_USAGE_BROKEN_EQUIP;
+                        else if (shouldEquipInSlot)
+                            return ITEM_USAGE_REPLACE;
+                        else
+                            return ITEM_USAGE_BAD_EQUIP;
+
+                        break;
+                    }
```
SmashingQuasar, line 1255: This can be simplified as follows:
```cpp
                    if (oldItemProto->SubClass <= itemProto->SubClass)
                    {
                        // Need to add some logic to check second slot before returning, but as it happens, all three of these
                        // return vals will result in an attempted equip action so it wouldn't have much effect currently
                        if (itemIsBroken && !oldItemIsBroken)
                            return ITEM_USAGE_BROKEN_EQUIP;
                        if (shouldEquipInSlot)
                            return ITEM_USAGE_REPLACE;
                        
                        return ITEM_USAGE_BAD_EQUIP;
                    }
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+                default:
+                {
+                    if (itemIsBroken && !oldItemIsBroken)
+                        return ITEM_USAGE_BROKEN_EQUIP;
+                    else if (shouldEquipInSlot)
+                        return ITEM_USAGE_EQUIP;
+                    else
+                        return ITEM_USAGE_BAD_EQUIP;
+                }
```
SmashingQuasar, line 1262: This can be simplified as follows:
```cpp
                default:
                {
                    if (itemIsBroken && !oldItemIsBroken)
                        return ITEM_USAGE_BROKEN_EQUIP;
                    if (shouldEquipInSlot)
                        return ITEM_USAGE_EQUIP;
                    
                    return ITEM_USAGE_BAD_EQUIP;
                }
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+                            return true; // Item is useful because it creates a required quest item
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    return false; // Item is not useful for any active quests
+}
```
SmashingQuasar, line 1417: You do not need this method at all, the core provides it for you.

Here is [an usage example](https://github.com/SmashingQuasar/mod-playerbots/blob/dev/SmashingQuasar/randombot-persistence/src/domain/player/inspector/global/GlobalPlayerInspector.h#L120):
```cpp
	bool requiresItemForActiveQuest(const uint32_t itemTemplateId) const
	{
		const Player* const player = this->getCurrentPlayer();

		// If the player is offline, we consider the item as being in use to avoid modifying an offline inventory.
		if (player == nullptr)
			return true;

		return player->HasQuestForItem(itemTemplateId);
	}
```

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -237,679 +339,2112 @@ ItemUsage ItemUsageValue::Calculate()
...
+        case 12709:
+            return botAI->HasSkill(SKILL_SKINNING);
+        case 19901:
+            return botAI->HasSkill(SKILL_SKINNING);
+        case 4471:  // Flint and Tinder
+            return botAI->HasSkill(SKILL_COOKING);
+        case 4470:  // Simple Wood
+            return botAI->HasSkill(SKILL_COOKING);
+        case 6256:  // Fishing Rod
+            return botAI->HasSkill(SKILL_FISHING);
```
SmashingQuasar: It would be nice to avoid an enum class instead of magic numbers. This way you can remove your comments and have them directly within the code for everyone to enjoy.

SmashingQuasar (commented):
I have done a first pass on the first half of the PR as requested by @hermensbas .

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: You’re right, but my PR is about a smart‑roll system for loot, not about refactoring ItemUsageValue. And as you can see, this is the original code not my code nor my commens. The day I have time to spend refactoring files, we’ll see.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Agreed on the early return; applied. Your snippet had a small typo (ParsedItemUsage parsed.itemId) which wouldn’t compile, but the intent is implemented.

src/Ai/Base/Actions/LootRollAction.cpp
Wishmaster117: Done

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Thanks, agreed on the concern. The original “preferred armor tier” logic can indeed be too strict during leveling (especially with classic/BC itemization), and may lead to odd prioritisation for healer specs.

I’ve adjusted the cross-armor handling to be more permissive while leveling (only when it’s a clear upgrade based on the stat-weight score / margin, and respecting the “primary armor user in group” checks), but kept it strict at level cap and in raid context to avoid endgame etiquette issues (e.g. plate/mail users needing on cloth/leather in raids).

So the armor tier now behaves as a leveling heuristic rather than a hard endgame rule, while still preventing “bizarre” raid loot behaviour.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Good point. I widened the caster detection to include CRIT/HASTE/HIT as caster-relevant stats (Holy Paladins in particular value spell crit), but I kept it safe by only treating “rating-only” items as caster when there are no physical primary signals (STR/AGI/AP/ARP). This avoids misclassifying typical physical gear that also has crit/haste.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Hi yes that not work for ptot paladin and it's what i want.
A Protection Paladin in WotLK doesn’t need to look for Spell Power weapons.
All our main abilities scale automatically from Attack Power, which comes from Strength, so tank weapons with Strength and defensive stats give better threat and survivability.

Spell Power weapons were useful in TBC, but in 3.3.5 they’re no longer optimal for Prot Paladins.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Off-hands (HOLDABLE / SHIELD) are ITEM_CLASS_ARMOR in AzerothCore/WotLK, so they don’t hit that proto->Class != ITEM_CLASS_ARMOR early return. That branch mainly covers non-armor equipables like bags/quivers, and IsPrimaryForSpec() is only consulted when the item was already evaluated as EQUIP/REPLACE.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Good catch , dest is now initialized (set to 0) before calling CanEquipItem.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Done

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: done

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: The scope of this PR is to add an intelligent loot‑rolling system for bots, not to refactor the ItemUsageValue.cpp file.
This code is the file’s original implementation; indeed, a lot of it could be simplified, but that is not within the scope of this PR for now.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Or like that:
```
                default:
                {
                    if (itemIsBroken && !oldItemIsBroken)
                        return ITEM_USAGE_BROKEN_EQUIP;

                    return shouldEquipInSlot ? ITEM_USAGE_EQUIP : ITEM_USAGE_BAD_EQUIP;
                }
```
But as i said this is the original code and i'm not refactoring this file just implementing a smart roll loot system

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: You’re absolutely right, but I’ll give you the same answer as for the others below: this is not within the scope of this PR.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Changed but same answer as the other, this is the original code

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Modified and same comment, old code

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: changed, but it's the old code don't touched it

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Done

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: You are right, corrected.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Split into two helpers, no functional change.

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Refactored for readability only, no behavior change

src/Ai/Base/Actions/LootRollAction.cpp
Wishmaster117: The sign inversion is intentional: we encode the “random enchant” into a single int32 where >0 represents a RandomPropertyId and <0 represents a RandomSuffixId (stored as negative). This matches how downstream code (notably StatsWeightCalculator::CalculateRandomProperty) distinguishes properties vs suffixes (LookupEntry(-id) for suffixes). I added inline documentation to make this convention explicit.

src/Ai/Base/Actions/LootRollAction.cpp
Wishmaster117: Yes, that’s intended. This action is scheduled under the “very often” trigger as a lightweight poll for pending group rolls. Returning false here simply means “no vote was sent / nothing to process”, and it allows other actions to run; it doesn’t cause an internal retry loop within the same tick. We return true only when we actually process and send a roll vote (one per tick).

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: But ok i'll do it

src/Ai/Base/Value/ItemUsageValue.cpp
Wishmaster117: Just to confirm: the intent was to rely on the core helper for quest-required items, so I replaced the manual quest-log scan with Player::HasQuestForItem(proto->ItemId).
I kept a small extra check for items that create a quest-required item (SPELL_EFFECT_CREATE_ITEM), but still delegate the actual “is it needed / do we have enough already” logic to HasQuestForItem(createdItemId).
Does this match what you had in mind?

```
bool ItemUsageValue::IsItemUsefulForQuest(Player* player, ItemTemplate const* proto)
{
    if (!player || !proto)
        return false;

    // Core helper: checks active quest requirements and counts (quest log).
    if (player->HasQuestForItem(proto->ItemId))
        return true;

    // Extra heuristic: items that create a quest-required item (e.g. "Create Item" effects).
    // We delegate requirement/count checks to the same core helper.
    for (uint8 i = 0; i < MAX_ITEM_SPELLS; ++i)
    {
        uint32 const spellId = proto->Spells[i].SpellId;
        if (!spellId)
            continue;

        SpellInfo const* spellInfo = sSpellMgr->GetSpellInfo(spellId);
        if (!spellInfo)
            continue;

        for (uint8 eff = 0; eff < MAX_SPELL_EFFECTS; ++eff)
        {
            if (spellInfo->Effects[eff].Effect != SPELL_EFFECT_CREATE_ITEM)
                continue;

            uint32 const createdItemId = spellInfo->Effects[eff].ItemType;
            if (createdItemId && player->HasQuestForItem(createdItemId))
                return true;
        }
    }

    return false;
}
```

Ben19882:
This is not compiling anymore because of changes in the mod.

mod-playerbots/src/Ai/Base/Value/ItemUsageValue.cpp:2270:28: fatal error: no member named 'lootRollLevel' in 'PlayerbotAIConfig'; did you mean 'lootNeedRollLevel'?
 2270 |     if (sPlayerbotAIConfig.lootRollLevel == 0)
      |                            ^~~~~~~~~~~~~
      |                            lootNeedRollLevel
      
      https://github.com/mod-playerbots/mod-playerbots/commit/2ce89939860740d1731589ec7b5a43bacfe3be5b

Wishmaster117:
> This is not compiling anymore because of changes in the mod.
> 
> mod-playerbots/src/Ai/Base/Value/ItemUsageValue.cpp:2270:28: fatal error: no member named 'lootRollLevel' in 'PlayerbotAIConfig'; did you mean 'lootNeedRollLevel'? 2270 | if (sPlayerbotAIConfig.lootRollLevel == 0) | ^~~~~~~~~~~~~ | lootNeedRollLevel
> 
> ```
>   https://github.com/mod-playerbots/mod-playerbots/commit/2ce89939860740d1731589ec7b5a43bacfe3be5b
> ```

I'll try to take a look when i get some time

PR #2025, opened 2026-01-17
[FIX] Finalized structure! (do not start fixing PR merge structure conflict till this is merged)

hermensbas:
Finalized


hermensbas:
Done just waiting for the window green build, since my buildserver is one linux

PR #2027, opened 2026-01-17
Removed the expansion folder from dungeons

hermensbas:
In order to make consistent with raids but also to shorten max used length directory for windows builds


hermensbas:
No need to wait for build, build alrdy tested.

PR #2028, opened 2026-01-17
[CHORE] Util classes format and simple cleanup with generated resources

hermensbas:
- Did some basic formatting
- Some generated docs
- Cleaned header/impl Helper.css
- Moved PerfMonitor from util to bot/handler/command

Still a freaking mess though, but its a start i guess. Cant ask ppl to add more or make use of those when its so messy.


hermensbas:
No code changes, no logic changes, added docs, corrected head. No need for additional build check in this case.

As always tested locally even though there is no need, but just to be consistent.

PR #2032, opened 2026-01-18
[Chore] Added debug map for perfMonitor

hermensbas:


hermensbas:
No need to wait for build, locally tested only folder change with no references.

Regrad:
> No need to wait for build, locally tested only folder change with no references.

I finished working on my BG/Arena patch, but I couldn't transfer it to the new source version. After your recent file rearrangement patches, I can't compile the source code on Windows. Too huge length of the command line for the RC task
image

PR #2033, opened 2026-01-19
Normalize playerbots command input to enforce lowercase invariant

notOrrytrout:
Fix issues with playerbots command handling.  Implemented missing input normalization.


notOrrytrout:
ill fix and resubmit

PR #2041, opened 2026-01-20
Bear Alternative form for low level cat strat

Celandriel:
This adds the bear form as an alternative form for druids with the cat strategies. This applies to low level (<20) druids that have the cat strategy applied. Reset botAI would not normally give them the strat, because they dont have cat form, but if set, this allows more dps.


notOrrytrout:
This kind of stuff is so needed.

Celandriel:
Credit goes to Grimfeather for noticing it.

hermensbas (approved):

PR #2045, opened 2026-01-22
Correct comment for AutoEquipUpgradeLoot in config

brighton-chi:
The setting applies only to altbots. Rndbots always auto equip upgrades.

```
bool EquipUpgradesAction::Execute(Event event)
{
    if (!sPlayerbotAIConfig->autoEquipUpgradeLoot && !sRandomPlayerbotMgr->IsRandomBot(bot))
        return false;

```

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -3,8 +3,9 @@
 ##################################################
 
 # Overview
-# "Randombot": randomly generated bots that log in separately from players and populate the world. Depending on settings, randombots may automatically grind, quest, and upgrade equipment and can be invited to groups and given commands.
-# "Altbot": characters created on player accounts, which may be logged in by the player and invited to groups and given commands like randombots. Depending on settings, altbots can be limited to characters on the active player's account or in the active player's guild.
+# "Randombot": randomly generated bots that log in separately from players and populate the world. Randombots may automatically grind, quest, level up, and upgrade equipment and can be invited to groups and given commands.
+# "AddClass bot": bots from the AddClassAccountPoolSize accounts. They are used for quickly adding a leveled and geared bot of any class to your party. They are recommended for a quick formation of a party.
+# "Altbot": characters created on player accounts, which may be logged in by the player and invited to groups and given commands like randombots. They are best suited for long-progression playthroughs.
 # Information about commands to control bots and set their strategies can be found on the wiki at https://github.com/mod-playerbots/mod-playerbots/wiki/Playerbot-Commands.
 
 ####################################################################################################


Celandriel:
Technically addClass as well

brighton-chi:
Pfft nobody cares about AddClass bots.

Actually, weirdly, people care about them a lot more than I'd expect.

NoxMax:
I'm one of those people who loves addclass. They're so quick and easy for experiments and whatnot.

Celandriel:
That's it. Gonna nukem

brighton-chi:
> I'm one of those people who loves addclass. They're so quick and easy for experiments and whatnot.

Oh trust me, you were the first one to come to mind.

So who wants to add to this PR a description of addclass bots to the top of the config where we have altbots and randombots?

```
# "Randombot": randomly generated bots that log in separately from players and populate the world. Depending on settings, randombots may automatically grind, quest, and upgrade equipment and can be invited to groups and given commands.
# "Altbot": characters created on player accounts, which may be logged in by the player and invited to groups and given commands like randombots. Depending on settings, altbots can be limited to characters on the active player's account or in the active player's guild.
# "Addclass bot": some weird thing for weird people <- help?
```

Celandriel (approved):

brighton-chi:
OK, after talking to Kel, I've gone the opposite direction. There's only one existing reference to addclass bots as a standalone concept, and I'm proposing we remove it. If we really wanted to be comprehensive, we'd have to insert references to them everywhere. Since they're more of an advanced, experimental feature (and not really for creating a raid on the spot like it might seem), perhaps the better approach is not to describe them as another category of bots in the first place.

NoxMax:
They are not only distinct, but they also have their own distinct accounts. I would not describe them as an advanced feature, that would imply the need for some technical configuration to utilize them. Quite the contrary, they are meant to be something that's quick and easy. I am very much against removing them from the project.

> "Addclass bot": some weird thing for weird people <- help?

Weird people of the world unite!
But honestly, why is addclass still viewed as "experimental" at this stage? It's old and stable at this point. It's not as configurable as altbots, but that doesn't make it experimental. But they can be used to run experiments of course.

brighton-chi:
The current state of things is that they aren't really described at all. So I see three options:
1. Add them to the config with a description at the top as a third kind of bot and then include them throughout the config where applicable (this would obviously be a bit of a project for somebody).
2. Add them to the config with a description at the top but describe them as an rndbot added by a command, with a few words about how they might differ.
3. Delete the one independent reference to them in the config (randombot trading) because it simply doesn't make any sense to have that but nothing else in the document referencing them with respect to features. I don't think we should remove anything regarding the baseline addclass bot system (i.e., things like enabling them or the pool size).

Or I guess a combination of 1 and 2 works too.

NoxMax:
Yeah I think properly documenting them in the conf makes perfect sense. They weren't originally, because yes, they did use to be experimental and often unpredictable in what you get when there was account mixing. But I and other contributors have streamlined their workings since then.

[The wiki right now describes them as](https://github.com/mod-playerbots/mod-playerbots/wiki/Playerbot-Commands)

> AddClass bots utilize unused Rndbots in your system. They are currently recommended only to be used for testing purposes i.e. quickly forming a raid group to test certain bots in a given raid, whereas Altbots are strongly recommended for long-term progression playthroughs.

I think the conf should define them, and the wiki should be changed to

> AddClass bots utilize bots from the AddClassAccountPoolSize accounts. They are used for quickly adding a leveled and geared bot of any class to your party. They are recommended for a quick formation of a party, whereas the more configurable Altbots are more suited for long-term progression playthroughs.

brighton-chi:
I added them to the top. If this works, then perhaps an addclass bot enthusiastic can plug them into the rest of the .dist if desired.

NoxMax:
Perfectly concise. Longer description is probably best for the wiki.

Wishmaster117:
> Pfft nobody cares about AddClass bots.
> 
> Actually, weirdly, people care about them a lot more than I'd expect.

Don't get that, addclassbots are autoequip too.
Btw, i love addclassbots and i'll make to them a new feature

biship:
I use AddClass all the time for testing dungeons. Especially for testing the better smarter looting PRs.

Celandriel:
Yea, we were joking about nuking them, they are an integral part of the experience at this point.

notOrrytrout:
> They are not only distinct, but they also have their own distinct accounts. I would not describe them as an advanced feature, that would imply the need for some technical configuration to utilize them. Quite the contrary, they are meant to be something that's quick and easy. I am very much against removing them from the project.
> 
> > "Addclass bot": some weird thing for weird people <- help?
> 
> Weird people of the world unite! But honestly, why is addclass still viewed as "experimental" at this stage? It's old and stable at this point. It's not as configurable as altbots, but that doesn't make it experimental. But they can be used to run experiments of course.

We can call is something else if addclass hurts too much.

PR #2049, opened 2026-01-23
Summon Logic Tweaks

dillyns:
Issues:
- When you have selfbot enabled and use summon command, you will summon yourself. This causes odd movement if you summon while moving, and can sometimes lead to falling through the floor.
- When using the summon command on bots with pets/guardians from a medium distance (like jumping down a ledge then commanding summon), the pets will pathfind run to catch up. This causes them to aggro everything on the way.

Solution: 
Fix summon logic to prevent selfbot summon and ensure pets are teleported with bots.

### DIFF
--- a/src/Ai/Base/Actions/UseMeetingStoneAction.cpp
+++ b/src/Ai/Base/Actions/UseMeetingStoneAction.cpp
@@ -217,7 +217,6 @@ bool SummonAction::Teleport(Player* summoner, Player* player, bool preserveAuras
                     player->GetPet()->NearTeleportTo(x, y, z, player->GetOrientation());
                 if (player->GetGuardianPet())
                     player->GetGuardianPet()->NearTeleportTo(x, y, z, player->GetOrientation());
-                    
                 if (botAI->HasStrategy("stay", botAI->GetState()))
                 {
                     PositionMap& posMap = AI_VALUE(PositionMap&, "position");


src/Ai/Base/Actions/UseMeetingStoneAction.cpp
```diff
@@ -214,6 +214,14 @@ bool SummonAction::Teleport(Player* summoner, Player* player, bool preserveAuras
                                                           AURA_INTERRUPT_FLAG_CHANGE_MAP);
 
                 player->TeleportTo(mapId, x, y, z, 0);
+                if (player->GetPet())
+                {
```
noisiver: There shouldn't be curly brackets for single line ifs

src/Ai/Base/Actions/UseMeetingStoneAction.cpp
```diff
@@ -214,6 +214,14 @@ bool SummonAction::Teleport(Player* summoner, Player* player, bool preserveAuras
...
+                    player->GetPet()->NearTeleportTo(x, y, z, player->GetOrientation());
+                }
+                if (player->GetGuardianPet())
+                {
```
noisiver: There shouldn't be curly brackets for single line ifs

hermensbas:
Did you test it? especially around the pet summons

dillyns:
> Did you test it? especially around the pet summons

I have been using these changes in my local server for many months now and have never seen any issue. Bot pets get summoned along with the bots instead of running through dungeons trying to catch up.

brighton-chi:
Delayed +1 from me on this one. Small tweak, huge improvement. Love this kind of stuff.

PR #2056, opened 2026-01-23
Added some additional defense checks around isHostile and unit/target

hermensbas:
Needs second pair of eyes, they appear in crash logs here and there. Its merely a patch on a open wound.

----
As in aslong there multithreads in mapupdate, which we need for decent performance and core calls are not done correctly due various reasons. These type of issues remain.

Although i am planning to experiment a little with threadsafe execution of our strategies vs performance.

The most effective thing we could do is check every single action and check its stateless and where it does effect the state or read the state of a core object its done in the safest way. flags, worldthread where possible and/ot simply taking into account the state might be invalid.


hermensbas:
i will monitor the pr

PR #2057, opened 2026-01-23
DK Non Combat Strategy Fix

dillyns:
### Problem
The generic DK non-combat strategy has all DKs trying to use Raise Dead when out of combat. 
Blood and Frost DKs should NOT be doing this, as they do not have unlimited duration ghouls. They should only use Raise Dead in combat like other dps cooldowns. 
This issue is especially obvious when using the Minor Glyph of Raise Dead.

### Overarching Generic Noncombat Problems
This type of generic non combat problem exists in different ways for other classes too, most obvious in tank and dps classes that can heal.
Examples:
- Tank druid leaves bear form after every pull to heal the party, eventually running out of mana and not being able to return to bear form.
- Prot paladin heals the party in between pulls and goes out of mana. Especially annoying at level 60 without all the mana regen skills available yet.

### Possible Solution

Add Unholy Death Knight non-combat strat and remove Raise Dead from Generic DK non-combat strat

**This example PR showcases a working fix for DKs only in order to keep it succinct, but the intention is to get some feedback here on how to solve this issue in a way that can be eventually applied to other classes with this problem.**


src/Ai/Class/Dk/DKAiObjectContext.cpp
```diff
@@ -51,6 +51,18 @@ class DeathKnightCombatStrategyFactoryInternal : public NamedObjectContext<Strat
     static Strategy* blood(PlayerbotAI* botAI) { return new BloodDKStrategy(botAI); }
 };
 
+class DeathKnightNonCombatStrategyFactoryInternal : public NamedObjectContext<Strategy>
+{
+public:
+    DeathKnightNonCombatStrategyFactoryInternal() : NamedObjectContext<Strategy>(false, true)
+    {
+        creators["unholy"] = &DeathKnightNonCombatStrategyFactoryInternal::unholy;
```
dillyns: if blood or frost had specific non-combat strategies, they could be added here.
Only one would be in-use at a time, the same as how the combat strategies work.

src/Ai/Class/Dk/Strategy/UnholyDKStrategy.cpp
```diff
@@ -170,6 +170,16 @@ void UnholyDKStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
     );
 }
 
+void UnholyDKNonCombatStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
+{
```
kadeshar, line 174: Missing base class call GenericDKNonCombatStrategy::InitTriggers(triggers);

dillyns: It doesn't override GenericDKNonCombatStrategy

kadeshar: Its wrong pattern. If something in GenericDKNonCombatStrategy dont fit to all spec then shouldnt be there.

dillyns: I don't really get what you are trying to say. This PR does exactly what you said, it removes Raise Dead from generic DK strategy because it doesn't fit all specs, only Unholy.
The problem is that I did not see any existing pattern for how to add spec-specific non combat behavior, other than checking the talent trees (which is imo not a good idea).

kadeshar: If exists class non combat strategy then spec non combat strategy should inherit from their creating whole chain
spec -> class -> base

dillyns: Is the suggestion to make the spec specfic strategy exclusive with "nc", which is where are the base class triggers are already being added? Otherwise wouldn't the base class InitTriggers get called twice, once by "nc" and once by the spec-specific override

kadeshar: Whole repository is based on such concept. Example:
FireMageStrategy -> GenericMageStrategy -> RangedCombatStrategy -> CombatStrategy -> Strategy
I will ask another mod to check this PR to another PoV because im not expert :)

hermensbas (dismissed):

src/Ai/Class/Dk/Strategy/UnholyDKStrategy.cpp
Celandriel: I think the problem here is that by default non-combat strategy 'nc' is points to GenericDKNonCombatStrategy. This adds spec specific nc strategy. as 'unholy nc' since I dont think there is a way to share the nc registration between specs right now. 

So either we can set up a shared "nc" strategy where only unholy gets the pet triggers (in which case you need the     GenericDKNonCombatStrategy::InitTriggers(triggers) ), or you create a "unholy nc" which DOES NOT require calling the InitTriggers.

PR #2068, opened 2026-01-25
Fix LootRollLevel=1 to match documented 'greed' behavior

Hokken:
## Summary

Fixes `AiPlayerbot.LootRollLevel = 1` to actually behave as "greed" mode per the config documentation.

## Problem

The config documentation states:
```conf
# Bots' loot roll level (0 = pass, 1 = greed, 2 = need)
# Default: 1 (greed)
AiPlayerbot.LootRollLevel = 1
```

However, level 1 was converting **all GREED votes to PASS**, causing bots to pass on almost everything:

| Item Type | AI Decision | Level 1 Behavior (Before) | Expected |
|-----------|-------------|---------------------------|----------|
| Gear upgrade | NEED | GREED ✓ | GREED |
| Usable gear (not upgrade) | GREED | **PASS** ✗ | GREED |
| Crafting materials | GREED | **PASS** ✗ | GREED |
| Recipes, consumables | GREED | **PASS** ✗ | GREED |

The only items bots would greed on were direct gear upgrades (originally NEED, downgraded to GREED).

## Root Cause

In `LootRollAction.cpp`, lines 104-107 were converting GREED to PASS:

```cpp
else if (vote == GREED)
{
    vote = PASS;  // This breaks "greed" mode
}
```

## Fix

Remove the GREED→PASS conversion. Level 1 now only downgrades NEED to GREED (as intended), preserving GREED votes for useful items.

## Behavior After Fix

| Level | Description | Behavior |
|-------|-------------|----------|
| 0 | Pass | Always pass on all items |
| 1 | Greed | Greed on useful items, never need |
| 2 | Need | Full AI logic (need/greed/pass) |

## Test Plan

- [ ] Set `AiPlayerbot.LootRollLevel = 1`
- [ ] Kill mobs that drop crafting materials, recipes, or non-upgrade gear
- [ ] Verify bots greed on useful items instead of passing
- [ ] Verify bots still pass on junk items
- [ ] Verify bots never roll need (only greed)


kadeshar (dismissed):

Sakuragi85:
I think this was a bad change. now all bots are greeding on all raid drops instead. cery hard to distribute raid drops to bots now.

Hokken:
Just change the value in the config so they will pass.

Sakuragi85:
no becaue then eveyone passes. i want the bots who need to greed, and everyone who doesnt need to pass

Hokken:
"I want the bots who need to greed", greed and need are 2 different things, that's why there is two different button on the loot window. My changes just respect the original wow logic. Now if bots don't need when they should then a new fix is needed elsewhere but that's unrelated to mine. I will investigate.

Hokken:
I did a quick investigation and I confirm that my fix is correct, the issue you are seeing is that the AI for the greed behavior is too broad. The possible fixes are:

1. Tighten CalculateRollVote() to be more selective about what warrants a GREED in raid contexts
2. Add a new LootRollLevel option (e.g., level 3) for “smart” behavior where bots only need on upgrades and pass on everything else.

I will test those solutions and see what work best for a natural realistic immersion.

In any case the current greed behavior is too systematic, real players are not systematically greedy.

PR #2069, opened 2026-01-26
Refactor raid strategy framework

brighton-chi:
# Pull Request

The purposes of this PR are to (1) establish a general raid helper framework for the benefit of future raid strategies and (2) make some improvements to problematic areas of the raid strategy code.

List of changes:
1. Added new RaidBossHelpers.cpp and RaidBossHelpers.h files in the Raid folder.

3. Moved reused helpers from Karazhan, Gruul, and Magtheridon strategies to the new helper files.
4. Modified the prior function that assigned a DPS bot to store and erase timers and trackers in associative containers--the function now includes parameters for mapId (so a bot that is not in the instance will not be assigned) and for the ability to exclude a bot (useful for excluding particular important roles, such as a Warlock tank, so they are not bogged down by these extra tasks at critical moments). I also renamed it from IsInstanceTimerManager to IsMechanicTrackerBot.
5. Moved all helper files in raid strategies to Util folders (was needed for ICC, MC, and Ulduar).
6. Renamed and reordered includes of Ulduar files in AiObjectContext.cpp to match other raid strategies.
a. This initially caused compile errors which made me realize that the existing code had several problems with missing includes and was compiling only due to the prior ordering in AiObjectContext.cpp. Therefore, I added the missing includes to Molten Core, Ulduar, and Vault of Archavon strategies.
b. Ulduar and Old Kingdom were also using the same constant name for a spell--the reordering caused a compile error here as well, which just highlighted an existing problem that was being hidden. I renamed the constant for Ulduar to fix this, but I think the better approach going forward would be to use a namespace or enum class. But that is for another time and probably another person.
7. Several changes with respect to Ulduar files:
a. The position constants and enums for spells and NPCs and such were in the trigger header file. I did not think that made sense so moved them to existing helper files.
b. Since the strategy does not use multipliers, I removed all files and references to multipliers in it.
c. I removed some unneeded includes. I did not do a detailed review to determine what else could be removed--I just took some out that I could tell right away were not needed.
d. I renamed the ingame strategy name from "uld" to "ulduar," which I think is clearer and is still plenty short.
8. Partial refactor of Gruul and Magtheridon strategies:
a. I did not due a full refactoring but made some quick changes to things I did previously that were rather stupid like repeating calculations, having useless logic like pointless IsAlive() checks for creatures already on the hostile references list, and not using the existing Position class for coordinates.
b. There were a few substantive changes, such as allowing players to pick Maulgar mage and moonkin tanks with the assistant flag, but a greater refactoring of the strategies themselves is beyond this PR.
c. I was clearing some containers used for Gruul and Magtheridon strategies; the methods are now fixed to erase only the applicable keys so that in the unlikely event that one server has multiple groups running Gruul or Magtheridon at the same time, there won't be timer or position tracker conflicts.

## How to Test the Changes

1. Enter any raid instance that has any code impacted by this PR
2. Engage bosses and observe if any strategies are now broken

I personally tested Maulgar, Gruul, and Magtheridon and confirmed that they still work as intended. 

## Complexity & Impact

I do not expect this PR to have any relevant changes to in-game performance, but I will defer to those more knowledgeable than I if there are concerns in this area. As I've mentioned before, you can consider me to be like a person who has taken half an intro C++ course at best.

## AI Assistance

None beyond autocomplete of repetitive changes.

### DIFF
--- a/src/Ai/Raid/RaidBossHelpers.cpp
+++ b/src/Ai/Raid/RaidBossHelpers.cpp
@@ -94,18 +94,22 @@ bool IsMechanicTrackerBot(PlayerbotAI* botAI, Player* bot, uint32 mapId, Player*
     return false;
 }
 
-// Return the first matching alive, in-combat hostile unit from a cell search of nearby units
+// Return the first matching alive unit from a cell search of nearby npcs
 // More responsive than "find target," but performance cost is much higher
-// Depending on usage, other filtering criteria may be more appropriate (e.g., "possible targets")
-Unit* GetFirstAliveUnitByEntry(PlayerbotAI* botAI, uint32 entry)
+// Re: using the third parameter (false by default), some units are never considered
+// to be in combat (e.g., totems)
+Unit* GetFirstAliveUnitByEntry(PlayerbotAI* botAI, uint32 entry, bool requireInCombat)
 {
     auto const& npcs =
-        botAI->GetAiObjectContext()->GetValue<GuidVector>("nearest hostile npcs")->Get();
+        botAI->GetAiObjectContext()->GetValue<GuidVector>("nearest npcs")->Get();
     for (auto const& npcGuid : npcs)
     {
         Unit* unit = botAI->GetUnit(npcGuid);
-        if (unit && unit->IsAlive() && unit->IsInCombat() && unit->GetEntry() == entry)
-            return unit;
+        if (unit && unit->IsAlive() && unit->GetEntry() == entry)
+        {
+            if (!requireInCombat || unit->IsInCombat())
+                return unit;
+        }
     }
 
     return nullptr;


src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
```diff
@@ -0,0 +1,341 @@
...
+    // A map to track the last role swap *per bot* by their GUID
+    static std::unordered_map<ObjectGuid, std::time_t> _lastRoleSwapTime;
+
+    // The cooldown that applies to every bot
+    static const std::time_t _roleSwapCooldown = 10;
+
+    static std::unordered_map<ObjectGuid, time_t> _harpoonCooldowns;
+};
+
+// template <class BossAiType>
```
kadeshar, line 278: What is purpose of this commented code?

src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
```diff
@@ -0,0 +1,341 @@
...
+#include "AiObject.h"
+#include "AiObjectContext.h"
+#include "EventMap.h"
+#include "ObjectGuid.h"
+#include "Player.h"
+#include "PlayerbotAI.h"
+#include "Playerbots.h"
+#include "ScriptedCreature.h"
+
+const uint32 ULDUAR_MAP_ID = 603;
```
kadeshar: Image
I think should be the same type

src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
brighton-chi: I don't know; didn't you write that? I did not mess with the content of any of the Ulduar code. I just addressed some compile issues, move some code between files, and deleted some includes. Maybe another PR can do a more comprehensive review of the Ulduar files.

src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
brighton-chi: Yeah. I can change it. I just didn't mess with the content of the Ulduar code. There are tons of constants throughout the module that should technically be constexpr so there's work that could be done on that point for sure.

src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
kadeshar: I see only addition in changes. I dont see deletion (or im blind what is also possible)

src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
kadeshar: Now when you creating some structure on that best unified that to make good example for others :)

src/Ai/Raid/Ulduar/Util/RaidUlduarBossHelper.h
brighton-chi: I moved the file inside the util folder, so it's showing up as the whole file was deleted and recreated.

kadeshar (dismissed):

hermensbas (dismissed):

kadeshar:
Require conflict resolve

brighton-chi:
> Require conflict resolve

Seems it was just the deletion of the Ulduar multiplier files it was calling a conflict, I don't know why that is. Anyway, it should be OK now.

brighton-chi:
What needs to be tested? Ulduar? If so, maybe we should try on Discord because I have not played any WotLK content and have now idea how the fights should work (also I don't want to spoil the content for myself).

brighton-chi:
Sorry to dismiss reviews. I was running Karazhan yesterday, and the Illhoof marking was not working. Kil'rek and Demon Chains were not getting picked up so they could not be marked since the bot would repeatedly mark Illhoof (which is very bad).

This wasn't a problem before. But I do notice issues sometimes with bot responsiveness when using "find target." I switched to using a direct entry lookup via "nearest hostile npcs," which solved the issue. I am going to test completely changing a raid strategy to rely on constants for every type of NPC reference, except for when a combat check is explicitly necessary (e.g., for the trigger), in which case I'd still use "find target." But anyway, that is outside the scope of this PR. I just wanted to fix the non-working strategy right now.

PR #2079, opened 2026-01-29
Refactor newrpginfo data union to std::variant

Celandriel:
# Pull Request

As I began modifying the newrpginfo to change the types of data it stored, or add new data I found myself with the issue of ending up either with garbage memory if the information wasnt properly stored on status change, or needing complicated destructor patterns for non trivial data sets. 

---

## Design Philosophy

Make rpginfo able to handle more complicated information in a strongly 

---

## Feature Evaluation

No Feature changes

---

## How to Test the Changes

-  Server should be stable for an extended period of time. 
- Bots should be able to complete quests, fly, etc as they did before.

## Complexity & Impact

    - [X ] No
    - [ ] Yes (**explain below**)

    - [ ] No
    - [ X] Yes (**describe and justify impact**)
Potentially as there can be more memory involved in the object.

    - [X ] No
    - [ ] Yes (**explain why**)

---

## Defaults & Configuration

    - [ X] No
    - [ ] Yes (**explain why**)

---

## AI Assistance

    - [ ] No
    - [ X] Yes (**explain below**)

If yes, please specify:

- Gemini suggested the use of std::variant as an alternative data structure. I found additinal external references that correlated with the same suggestion of moving away from a union.
- Implementation was performed manually with Co-pilot auto-complete

---

## Final Checklist 
In progress.
- [ ] Stability is not compromised
- [ ] Performance impact is understood, tested, and acceptable
- [ ] Added logic complexity is justified and explained
- [ ] Documentation updated if needed

---

## Notes for Reviewers

Im not 100% sure if this is a good design choice. There are some things I didnt quite like by the end of this, specifically having to double check whenever accessing data whether exists or not even though an action has already been triggered. But I have a PR in the works where I want to store a full flight path vector, and the union was giving me issues. (It appears that state changes may be occuring in the same tick between RPG status update and the stated action, leading to incorrect data gathering. 

I ended up solving it by first checking a pointer to the object, and then getting the reference. 
```c++
    auto* dataPtr = std::get_if<NewRpgInfo::DoQuest>(&info.data);
    if (!dataPtr)
        return false;
    auto& data = *dataPtr;
```

### DIFF
--- a/src/Ai/World/Rpg/Action/NewRpgAction.cpp
+++ b/src/Ai/World/Rpg/Action/NewRpgAction.cpp
@@ -430,7 +430,7 @@ bool NewRpgTravelFlightAction::Execute(Event event)
     if (!dataPtr)
         return false;
 
-        auto& data = *dataPtr;
+    auto& data = *dataPtr;
     if (bot->IsInFlight())
     {
         data.inFlight = true;


Celandriel:
Just to share how to review this. 
The core change is in Line 75 of NewRpgInfo.h moving from a union to a std::variant. All other changes are based on this one. 

The implications of this change is that now the status of  NewRpgInfo can be determined by std::visit. (See new GetStatus function in NewRpgInfo.cpp)

Now when calling on the data stored in info, you specify the variant type that you are looking for. 
Example:
`auto& data = std::get<NewRpgInfo::GoGrind>(info.data);`

If you are unsure what variant you are in, or if there is a risk that the status change after triggers are processed, you can use `std::get_if` to check first if the variant is there, and then use the data within it. 

This can happen because triggers are all processed first, therefore the RpgStatusUpdate  can run before the one of the RpgAction and change the info type, causing an already triggered GoCamp action to run, but finding an invalid data type. 

Its a little more verbose, but should be overall safer.

hermensbas:
It looks a bit silly going from pointer back to reference but i think this is valid pattern.

Celandriel:
I agree, I dont think its necessary to do that in this case. There shouldnt be any risk in it changing. Maybe I keep as pointer?

hermensbas:
> ```c++
>   auto* dataPtr = std::get_if<NewRpgInfo::DoQuest>(&info.data);
>     if (!dataPtr)
>         return false;
>     auto& data = *dataPtr;
> ```

This pattern is alright, looks silly but it isn't.

src/Ai/World/Rpg/Action/NewRpgAction.cpp
```diff
@@ -410,29 +425,33 @@ bool NewRpgDoQuestAction::DoCompletedQuest()
 
 bool NewRpgTravelFlightAction::Execute(Event event)
 {
+    NewRpgInfo& info = botAI->rpgInfo;
+    auto* dataPtr = std::get_if<NewRpgInfo::TravelFlight>(&info.data);
+    if (!dataPtr)
+        return false;
+
+        auto& data = *dataPtr;
```
kadeshar: 433 have additional tab

This code is repeated 3 times (diffrence is only type of info). Consider optimizing that.

src/Ai/World/Rpg/NewRpgInfo.cpp
```diff
@@ -82,58 +76,83 @@ void NewRpgInfo::SetMoveFarTo(WorldPosition pos)
...
-            out << "\npoiPos: " << do_quest.pos.GetMapId() << " " << do_quest.pos.GetPositionX() << " "
-                << do_quest.pos.GetPositionY() << " " << do_quest.pos.GetPositionZ();
-            out << "\nlastReachPOI: " << do_quest.lastReachPOI ? GetMSTimeDiffToNow(do_quest.lastReachPOI) : 0;
-            break;
-        case RPG_TRAVEL_FLIGHT:
+            out << "\nquestId: " << arg.questId;
+            out << "\nobjectiveIdx: " << arg.objectiveIdx;
+            out << "\npoiPos: " << arg.pos.GetMapId() << " " << arg.pos.GetPositionX() << " "
+                << arg.pos.GetPositionY() << " " << arg.pos.GetPositionZ();
+            out << "\nlastReachPOI: " << arg.lastReachPOI ? GetMSTimeDiffToNow(arg.lastReachPOI) : 0;
```
kadeshar: The << operator has higher precedence than the ternary ?: operator
https://en.cppreference.com/w/cpp/language/operator_precedence.html

Should be
out << "\nlastReachPOI: " << (arg.lastReachPOI ? GetMSTimeDiffToNow(arg.lastReachPOI) : 0);

src/Ai/World/Rpg/Action/NewRpgAction.cpp
Celandriel: I guess what Im not sure of is whether I should be using the pointer or use the reference. In 160 and 171 I just use the pointer directly since its a 1 line usage of the data, but for these I felt it more correct to go back to the reference.
Is it style? Function?

kadeshar: Ok, then only fix tab :)

PR #2086, opened 2026-01-30
Fix WSG graveyard camping by flag carrier

brighton-chi:
Quick fix for a very annoying error identified by SmashingQuasar. In WSG, bots will camp the opposing graveyard if up 2-0. This is supposed to exclude the flag carrier, but a logical error has resulted in the flag carrier being excluded for Alliance camping only, meaning the Horde flag carrier will camp the GY with the rest of the team if up 2-0 and thus refuse to end the game.

### DIFF
--- a/src/Ai/Base/Actions/BattleGroundTactics.cpp
+++ b/src/Ai/Base/Actions/BattleGroundTactics.cpp
@@ -2227,8 +2227,8 @@ bool BGTactics::selectObjective(bool reset)
             }
             // Graveyard Camping if in lead
             else if (!hasFlag && role < 8 &&
-                (team == TEAM_ALLIANCE && allianceScore == 2 && hordeScore == 0) ||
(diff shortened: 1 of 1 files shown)


Celandriel:
Is the point of camping to farm rep?

brighton-chi:
The only way to get WSG rep is to cap the flag. You could do it to farm honor, but it will be very limited. Camping is mainly to be an asshat and grief the other team with a premade.

Celandriel:
What a crap strat lol

brighton-chi:
If the flag carrier doesn't do it, it's probably fine, since presumably if your team locks down the other team at their GY, the flag carrier can easily end the game. Since the strat has the Alliance exclude the flag carrier, I assume it is supposed to exclude the flag carrier for Horde also (and in any case, intentional or not, it is terrible for the flag carrier to refuse to cap). The strat also excludes healers so I don't know how successful the camping would be anyway. I've never played Alliance so I didn't even have any idea this was a bug.

PR #2104, opened 2026-02-02
Fix rest of trainers' related stuff + codestyle changes and corrections

privatecore:
# Pull Request

* Fix the rest of the trainer-related functionality: list spells and learn (cast vs. direct learn) spells.
* Rewrite `TrainerAction`: split the logic between appropriate methods (`GetTarget`, `isUseful`, `isPossible`) instead of pushing everything inside a single `Execute` method.
* Change method definitions to remove unnecessary declarations and parameters overhead.
* Move the `Trainer` header into the implementation. Rewrite `RpgTrainTrigger` to fit the original logic and move all validation to `RpgTrainAction` (`isUseful` + `isPossible`).
* Implement "can train" context value calculation to use with `RpgTrainTrigger`.
* Update and optimize "train cost" context value calculation -- it should be much faster.
* Replace `AiPlayerbot.AutoTrainSpells` with `AiPlayerbot.AllowLearnTrainerSpells` and remove the "free" value behavior — please use `AiPlayerbot.BotCheats` if you want bots to learn trainer's spells for "free".
* Add `nullptr` checks wherever necessary (only inside targeted methods/functions).
* Make some codestyle changes and corrections based on the AC codestyle guide.

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## How to Test the Changes

Force bots to learn spells from trainers using the chat command `trainer learn` or `trainer learn <spellId>`. Bots should properly list available spells (`trainer` command) or learn them (based on configuration and command).

## Complexity & Impact

    - [x] No
    - [ ] Yes (**explain below**)

    - [x] No

    - [x] No
    - [ ] Yes (**explain why**)

---

## Defaults & Configuration

    - [x] No
    - [ ] Yes (**explain why**)

- [x] Lightweight mode remains the default

---

## AI Assistance

    - [x] No
    - [ ] Yes (**explain below**)

If yes, please specify:

---

## Final Checklist

---

## Notes for Reviewers

before merging.

### DIFF
--- a/src/Ai/Base/Actions/TrainerAction.cpp
+++ b/src/Ai/Base/Actions/TrainerAction.cpp
@@ -64,11 +64,7 @@ bool TrainerAction::isPossible()
         return false;
 
     if (trainer->GetSpells().empty())
-    {
-        // Not sure why we need to say anything, especially w/o master
-        botAI->TellError("No spells can be learned from this trainer");
         return false;
-    }
 
     return true;
 }


hermensbas (approved):

src/Ai/Base/Trigger/RpgTriggers.cpp
```diff
@@ -161,52 +161,19 @@ bool RpgRepairTrigger::IsActive()
...
-
-        uint32 cost = uint32(floor(spell.MoneyCost * fDiscountMod));
-
-        if (cost > AI_VALUE2(uint32, "free money for", (uint32)NeedMoneyFor::spells))
-            continue;
-
-        return true;
-    }
+    if (!AI_VALUE2(uint32, "free money for", (uint32)NeedMoneyFor::spells))
+        return false;
```
Wishmaster117, line 174: If autoTrainSpells is set to "free" (or the bot has the gold cheat), could AI_VALUE2("free money for", NeedMoneyFor::spells) still be 0 and therefore prevent the RPG train trigger from ever activating, even though training should be allowed without budget constraints?

src/Ai/Base/Actions/RpgSubActions.cpp
```diff
@@ -250,6 +255,65 @@ Event RpgSellAction::ActionEvent(Event event) { return Event("rpg action", "vend
...
+    {
+        Trainer::Spell const* trainerSpell = trainer->GetSpell(spell.SpellId);
+        if (!trainerSpell)
+            continue;
+
+        if (!trainer->CanTeachSpell(bot, trainerSpell))
+            continue;
+
+        if (currentGold < static_cast<uint32>(floor(trainerSpell->MoneyCost * reputationDiscount)))
+            continue;
```
Wishmaster117, line 302: Should RpgTrainAction::isPossible() consider the "free" auto-train mode (or gold cheat) and bypass the budget check, so that bots don’t refuse to train just because the “free money for spells” budget is 0?

src/Ai/Base/Trigger/RpgTriggers.cpp
privatecore: Hmm, so the current behavior is wrong in the case of an enabled gold cheat?

Wishmaster117: In FreeMoneyForValue::Calculate() we return 10000000 when botAI->HasCheat(BotCheatMask::gold), so AI_VALUE2("free money for", NeedMoneyFor::spells) shouldn’t be 0 and the trigger shouldn’t be blocked. Isn’t the real problematic case autoTrainSpells == "free" (no-budget training), where the value can still be 0 and prevent the trigger?

src/Ai/Base/Actions/RpgSubActions.cpp
privatecore: Dunno if it was designed like this, but `FreeMoneyForValue::Calculate` actually includes check for `HasCheat(BotCheatMask::gold)`. Also, the `sPlayerbotAIConfig.autoTrainSpells` config is only used inside `TrainerAction`.

Wishmaster117: Yep, agreed on the gold cheat part: FreeMoneyForValue::Calculate() already handles it (if (botAI->HasCheat(BotCheatMask::gold)) return 10000000;), so AI_VALUE2("free money for", NeedMoneyFor::spells) shouldn’t be 0 in that case.

My concern is the "free" auto-train mode: sPlayerbotAIConfig.autoTrainSpells is only respected inside TrainerAction::Learn(), but the PR introduces RPG gating earlier (e.g. RpgTrainTrigger::IsActive() and RpgTrainAction::isPossible()) based on "free money for spells".

Could autoTrainSpells == "free" still end up blocked by those budget checks before TrainerAction runs, even though training is intended to ignore cost in that mode?

privatecore: Yes, I see, but this is the current behavior without the PR's changes: if the bot does not have money to learn spells, it will not learn anything. Are you sure we need these changes (`autoTrainSpells`)?

privatecore: Prev `RpgTrainTrigger::IsActive` code:
```cpp
    ...

    for (auto& spell : trainer->GetSpells())
    {
        if (!trainer->CanTeachSpell(bot, trainer->GetSpell(spell.SpellId)))
            continue;

        uint32 cost = uint32(floor(spell.MoneyCost * fDiscountMod));

        if (cost > AI_VALUE2(uint32, "free money for", (uint32)NeedMoneyFor::spells))
            continue;

        return true;
    }

    return false;
```
If the bot does not have money/gold for at least one spell, the `RpgTrainTrigger::IsActive` check will return `false`. We can add an additional check against `autoTrainSpells` inside `RpgTrainAction::isPossible`, but this would introduce new behavior. Is that what we need?

privatecore:
I need to solve some problems with `AutoTrainSpells = "free"`. My current idea is to separate the availability to learn spells from their cost:
```cfg
#
#    AllowLearnTrainerSpells
#        Description: Allow the bot to learn trainers' spells as long as it has the money.
#        Default:     1 - (Enabled)
#                     0 - (Disabled)
AiPlayerbot.AllowLearnTrainerSpells = 1

#
#    NoTrainerSpellsCost
#        Description: Learning trainers' spells doesn't cost anything.
#        Default:     0 - (Disabled)
#                     1 - (Enabled)
AiPlayerbot.NoTrainerSpellsCost = 0
```
But `AiPlayerbot.NoTrainerSpellsCost = 1` and `AiPlayerbot.AutoTrainSpells = "free"` both seem like duplicates of `AiPlayerbot.BotCheats` = "gold".

privatecore:
So, the idea behind removing `AiPlayerbot.AutoTrainSpells = "free"` is that both this option and `BotCheats` are "cheats" after all. If anyone wants to use separate cheats for each "gold" action -> we can always impl additional `BotCheats` masks (replacing the "gold" one):
```
"repair" -- skip repair cost  
"travel" -- skip travel cost  
"train" -- skip train cost  
etc.
```

src/Ai/Base/Actions/TrainerAction.cpp
```diff
@@ -9,133 +9,141 @@
...
+Unit* TrainerAction::GetTarget()
+{
+    // There are just two scenarios: the bot has a master or it doesn't. If the
+    // bot has a master, the master should target a unit; otherwise, the bot
+    // should target the unit itself.
+    if (Player* master = GetMaster())
+        return master->GetSelectedUnit();
 
-    msg << " - learned";
+    return bot->GetSelectedUnit();
```
Wishmaster117: why we don't use this masterless option?
    return botAI->GetUnit(bot->GetTarget());

src/Ai/Base/Actions/RpgSubActions.cpp
```diff
@@ -250,6 +255,57 @@ Event RpgSellAction::ActionEvent(Event event) { return Event("rpg action", "vend
...
+{
+    if (!rpg->InRange())
+        return false;
+
+    Creature* creature = rpg->guidP().GetCreature();
+    if (!creature)
+        return false;
+
+    if (!creature || !creature->IsInWorld() || !creature->IsAlive())
+        return false;
```
kadeshar, line 268: Duplicated check

src/Ai/Base/Actions/RpgSubActions.cpp
```diff
@@ -250,6 +255,57 @@ Event RpgSellAction::ActionEvent(Event event) { return Event("rpg action", "vend
...
+
+    return true;
+}
+
+bool RpgTrainAction::isPossible()
+{
+    Creature* creature = rpg->guidP().GetCreature();
+    if (!creature)
+        return false;
```
kadeshar: Duplicated check from isUseful. We cant pass this creature from check which is executed first?

src/Ai/Base/Actions/TrainerAction.cpp
```diff
@@ -9,133 +9,141 @@
...
+                        (trainer->GetTrainerType() != Trainer::Type::Tradeskill || !botAI->HasActivePlayerMaster()));
+
+    Iterate(target, learnSpells, spellId);
+
+    return true;
+}
+
+bool TrainerAction::isUseful()
+{
+    Creature* target = GetTarget()->ToCreature();
```
kadeshar: GetTarget() cant return null?

src/Ai/Base/Actions/TrainerAction.cpp
```diff
@@ -9,133 +9,141 @@
...
-        {
-            uint32 learnedSpell = spellInfo->Effects[j].TriggerSpell;
-            if (!bot->HasSpell(learnedSpell))
-            {
-                bot->learnSpell(learnedSpell);
-                learned = true;
-            }
-        }
+        // Not sure why we need to say anything, especially w/o master
+        botAI->TellError("No spells can be learned from this trainer");
```
kadeshar: That can spam chat if used with wrong trigger. I would avoid such apporach and keep in Execute

src/Ai/Base/Actions/RpgSubActions.cpp
privatecore: This is kind of tricky since `Action` does not have proper attributes to store the target and target GUID, but `RpgHelper` actually has a helper method/getter that returns `GuidPosition`. So, I rewrote the `isPossible` checks to use `GuidPosition` -> `CreatureTemplate` instead.

src/Ai/Base/Actions/TrainerAction.cpp
privatecore: The problem is that this can't happen inside the `Execute` method since `isPossible` already did all the validations (`TrainerAction` & `RpgTrainAction`).

PR #2105, opened 2026-02-02
Fix bug with GetCreature + GetGameObject = use ObjectAccessor's methods instead

privatecore:
# Pull Request

https://en.cppreference.com/w/cpp/algorithm/equal_range.html
> second is an iterator to the first element of the range [first, last) ordered after value (or last if no such element is found). 

The original code uses `return bounds.second->second`, which causes the wrong creature/gameobject to be returned. Instead, both methods (`GetCreature` and `GetGameObject`) now utilize ObjectAccessor's methods to retrieve the correct entities. These built-in methods offer a safer way to access objects. Additionally, `GetUnit` no longer includes redundant creature processing before checks and now has the same logic as the `ObjectAccessor::GetUnit` method.

Furthermore, `GuidPosition::isDead` method has been renamed to `GuidPosition::IsCreatureOrGOAccessible` and updated, as it is used only for creatures (NOT units) and gameobjects.

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## How to Test the Changes

The behavior has not changed after all.

## Complexity & Impact

    - [x] No
    - [ ] Yes (**explain below**)

    - [x] No

    - [x] No
    - [ ] Yes (**explain why**)

---

## Defaults & Configuration

    - [x] No
    - [ ] Yes (**explain why**)

---

## AI Assistance

    - [x] No
    - [ ] Yes (**explain below**)

If yes, please specify:

---

## Final Checklist

---

## Notes for Reviewers

before merging.

### DIFF
--- a/.github/workflows/check_pr_source.yml
+++ b/.github/workflows/check_pr_source.yml
@@ -1,0 +1,19 @@
+name: Enforce test-staging → main
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  require-test-staging:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Ensure PR source is test-staging
+        run: |
+          echo "Base: ${{ github.event.pull_request.base.ref }}"
+          echo "Head: ${{ github.event.pull_request.head.ref }}"
+          if [ "${{ github.event.pull_request.head.ref }}" != "test-staging" ]; then
+            echo "✖ Pull request must come from test-staging"
+            exit 1
+          fi
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -66,38 +66,47 @@ Please answer the following:
 
 ## Complexity & Impact
 
-- Does this change add new decision branches?
-    - [ ] No
-    - [ ] Yes (**explain below**)
+Does this change add new decision branches?
+```
+[ ] No
+[ ] Yes (**explain below**)
+```
 
-- Does this change increase per-bot or per-tick processing?
-    - [ ] No
-    - [ ] Yes (**describe and justify impact**)
+Does this change increase per-bot or per-tick processing?
+```
+[ ] No
+[ ] Yes (**describe and justify impact**)
+```
 
-- Could this logic scale poorly under load?
-    - [ ] No
-    - [ ] Yes (**explain why**)
-
+Could this logic scale poorly under load?
+```
+[ ] No
+[ ] Yes (**explain why**)
+```
 ---
 
 ## Defaults & Configuration
 
-- Does this change modify default bot behavior?
-    - [ ] No
-    - [ ] Yes (**explain why**)
+Does this change modify default bot behavior?
+```
+[ ] No
+[ ] Yes (**explain why**)
+```
 
 If this introduces more advanced or AI-heavy logic:
-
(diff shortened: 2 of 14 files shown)


hermensbas (approved):

src/Ai/Base/Value/QuestValues.cpp
```diff
@@ -173,7 +173,7 @@ std::vector<GuidPosition> ActiveQuestGiversValue::Calculate()
                     continue;
             }
 
-            if (guidp.isDead())
+            if (guidp.IsCreatureOrGOAccessible())
                 continue;
```
Wishmaster117, line 177: I’m not sure about the polarity here: previously we skipped when guidp.isDead() was true. With the change, we now skip when guidp.IsCreatureOrGOAccessible() is true, which seems like it would skip the accessible creature/GO and keep the inaccessible ones. Is that intended? If yes, can you explain the behavior change you’re aiming for (especially when the grid is loaded)?

src/Mgr/Travel/TravelMgr.cpp
```diff
@@ -881,61 +881,79 @@ std::vector<GameObjectData const*> WorldPosition::getGameObjectsNear(float radiu
...
 {
     if (!*this)
         return nullptr;
 
     if (loadedFromDB)
-    {
-        auto creatureBounds = getMap()->GetCreatureBySpawnIdStore().equal_range(GetCounter());
-        if (creatureBounds.first != creatureBounds.second)
-            return creatureBounds.second->second;
+        return ObjectAccessor::GetSpawnedGameObjectByDBGUID(getMapId(), GetCounter());
```
Wishmaster117: Here ou call GetSpawnedGameObjectByDBGUID(getMapId(), ...), but in GuidPosition::GetCreature() you call GetSpawnedCreatureByDBGUID(GetMapId(), ...). Are both getMapId() and GetMapId() intentionally available here (WorldPosition vs ObjectGuid), or should we standardize to one to avoid build issues / accidental different semantics?

src/Ai/Base/Value/QuestValues.cpp
privatecore: Yeah, my bad -- fixed!

src/Mgr/Travel/TravelMgr.cpp
privatecore: Damn, those method's names...

```cpp
uint32 WorldPosition::getMapId() { return GetMapId(); }
float WorldPosition::getX() { return GetPositionX(); }
float WorldPosition::getY() { return GetPositionY(); }
float WorldPosition::getZ() { return GetPositionZ(); }
float WorldPosition::getO() { return GetOrientation(); }
```
All these methods are useless. I will use `GetMapId()` in the updated methods, but we rly need to clean this up -- it's not defined and used as an adapter anywhere.

Wishmaster117: Fixing this module is harder than clearing ICC in full grey gear :D

privatecore: > Fixing this module is harder than clearing ICC in full grey gear :D

Whoever created `GuidPosition` is pure evil -_- Combining together object and position is not good at all.

Celandriel (approved):

PR #2106, opened 2026-02-02
Warnings PR 1: Event warnings and headers

Celandriel:
# Pull Request

This is the first in a series of PRs intended to eliminate warnings in the module. The design intent is to eliminate the calling event when not needed in the body of the function. Based off of SmashingQuasars work.

---

## How to Test the Changes

## Complexity & Impact

    - [x] No
    - [ ] Yes (**explain below**)

    - [x] No

    - [x] No
    - [ ] Yes (**explain why**)

---

## Defaults & Configuration

    - [x] No
    - [ ] Yes (**explain why**)

---

## AI Assistance

    - [x] No
    - [ ] Yes (**explain below**)

---

## Final Checklist

---

## Notes for Reviewers

before merging.

### DIFF
--- a/src/Ai/Raid/GruulsLair/Action/RaidGruulsLairActions.cpp
+++ b/src/Ai/Raid/GruulsLair/Action/RaidGruulsLairActions.cpp
@@ -491,7 +491,7 @@ bool HighKingMaulgarMisdirectOlmAndBlindeyeAction::Execute(Event /*event*/)
 // Gruul the Dragonkiller Actions
 
 // Position in center of the room
-bool GruulTheDragonkillerMainTankPositionBossAction::Execute(Event /*event*/)
+bool GruulTheDragonkillerTanksPositionBossAction::Execute(Event /*event*/)
 {
     Unit* gruul = AI_VALUE2(Unit*, "find target", "gruul the dragonkiller");
     if (!gruul)


hermensbas:
please dont use (Event), rather use (Event /*event*/) as in general when dealing with unused params or even unused prefix. 

Removing the param name can harm signature of the method, even when not used. Since the nature of this port project i think i would be wise to keep them as context and understanding about the intend.

src/Ai/Base/Value/DpsTargetValue.cpp
```diff
@@ -53,7 +52,7 @@ class CasterFindTargetSmartStrategy : public FindTargetStrategy
         result = nullptr;
     }
 
-    void CheckAttacker(Unit* attacker, ThreatMgr* threatMgr) override
```
kadeshar, line 56: That shouldnt be commented variable?

src/Bot/PlayerbotMgr.cpp
```diff
@@ -1700,7 +1698,7 @@ void PlayerbotMgr::TellError(std::string const botName, std::string const text)
     errors[text] = names;
 }
 
-void PlayerbotMgr::CheckTellErrors(uint32 elapsed)
+void PlayerbotMgr::CheckTellErrors(uint32)
```
kadeshar: If that method in not somehow overriden or something then this parameter should be deleted. If is then should be commented.

src/Script/Playerbots.cpp
```diff
@@ -441,49 +417,37 @@ class PlayerbotsScript : public PlayerbotScript
...
-        }
 
         if (PlayerbotMgr* playerbotMgr = GET_PLAYERBOT_MGR(player))
-        {
             playerbotMgr->HandleMasterOutgoingPacket(*packet);
-        }
     }
 
-    void OnPlayerbotUpdate(uint32 diff) override
+    void OnPlayerbotUpdate(uint32) override
```
kadeshar: Parameter should be commented

src/Ai/Raid/Ulduar/Trigger/RaidUlduarTriggers.cpp
```diff
@@ -1,19 +1,17 @@
...
 #include "SharedDefines.h"
 #include "Trigger.h"
 #include "Vehicle.h"
-#include <MovementActions.h>
-#include <FollowMasterStrategy.h>
-#include <RtiTargetValue.h>
+#include "MovementActions.h"
+#include "FollowMasterStrategy.h"
+#include "RtiTargetValue.h"
+#include "../../../../../../src/server/scripts/Northrend/Ulduar/Ulduar/ulduar.h"
```
Wishmaster117: Are they another way?

src/Ai/Raid/Ulduar/Trigger/RaidUlduarTriggers.cpp
```diff
@@ -1839,10 +1828,7 @@ Unit* YoggSaronTrigger::GetNextIllusionRoomRtiTarget()
         for (const ObjectGuid& guid : targets)
         {
             Unit* unit = botAI->GetUnit(guid);
-            if (unit && unit->IsAlive() && unit->GetEntry() == NPC_LAUGHING_SKULL)
-            {
```
Wishmaster117, line 1843: Here you’re removing the filter on the mob’s entry (a specific NPC) and returning unconditionally the first GUID you iterate over.
The bot may end up targeting any NPC returned first by “nearest npcs” (or even a nullptr if the first GUID doesn’t resolve correctly), instead of looking for the intended mob.

This could break the Yogg-Saron behavior (cheat raid) and cause inconsistent targeting, absurd movement, or even wipes, right?

src/Ai/Base/Value/DpsTargetValue.cpp
Celandriel: Mistakenly removed.

src/Bot/PlayerbotMgr.cpp
```diff
@@ -1698,7 +1698,7 @@ void PlayerbotMgr::TellError(std::string const botName, std::string const text)
     errors[text] = names;
 }
 
-void PlayerbotMgr::CheckTellErrors(uint32)
+void PlayerbotMgr::CheckTellErrors(uint32 elapsed)
```
kadeshar, line 1701: If you cannot delete parameter then simply comment variable otherwise simple delete parameter

src/Script/Playerbots.cpp
```diff
@@ -447,7 +447,7 @@ class PlayerbotsScript : public PlayerbotScript
             playerbotMgr->HandleMasterOutgoingPacket(*packet);
     }
 
-    void OnPlayerbotUpdate(uint32) override
+    void OnPlayerbotUpdate(uint32 diff) override
```
kadeshar, line 450: If you cannot delete parameter then simply comment variable otherwise simple delete parameter

Celandriel: To fix in another PR.

hermensbas: You might wanna be careful with removing parameters when its not used. This is not some finished production code, its coming from half finished port. Changing the signatues, just because its not currently current used which will include the calling structures through the code, will harm the process.

hence they invented the maybe_unused flag or just commenting out the parameter. Changing the context or altering method and calling signatures will have a negative impact on the project in the long run. Clean code principles aint there to blindly follow.

Once the project can state the port has been completed and we are simply and only extending logic for a while then we should consider removing unused params and signatures. And yes git holds the truth until it doesnt or is obfuscared by the sheer amount of history mutations or force push/squash is applied for whatever reason.

Celandriel: Yea, those were unintentional at this stage. I just missed them when splitting from origin pr

hermensbas: In a few weeks i can help again

hermensbas (approved):

PR #2110, opened 2026-02-03
Fix RandomItemMgr performance issues - macro and micro optimizations

privatecore:
### Summary

This refactor makes `RandomItemMgr` faster, safer, and easier to maintain. It reduces cache scans, cuts extra copies, improves DB work, and makes item checks more direct across equip, ammo, rarity, and random item logic.

### What was done

- **Switched `std::map` to `std::unordered_map`** - internal caches now use hash maps, and `BotEquipKey` has a proper hash function, so most lookups are faster.
- **Switched `std::set` to `std::unordered_set`** - item and slot containers now use hash sets for faster insert, find, and erase operations.
- **Fixed argument order** - many methods had arguments in the wrong order, for example `CanEquipWeapon(class, item)` is now `CanEquipWeapon(item, class)`, and the same style is used everywhere.
- **Aligned method signatures with header definitions** - method order and parameter types were cleaned up to match the header and make the code easier to follow.
- **Used one DB transaction for cache build** - `BuildCacheEquip`, `BuildCacheRandomItem`, and `BuildCacheRarity` now add all records in one transaction instead of one by one, so there are fewer DB round-trips.
- **Split cache load and cache build** - each cache now tries to load data from DB first, and only rebuilds if the cache is empty, so startup does less work.
- **Fixed cache bucket logic** - food and trade caches now use the correct level buckets, and rarity loading now reads query results in the right way.
- **Reduced cache lookup cost** - old code used linear scans and extra loops over item lists, while new code uses direct hash map lookups.
- **Reversed `viableSlots` mapping** - it now goes from `InventoryType` to `EquipmentSlot`, so code can find needed slots directly without scanning the whole map.
- **Pre-computed weapon use rules at init** - weapon checks are now prepared once in `InitWeaponProficiency`, so class-to-weapon checks are just a fast bitwise check.
- **Simplified `CanEquipArmor`** - armor rules were cleaned up and moved into one clear place instead of a long `if/else` chain.
- **Rewrote `CanEquipItem`** - the method now uses the new cache layout and simpler checks, with better defensive handling for invalid data.
- **Renamed and improved `IsTestItem` to `IsInternalItem`** - it now checks item flags and names in a better way, so test and internal items are skipped more reliably.
- **Renamed and improved `GetCachedEquipments` to `GetEquipmentNew`** - it now returns a reference to avoid copying the whole item list every time the method is called.
- **Added `GetEquipment`** - equipment lookup now uses `equipCache` directly and has extra checks for null or invalid item data.
- **Removed extra trainer processing** - `IsCraftedBySkill` now uses a direct `SkillLineAbility` lookup instead of looping through all trainers, which saves time.
- **Added `IsValidItem`** - all item validation rules were moved into one shared method, so other code can reuse the same checks.
- **Changed `GetAmmo`** - it now returns a single item id instead of a vector, which makes the logic simpler and faster.
- **Used `IsInternalItem` in bot item setup** - test and internal items are filtered with the shared check during item init.
- **Made `Query` safer** - it is now `const` and uses direct map lookups instead of default inserts and full scans.
- **Made item getters return from cached buckets** - `GetAmmo`, `GetRandomPotion`, `GetRandomFood`, and `GetRandomItem` now use cached data directly and avoid extra copying.
- **Improved cache load logs** - `LoadCacheEquip`, `LoadCacheRandomItem`, and `LoadCacheRarity` now clearly report empty cache cases.
- **Used `IsValidItem` before random item cache fill** - item validity checks are now handled in one place before the cache is filled.
- **Kept most of `BuildCacheItemInfo` unchanged** - the heavy weight calculation and the commented quest or vendor code were mostly left for a separate refactor.
- **Initialized helper tables in one place** - weapon proficiency, predicates, viable slots, and weight links are now set during `RandomItemMgr` init.
- **Moved predicate setup into `InitPredicates`** - this keeps constructor code smaller and makes cache rules easier to follow.
- **Simplified `GetRandomItem`** - random item selection now uses the cached list directly, without manual index handling.
- **Improved `BuildCacheRandomItem` filtering** - temporary, invalid, and internal items are skipped earlier, before the cache is filled.
- **Removed some dead code** - old vendor and loot drop scan code that was already unused or commented out was cleaned up where it was safe.
- **Improved logging** - cache load and build functions now show timing and record counts in a consistent way.
- **Added better defensive checks** - several methods now guard against null pointers and invalid templates before using item data.
- **Fixed server loading messages and comments** - load logs and code comments were updated for clearer output and easier maintenance.
- **Fixed warnings and broken trinket logic** - warning cleanup also corrected a bad trinket-related condition.

### Profit

The main result is faster and cleaner item processing for bots. Cache lookups are cheaper, startup work is smaller, and item checks are more direct, which should help both performance and stability. This also makes the code easier to change later. The new structure gives clearer methods, better cache flow, and fewer repeated checks across equip, ammo, rarity, and random item logic.

### Notes

- Did not touch anything related to item weight calculations, loading, and building - this part should be refactored separately.
- Did not rewrite `PlayerbotFactory` methods, as they are clearly duplicating `RandomItemMgr` ones and also have their own logic problems with loop order and iteration - that is a separate topic.
- Did not clean up most of `BuildCacheItemInfo` and did not remove dead code or methods used inside commented-out blocks (`GetQuestIdForItem`, `GetQuestIdsForItem`).

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## How to Test the Changes

## Complexity & Impact

    - [x] No
    - [ ] Yes (**explain below**)

    - [x] No

    - [x] No
    - [ ] Yes (**explain why**)

---

## Defaults & Configuration

    - [x] No
    - [ ] Yes (**explain why**)

---

## AI Assistance

    - [x] No
    - [ ] Yes (**explain below**)

If yes, please specify:

---

## Final Checklist

---

## Notes for Reviewers

before merging.


privatecore:
I'm gonna to focus on more optimizations: caching, inline methods, utility methods, etc. So, ill only check compile + raw performance for now. Anyone interested in testing -- plz use the workaround from this [comment](https://github.com/mod-playerbots/mod-playerbots/issues/1987#issuecomment-3721670841).

elderbit:
Great work! Unfortunately, I can't approve anything here, but I'm currently using the code. No problems so far. To be honest, I can't really see a big improvement in performance, but I think that many small optimizations add up to make a big difference.

privatecore:
Loading from DB is much faster, so I only highlighted the build-cache items. Not sure we actually need to store the cache in the DB -- it creates pitfalls with bots' max levels, playable classes/races, etc.
```log
Loading equipment cache...
>> Loaded 0 equipment items. DB table `playerbots_equip_cache` is empty!
 
Building equipment cache for 80 levels...
>> Cached total 1628586 equipment entries in 419 ms
 
Building equipment cache (new) for 80 levels...
>> Cached total 24676 equipment entries in 6 ms
 
Building ammo cache for 80 levels...
>> Cached total 1465 ammo entries in 1 ms
 
Building food cache for 80 levels...
>> Cached total 556 food entries in 0 ms
 
Building potion cache for 80 levels...
>> Cached total 575 potion entries in 0 ms
 
Building trade cache for 80 levels...
>> Cached total 2981 trade entries in 1 ms
 
Loading random item cache...
>> Loaded 0 random items. DB table `playerbots_rnditem_cache` is empty!
 
Building random item cache from 46096 items
>> Cached total 12637 random items in 6 ms
 
Loading items rarity cache...
>> Loaded 0 rarity items. DB table `playerbots_rarity_cache` is empty!
 
Building items rarity cache...
>> Cached total 7157 rarity entries in 2659 ms
```

privatecore:
SQL queries used during testing are attached.

[build_cache_item_class_subclass.sql](https://github.com/user-attachments/files/28016089/build_cache_item_class_subclass.sql)
[build_cache_item_rarity.sql](https://github.com/user-attachments/files/28016091/build_cache_item_rarity.sql)
[check_internal_item_name.sql](https://github.com/user-attachments/files/28017772/check_internal_item_name.sql)

ThePenguinMan96:
What is your scope of testing within the client? What characters, levels, and classes have you tried, and do the results vary from the current master branch?

privatecore:
> What is your scope of testing within the client? What characters, levels, and classes have you tried, and do the results vary from the current master branch?

All getters are still in `PlayerbotFactory` and I didn’t change them. Checked the cache against the master branch and got diff results: a lot of event items (duration > 0), quest items (item level = 0), and items players can’t use were removed from the cache.

`equipCache` isn’t used right now, probably because it’s slow to build, but we could bring it back if `PlayerbotFactory` gets refactored. Same applies to `rarityCache`.

Used the code snippet from the attachment to compare the results.

[debug_rnditem_cache.patch](https://github.com/user-attachments/files/28027778/debug_rnditem_cache.patch)

Celandriel:
> 🤖 **Automated review by Claude (model: Fable)** via Claude Code. Verify findings before acting; this is advisory, not a maintainer approval.

# PR #2110 Review — RandomItemMgr performance refactor

## Verdict: REQUEST CHANGES

The refactor is substantially good work — the performance claims are mostly real, several latent thread-safety and logic bugs are genuinely fixed, and no hallucinated APIs exist (`GetSkillLineAbilitiesBySkillLine`, `GetSkillRaceClassInfo`, `SPELL_CATEGORY_FOOD/DRINK`, `SelectRandomContainerElement` all verified in the core fork). All call sites of the swapped-argument methods (`CanEquipWeapon`, `CanEquipArmor`) and changed signatures (`GetAmmo`, `GetCachedEquipments`→`GetEquipmentNew`, `IsTestItem`→`IsInternalItem`) were updated — only `ItemUsageValue.cpp` and `PlayerbotFactory.cpp` call them, and both are in the PR. However, there are blocking issues around merge staleness, a level-cap regression, stale-DB-cache semantics, dead-code infrastructure, and an exception-on-bad-data hazard. The PR's claim of "no default behavior change" is **inaccurate** — this PR meaningfully changes which items bots equip.

---

## Blocking issues

### B-1. PR is stale vs master — wholesale header rewrite will drop `GetEnchantmentPool` (compile break on merge-forward)
Current master's `RandomItemMgr.h` declares `GetEnchantmentPool(uint32) const` (impl in `RandomItemMgr.cpp`, live caller at `src/Mgr/Item/StatsWeightCalculator.cpp:226`). The PR's rewritten header/cpp (based on `test-staging`) contains no trace of it. Because this PR replaces the file's structure rather than editing in place, a merge to master will conflict on ~2,400 lines or silently drop the method, causing a compile failure in `StatsWeightCalculator.cpp`. **Fix:** rebase onto current master (or sync test-staging) before merge; re-verify the header against all `sRandomItemMgr.` call sites.

### B-2. Cache level cap silently switched from server max level to `randomBotMaxLevel`
`BuildCacheAmmo`/`BuildCachePotion`/`BuildCacheEquip` and `IsValidItem` (`proto->RequiredLevel > randomBotMaxLevel` returns false) now size/filter by `randomBotMaxLevel`; old code used `CONFIG_MAX_PLAYER_LEVEL`. On a server with `RandomBotMaxLevel = 60` and real players running level-80 self/alt-bots, `GetAmmo(80, …)` and `GetRandomPotion(80, …)` miss and return 0, and `equipCacheNew` has no level-61+ entries. Old behavior served these bots. Silent functional regression for a supported config (rule G2). **Fix:** cap at `std::max(randomBotMaxLevel, CONFIG_MAX_PLAYER_LEVEL)` or restore the old per-cache sources; drop the `RequiredLevel > randomBotMaxLevel` clause from `IsValidItem`.

### B-3. Load-vs-build split trusts a stale DB cache with changed bucket semantics
Old builder wrote buckets as `ItemLevel / 10`; new builder and `Query()` use `(ItemLevel - 1) / 10`. `LoadCacheRandomItem` loads whatever rows exist and only rebuilds if the table is empty. An existing `playerbots_rnditem_cache` populated by old code loads with off-by-one buckets for every item whose ItemLevel is a multiple of 10, and retains items the new `IsInternalItem`/`IsValidItem` filters would now exclude. No cache version stamp, no truncation SQL. **Fix:** ship a SQL update truncating `playerbots_rnditem_cache` (and `playerbots_equip_cache` if B-4 is resolved), or add a cache-version row checked at load.

### B-4. Dead infrastructure added/rewritten with zero live callers (rules D1/D3)
`GetEquipment()`, `LoadCacheEquip()`, `BuildCacheEquip()`, `equipCache`, `BotEquipKey` + the new `std::hash<BotEquipKey>` specialization — grepping the whole module, `GetEquipment(` has **zero callers**, and `LoadCacheEquip`/`BuildCacheEquip` are referenced only from commented lines in `Init()`. So `equipCache` is never populated and `GetEquipment` always returns the static empty list. Likewise `ShouldEquipArmorForSpec`/`ShouldEquipWeaponForSpec` were fully rewritten (~300 lines) yet their only callers are commented-out blocks. Worse, the rewritten (dead) `BuildCacheEquip` applies `CanEquipArmor` to *all* armor slots (old code applied it only to the 8 body-armor slots), which would exclude all rings/necks/trinkets if re-enabled — a logic landmine. **Fix:** wire it in (with the jewelry-slot fix and a truncation story) or delete it.

### B-5. `m_weightScales[clazz].at(spec)` throws on missing spec — crash where old code returned false
In `ShouldEquipArmorForSpec`/`ShouldEquipWeaponForSpec`: `WeightScale const& scale = m_weightScales[clazz].at(spec);`. If `playerbots_weightscales` is empty/partial (the code even logs an error and continues), or a caller passes an unregistered spec, `unordered_map::at` throws and the server terminates. Old code used `operator[]` (default-constructs, returns false gracefully). Mitigated today only by these methods being dead (B-4); if kept, must be `find()` with a null-id fallback.

### B-6. Cache re-initialization on `.playerbot reload` corrupts caches and now races on returned references (rule B11)
`.playerbot reload` re-runs `Init()` on a **map thread**. Re-running `Init()` appends into already-populated caches (duplicate entries — pre-existing) and, with this PR, rebuilds `unordered_map`s **while other map threads hold `RandomItemList const&` references returned by `GetEquipmentNew`** (PlayerbotFactory.cpp:2271 iterates the reference). A rehash during rebuild invalidates those references, causing use-after-free. The old copy-returning `GetCachedEquipments` had a much narrower window. **Fix:** guard `Init()` against re-entry (`if (initialized) return;`), or `clear()` + rebuild only when bots are inactive.

---

## Non-blocking issues

1. **Undeclared default-behavior changes (G2) — list in the PR description and smoke-test per class:** `CanEquipWeapon` is now DBC-proficiency-driven (rogues lose 1H axes, druids lose polearms — correct, but changes equip decisions); `IsInternalItem` adds `"NPC "` items (~4,859) to the exclusion list and `IsValidItem` rejects any `Duration != 0` (conjured/event items now excluded); `INVTYPE_WEAPONOFFHAND` is now a viable OFFHAND type (previously invisible), `INVTYPE_2HWEAPON`/`WEAPONMAINHAND` removed from OFFHAND; `GetUpgrade`'s "skip no-stats trinkets" operator-precedence bug (`a && b || c || d`) is fixed; `BuildCacheEquipNew` now dedups quest-reward items; `CanEquipItem` fixes a uint32 underflow that rejected all uncommon+ gear for bots below level 15. Mostly improvements — but they ARE behavior changes.
2. **`IsUsedBySkill` static cache mutated from map threads without sync** (the documented hottest path, PR #2190). Pre-existing race, but the PR switched `std::set` to `std::unordered_set` where a concurrent insert-during-rehash is a harder crash. Also keyed by itemId only (positive for one skill answers true for every skill), and negative results aren't cached, so the expensive `IsCraftedBySkill` reruns on every miss on the map thread. Recommend precomputing per-skill item sets at startup.
3. **`GetStatWeight` carries a moved-verbatim bug**: two consecutive identical `if (name == specName)` checks, the second assigning `specId = specNum` (loop counter) instead of `info.id`. Harmless only because `itemInfoCache` is never populated — fix or remove.
4. **`GetUpgradeList` log format string still contains `%d`** (`"...Old item:%d..."`) — violates the `{}`-only fmt rule; prints a literal `%d`.
5. **Dead public API retained:** `GetUpgrade`, `GetUpgradeList`, `GetStatWeight`, `GetLiveStatWeight`, `HasStatWeight`, `GetQuestIdForItem`, `GetQuestIdsForItem`, `GetRandomFood`, `GetItemRarity`, `LoadCacheRarity`/`BuildCacheRarity` all have zero live callers and operate on caches that are empty or never built. The rewritten `BuildCacheRarity` SQL (now AC-schema-correct with window functions — a real fix) remains unreachable. Per D2, wire or remove.
6. **Style (AC standards):** same-line conditionals (`if (slot == EQUIPMENT_SLOT_MAINHAND) hasMH = true;`); the `inSlot` lambda's `for (...) if (...) return true;` on one line; `// validatoin` typo x6; misplaced "junk (poor/normal)" comment inside the `Quality > NORMAL` branch; `BuildCacheEquip` comment claims to skip poor quality but code only skips `> EPIC`.
7. **`m_weaponProficiency` race placeholder:** `GetSkillRaceClassInfo(skillId, RACE_HUMAN, cls)` assumes weapon-skill availability is race-independent — true for WotLK, and the author documents it. Acceptable.

---

## Positive notes

- **All argument-order swaps are consistent and complete.** Only two files outside RandomItemMgr call the reordered methods, both updated correctly; PlayerbotFactory's same-named `CanEquipArmor/CanEquipWeapon` are its own members and unaffected. The `IsTestItem(gemId)` to `IsInternalItem(proto)` swap at PlayerbotFactory.cpp:500 is correct (the new check is a strict superset).
- **Genuine thread-safety fixes:** old `GetItemRarity`, `GetRandomFood/Trade`, `GetAmmo`, `Query` all used `operator[]`, *inserting into shared maps from map threads* — a real B11 violation now eliminated by const `find()` accessors.
- **Real performance wins:** `BuildCacheAmmo` replaces ~160 synchronous startup queries with one item-store pass plus a sort that reproduces the old `ORDER BY stackable DESC, ItemLevel DESC` semantics exactly (including TBC/WotLK entry-ID cutoffs); cache builds wrapped in single transactions (B10); `IsCraftedBySkill` drops the all-trainer scan for a direct SkillLineAbility lookup; `GetEquipmentNew` avoids a vector copy per call.
- **Correctness fixes verified equivalent or better:** `AddItemStats`/`CanUseItemStats` consolidation is bit-for-bit equivalent to the old three-switch logic; food/trade/potion bucket math equivalent at all boundaries; `BotEquipKey::operator<` inversion bug fixed; null-proto guards added where old code dereferenced unchecked.
- On the AI-assistance question: the uniform per-line commenting, `[[nodiscard]]` saturation, and tabular alignment across 2,500 lines are stylistically machine-like, but everything checks out against real core APIs and real game rules — judged on substance, which is solid.

**Bottom line:** approve after (1) rebase onto master preserving `GetEnchantmentPool`, (2) the `randomBotMaxLevel` cap regression fix, (3) a cache-invalidation SQL or version stamp, (4) a wire-or-delete decision on the equip-cache/`ShouldEquip*` dead infrastructure (with `.at()` to `find()` if kept), (5) a re-init guard for reload, and (6) an honest G2 declaration of the equip-behavior changes.

privatecore:
> Bottom line: approve after (1) rebase onto master preserving GetEnchantmentPool, (2) the randomBotMaxLevel cap regression fix, (3) a cache-invalidation SQL or version stamp, (4) a wire-or-delete decision on the equip-cache/ShouldEquip* dead infrastructure (with .at() to find() if kept), (5) a re-init guard for reload, and (6) an honest G2 declaration of the equip-behavior changes.

1. The `test-staging` branch is merged and conflicts are resolved.  
2. The logic is changed completely: we now build the cache using `DEFAULT_MAX_LEVEL`(80). Getters also normalize the level when `randomBotMaxLevel > DEFAULT_MAX_LEVEL`. DBC files and item templates do not contain required levels above 80. **Note:** `randomBotMaxLevel` is already normalized when loading config files with `CONFIG_MAX_PLAYER_LEVEL`.  
3. SQL files are added with `TRUNCATE TABLE` statements.  
4. Fixed `at()` -> `find()` and added a fallback.  
5. Re-initialization guard is added, with a comment for future updates/changes to properly support rebuild/recache actions.  
6. Dunno, if I need to explain something to an AI...

ThePenguinMan96:
Have you actually tested this in the client? Could you show screenshots of the test results? These seem like significant changes. 

Anytime anything is changed or introduced, lots of testing needs to go into it. Speaking from experience. Otherwise it will introduce new bugs/issues.

Celandriel:
> 6. Dunno, if I need to explain something to an AI...

The AI can be pretty dumb, but is useful for some stuff. 
I had limited mental energy when reviewing this so I fell back onto it for some review.


 One question, was this copied from any other project where author attribution is necessary?

privatecore:
> Have you actually tested this in the client? Could you show screenshots of the test results? These seem like significant changes.

Have created 80 bots (levels 1-80) and initialized equipment multiple times. I’m not sure how you want to test this on the client side, but you can freely merge this PR and test it yourself.

privatecore:
> The AI can be pretty dumb, but is useful for some stuff. I had limited mental energy when reviewing this so I fell back onto it for some review.
> 
> One question, was this copied from any other project where author attribution is necessary?

To be honest, I don't know any other project except this one. All this changes are based on AC/TC codebase, DB data and my logic.

> B-6. Cache re-initialization on .playerbot reload corrupts caches and now races on returned references (rule B11)

Check `RandomPlayerbotMgr::HandlePlayerbotConsoleCommand` -> `sPlayerbotAIConfig.Initialize()` = all caches right now have this bug... Regarding RandomItemMgr: if we decide to drop the rarity cache (took >2 seconds), then all caches can be built on loading w/o DB.

Celandriel:
> To be honest, I don't know any other project except this one. All this changes are based on AC/TC codebase, DB data and my logic.

No worries. It's something we will need to start verifying.

> 
> > B-6. Cache re-initialization on .playerbot reload corrupts caches and now races on returned references (rule B11)
> 
> Check `RandomPlayerbotMgr::HandlePlayerbotConsoleCommand` -> `sPlayerbotAIConfig.Initialize()` = all caches right now have this bug... Regarding RandomItemMgr: if we decide to drop the rarity cache (took >2 seconds), then all caches can be built on loading w/o DB.

If this is true then we should add in a clear before init just in case

privatecore:
> If this is true then we should add in a clear before init just in case

Or guard all cache re-initializations. Again, regarding `RandomItemMgr`: with these changes, caches are created for all levels 1..80, and the only parameter/config value that can affect their content is `unobtainableItems`. So it’s OK to skip re-init of random items caches if we rebuild them on every start-up, OR to remove DB storage for caches and add clear command.

Celandriel:
True. It depends on whether sthere something that is modifieable in the config with respect to the cache

privatecore:
If someone wants and can help with testing, there are only two viable options (as far as I know) that do not require hacking the bot creation process:

1. Set `RandomBotMinLevel` = `RandomBotMaxLevel` to 1..80, and set `DeleteRandomBotAccounts` = 0/1 after each check.  
2. Use the `.playerbot bot initself` command (more precise, since you can test all class/level combs) plus the `.levelup 1` command.

Celandriel (approved):

PR #2117, opened 2026-02-03
Add aggressive non combat targeting strategy 

dillyns:
# Pull Request

Tired of failing that escort quest because your bots stood and watched while the escort npc got swarmed and killed?
Tired of your bots standing around doing nothing while the npc you are supposed to be guarding for 5 minutes is getting attacked?
Don't want to use the grind strategy because it is too heavy-handed and has too many restrictions?

Look no further! Just do "nc +aggressive" and your bots will pick a fight with anything they can in a 30 yard radius.

The aggressive targetting is a stripped down version of the grind target.

## Feature Evaluation

Please answer the following:

Add a strategy, action, and targetting that will cause bots to attack nearby enemies when out of combat.

Hopefully this is the cheapest.

Minimal runtime cost as this strategy needs to be added specifically to bots.

---

## How to Test the Changes

- Add a bot to party, or use selfbot
- Give them the aggressive strategy via "nc +aggressive"
- They should attack anything within 30 yards.
- If it is a bot with a master, the 30 yards should be centered around the master not the bot (prevent chaining from enemy to enemy)

## Complexity & Impact

```
[] No
[x] Yes (**explain below**)
Only for bots that have the added strategy, adds decision to attack nearby targets when out of combat.
```

```
[] No
[x] Yes (**describe and justify impact**)
Minimal increase to only bots that have this strategy added.
```

```
[x] No
[ ] Yes (**explain why**)
```
---

## Defaults & Configuration

```
[x] No
[ ] Yes (**explain why**)
```

```
[x] Lightweight mode remains the default
[ ] More complex behavior is optional and thereby configurable
```
---

## AI Assistance

```
[ ] No
[x] Yes (**explain below**)
```
Claude is used to explore the codebase to find similar implementations to be used for examples.

---

## Final Checklist

---

## Notes for Reviewers

before merging.


src/Ai/Base/Value/AggressiveTargetValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    if (master && (master == bot || master->GetMapId() != bot->GetMapId() || master->IsBeingTeleported() ||
+                   !GET_PLAYERBOT_AI(master)))
+        master = nullptr;
+
+    GuidVector targets = *context->GetValue<GuidVector>("possible targets");
+    if (targets.empty())
+        return nullptr;
+
+    float aggroRange = 30.0f;
```
dillyns: Is this worth being added to config?

Wishmaster117:
Hi, thanks for the PR,  if merged, Multibot needs to be updated.

NoxMax:
> Tired of failing that escort quest because your bots stood and watched while the escort npc got swarmed and killed?
Tired of your bots standing around doing nothing while the npc you are supposed to be guarding for 5 minutes is getting attacked?

If you attack whatever is attacking the NPC and/or the attacker aggros on you, the bots in your party would attack the attacker automatically.

If they start fights on their own, this could lead to them attacking random NPCs not related to the escort to begin with, and you will be left to protect the escort on your own. It's counterintuitive to what your intent is.

Current default behaviour is already good as far as this, but if escort quests are an issue that people think should be resolved for bots, a better approach would be for them to recognize they are on an escort quest, and if they are, they would defend the escort NPC when anything aggros against it, regardless of player master.

But again, I have done many escort quests with bots in my party, and I haven't found their behavior lacking. They simply attack what I attack.

brighton-chi:
> > Tired of failing that escort quest because your bots stood and watched while the escort npc got swarmed and killed?
> > Tired of your bots standing around doing nothing while the npc you are supposed to be guarding for 5 minutes is getting attacked?
> 
> If you attack whatever is attacking the NPC and/or the attacker aggros on you, the bots in your party would attack the attacker automatically.
> 
> If they start fights on their own, this could lead to them attacking random NPCs not related to the escort to begin with, and you will be left to protect the escort on your own. It's counterintuitive to what your intent is.
> 
> Current default behaviour is already good as far as this, but if escort quests are an issue that people think should be resolved for bots, a better approach would be for them to recognize they are on an escort quest, and if they are, they would defend the escort NPC when anything aggros against it, regardless of player master.
> 
> But again, I have done many escort quests with bots in my party, and I haven't found their behavior lacking. They simply attack what I attack.

I guess I could potentially see some uses out of being able to have bots automatically attack targets within a distance that is more limited than their sight range, but I also am not sure the relevance to the escort quest context. IMO the most frustrating thing about escort quests is that bots do not see friendly NPCs as having healing target values. Having bots heal friendly NPCs could help a lot with certain quests (and Heroic Escape From Durnholde, looking at you, baby Thrall). 

Edit: Or is the idea that this strategy would function like an automated flee? Once the player gets 30+ yards away, targets outside that radius become invalid and thus the bot stops attacking and returns to the player?

src/Ai/Base/Value/AggressiveTargetValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+Unit* AggressiveTargetValue::FindTargetForAggressive()
+{
+    Player* master = GetMaster();
+
+    if (master && (master == bot || master->GetMapId() != bot->GetMapId() || master->IsBeingTeleported() ||
+                   !GET_PLAYERBOT_AI(master)))
+        master = nullptr;
+
+    GuidVector targets = *context->GetValue<GuidVector>("possible targets");
```
privatecore: `AI_VALUE(GuidVector, "possible targets")` ?

src/Ai/Base/Value/AggressiveTargetValue.cpp
privatecore:
```cpp
#define MAX_AGGRO_RADIUS 45.0f  // yards
```
Also, it should be noted that the aggro radius can be configured with `Rate.Creature.Aggro`. The calculations for the aggro range can be found in `Creature::GetAggroRange`.

src/Ai/Base/Value/AggressiveTargetValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+    float distance = 0;
+    Unit* result = nullptr;
+
+    for (ObjectGuid const guid : targets)
+    {
+        Unit* unit = botAI->GetUnit(guid);
+        if (!unit)
+            continue;
+
+        if (!unit->IsInWorld() || unit->IsDuringRemoveFromWorld())
```
privatecore, line 34: `IsAlive` ?

dillyns:
The intention is to have something that works like "grind" but without:
1. Level restrictions
2. Added food&drink actions

I needed a way to have the bots attack everything in the vicinity on sight and the grind strategy did not fulfill that need, so I created and used this option.

src/Ai/Base/Value/AggressiveTargetValue.cpp
kadeshar: Yes, you should do that. That is even in config right now. Only use in this value :)

Image

src/Ai/Base/Value/AggressiveTargetValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "AggressiveTargetValue.h"
+
+#include "Playerbots.h"
+#include "ServerFacade.h"
+#include "SharedDefines.h"
+
+Unit* AggressiveTargetValue::Calculate()
+{
+    if (bot->IsInCombat())
+        return nullptr;
```
kadeshar: For me that should be in Action isUseful rather than in value.

src/Ai/Base/Value/AggressiveTargetValue.cpp
privatecore: I didn't check the code -- thought it used the creature aggro radius, lol

src/Ai/Base/Value/AggressiveTargetValue.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+        if (!bot->IsHostileTo(unit) && unit->GetNpcFlags() != UNIT_NPC_FLAG_NONE)
+            continue;
+
+        if (abs(bot->GetPositionZ() - unit->GetPositionZ()) > INTERACTION_DISTANCE)
+            continue;
+
+        if (!bot->InBattleground() && master && botAI->HasStrategy("follow", BotState::BOT_STATE_NON_COMBAT) &&
+            ServerFacade::instance().GetDistance2d(master, unit) > aggroRange)
+            continue;
```
privatecore, line 49: > If it is a bot with a master, the 30 yards should be centered around the master not the bot (prevent chaining from enemy to enemy)

Not sure if I get it right: all checks except this one are made against `bot` -> `unit` positions, but according to your description, it should be `master` -> `unit`.

src/Ai/Base/Value/AggressiveTargetValue.cpp
hermensbas: Think so as well

src/Ai/Base/Value/AggressiveTargetValue.cpp
dillyns: This is copied from the grind strategy so it follows the same behavior. The bot will pick the closest thing TO THE BOT it can attack that is also within aggrorange of a master when on "follow". It does not attack whoever is closest to its master, it just has to be in the master's range.
The reason this is necessary is because of the following sequence:
Bot goes to attack something 25 yards away -> kills it -> once of combat it detects another enemy 25 yards away -> goes to attack it -> kills it -> potentially repeat as long as an enemy is in range.
The bots can just end up running off from the master if this chain continues.

privatecore: Hmm, this makes sense, but why do we only use this logic outside of the BG?

kadeshar (approved):

PR #2131, opened 2026-02-11
Fix: Logout refactor and RNDbot logout error handling

NoxMax:
# Pull Request

* Improper crash fix by 297f11d3e59f52a29f68188245e6b786c9fa838e. That fix worked not because setting `logout = true` stops the server from crashing, but because it stopped the following `if (!logout)` block from executing at all. In that block `delete target;` was the actual cause of the crashing and this was recreated in testing. `botWorldSessionPtr->LogoutPlayer(true);` already deleted target internally, then comes `delete target;` to delete already freed memory and the whole thing crashes.

* Players had the ability to logout anyone's alt/addClass bots. Now there's a check to make sure command is from master.

* When commanded to logout, RNDbots in player's party used to reply with "I'm logging out!", but they don't, because they shouldn't, because they are not alt/adClass bots. Now they say "You can't command me to logout!"

* Added early exits for the "logout cancel" block, then I remembered bots were made to always instantly logout because of past issues with timed logout. Need to review whether or not we should re-implement timed logout, or if it's not worth it and its dead code removed with instant logout remaining the only option.

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

This PR removes more code than it adds, and makes sure that exits happen as early as possible. It has no effect on processing power and makes the code slightly more maintainable.

---

## How to Test the Changes

1. Whisper `logout` to any bot whose master is not you. The bot can be RND/alt/addClass. The bot may have someone else as a master or may not have a master at all. The bot may be part of a party or not. Regardless, you are not its master. It should tell you "You are not my master!". Two players or two instances of the client from two different accounts are needed for this test, in order for Player A to command a bot to logout, when the bot's master is Player B.

2. Invite an RND bot to your party. As along as it's in your party, you are it's master, but RND bots cannot be logged out through chat commands. If you whisper `logout` to it, it should say "You can't command me to logout!", and not logout.

3. Whisper to an alt/addClass bot `logout`. The bot can be in your party or could've been uninvited. All that matters is that you are its master. It should reply "I'm logging out!"

## Complexity & Impact

- - [x] No

- - [x] No

- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [ ] No
- - [x] Yes (**explain why**)
In that it fixes wrong behavior.

---

## AI Assistance

- - [ ] No
Used Claude for code review, and translations.

If yes, please specify:

---

## Lines to Translate

These are keys and defaults of lines that were added/edited, and to be translated at a later SQL update.

| Key | Default line |
| --- | --- |
| bot_not_your_master | You are not my master! |
| bot_rndbot_no_logout | You can't command me to logout! |
---

## Final Checklist

---

## Notes for Reviewers

before merging.

### DIFF
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -1015,28 +1015,42 @@ void PlayerbotAI::HandleCommand(uint32 type, std::string const text, Player* fro
     }
     else if (filtered == "logout")
     {
-        if (!bot->GetSession()->isLogingOut())
+        if (bot->GetSession()->isLogingOut())
+            return;
+
+        // Verify the command came from this bot's master. Also handles nullptr
+        if (fromPlayer != master)
+        {
+            if (type == CHAT_MSG_WHISPER)
+                bot->Whisper("You are not my master!", LANG_UNIVERSAL, fromPlayer);
+            return;
+        }
+
+        PlayerbotMgr* masterBotMgr = GET_PLAYERBOT_MGR(master);
+        if (!masterBotMgr)
+            return;
+
+        // Only respond if this bot is in master's collection (alt/addclass)
+        if (masterBotMgr->GetPlayerBot(bot->GetGUID()))
         {
             if (type == CHAT_MSG_WHISPER)
                 TellMaster("I'm logging out!");
 
-            PlayerbotMgr* masterBotMgr = nullptr;
-            if (master)
-                masterBotMgr = GET_PLAYERBOT_MGR(master);
-            if (masterBotMgr)
-                masterBotMgr->LogoutPlayerBot(bot->GetGUID());
+            masterBotMgr->LogoutPlayerBot(bot->GetGUID());
         }
+        else if (type == CHAT_MSG_WHISPER)
+            TellMaster("You can't command me to logout!");
     }
     else if (filtered == "logout cancel")
     {
-        if (bot->GetSession()->isLogingOut())
-        {
-            if (type == CHAT_MSG_WHISPER)
-                TellMaster("Logout cancelled!");
+        if (!bot->GetSession()->isLogingOut())
+            return;
 
-            WorldPackets::Character::LogoutCancel data = WorldPacket(CMSG_LOGOUT_CANCEL);
-            bot->GetSession()->HandleLogoutCancelOpcode(data);
-        }
+        if (type == CHAT_MSG_WHISPER)
+            TellMaster("Logout cancelled!");
+
+        WorldPackets::Character::LogoutCancel data = WorldPacket(CMSG_LOGOUT_CANCEL);
+        bot->GetSession()->HandleLogoutCancelOpcode(data);
     }
     else
     {
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -1022,7 +1022,11 @@ void PlayerbotAI::HandleCommand(uint32 type, std::string const text, Player* fro
         if (fromPlayer != master)
         {
             if (type == CHAT_MSG_WHISPER)
-                bot->Whisper("You are not my master!", LANG_UNIVERSAL, fromPlayer);
+            {
+                std::string message = PlayerbotTextMgr::instance().GetBotTextOrDefault(
+                    "bot_not_your_master", "You are not my master!", {});
+                bot->Whisper(message, LANG_UNIVERSAL, fromPlayer);
+            }
             return;
         }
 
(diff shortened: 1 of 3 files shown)


NoxMax:
Sidenote for your consideration, additional error handling options for telling RNDbots to logout ;)

```
else if (type == CHAT_MSG_WHISPER)
{
    static const std::vector<std::string> messages = 
    {
        "You can't command me to logout!",
        "no u.",
        "Make me!",
        "15 more minutes!",
        "How about you logout!",
        "No. No I don't think I will.",
        "You don't get to tell me what to do!",
        "Mom said I can play for one more hour."
    };
    TellMaster(messages[urand(0, messages.size() - 1)]);
}
```

Wishmaster117:
> Sidenote for your consideration, additional error handling options for telling RNDbots to logout ;)
> 
> ```
> else if (type == CHAT_MSG_WHISPER)
> {
>     static const std::vector<std::string> messages = 
>     {
>         "You can't command me to logout!",
>         "no u.",
>         "Make me!",
>         "15 more minutes!",
>         "How about you logout!",
>         "No. No I don't think I will.",
>         "You don't get to tell me what to do!",
>         "Mom said I can play for one more hour."
>     };
>     TellMaster(messages[urand(0, messages.size() - 1)]);
> }
> ```

Love this part, maybe you can add it to database instead off hardcoding them?

src/Bot/PlayerbotAI.cpp
```diff
@@ -1017,28 +1017,42 @@ void PlayerbotAI::HandleCommand(uint32 type, std::string const text, Player* fro
...
     {
-        if (!bot->GetSession()->isLogingOut())
+        if (bot->GetSession()->isLogingOut())
+            return;
+
+        // Verify the command came from this bot's master. Also handles nullptr
+        if (fromPlayer != master)
+        {
+            if (type == CHAT_MSG_WHISPER)
+                bot->Whisper("You are not my master!", LANG_UNIVERSAL, fromPlayer);
```
kadeshar: All text spoke by bots should be translatable.

NoxMax:
> Love this part, maybe you can add it to database instead off hardcoding them?

I honestly added it as a joke comment for the PR post only, but if people want to, I can actually include it.

src/Bot/PlayerbotMgr.cpp
```diff
@@ -391,44 +394,14 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
...
-
-        // Peiru: Allow bots to always instant logout to see if this resolves logout crashes
-        logout = true;
-
-        // if no instant logout, request normal logout
-        if (!logout)
+        if (bot)
         {
-            if (bot->GetSession()->isLogingOut())
-                return;
```
kadeshar, line 401: After this deletion logout is never read. That most above code is useless.

src/Bot/PlayerbotMgr.cpp
```diff
@@ -391,44 +394,14 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
...
-            }
-            else
-            {
-                RemoveFromPlayerbotsMap(guid);     // deletes bot player ptr inside this WorldSession PlayerBotMap
-                delete botWorldSessionPtr;  // finally delete the bot's WorldSession
-                if (target)
-                    delete target;
-            }
-            return;
-        }  // if instant logout possible, do it
```
kadeshar, line 424: After this deletion target is never read.

src/Bot/PlayerbotMgr.cpp
```diff
@@ -391,44 +394,14 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
         {
             target = botAI->GetAiObjectContext()->GetValue<TravelTarget*>("travel target")->Get();
         }
-
-        // Peiru: Allow bots to always instant logout to see if this resolves logout crashes
-        logout = true;
-
-        // if no instant logout, request normal logout
-        if (!logout)
+        if (bot)
```
kadeshar: This condition is good? Its duplicate condition from 346 line

NoxMax:
@kadeshar I had addressed those in an unpushed commit that puts timed logout back, though I haven't tested and we need to discuss whether or not timed logouts should be there at all.
Anyway, I cherry picked those changes that address your points. I kept the logout lines though but commented them out; if timed logout is not added in the short term, that segment will be useful in the future if and when it's added.

kadeshar:
I only confirm how we handle now new translations. Then i approve.

kadeshar:
I consulted that with others and we require script which first DELETE all rows and then INSERT rows to playerbots text tables which are introduced by PR similar way like is handled in azerothcore scripts. Scripts should be based on id.

NoxMax:
Would the translations be consolidated? I think there should a single continuously updated translation file in the updates, per language of course to make things clear.

kadeshar:
Decision was to no maintain single translation file.

Celandriel:
Whats the status of this?

NoxMax:
Good to go. I might actually add the snarky bot refusal lines in another PR. And more importantly, bring back timed logouts if people want that option for bots.

kadeshar:
Like i wrote before in PR required is update sql script if you adding translation

Celandriel (approved):

PR #2133, opened 2026-02-11
Added unobtainable items to config

kadeshar:
# Pull Request

Moving hardcoded values to config

## How to Test the Changes

- use maintenance command
- unequip and destroy item get from this command
- turn off server
- add item to config
- turn on server
- use maintenace command
- check that different item was provided

## Complexity & Impact

- - [X] No

- - [X] No

- - [X] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [X] No
- - [ ] Yes (**explain why**)
---

## AI Assistance

- - [X] No

---

## Final Checklist

- - [X] Stability is not compromised
- - [X] Performance impact is understood, tested, and acceptable
- - [X] Added logic complexity is justified and explained
- - [X] Documentation updated if needed

---


brighton-chi:
Why is Sunwell Orb excluded? Isn't it obtainable?

NoxMax:
It's a quest item, but otherwise has +3 INT, somewhat comparative with items of its ilvl. Probably shouldn't be de-listed. Regardless, it is from a much older commit. No context was given about it https://github.com/mod-playerbots/mod-playerbots/pull/287/changes/009be18646b5acb7c0b126ee9d79a14f6d6cf9a9#diff-cb0f593cc88de36ed662158afbc2232a686075ccee93d6e4e26dbb3a732e7b88

kadeshar:
> Why is Sunwell Orb excluded? Isn't it obtainable?

Its not my addition. But based on wowhead its quest item which loose power after quest completition. And its horde only item?

NoxMax:
the +3 INT is there regardless of the quest.

brighton-chi:
I think default exclusions should only be for items that aren't normally obtainable like GM and test items.

hermensbas (approved):

Issue #1053, opened 2025-03-03
Mounted bots with Path of Frost aura unaffected


Vortikai:
**Bug Description**
A clear and concise description of what the bug is. If the bug is a crash, a crash log must be posted or the issue will be removed.

A few months ago, there was an update that allowed the bots to swim. I suspect they are still using this swim state while mounted even if Path of Frost, the DK ability that allows the character to walk on water while mounted, is on them.

**Commit Hash**
The hash of the current commit.

**How To Reproduce Bug**
Detailed steps to reproduce the behavior.

1. Have an altbot.
2. Cast Path of Frost
3. Mount
4. Walk on water

**Expected Behavior**

When this aura is active, bots should not change to a swimming speed state while crossing water, and instead walk on the water while mounted.

**Screenshots**

**Modules**
Please list all modules used as many are known to cause conflicts with Playerbots.

**Playerbot Settings**
Number of bots, scaling settings, etc if performance related.

**System**
OS: [e.g. Windows, Linux]
Hardware: [e.g. CPU if performance related]

Linux

**Additional Context**

Vortikai:
I'm upgrading now.

Vortikai:
Using the latest commit now. Yeah they're walking on water, but not running or normal ground mount speeds.

Vortikai:
This isn't just for mounted states, so I am changing the title and bug description

NoxMax:
Have you tested things recently? I just did, and I got the exact inverse of your issue. I cast Water Walking on myself and one of the alts. Me and the casted on alt used appropriate walk/running speed. Meanwhile the alt without the buff, could also keep up with us mounted or running.

![Image](https://github.com/user-attachments/assets/c3e0b126-c7a8-4f02-ae16-1ac21d50953d)

PR #2134, opened 2026-02-12
Flying, waterwalking, swimming movement optimizations and transition fixes.

hermensbas:
# Pull Request

**Fixes and optimizations for flying, water walking, swimming**:
* optimized triggers
* ensuring movement flag updates only happen between actual transitions states 
* fly bug fix; fly with bots following with stay command midair, fly down and dismount yourself, follow command and now the bots fall instead of lingering around in the air)
* updated z-axes correction for water walking and bots (for real players this is handled client-side)
* added lift off movement for more stabile transition from ground(level) to flying 

**Tested**:
* Test all transitions; water walk, swimming, swimming, walking, mounting while water walking etc.
* Flying with bots and fly master routes
* Movement flag updates only occur during transitions

**Known issues**: transition between water walking, swimming and back again, in most cases the bots will stay under the waterline instead of jumping on the z axes on water level. (will fix that another time)

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

---

## How to Test the Changes

Apply water walking effect on your bots, shaman or dk, and test all possible transitions and follow actions 
of the bots. water walking, swim, walk on land, swimming and walk without water walking effect/aura, fly mount from water, from ground, etc.

## Complexity & Impact

- - [x] No

- - [x] No

- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [x] No
- - [ ] Yes (**explain why**)

---

## AI Assistance

- - [x] No

If yes, please specify:

---

## Final Checklist

---

## Notes for Reviewers

before merging.

### DIFF
--- a/src/Ai/Base/Actions/MovementActions.cpp
+++ b/src/Ai/Base/Actions/MovementActions.cpp
@@ -953,11 +953,8 @@ bool MovementAction::Follow(Unit* target, float distance) { return Follow(target
 
 void MovementAction::UpdateMovementState()
 {
-    const bool isCurrentlyRestricted = // see if the bot is currently slowed, rooted, or otherwise unable to move
-        bot->HasUnitState(UNIT_STATE_LOST_CONTROL) ||
-        bot->IsRooted() ||
-        bot->isFrozen() ||
-        bot->IsPolymorphed();
+    const bool isCurrentlyRestricted =  // see if the bot is currently slowed, rooted, or otherwise unable to move
+        bot->HasUnitState(UNIT_STATE_LOST_CONTROL) || bot->IsRooted() || bot->isFrozen() || bot->IsPolymorphed();
 
     // no update movement flags while movement is current restricted.
     if (!isCurrentlyRestricted && bot->IsAlive())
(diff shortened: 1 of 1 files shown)


src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -953,84 +953,89 @@ bool MovementAction::Follow(Unit* target, float distance) { return Follow(target
...
-        if (wantsToFly && !isFlying && masterIsFlying)
+        // handle flying
+        if (wantsToFly && !isFlying && isMasterFlying)
         {
             bot->AddUnitMovementFlag(MOVEMENTFLAG_CAN_FLY);
             bot->AddUnitMovementFlag(MOVEMENTFLAG_DISABLE_GRAVITY);
             bot->AddUnitMovementFlag(MOVEMENTFLAG_FLYING);
-            movementFlagsUpdated = true;
+
+            // required for transision and state monitoring.
```
brighton-chi: transition!

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -1813,12 +1818,11 @@ void MovementAction::DoMovePoint(Unit* unit, float x, float y, float z, bool gen
     if (!mm)
         return;
 
-    // enable water walking
-    if (unit->HasUnitMovementFlag(MOVEMENTFLAG_WATERWALKING))
+    // bot water collission correction
```
brighton-chi: "collision"

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -953,84 +953,87 @@ bool MovementAction::Follow(Unit* target, float distance) { return Follow(target
...
-            // swimming
-            else if (wantsToSwim && !isSwimming && masterIsSwimming)
+            else if ((!canWaterWalk || isMasterSwimming) && isWaterWalking)
             {
                 bot->RemoveUnitMovementFlag(MOVEMENTFLAG_WATERWALKING);
-                bot->AddUnitMovementFlag(MOVEMENTFLAG_SWIMMING);
+                if (canSwim)
+                    bot->SetSwim(true);
+                movementFlagsUpdated = true;
+            }
```
NoxMax: While you are here, could you add
```
else if (canSwim && (!canWaterWalk || isMasterSwimming) && !isSwimming)
{
    bot->SetSwim(true);
    movementFlagsUpdated = true;
}
```
This should stop bots from keeping up with master's speed, if they don't have Water Walk (I found out that they can these days after checking the status of #1053)

Lastly, they can also keep up with master if they are underwater and have the buff. Adding `liquidState == LIQUID_MAP_UNDER_WATER` to the if/else if before that should fix this. But, not sure that should be fixed, because then you need to have bots jump out of water so they get above it. Not worth it.

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -953,84 +953,87 @@ bool MovementAction::Follow(Unit* target, float distance) { return Follow(target
...
         {
             bot->AddUnitMovementFlag(MOVEMENTFLAG_CAN_FLY);
             bot->AddUnitMovementFlag(MOVEMENTFLAG_DISABLE_GRAVITY);
             bot->AddUnitMovementFlag(MOVEMENTFLAG_FLYING);
-            movementFlagsUpdated = true;
+
+            // required for transition and state monitoring.
+            if (MotionMaster* mm = bot->GetMotionMaster())
+                mm->MoveTakeoff(0, {bot->GetPositionX(), bot->GetPositionY(), bot->GetPositionZ() + 1.F}, 0.F, true);
         }
```
kadeshar, line 1018: missing sets on movementFlagsUpdated. Its ok?

hermensbas: yes, the moveTakeOff handles that. We only need to add it again if for some reason i wanna remove moveTakeOff. Sharp notice though but this was intensional since we dont wanna interupt the takeoff effect and animation.

hermensbas: In general when sending duplicate or multiple movement flags can interupt/break the current effect. For example waterwalk effect breaks or could be as small having strange little animations interupts when entering the water. Hence  its kinda fragile. 

Most of these effects are either intended for server-side creature handling or handled in total or partial in client througt ops calls. Thats one of the reasons i am not using player->setWaterWalk since it sends RPC/opscode to the client which doesnt exists in case of a bot.

There are still some effects that dont with bots yet, for example levitate of a priest.

PR #2138, opened 2026-02-13
Updates to SSC Strategies

brighton-chi:
# Pull Request

Most of the changes are not functional but are to modify style based on comments received to the TK PR (e.g., eliminate nesting of if statements) and leverage general boss helpers. There is some reordering of returns and other changes to try to consolidate or clean-up the code (such as removing unnecessary parameters).

The strategies themselves have only minor changes.

- Main tank no longer uses tangential movement for Lurker Spout, unlike other bots. The MT will just call moves directly to a position behind Lurker. This is because I found tangential movement was taking too long for the MT to get in place since it starts right in front of Lurker.
- Vashj MT group Shaman will now use Grounding Totem without actually switching to the Grounding Totem strategy. I have now eliminated all strategy swaps, which I dislike because they persist after the encounter, and it's better not to mess with player strategies since presumably people are generally using Windfury or Wrath of Air for the Air Totem.
- I made a ton of changes to Vashj core passing as I noticed the existing logic is nonfunctional in several ways. It generally works fine ingame, but the changes should make things much smoother. For example, the storing of the nearest trigger NPC for generators in the existing strategy is useless because it relies on insert_or_assign for an unordered map that will continue to run during the course of the core passing logic, and a similar issue exists with respect to the map to store the last time a bot held a core. The result is that there is slight movement of bots when the core is flying through the air and not held by any bot because the trigger for core passing does not fire during that period. In practice, the time is brief enough that the sequence works OK, but it looks stupid because the bots should not be moving at all. So that should be fixed.

There is a known issue re: core passing that would take extreme effort to fix and I am not going to do it because it is a fringe situation. There are a couple of spots where the Tainted Elemental can rarely spawn that can result in a straight-line passing sequence to the nearest generator that is blocked by LoS. Fixing this would be extremely difficult and niche, so what you will need to do if this happens is to just command your bots to destroy the core and try again with the next spawn.

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

I have attempted to streamline methods and even remove some. The strategy is admittedly somewhat performance heavy due to the need for function calls such as iterating over inventory items. However, the new version should be less performance intensive than the merged strategy--for example, there were places where all members of the raid would have their inventory checked, but I've now limited the check to only the 5 core handler bots. I've run the instance with pmon on, and there are no methods that stand out as particularly resource intensive when not in a boss encounter, and I view that as the most important thing (though I make effort to reduce performance impact during encounters also). Expensive checks that are unavoidable for the strategy to work such as grid searches are gated behind cheaper checks.

---

## How to Test the Changes

Enter SSC with a raid group and run the instance, including all bosses. Every boss should be killable, and every major mechanic should be addressed by bots. I will work with Dreathean to get the Wiki up soon so that should be a reference for testing strategies.

## Complexity & Impact

- - [ ] No

Only within the context of strategies, which are basically all new decision branches, and there are some tweaks here to what is currently merged.

- - [x] No

- - [ ] No
- - [x] Yes (**explain why**)

I'm sure if you have a large server, with multiple raid groups running the instance at the same time, the performance impact could be significant. But I have done my best to limit it, and I think some notable performance impact is unavoidable with the current framework for raid strategies.

---

## Defaults & Configuration

- - [ ] No
- - [x] Yes (**explain why**)

Only for strategies in the instance.

- - [ ] Lightweight mode remains the default

There should be no impact if co +ssc and nc +ssc are not added to bots. Because raid strategies are currently not removed after leaving an instance, players should manually remove them (or reset botAI). This is a general issue that needs to be addressed with the module.

---

## AI Assistance

- - [ ] No

If yes, please specify:

GPT-4 because I don't like to use up my premium requests in CoPilot and I generally like it better than GPT-5 =P

I use LLMs to draft code snippets but do review everything and have become less-and-less reliant over time. I don't use agent mode, only ask. For this PR, I had it do the updated version of AnyRecentCoreInInventory(), which is more complicated than before and uses indexing for each bot to consider status of only prior bots in the passing chain. Everything else either I wrote or could have written but had the AI help and just edited afterward to save time.

---

## Final Checklist

---

## Notes for Reviewers

before merging.


src/Ai/Raid/SerpentshrineCavern/Action/RaidSSCActions.cpp
```diff
@@ -19,42 +19,40 @@ bool SerpentShrineCavernEraseTimersAndTrackersAction::Execute(Event /*event*/)
...
-            erased = true;
+        erased = true;
     }
-    if (!AI_VALUE2(Unit*, "find target", "lady vashj"))
+
+    if (!AI_VALUE2(Unit*, "find target", "lady vashj") &&
+         hasReachedVashjRangedPosition.erase(guid) > 0)
     {
-        if (vashjRangedPositions.erase(guid) > 0)
-            erased = true;
```
trolloks, line 55: Seems like you missed this case?

```
vashjRangedPositions.erase(guid) > 0
```

brighton-chi: Oops, I deleted that map but missed the declaration in the helper header actually.

src/Ai/Raid/SerpentshrineCavern/Action/RaidSSCActions.cpp
```diff
@@ -2408,119 +2150,99 @@ bool LadyVashjPassTheTaintedCoreAction::Execute(Event /*event*/)
...
         }
         // First core passer: receive core from looter at the top of the stairs,
         // pass to second core passer
-        else if (bot == firstCorePasser)
+        else if (bot == firstCorePasser &&
+                 IsSecondCorePasserInPosition(firstCorePasser, secondCorePasser, closestTrigger))
         {
             const time_t now = std::time(nullptr);
             auto it = lastImbueAttempt.find(instanceId);
-            if (it == lastImbueAttempt.end() || (now - it->second) >= 2)
```
kadeshar, line 2462: There is no possibility to crash due this deletion and UseCoreOnNearestGenerator?

brighton-chi: I was thinking it's not really possible that there wouldn't be an imbue attempt because the order is fixed so the designated core looter always has to imbue before any passers can.

But also there's not any reason not to have the check I suppose in case something really weird happens. So I'll add it back.

kadeshar (dismissed):

brighton-chi:
Sorry to dismiss review. I noticed a misspelling of an ability. Funny that Blizzard, an American company, would use the UK English "judgement" instead of the American English "judgment."

Edit: Apparently they actually changed the spelling to "judgment" in MoP. Hah.

kadeshar (approved):

PR #2147, opened 2026-02-16
Paladin Seal of wisdom fallback fix for Ret/Prot Paladins

dillyns:
# Pull Request

This PR removes the fallback for Seal of Wisdom action from generic Paladin strategy and moves it to Holy Paladin only. This is necessary because a paladin who does not have Seal of Wisdom yet who goes low mana and triggers the seal of wisdom action will end up falling back to Seal of Righteousness, even though a better seal may be available, such as Seal of Command.
The fallback is added to Holy Paladin so that low level holy paladins still use Righteousness until they get Wisdom

---

## Feature Evaluation

Please answer the following:

Ret paladin without Seal of Wisdom shouldn't change seals on low mana.
Cheapest implementation is to remove seal of wisdom fallback for non-holy paladins.
No difference in cost compared to existing logic.
---

## How to Test the Changes

Use a ret paladin bot who has Seal of Command but who does not have Seal of Wisdom. A paladin under level 38 will do.
Order them to attack something, like a test dummy, until they eventually run low on mana.

Before this change: The paladin will switch to Seal of Righteousness when they get low mana.
After this change: The paladin leaves Seal of Command on when they get low mana.

## Complexity & Impact

- - [x] No

- - [x] No

- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [x] No
- - [ ] Yes (**explain why**)

---

## AI Assistance

- - [x] No

If yes, please specify:

---

## Final Checklist

---

## Notes for Reviewers

before merging.


Celandriel (dismissed):

Celandriel:
Adding it to the strategy isnt the right way to do this. The strategy will try to execute both actions every time this gets triggered.

dillyns:
> Adding it to the strategy isnt the right way to do this. The strategy will try to execute both actions every time this gets triggered.

Whoops, yes my mistake should be a fallback on the action not the strategy. Will fix this when I have some time

Celandriel (approved):

Issue #2148, opened 2026-02-16
After deleting all random bot guilds and arena teams using the configuration options, the server fails to recreate new guilds


Wishmaster117:
**Bug Description**  
After deleting all random bot guilds and arena teams using the configuration options, the server fails to recreate new guilds after restoring the settings. Upon server restart, the message **"No guilds found in database, resetting guild cache"** appears, and no new guilds are generated.

**Commit Hash**  
AzerothCore rev. 8b0c82c3147d 2026-01-30 21:54:26 +0100 (Playerbot branch) (Win64, RelWithDebInfo, Static)

**How To Reproduce Bug**  
1. Start from a fresh server installation.  
2. Configure and run the server with:  
   - 200 bots  
   - 14 guilds created  
   - 200 bots assigned to guilds  
3. Use the following configuration:  
   ```
   AiPlayerbot.RandomBotGuildCount = 20
   AiPlayerbot.RandomBotGuildSizeMax = 15
   ```  
   → Guild creation works correctly.  
4. Change configuration to:  
   ```
   AiPlayerbot.DeleteRandomBotArenaTeams = 1
   AiPlayerbot.DeleteRandomBotGuilds = 1
   ```  
   → All guilds and arena teams are deleted as expected.  
5. Revert the configuration back to:  
   ```
   AiPlayerbot.DeleteRandomBotArenaTeams = 0
   AiPlayerbot.DeleteRandomBotGuilds = 0
   ```  
6. Restart the server.  
7. Observe the error message:  
   ```
   No guilds found in database, resetting guild cache
   ```  
   → No new guilds are created.

**Expected Behavior**  
When the deletion settings are disabled again, the server should recreate new random bot guilds according to the configured values (`RandomBotGuildCount` and `RandomBotGuildSizeMax`).

**Screenshots**  
(Attach screenshots if relevant.)

**Modules**  
(Please list all modules you are using, as some may conflict with Playerbots.)

**Playerbot Settings**  
- 200 bots  
- Random bot guild count: 20  
- Random bot guild size max: 15  
- DeleteRandomBotArenaTeams toggled from 1 back to 0  
- DeleteRandomBotGuilds toggled from 1 back to 0  

**System**  
OS: Windows 11

**Additional Context**  
After deleting random bot guilds and arena teams, the system does not recreate them even when the deletion settings are disabled and the server is restarted. The guild cache resets, but no new guilds are generated.

Celandriel:
[...]

PR #2149, opened 2026-02-16
Restore bot guilds and arena teams creation logic when guild data is wiped

Wishmaster117:
# Pull Request

This change improves bot guild and arena  teams recovery and assignment after guild/arena teams, resets/restarts, with a focus on keeping logic simple, centralized, and low-cost.

### What changed
- Added a public `EnsureGuild()` entry point in `PlayerbotFactory` to centralize the guild initialization gate (`randomBotGuildCount > 0`) and avoid duplicated checks.
- Kept `InitGuild()` as the single place for guild assignment/creation behavior.
- On bot login, call `factory.EnsureGuild()` so bots can recover guild state after restart (including cases where guild data was reset).
- Centralized orphan cleanup in `PlayerbotGuildMgr::ValidateGuildCache()`:
  - `DELETE FROM guild_member WHERE guildid NOT IN (SELECT guildid FROM guild)`
- Simplified runtime cache reset in `ResetGuildCache()` by clearing cache snapshots and marking guild names available before rebuilding from DB.

### Why it is needed
- Prevent bots from remaining in stale guild state after DB resets/restarts.
- Avoid duplicated cleanup logic across multiple call paths.
- Keep behavior predictable and easier to maintain upstream.

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

  Add one centralized entrypoint (`EnsureGuild`), one centralized orphan cleanup query in guild cache validation, and one login-time call to trigger existing guild assignment flow.

  Reuse existing `InitGuild()` and `AssignToGuild()` logic, avoid new systems, and run a single bounded cleanup query on guild validation.

  Low: one cheap config-gated function call per bot login and one simple cleanup SQL during guild cache validation; no per-tick heavy logic added.

---

## How to Test the Changes

  1. Set:
     - `AiPlayerbot.RandomBotGuildCount` > 0
     - (Optional) `AiPlayerbot.DeleteRandomBotGuilds = 1`
  2. Start worldserver and allow random bots to log in.
  3. Verify bots receive/restore guild/arena teams assignment after login.
  4. Simulate reset case:
     - Empty or remove `guild` rows (and/or leave orphan `guild_member` rows) and do the same fror 'anena_team' and 'arena_team_member'.
     - Restart worldserver.
     - Verify orphan memberships are cleaned and bots get reassigned/created into guilds and arena teams.
  5. Confirm no compile errors and no guild/arena-related startup regressions.

  - Random bot accounts enabled.
  - Guild names table populated (`playerbots_guild_names`).

  - Expected: bots no longer remain stuck in stale guild/arena state after restart/reset.
  - Verify via startup/log traces and in-game guild membership checks.

## Complexity & Impact

- - [x] No

- - [x] No

- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [ ] No
- - [x] Yes (**explain why**)

Explanation: guild recovery is now explicitly triggered on bot login via `EnsureGuild()`, but remains config-gated by existing `RandomBotGuildCount`.

---

## AI Assistance

- - [x] No

---

## Final Checklist

---

## Notes for Reviewers

This patch intentionally favors centralized, low-cost recovery behavior over broader behavioral changes.  
No heavy per-tick logic was introduced; the change remains localized to guild recovery and cache validation paths.

Close : https://github.com/mod-playerbots/mod-playerbots/issues/2148


NoxMax:
Do arena teams deal with the same issue when wiped? I have not checked at all to be honest, but if they do, should there be a recreation process for them same as guilds?

src/Mgr/Guild/PlayerbotGuildMgr.cpp
```diff
@@ -189,6 +188,10 @@ void PlayerbotGuildMgr::LoadGuildNames()
 
 void PlayerbotGuildMgr::ValidateGuildCache()
 {
+    // Defensive cleanup for DB states where guild rows were removed manually.
+    // Orphaned guild_member entries can otherwise keep bots in a stale guild state.
+    CharacterDatabase.Execute("DELETE FROM guild_member WHERE guildid NOT IN (SELECT guildid FROM guild)");
```
Celandriel: I prefer not modifying tables that are owned by the core.

Wishmaster117:
> Do arena teams deal with the same issue when wiped? I have not checked at all to be honest, but if they do, should there be a recreation process for them same as guilds?

No arena teams are ok

Celandriel:
> No arena teams are ok

Do you know why its not an issue with arenas? I cant tell from the code. It looks to me that they would both require bots to be initialized again, and actually the arena code states as much
```
src\Bot\Factory\PlayerbotFactory.cpp:
  4070      // A manual reinitalization (.playerbots rndbot init) is also required after the teams have been deleted.
  ```

**Comment by Wishmaster117** [comment]
> > No arena teams are ok
> 
> Do you know why its not an issue with arenas? I cant tell from the code. It looks to me that they would both require bots to be initialized again, and actually the arena code states as much
> 
> ```
> src\Bot\Factory\PlayerbotFactory.cpp:
>   4070      // A manual reinitalization (.playerbots rndbot init) is also required after the teams have been deleted.
> ```

I admit I didn’t check the arena part at first, because when I looked at the table it was already filled. After testing more thoroughly, it seems that setting "AiPlayerbot.DeleteRandomBotArenaTeams = 1" doesn’t actually delete them, since I still have the same arena team names after restarting the server. That’s why I initially thought there was no issue with arena team creation.
But if we manually truncate the arena_team and arena_team_member tables, they are indeed not recreated either.
So for arenas, there are two problems:
1) The tables are not deleted when "AiPlayerbot.DeleteRandomBotArenaTeams = 1" is set.
2) If we delete them manually, they are not recreated.

**Comment by Wishmaster117** [comment]
Ok arena patched too; have a good night.

File: src/Bot/Factory/PlayerbotFactory.cpp [hunk 2]
```diff
@@ -3924,13 +3925,32 @@ void PlayerbotFactory::InitInventoryEquip()
...
+
 void PlayerbotFactory::InitGuild()
 {
-    if (bot->GetGuildId())
+    if (uint32 guildId = bot->GetGuildId())
     {
-        if (!bot->HasItemCount(5976, 1) && bot->GetLevel() > 9)
-            StoreItem(5976, 1);
-        return;
+        if (!sGuildMgr->GetGuildById(guildId))
```
Celandriel, line 4549: In this code checking for GetGuildById isnt necessary since we dont interact with it. I think its unlikely that the core make a mistake when loading a bot's membership in a guild.
I suggest restoring this block.

src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -3924,13 +3925,32 @@ void PlayerbotFactory::InitInventoryEquip()
     }
 }
 
+void PlayerbotFactory::EnsureGuild()
+{
+    // Keep the config gate in a single place so external callers do not duplicate checks.
+    if (sPlayerbotAIConfig.randomBotGuildCount > 0)
```
Celandriel: Instead of this approach, what if we created a `bool needMoreGuild` in playerbotguildmgr, and if true, then on login send an initguild?

src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -4047,12 +4067,35 @@ void PlayerbotFactory::InitImmersive()
     }
 }
 
+void PlayerbotFactory::EnsureArenaTeams()
+{
+    // Keep the config gate centralized to avoid duplicated checks at call sites.
+    if (sPlayerbotAIConfig.randomBotArenaTeam2v2Count ||
```
Celandriel: Can we check size of sPlayerbotAIConfig.randomBotArenaTeams as a gate?

src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -4047,12 +4067,35 @@ void PlayerbotFactory::InitImmersive()
...
+}
+
 void PlayerbotFactory::InitArenaTeam()
 {
 
     if (!sPlayerbotAIConfig.IsInRandomAccountList(bot->GetSession()->GetAccountId()))
         return;
 
+    // Defensive cleanup for manual DB edits that removed teams but left member rows behind.
+    CharacterDatabase.Execute("DELETE FROM arena_team_member WHERE arenaTeamId NOT IN (SELECT arenaTeamId FROM arena_team)");
```
Celandriel: I dont think this is the place to do this. 
Right now account deletions are handled by  RandomPlayerbotFactory::CreateRandomBots. I had placed the guild account management into the guild manager, but since arenas dont have that right now, a  unified place that manages resetting accounts, arenas etc. and then shuts down the server makes the most sense. 

The bigger issue generally is knowing what information is 'true'. What I mean is, when the server is not running, then the database is the current state of the world, however, when it is running, then the database no longer represents the latest state in the world.

src/Bot/Factory/PlayerbotFactory.h
```diff
@@ -85,6 +85,8 @@ class PlayerbotFactory
     void InitKeyring();
     void InitReputation();
     void InitAttunementQuests();
+    void EnsureGuild();
```
Celandriel: Im not sure these are needed at all.

src/Mgr/Guild/PlayerbotGuildMgr.cpp
Wishmaster117: Ok removed

src/Bot/Factory/PlayerbotFactory.cpp
Wishmaster117: I get the point about trusting core guild loading in the nominal path, but this check is not about second-guessing core logic, it is about handling a cross-layer desync scenario safely.

bot->GetGuildId() is a persisted player field, while sGuildMgr->GetGuildById(guildId) depends on the in-memory guild registry being consistent with DB state at runtime. Those are not strictly equivalent invariants. 
If GetGuildId() is non-zero but the guild object is missing, early-returning from InitGuild() leaves the bot in a stale membership state and prevents reassignment flow. 
The current guard explicitly converts that invalid state into a recoverable one (SetInGuild(0), SetGuildIdInvited(0)), then continues through the standard assignment/create path. That is a local, idempotent, low-cost sanity gate.

From a module-resilience perspective, this is preferable to assuming global consistency forever, especially since we intentionally removed SQL-side cleanup in ValidateGuildCache() to keep module code read-only on core tables. In other words, this runtime check is now the minimal defensive layer preventing stale guild IDs from short-circuiting bot guild initialization.

So I would keep the GetGuildById validation in InitGuild(): it does not change nominal behavior, and it avoids a hard-to-debug stuck state in non-nominal data conditions.

src/Bot/Factory/PlayerbotFactory.cpp
Wishmaster117: I agree that introducing a new needMoreGuild state in PlayerbotGuildMgr would be a broader design change.
For this PR, the goal is to restore broken runtime behavior with the smallest safe patch, not to redesign guild orchestration.
EnsureGuild() is intentionally a thin wrapper:
it keeps the existing config gate in one place (randomBotGuildCount > 0),
it reuses current InitGuild() logic,
it is already called on login, so recovery happens where the issue is observed.

Adding a global needMoreGuild flag would require defining and maintaining extra state transitions (guild create/delete/member updates/login/reload paths), which is outside the scope of this bugfix and increases regression risk.

So for this change set, I prefer keeping the local, minimal, operational fix and avoiding system-level refactor.

src/Bot/Factory/PlayerbotFactory.cpp
Wishmaster117: yeah but my take is: no, we shouldn’t gate this solely on randomBotArenaTeams.size().

randomBotArenaTeams is a mutable runtime cache, not the source of truth for the functional intent.
The actual intent behind “enable arena bots” is expressed by the config counts (2v2 / 3v3 / 5v5).

Inside InitArenaTeam, there’s already an explicit comment explaining that this vector can legitimately be empty after a restart, and that the recreation logic is supposed to trigger precisely when it’s empty (empty() -> create).
So if we gate EnsureArenaTeams() only on randomBotArenaTeams.size(), we risk short‑circuiting the recovery path after a restart or a purge.

The current implementation of EnsureArenaTeams() — gated by the config counts — is consistent with a “feature enabled by config” behavior. It’s more stable than relying on a runtime container that can be temporarily empty or incomplete.

I would keep the current config‑based gate to avoid breaking operational recovery.

To incorporate this idea, the only safe variant would be a hybrid gate (config first, then runtime state), but not size() alone.

src/Bot/Factory/PlayerbotFactory.h
Wishmaster117: In this case, EnsureGuild() / EnsureArenaTeams() are intentional thin wrappers, not duplicate logic.
They centralize the config gating (policy) at a single entry point, while InitGuild() / arena init keep the operational logic.
This avoids duplicating conditional checks at call sites and keeps the login flow explicit and maintainable.
They are also actively used on bot login (factory.EnsureGuild(); and factory.EnsureArenaTeams();), so they are not unused methods.
Given the current bugfix scope operational recovery, minimal refactor, keeping these wrappers is the safer and cleaner option.

src/Bot/Factory/PlayerbotFactory.cpp
Wishmaster117: You’re right about placement: this SQL cleanup does not belong in PlayerbotFactory::InitArenaTeam().
I removed the DELETE FROM arena_team_member ... from this path so arena init remains runtime-focused in-memory team consistency and creation flow, without mutating core-owned tables during bot login/init.

Celandriel:
So I would like to make it so the arena cache is properly initialized on startup before. We can pull it from the sql or it would be better if the core has an arena holder that we can parse. Check if the captain is a bot, and if so add to the cache. 

Its my preference to eliminate the ensure functions and just check it in the init, so it covers tracking for both the factory and the login process.

I know it's beyond a quick fix, but it's a more complete solution.

hermensbas:
> So I would like to make it so the arena cache is properly initialized on startup before. We can pull it from the sql or it would be better if the core has an arena holder that we can parse. Check if the captain is a bot, and if so add to the cache.
> 
> Its my preference to eliminate the ensure functions and just check it in the init, so it covers tracking for both the factory and the login process.
> 
> I know it's beyond a quick fix, but it's a more complete solution.

That can still be done, but for now wont this fix the current issues on hand?

Wishmaster117:
> So I would like to make it so the arena cache is properly initialized on startup before. We can pull it from the sql or it would be better if the core has an arena holder that we can parse. Check if the captain is a bot, and if so add to the cache.
> 
> Its my preference to eliminate the ensure functions and just check it in the init, so it covers tracking for both the factory and the login process.
> 
> I know it's beyond a quick fix, but it's a more complete solution.

Hi, is taht what You 're talking about? https://github.com/mod-playerbots/mod-playerbots/pull/2149/changes/41b8267577a3f1fe7a6b6b8de0b634cd376f37c9

PR #2150, opened 2026-02-16
Master update from Test staging

Celandriel:
Candidate PR for updating master. 

AI Summary of changes:
Thread-Safe PVP Remake, Movement Optimizations, and Action Validation Fixes

This PR merges a series of critical stability updates, technical refactors, and movement optimizations from the test branch to master.
## Summary of Changes
Core Engine & Stability

* Thread-Safe PVP Remake: Re-implemented the PVP probability system using deterministic hashing.
* Action Validation Logic: Fixed flaw in action processing. The engine now correctly evaluates isPossible before isUseful, resolving validation order issues for triggers and specific actions.
* Modern C++ Refactor: Replaced the newrpginfo data union with std::variant. This prevents memory corruption and simplifies the handling of non-trivial data sets like flight paths.

Movement & World Interaction
* Movement Flag Optimizations: Optimized triggers for flying, swimming, and water walking. Transition updates now only occur during actual state changes to reduce overhead.
* Z-Axis & Fly Bug Fixes: Added "lift-off" movement for stable ground-to-air transitions and fixed a bug where following bots would "linger" in the air after the master dismounted.
* Standardization: Refactored WorldPosition method names (e.g., getX → GetPositionX) to align with AzerothCore coding standards.

Combat & Strategy Refinement
* Magtheridon Fix: Resolved an issue where cube clickers would stop attacking between Blast Novas due to a state-tracking timer failure.
* Equip Action Registry: Fixed a registry bug where the correct equipment action was not triggering; renamed the action for better clarity.

Code Quality & Maintenance
* Warning Elimination: Conducted a multi-stage cleanup to remove unused event variables in function bodies and removed unnecessary std::move calls on return vectors to allow for RVO (Return Value Optimization).

### DIFF
--- a/src/Ai/Base/Actions/BuyAction.cpp
+++ b/src/Ai/Base/Actions/BuyAction.cpp
@@ -175,7 +175,7 @@ bool BuyAction::Execute(Event event)
 
                     if (needMoneyFor == NeedMoneyFor::gear)
                     {
-                        botAI->DoSpecificAction("equip upgrades");
+                        botAI->DoSpecificAction("equip upgrades packet action");
                     }
                 }
             }
@@ -206,7 +206,7 @@ bool BuyAction::Execute(Event event)
                 if (usage == ITEM_USAGE_REPLACE || usage == ITEM_USAGE_EQUIP ||
                     usage == ITEM_USAGE_BAD_EQUIP || usage == ITEM_USAGE_BROKEN_EQUIP)
                 {
-                    botAI->DoSpecificAction("equip upgrades");
+                    botAI->DoSpecificAction("equip upgrades packet action");
                     break;
                 }
             }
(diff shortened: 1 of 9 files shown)


NoxMax:
Bash's revert is already in master. Should be pushed onto test-staging as well though.

Celandriel:
> Bash's revert is already in master. Should be pushed onto test-staging as well though.

Its in both, I didnt mention it in the PR because of that.

NoxMax:
Wait I just double checked again. It shows it's not in test-staging 
Untitled

Celandriel:
Maybe the comparison is being weird, but each of them has a commit that removed the db changes.

NoxMax:
Oh sorry you're right. It just has a different commit hash. I will never wrap my head around git.

Untitled

hermensbas (approved):

PR #2155, opened 2026-02-19
Fix quest links triggering trade window

Hokken:
## Summary

`ChatHelper::parseable()` matched any hyperlink containing `|H`, including quest links (`|Hquest:`), achievement links, spell links, etc. This caused bots to interpret quest links shared in party chat as item trade requests, opening the trade window instead of ignoring them.

Narrowed the check from `"|H"` to `"|Hitem:"` so only actual item links trigger the parseable/trade logic.

**One-line change** in `src/Bot/Cmd/ChatHelper.cpp:603`

## Root Cause

The WoW client uses `|H<type>:<id>|h[Name]|h` hyperlinks for many object types:
- `|Hitem:12345|h[Item Name]|h` — items
- `|Hquest:678|h[Quest Name]|h` — quests  
- `|Hspell:890|h[Spell Name]|h` — spells
- `|Hachievement:...|h` — achievements

The old check `text.find("|H")` matched ALL of these, so sharing a quest link in party chat would cause the bot to enter the item parsing/trade flow.

## Test Scenarios

| Scenario | Before | After |
|----------|--------|-------|
| Share `[Quest Name]` in party chat | Trade window opens | No reaction (correct) |
| Share `[Item Name]` in party chat | Trade window opens | Trade window opens (unchanged) |
| Say "questitem" in chat | Parsed correctly | Parsed correctly (unchanged) |
| Share `[Spell Name]` in party chat | Trade window opens | No reaction (correct) |

Tested on AzerothCore 3.3.5a with mod-playerbots, confirmed fix resolves the issue.

### DIFF
--- a/src/Bot/Cmd/ChatHelper.cpp
+++ b/src/Bot/Cmd/ChatHelper.cpp
@@ -598,7 +598,7 @@ uint32 ChatHelper::parseSlot(std::string const text)
     return EQUIPMENT_SLOT_END;
 }
 
-bool ChatHelper::parseable(std::string const text)
+bool ChatHelper::parseableItem(std::string const text)
 {
     return text.find("|Hitem:") != std::string::npos || text == "questitem" || text == "ammo" ||
            substrContainsInMap<uint32>(text, consumableSubClasses) ||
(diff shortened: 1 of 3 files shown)


Celandriel (dismissed):

Celandriel:
What I dont like is that the function name is parseable, and now doesnt reflect whether there are parseable objects in the call, but only items. 
Its true that it is only called in one place, https://github.com/mod-playerbots/mod-playerbots/blob/295f2d2784cfb11b34dc48ff356848ef10b107fd/src/Bot/Engine/ExternalEventHelper.cpp#L33
But the loss of functionality is an issue for me with such a broadly named command.

PR #2170, opened 2026-02-25
PR template proposal, using the proposed template itself

NoxMax:
## Pull Request Description

The addition of a PR template last month was a great idea. We had nothing before and just let people type whatever they thought was relevant. Some wrote a whole article with too many details, and some just wrote the title and didn't explain any of the important details.

So the addition of the PR template makes sure contributors know what's most important to this project. However, several people thought the template was... a lot. A lot of the information it showed, while useful to the contributor, made it a bit confusing to reviewer to know what the contributor wrote, and what is part of the PR template, so a lot of these guidelines have now been put behind `<!-- -->`. 

Moreover, even what has been hidden, has been truncated. The main message of the guidelines is stability is our top priority. It is a critical message, but it was repeated several more times than it should've been. Less is more here, and if an important message is repeated in a verbose manner, people would gloss over it like scrolling down long terms and conditions. The questions were also made more concise and explicit. We don't want contributors to question the questions themselves and how do they even apply to their code. 

The process of PR submission itself should not feel like submitting a long bureaucratic form. Think of it like a scientific paper abstract: It gives a reasonably short summary explaining the work, as clearly as possible.

## Feature Evaluation

These are the core questions that are important to know, but even then, not always relevant. So a note was added to the contributor that they can obviously skip these if their PR is something like a comment edit or whatever else that clearly doesn't add processing.

## How to Test the Changes

You are already testing whether or not this template is effective by looking at it.

1. See the"[Preview](https://github.com/NoxMax/mod-playerbots/blob/PR-template-proposal/PULL_REQUEST_TEMPLATE.md)" of the file just so it's clear what the template is like without any of the comments I made here.
2. See the "[Code](https://github.com/NoxMax/mod-playerbots/blob/PR-template-proposal/PULL_REQUEST_TEMPLATE.md?plain=1)" section of the template to see how it would actually look to a contributor. The only difference is the alignment of the translation table; it looks weird in the .md file, but it would look properly aligned to the contributor when submitting.

## Impact Assessment

    - [x] No, not at all
    
Broke down processing impact into minimal and moderate. It is not uncommon that we have changes that add some minimal processing, and yes, collectively they can become an issue, but we also need to distinguish them from the rare changes that have a moderate impact, and how critical those changes are.

    - [x] No
    - [ ] Yes (**explain why**)

    - [x] No
    - [ ] Yes (**explain below**)
    
This question merges two previous one, because it's really asking the same thing: Will your change be a headache to maintain down the line?

## Messages to Translate

- [x] No
- [ ] Yes (**list messages in the table**)

|			 |			      |
|			 |			      |

This is a new section, based on an idea that was discussed to not have everyone (a) add their SQL translation files to their PR, (b) figure out file date name based on merge order, and (c) coordinate who's using which message key. No. The hidden instruction instead tell the contributor to prep the code to be translatable, by looking up GetBotTextOrDefault in the codebase for examples, and leave it that.
When merged, it would just use the default English fallback, then a monthly PR can be made containing translations for all the recently merged commits that have bot messages. The code would then automatically pickup the translated lines for that.
This section of the template would remain if there's a consensus that this is how the translation workflow should be.

## AI Assistance

- [x] No
- [ ] Yes (**explain below**)

## Final Checklist

- [x] Documentation updated if needed (Code comments, Conf comments, Commands in the Wiki).

Final checklist remains the same, only clarifies to the contributors what sort of documentations that need updating.

## Notes for Reviewers

This is a literal draft of of what the template should be, in that I look forward to your ideas to any ways that can further improve this.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,23 +1,22 @@
 <!--
 Thank you for contributing to mod-playerbots, please make sure that you...
-1. Submit your PR to the test-staging branch, not master.  
+1. Submit your PR to the test-staging branch, not master.
 2. Read the guidelines below before submitting.
 3. Don't delete parts of this template.
 
 DESIGN PHILOSOPHY: We prioritize STABILITY, PERFORMANCE, AND PREDICTABILITY over behavioral realism.
 
-Every action and decision executes PER BOT AND PER TRIGGER. Small increases in logic complexity
-scale poorly across thousands of bots and negatively affect all participants. We prioritize 
-stability over intelligence: a stable system is always preferred over a smarter one. Complexity 
-must justify its cost. Bots don't need to behave perfectly; believable behavior is the goal, 
-not human simulation. Default behavior must be cheap; expensive behavior must be opt-in.
+Every action and decision executes PER BOT AND PER TRIGGER. Small increases in logic complexity scale
+poorly across thousands of bots and negatively affect all. We prioritize a stable system over a smarter
+one. Bots don't need to behave perfectly; believable behavior is the goal, not human simulation.
+Default behavior must be cheap in processing; expensive behavior must be opt-in.
 
 Before submitting, make sure your changes aligns with these principles.
 -->
 
-
 ## Pull Request Description
 <!-- Describe what this change does and why it is needed -->
+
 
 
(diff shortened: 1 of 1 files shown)


Celandriel:
I support the improved PR change. Not sure if just in draft awaiting feedback.

PR #2179, opened 2026-03-01
Added all TBC attunement quests

trolloks:
# Pull Request

Added all TBC attunement quests to conf

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

---

## How to Test the Changes

Run maintenance on bots
This only applies to Individual Progression mod, since attunements aren't required in base AC
Bots should be able to enter:
- The Eye (Tempest Keep)
- Mount Hyjal
- Black Temple

## Complexity & Impact

- - [x] No

- - [x] No

- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [ ] No
- - [x] Yes (**explain why**)
All attunements for TBC are now added on 'maintenance' command

- - [ ] Lightweight mode remains the default
---

## AI Assistance

- - [x] No

If yes, please specify:

---

## Final Checklist

---

## Notes for Reviewers

before merging.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -566,8 +566,8 @@ AiPlayerbot.AutoGearScoreLimit = 0
 # Default: food, taxi, and raid are enabled
 AiPlayerbot.BotCheats = "food,taxi,raid"
 
-# List of attunement quests (comma-separated list of quest IDs) that are automatically completed for all bots via the 'maintenance' chat command.  
-# mod-playerbots does not restore removed attunement requirements, although other mods, such as mod-individual-progression, may do so.
+# List of attunement quests (comma-separated list of quest IDs) that are automatically completed for all bots.  
+# While mod-playerbots does not restore removed attunement requirements, although other mods, such as mod-individual-progression, may do so.
 # This is meant to exclude bots from such requirements.
 # 
 # Default:


conf/playerbots.conf.dist
```diff
@@ -592,7 +592,17 @@ AiPlayerbot.BotCheats = "food,taxi,raid"
...
+# The Eye
+# - 10888, Trial of the Naaru: Magtheridon
+#
+# Mount Hyjal
+# - 10445, The Vials of Eternity
+#
+# Black Temple
+# - 10985, A Distraction for Akama
+#
+AiPlayerbot.AttunementQuests = 10279,10277,10282,10283,10284,10285,10296,10297,10298,11481,11482,11488,11490,11492,10901,10888,10445,10985
```
kadeshar, line 608: Changes in config file should be reflected in PlayerbotAIConfig.cpp

conf/playerbots.conf.dist
brighton-chi: It's probably helpful to add a comment that says what is the purpose of listing these attunement quests.

kadeshar (approved):

conf/playerbots.conf.dist
trolloks: Are you referring to it only being used with IP?

brighton-chi: I mean that right now it just says "attunement quests" but not why the attunement quests are listed there, like what's the purpose.

trolloks: How about 'List of TBC attunement quests for bots to complete when using Individual Progression Mod'?

Any suggestions welcome. Copy has never been my strong suit XD.

kadeshar: Its list of quests which bot completes (all randombots, altbots, addbots) without rewards.

Its important because there is AiPlayerbot.RandomBotQuestIds which apply only to randombots and bots also gets rewards

brighton-chi: I understand that. But I have the context; I think a typical player would not understand from looking at the config.

So maybe something like "List of attunement quests that are automatically completed for all bots. While mod-playerbots does not restore removed attunement requirements, other mods such as mod-individual-progression may do so, and this config option excludes bots from such requirements." That's kind of long and is a multi-line comment for sure, but I'm not sure how much more can be trimmed without giving proper context.

trolloks: I'm happy with that. @kadeshar what do you think?

kadeshar: For me that will be good :)

conf/playerbots.conf.dist
```diff
@@ -566,7 +566,10 @@ AiPlayerbot.AutoGearScoreLimit = 0
 # Default: food, taxi, and raid are enabled
 AiPlayerbot.BotCheats = "food,taxi,raid"
 
-# Attunement quests (comma-separated list of quest IDs)
+# List of attunement quests (comma-separated list of quest IDs) that are automatically completed for all bots via the 'maintenance' chat command.  
+# mod-playerbots does not restore removed attunement requirements, although other mods, such as mod-individual-progression, may do so.
```
kadeshar: While mod-playerbots

conf/playerbots.conf.dist
```diff
@@ -566,7 +566,10 @@ AiPlayerbot.AutoGearScoreLimit = 0
 # Default: food, taxi, and raid are enabled
 AiPlayerbot.BotCheats = "food,taxi,raid"
 
-# Attunement quests (comma-separated list of quest IDs)
+# List of attunement quests (comma-separated list of quest IDs) that are automatically completed for all bots via the 'maintenance' chat command.
```
kadeshar: This text changed and im not sure about that. Attunment quest are applied not only via maintenance command

brighton-chi:
Just thought of something--IIRC at least some of the quests have two versions because the pre-WotLK version awards a title (Champion of the Naaru and Hand of A'dal). I don't know what the right quest to use here would be. On one hand, the title would not be available in WotLK. On the other hand, the purpose here is to complete attunements for IP, and the applicable quests for IP would be the pre-WotLK versions.

I don't think it matters for rndbots, but if the automatically awarded version locks altbots out of the title, maybe that's not good. Or maybe that's just a fair consequence of using autocompletion, though the player might not expect that consequence, and completion is included by default.

trolloks:
> Just thought of something--IIRC at least some of the quests have two versions because the pre-WotLK version awards a title (Champion of the Naaru and Hand of A'dal). I don't know what the right quest to use here would be. On one hand, the title would not be available in WotLK. On the other hand, the purpose here is to complete attunements for IP, and the applicable quests for IP would be the pre-WotLK versions.
> 
> I don't think it matters for rndbots, but if the automatically awarded version locks altbots out of the title, maybe that's not good. Or maybe that's just a fair consequence of using autocompletion, though the player might not expect that consequence, and completion is included by default.

Yeah that is true, but there are a couple of side effects to running the maintenance command that is not ideal for altbots. Would it not be better to add a safety (in another pr), like there is for autogear?

brighton-chi:
Just so everybody is clear, the included quest IDs are for the TBC versions of the quests. Therefore, bots will be given the "Champion of the Naaru" and "Hand of A'dal" titles, which are not obtainable in WotLK (the same quests are available but with different IDs and do not reward the titles). I do think using the TBC versions is ultimately the better approach.

brighton-chi:
FYI there is a typo in the config (see asterisks), but I suppose a fix can just be thrown in during some random future PR. I can even do it in a raid strat one.

```
# While mod-playerbots does not restore removed attunement requirements, **although** other mods, such as mod-individual-progression, may do so.
```

PR #2176, opened 2026-02-28
Requirement fix to use bigobj parameter to compile

kadeshar:
# Pull Request
Compilation fix which making possible compiling without bigobj parameter

---

## How to Test the Changes

- compile using Visual Studio without bigobj parameter

## Complexity & Impact

Does this change add new decision branches?
- - [x] No
- - [ ] Yes (**explain below**)

Does this change increase per-bot or per-tick processing?
- - [x] No
- - [ ] Yes (**describe and justify impact**)

Could this logic scale poorly under load?
- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

Does this change modify default bot behavior?
- - [x] No
- - [ ] Yes (**explain why**)

## AI Assistance

Was AI assistance (e.g. ChatGPT or similar tools) used while working on this change?
- - [ ] No
- - [x] Yes (**explain below**)

Automate file creation

---

## Final Checklist

- - [x] Stability is not compromised
- - [x] Performance impact is understood, tested, and acceptable
- - [x] Added logic complexity is justified and explained
- - [x] Documentation updated if needed

PR #2186, opened 2026-03-04
Implement Zul'Aman Strategies

brighton-chi:
# Pull Request

_Implement strategies for all bosses in Zul'Aman. See next post for overview of implemented strategies._

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

_I have attempted to order checks while taking into account cost and likelihood and have opted to find lower-cost methods where possible. I have also not gone as in depth as I have with other strategies, partially because it is not necessary to complete encounters but also to try to limit the performance impact. From my observation, including with pmon, none of the methods should be overly taxing._

---

## How to Test the Changes

_Run Zul'Aman. See next post for strategies to test._

## Complexity & Impact

- - [ ] No

_Only in the context of raid strategies, with new methods to consider and new multipliers to evaluate when bots perform actions with the instance strategy active._

- - [ ] No
- - [x] Yes (**describe and justify impact**)

_The impact is only with the "zulaman" strategy active, which will be applied only in the instance. There is currently a PR open to also remove instance strategies when leaving the map to get rid of the residual performance impact._

- - [ ] No
- - [x] Yes (**explain why**)

_Technically yes, but I think it is unlikely to have an appreciable difference unless there are many groups running the instance at the same time on a large server._

## Defaults & Configuration

- - [ ] No
- - [x] Yes (**explain why**)

_Only in the instance. It is a necessary trade-off to consider with raid strategies that I always keep in mind (degree of automation vs. player choice)._

_Not exactly sure how to address this question in the context of this PR, but there aren't any techniques or methods for this strategy that I have not tried before (or something very similar)._

## AI Assistance

- - [ ] No

If yes, please specify:

_Gemini and GPT for some questions about the codebase and C++ and assistance with drafting a few things like containers that I find more tedious to do myself._

---

## Final Checklist

---

## Notes for Reviewers

before merging.


brighton-chi:
Same thing as usual, written with 50% damage and 50% healing, plus the reversion of 3.0.2’s 30% HP nerf. Below doesn't describe minor strategies like multipliers that exist just to keep bots from doing stupid-looking movements.

**Trash – Aman’ishi Medicine Man**

- Bots will mark and kill (1) Protective Wards and then (2) Healing Wards.

 
**Akil’zon**

- Misdirect boss to main tank upon pull
- Tanks position the boss at a fixed waypoint in the middle of the platform
- Ranged spread 13 yards from each other to avoid chaining Static Disruption
- When Electrical Storm is cast, bots run to stack under the player that is targeted—this generally works pretty well for me, but I do sometimes have a bot that is late arriving and is killed. If the lack of preemptive movement proves to be a problem, I will try a timer to get the bots stacked before Electrical Storm is cast. I’d like to avoid it because timers are kind of annoying to do, especially with events that have a time range during which they occur instead of a fixed recurrence, and stacking bots too early can be problematic due to Static Disruption.

**Nalorakk**

- Misdirect boss to main tank upon pull
- Main tank and first assist tank position the boss at a fixed waypoint toward the back of the platform, and they will stack to split the Brutal Swipe damage
- Main tank and first assist tank swap taunts so the main tank handles the boss in troll form, and the first assist tank handles the boss in bear form
- Ranged spread out 11 yards (the boss uses Surge to knock back players and anyone within 10 yards)

**Jan’alai**
This is by far the hardest boss in the instance, and the strategy takes a little bit of management by the player.

- Misdirect boss to main tank upon pull
- Tanks position the boss at a fixed waypoint in the middle of the platform
- Ranged have fixed positions around the boss. They spread in a circle with a 15-yard radius to avoid too many getting hit by the Flame Breath cone.
- When Fire Bombs appear, bots will find a safe spot and wait out the explosion
- When Hatchers spawn, bots will mark one with Skull and kill it. They will mark the other with Moon and ignore it (and also will not cc it). The intent is to allow one Hatcher to hatch all of the eggs on one side, which is the standard strategy. After one side is hatched, the Hatcher will run to the other side. At that point, you can just let them hatch the other side as well, or you can override the Moon mark and kill the Hatcher (any remaining eggs will automatically hatch when the boss gets to 30%).
- Bots will save Bloodlust/Heroism until Dragonhawk Hatchlings spawn

**Halazzi**

- Misdirect boss to main tank upon pull
- Main tank and first assist tank position the boss at a fixed waypoint toward the back of the room, and they will stand in front of him to split the Saber Lash damage. Don’t try to be a hero like @flashtate and take the whole hit by yourself. It’s 47,500 base damage.
- First assist tank picks up the Lynx when it spawns
- DPS will focus Corrupted Lightning Totems and then Halazzi—they will ignore the Lynx

**Hex Lord Malacrass**
Make sure to have the Priest or Paladin Shadow Resistance buff up.

- Misdirect boss to main tank upon pull
- The 10 possible adds are ordered in a vector from most to least dangerous and marked and killed in such order
- There is a multiplier to disable dispelling of Unstable Affliction; I’m not positive this works, however
- If Malacrass siphons a Warrior soul, bots will run out of Whirlwind. They will also stop casting at him when Spell Reflection is up.
- If Malacrass summons a Hunter soul, bots will run away from the Freezing Trap.

**Zul’jin**

- Misdirect boss to main tank upon pull and each phase change—I’m not 100% sure whether it is functioning after phase changes
- Tanks position the boss at a fixed waypoint in the middle of the platform, except for in Phase 3 (when he is untankable) and Phase 5 (when it is better for tanks to be flexible with positioning)
- Bots run out of Whirlwind in Phase 1
- Bots avoid Cyclones in Phase 3—they aren’t fantastic at it, but I found implementing a custom method to still be worth it. Compared to relying on the standard AvoidAoeAction(), bots take about 40% less damage over the course of the phase. For sure AvoidAoeAction() should be disabled, but I’m wondering still if the best approach is to use no avoidance at all.
- Bloodlust/Heroism is saved for Phase 3
- Ranged will spread 6 yards in Phase 5 to avoid stacking and all eating the Flame Breath cone

brighton-chi:
Keeping on draft until https://github.com/mod-playerbots/mod-playerbots/pull/2176 merge into test-staging so I can update based on the AiObjectContext refactor.

src/Ai/Raid/ZulAman/Strategy/RaidZulAmanStrategy.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    // Trash
+    triggers.push_back(new TriggerNode("amani'shi medicine man summoned ward", {
+        NextAction("amani'shi medicine man mark ward", ACTION_RAID + 1) }));
+
+    // Akil'zon <Eagle Avatar>
+    triggers.push_back(new TriggerNode("akil'zon pulling boss", {
+        NextAction("akil'zon misdirect boss to main tank", ACTION_RAID + 2) }));
+
+    triggers.push_back(new TriggerNode("akil'zon boss engaged by main tank", {
```
Celandriel, line 19: Should it be the case that trigger and action have the same name for easy of correlation?

Celandriel (commented):
Should this be updated? 
https://github.com/mod-playerbots/mod-playerbots/blob/545b21ec0c60d6ef01f6e81f2921104a49ffc624/src/Ai/Raid/TempestKeep/Strategy/RaidTempestKeepStrategy.cpp#L108-L109

src/Ai/Raid/ZulAman/Strategy/RaidZulAmanStrategy.cpp
brighton-chi: I don't know what it should be, but that hasn't always been the case, even predating me. I'm looking at ICC and Ulduar, and some of them do match, some don't. In some cases, one trigger is multiple actions so at least in those instances it wouldn't be possible to have an exact match.

brighton-chi:
> Should this be updated?
> 
> https://github.com/mod-playerbots/mod-playerbots/blob/545b21ec0c60d6ef01f6e81f2921104a49ffc624/src/Ai/Raid/TempestKeep/Strategy/RaidTempestKeepStrategy.cpp#L108-L109

Thanks, dunno what happened there.

Celandriel:
We should be using enum class instead of enum (I know we havent cleaned up all cases of it in code currently, and even AC is mixed. 

Also Consider trying to simplify the complexity here (If possible)
https://sonarcloud.io/project/issues?pullRequest=2186&open=AZy3UNzvJFclLl1H4fkI&id=mod-playerbots_mod-playerbots

brighton-chi:
> We should be using enum class instead of enum (I know we havent cleaned up all cases of it in code currently, and even AC is mixed.
> 
> Also Consider trying to simplify the complexity here (If possible) https://sonarcloud.io/project/issues?pullRequest=2186&open=AZy3UNzvJFclLl1H4fkI&id=mod-playerbots_mod-playerbots

AC almost never uses it, despite it being their codestyle, ostensibly. I can update this one, but updating prior ones is for another day since I'll have to cast everything.

I'll see what I can do about the "cognitive complexity." I've never heard that term before, but it's hilarious.

Celandriel:
> I'll see what I can do about the "cognitive complexity." I've never heard that term before, but it's hilarious.
Make Brain Hurt. 

Dont worry about priors on enum class.

Celandriel:
Honestly im not entirely sure that there is anything to do specifically here, but just something to keep in mind.

Celandriel (approved):

Issue #1788, opened 2025-10-30
Can't get a cat druid to buff me thorns


skirmess:
Can't get a cat druid to buff me thorns if the druid is in cat form. And, it seems, there is no command to force them out of cat form.

I would like to whisper them `cast Thorns on Charname` to get them to buff me even when they're in cat form.

kadeshar:
Did you tried:

1. Switching strategy

<img width="897" height="128" alt="Image" src="https://github.com/user-attachments/assets/70451b7f-07b2-4b8e-956b-4a1831785c12" />

2. Manually cast spell via select character which you want to and command `cast thorns`
3. Manually cast spell via command `do thorns on party`

skirmess:
1. Switching strategy. 

I'm not really sure how that works or how I would do that. I tried `co +caster` and it looks like that but she stays in cat.

<img width="1054" height="524" alt="Image" src="https://github.com/user-attachments/assets/625ac6e0-b1fc-4315-82a5-353fb45e5cb3" />


2. Manually cast spell via select character which you want to and command cast thorns<img width="1140" height="574" alt="Image" src="https://github.com/user-attachments/assets/f2a258b2-c162-4fa9-955c-77345f8ec213" /><img width="1101" height="548" alt="Image" src="https://github.com/user-attachments/assets/01b60d75-c3cd-4ca2-bf7b-8eac4eef720d" />3. Manually cast spell via command do thorns on party

<img width="1039" height="527" alt="Image" src="https://github.com/user-attachments/assets/5a0c441d-a2b0-4bc9-9f8a-d5b944fc2c21" />

<img width="240" height="45" alt="Image" src="https://github.com/user-attachments/assets/e2e1c4ea-bf7d-45c2-a87a-8007d0f77c86" />

kadeshar:
Cant reproduce.

<img width="316" height="94" alt="Image" src="https://github.com/user-attachments/assets/ca70780f-78b3-48f8-b40e-89c91a54c5af" />
<img width="344" height="181" alt="Image" src="https://github.com/user-attachments/assets/50562943-931f-484a-aa5b-77ae0c758c3e" />

Character on which you try cast thorns is not in gm mode?

skirmess:
No GM mode, it's a normal account.

[...]

PR #2194, opened 2026-03-07
Added cancellation druid form actions

kadeshar:
# Pull Request

Added new (for now manual) actions to cancel druid forms. 
Resolve: https://github.com/mod-playerbots/mod-playerbots/issues/1788

---

## Feature Evaluation

Please answer the following:

- order bot enter some form like `do travel form`
- order bot cancel form like `do cancel travel form`

---

## How to Test the Changes

- order bot enter some form like `do travel form`
- order bot cancel form like `do cancel travel form`

## Complexity & Impact

- - [x] No

- - [x] No

- - [x] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [x] No
- - [ ] Yes (**explain why**)
- 
---

## AI Assistance

- - [x] No

---

## Final Checklist

---

### DIFF
--- a/src/Ai/Base/ChatActionContext.h
+++ b/src/Ai/Base/ChatActionContext.h
@@ -20,7 +20,6 @@
 #include "CustomStrategyEditAction.h"
 #include "DebugAction.h"
 #include "DestroyItemAction.h"
-#include "DruidShapeshiftActions.h"
 #include "DropQuestAction.h"
 #include "EquipAction.h"
 #include "FlagAction.h"


noisiver:
I think it is important to include instructions without the `do` part. Many find it very annoying because it makes them use voice lines, and it seem no one knows commands can be issued without `do` and instead ask how to disable those voice lines.

kadeshar:
Added to chat commands

src/Ai/Base/ChatActionContext.h
```diff
@@ -168,6 +169,13 @@ class ChatActionContext : public NamedObjectContext<Action>
         creators["tell target"] = &ChatActionContext::tell_target;
         creators["formation"] = &ChatActionContext::formation;
         creators["stance"] = &ChatActionContext::stance;
+        creators["cancel tree form"] = &ChatActionContext::cancel_tree_form;
+        creators["cancel travel form"] = &ChatActionContext::cancel_travel_form;
```
Celandriel: I wonder if a generic "cancel form" command could be used and then the form type is handled by a switch depending on the stated form. 
So 'cancel form cat' gets processed as a 'cancel form' trigger and action, then using event.getParam() you can get 'cat' and target the correct cancellation that way?

Celandriel: Another thing. I dont think we have to register the action here, it may be possible to register only in the DruidAiObjectContext.

kadeshar:
Fixed uneccessary action declaration in chatactioncontext.
About unified command im tried made some implementation but that make code complicated. I would stay on current approach which is much cleaner

Celandriel:
see DebugAction.cpp has a good example of it. 
Ok. No worries.

kadeshar:
Yeap, exactly that was the result. if spaghetti. I see other possible improvement but i will make that in separate PR when i test all possibilities.

Issue #2119, opened 2026-02-06
WLK shaman totem quest vs relic totems: avoid keeping 4 totem items when relic exists


XYUU:
## Background

In WLK there are two quest IDs: 14100 and 14111. These quests combine the original four undestroyable totem items into a single destroyable totem item.

At the same time, all shaman relics (relic type, totem subclass) already provide the functionality of the four totems. Therefore, when a shaman has a relic in equipment or bags, the original four totem items are no longer required.

## Proposal

In random bot / inventory maintenance logic, when a shaman relic is detected:

* Do not add the four totem items 5175/5176/5177/5178
* Optionally remove existing totem items from bags/equipment/bank
* 
## Scope
* Only affects shaman totem item maintenance
* Does not change relic functionality

If needed, I can provide exact code locations and implementation details.

kadeshar:
Fill template. Missing important information for this issue.

noisiver:
Fill out the template for the original message, not as a comment.

kadeshar:
PR merged to master. If something is not right please let us know

PR #2197, opened 2026-03-08
Fix: WLK shaman totem quest vs relic totems: avoid keeping 4 totem items when relic exists #2119

XYUU:
## Summary
* Detects shaman relics (relic type, totem subclass) in bags/equipment.
* Skips adding the four classic totem items (5175–5178) when a relic exists.
* Cleans up any existing totem items from bags/equipment/bank when a relic exists, while keeping Ankh handling intact.
## Test plan
Verified manually (local environment).


src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -3342,18 +3342,36 @@ void PlayerbotFactory::InitReagents()
...
+                ItemIds totemIds = {5175, 5176, 5177, 5178};
+                FindItemByIdsVisitor totemVisitor(totemIds);
+                IterateItems(&totemVisitor, (IterateItemsMask)(ITERATE_ITEMS_IN_BAGS | ITERATE_ITEMS_IN_EQUIP | ITERATE_ITEMS_IN_BANK));
+                for (Item* item : totemVisitor.GetResult())
+                    bot->DestroyItem(item->GetBagSlot(), item->GetSlot(), true);
+            }
             if (level >= 30)
             {
-                items.push_back({5178, 1});  // Air Totem
+                if (!hasRelic)
```
brighton-chi, line 3369: I think the logic is easier to follow if the air totem is moved to be grouped with the other totems instead of having the separate !hasRelic check here.

brighton-chi: Also, if you do that, I don't think there is any reason to have the !hasRelic local variable anymore instead of inlining relicVisitor.found.

XYUU: When I first read your comment, I also felt that your suggestion made a lot of sense. However, when I tried to apply the change, I realized that inlining relicVisitor.found would require extracting level >= 30 into a separate variable. And it’s not only about totems — shamans also learn Reincarnation at level 30, which requires carrying Ankh. So even though moving the logic to the totem section improves readability for totem management, it actually makes future maintenance harder for people who play and maintain the bots. They would need to check more level‑based differences across bots, because level progression in gameplay doesn’t follow the same timing as the code structure. For that reason, I think keeping the current approach might be better overall.

kadeshar (approved):

PR #2203, opened 2026-03-13
Fix Hunter Aspect Switching + Trigger Cleanups

brighton-chi:
# Pull Request

Note: When I reference Aspect of the Hawk below, it also means Aspect of the Dragonhawk (the code will use Dragonhawk if the Hunter has it, Hawk if not, they share actions, triggers, and strategies).

Hunter Aspects are currently bugged. All Hunters, regardless of spec or strategy, are hardcoded to use Aspect of the Hawk when mana is at 70%+ and Aspect of the Viper when mana drops to "lowMana" from the config (default is 15%), divided by 2.

This means the following:

- Hawk (bdps) and Viper (bmana) strategies are useless
- Pack (bspeed) and Wild (rnature) strategies are applied, but bots will rapidly switch back and forth between Pack/Wild and Hawk/Viper, depending on strategy and mana level.

This PR addresses the issues by doing the following:

- Global Hawk strategy is removed. Now you need to set bdps for Hunters to use Hawk, but bdps remains the default Aspect strategy for all Hunters.
- Dedicated Viper strategy is removed, leaving the global strategy. However, Viper will be used (when lowMana/2) ONLY if the bot is set to bdps. If the bot has the Wild or Pack strategy, they will not switch to Viper at all. I did this because I am assuming if you are using Wild or Pack, you need them for reasons other than to pump DPS.
- The threshold to switch back to Hawk is lowered from 70% to 60%. The gap between lowMana/2 and 60% is now filled--if bdps is on, Hunters will switch to Hawk whenever above the Viper threshold, _except_ for when they have the Viper aura, in which case they will not switch to Hawk until 60% mana. This lets the Hunter build back mana before swapping back to Hawk (more like general player behavior) while still letting them swap from other Aspects to Hawk without needing to be all the way at 60% mana.
- Gets rid of a weird condition in the Hawk trigger that would make it so that Hunters would switch to Hawk when at exactly 0 mana. I'm not sure what the point of that is.

Also, I refactored the triggers a bit because I noticed there was some dead code in there. I didn't do a comprehensive refactor, but there was a lot of stuff that clearly didn't make sense even to my eyes, like back-to-back returns. I think there's more unnecessary code even just in the triggers, but I didn't want to get too into the weeds with this PR.

---

## Design Philosophy

long-term robustness.

Principles:

- **Defaults must be cheap**  

---

## Feature Evaluation

Please answer the following:

I don't expect there to be any impact on costs, and if anything this PR removes some unneeded checks from triggers.

---

## How to Test the Changes

The easiest way is to go shoot a dummy with Volley until low on mana and then toggle on selfbot. You can do this with various Aspects active to test.

## Complexity & Impact

- - [X] No

- - [X] No

- - [X] No
- - [ ] Yes (**explain why**)
---

## Defaults & Configuration

- - [ ] No
- - [X] Yes (**explain why**)

Described above. Default behavior is broken.

- - [X] Lightweight mode remains the default
---

## AI Assistance

- - [ ] No

I asked Claude some questions about the triggers to make sure I didn't screw anything up.

If yes, please specify:

---

## Final Checklist

- - [X] Stability is not compromised
- - [X] Performance impact is understood, tested, and acceptable
- - [X] Added logic complexity is justified and explained
- - [X] Documentation updated if needed

---

## Notes for Reviewers

before merging.


kadeshar (dismissed):

src/Ai/Class/Hunter/Strategy/HunterBuffStrategies.cpp
```diff
@@ -31,7 +31,7 @@ void HunterBuffDpsStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
     triggers.push_back(
         new TriggerNode("aspect of the hawk", { NextAction("aspect of the dragonhawk", 20.1f),
-                                                                NextAction("aspect of the hawk", 20.0f) }));
+                                                NextAction("aspect of the hawk", 20.0f) }));
```
Celandriel: This is not how to declare an alternative. Need to create an action node with an alternative.

brighton-chi: True, I was avoiding a refactor of the whole class strategy =P but this one I can fix

brighton-chi: Still don't want to do a complete refactor but can't ignore issues if I see them...

src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp
```diff
@@ -4,62 +4,42 @@
...
+        NextAction("trueshot aura", 2.0f) }));
     triggers.push_back(new TriggerNode("often", {
-                       NextAction("apply stone", 1.0f),
-                       NextAction("apply oil", 1.0f),
-                       }));
-    triggers.push_back(new TriggerNode("low ammo", { NextAction("say::low ammo", ACTION_NORMAL)}));
-    triggers.push_back(new TriggerNode("no track", { NextAction("track humanoids", ACTION_NORMAL)}));
-    triggers.push_back(new TriggerNode("no ammo", { NextAction("equip upgrades packet action", ACTION_HIGH + 1)}));
+        NextAction("apply stone", 1.0f),
+        NextAction("apply oil", 1.0f) }));
```
Celandriel: Are you going to hate me if I say I prefer fully tabbed like it was previously, or not at all? 

```cpp
 triggers.push_back(new TriggerNode("trueshot aura", {NextAction("trueshot aura", 2.0f) }));

or

triggers.push_back(new TriggerNode("trueshot aura", {
                                   NextAction("trueshot aura", 2.0f) }));
````

Comment (brighton-chi) [reply src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp]: I try to follow the 80-character line limit convention, or at least not go too far over. And I feel like the complete indenting looks stupid when it goes that far over, but I can go with that approach.

The way stuff is formatted now doesn't really follow any approach that makes sense to me.

Comment (Celandriel) [reply src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp]: Yea, I dont think weve ever settled on a max length.

Comment (brighton-chi) [reply src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp]: I mean the general programming convention.

**Review by Celandriel — DISMISSED** [review]

File: src/Ai/Class/Hunter/Strategy/HunterBuffStrategies.cpp [hunk 3]
```diff
@@ -10,9 +10,20 @@
...
+    }
 
 private:
+    static ActionNode* aspect_of_the_dragonhawk([[maybe_unused]] PlayerbotAI* botAI)
+    {
+        return new ActionNode("aspect of the dragonhawk",
+                              /*P*/ {},
+                              /*A*/ { NextAction("aspect of the hawk") },
+                              /*C*/ {});
+    }
```
kadeshar, line 27: That shouldnt be included in combat strategy?

src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp
```diff
@@ -4,62 +4,31 @@
...
-                              /*P*/ {},
-                              /*A*/ { NextAction("readiness")},
-                              /*C*/ {});
-    }
-
-    static ActionNode* aspect_of_the_pack([[maybe_unused]] PlayerbotAI* botAI)
-    {
-        return new ActionNode("aspect of the pack",
-                              /*P*/ {},
-                              /*A*/ { NextAction("aspect of the cheetah")},
```
kadeshar, line 33: This cheetah fallback is on purpose?

src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -3321,7 +3321,7 @@ void PlayerbotFactory::InitReagents()
             break;
         case CLASS_PALADIN:
             if (level >= 52)
-                items.push_back({21177, 80});   // Symbol of Kings
+                items.push_back({21177, 100});   // Symbol of Kings
```
kadeshar, line 3324: That is not hunter changes :P

brighton-chi: I know. I mentioned it to Keleborn, seems not worth its own PR, and there's no good reason not to make the change (the item stacks to 100). But if you prefer, it can be its own PR.

src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp
brighton-chi: I assume so. I didn't really give this much thought. I've never used bspeed. I don't really know when you would want a bot Hunter to use Cheetah since bots cannot kite anyway. I'd be fine with removing the alternative.

kadeshar: Why? When you remove that between level 20-39 bspeed will be useless without this alternative

src/Bot/Factory/PlayerbotFactory.cpp
kadeshar: Ok, makes sense :D

src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp
brighton-chi: I mean maybe bspeed is useless anyway before 40. Why would you ever want to increase only the Hunter's run speed? I can't think of anything useless about it for a bot.

I don't have any problem with keeping it, though.

src/Ai/Class/Hunter/Strategy/HunterBuffStrategies.cpp
brighton-chi: Hm, in practice it is added in combat also, with bdps being the default, but I'm not sure about the exact mechanic that makes that happen. The aspect strategies don't appear to be included in AddDefaultCombatStrategies. I'll have to look into it later.

src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp
kadeshar: For example when you druid and you leveling with hunter :P

brighton-chi: True, true, or maybe you have a mount and the hunter is too low level or too poor to afford one.

src/Ai/Class/Hunter/Strategy/HunterBuffStrategies.cpp
brighton-chi: Ok, I think they are not added in combat by default. I use MultiBot to change strategies, and it has a separate combat strategy button vs. non-combat for Hunter aspects. But I think actually by default there is no combat Hunter aspect strategy, it's just that there isn't any restriction on whether a non-combat strategy can be manually added as a combat strategy. I do that with worldbuff too, co +worldbuff will make bots refresh worldbuffs in combat.

So I think actually the Aspect strategies should be modified to be generic so there is a default combat strategy in addition to a default non-combat strategy (like the Paladin auras). I'll do that.

Celandriel:
Getting build errors.

brighton-chi:
> Getting build errors.

Oops, forgot to update the ICC multiplier in this branch after making a base class for Explosive Shot. Should be good now.

Celandriel:
> > Getting build errors.
> 
> Oops, forgot to update the ICC multiplier in this branch after making a base class for Explosive Shot. Should be good now.

macos error
/Users/runner/work/mod-playerbots/mod-playerbots/modules/mod-playerbots/src/Ai/Raid/Icecrown/Multiplier/RaidIccMultipliers.cpp:777:76: error: unknown type name 'CastExplosiveShotAction'
            dynamic_cast<CastExplosiveTrapAction*>(action) || dynamic_cast<CastExplosiveShotAction*>(action))

src/Ai/Class/Hunter/Strategy/GenericHunterNonCombatStrategy.cpp
kadeshar: What with that?

brighton-chi: Hah that seems like about the right reaction.

OK so Pack should be back in but moved to the buff strategies. Cheetah is a fallback. The isUseful() check for Cheetah is pointless so I removed it. I kept the Pack trigger, which won't put Pack up if Cheetah is up. This lets players manually switch to Cheetah, even after learning pack, in case they want only the Hunter to have the speed boost (and risk of stun) and not the rest of the party.

kadeshar (approved):

Issue #990, opened 2025-02-20
Give players the option to delay bots before they engage


danielcoh92:
**Is your feature request related to a problem? Please describe.**

Right now, bots will engage with full force as soon as you pull a pack of mobs in a dungeon / raid setting.
This makes tanking harder as you have to fight the bots for threat as soon as you pull.
**Describe the solution you'd like**
Please add an option to set a delay for bots so they wait x amounts of seconds before they engage. This will allow the tank to gain some aggro before the DPS bots engage.

**Additional context**
I've seen references to this command somewhere but I'm not sure this is implemented here as this command does nothing:
/t BOTNAME orders delay <0-10> (sets the amount of time the bot will wait before engaging your target or tank's target in combat)

PaoloBenigni:
Currently my approach has been to use two macros to control this. 

```
/p @dps co +passive
/p @dps co -passive
```

danielcoh92:
> Currently my approach has been to use two macros to control this.
> 
> ```
> /p @dps co +passive
> /p @dps co -passive
> ```

Yeah, I've been using /p @dps flee and /p @dps follow and it works, but having to use those 2 macros on every pull can get quite tedius.. If we could set the delay for x amount of seconds at the start of the dungeon/raid it would be so much better and automate this process instead of having to press those 2 macros over and over. The tank will get the chance to reach the mobs and grab aggro and dungeons would be less tedius to run and require less micro-managing of the dps bots.

kadeshar:
Workaround:
Requires slashin or elvui (not all versions) addon

Prepare macro (/in 8 mean after 8 second)
/p @dps co +passive
/p @heal co +passive
/p @tank do attack my target
/in 8 /p @dps co -passive
/in 8 /p @heal co -passive

Solution:
Migrate "wait for attack" strategy from cmangos bots or create something similar

Jaytapp:
The wait for attack was very useful and the "pull" also.

PR #2211, opened 2026-03-14
Wait for attack strategy migration

kadeshar:
## Pull Request Description
Migration of "wait for attack" strategy from cmangos playerbots.

Resolves: https://github.com/mod-playerbots/mod-playerbots/issues/990

Ported from [cmangos playerbots](https://github.com/cmangos/playerbots)

## Feature Evaluation
Optional strategy for bots which are in party with real player. 

## How to Test the Changes
- add strategy to bot "nc +wait for attack" and "co +wait for attack"
- set time via command "wait for attack time x" where x is time which they wait in seconds (you should get response from bot)
- attack any target (for example dummy in main city)(bot should wait with attack)

## Impact Assessment
    - [ ] No, not at all
    - [x] Minimal impact (**explain below**)

Performance wise only bots having this optinal strategy have additional cost in multiplier which check every attack action that should be execute.

    - [x] No
    - [ ] Yes (**explain why**)

    - [x] No
    - [ ] Yes (**explain below**)

## Messages to Translate
- [ ] No
- [x] Yes (**list messages in the table**)

1740 | Please provide a time to set (in seconds)
1741 | Please provide valid time to set (in seconds) between 0 and 99
1742 | Wait for attack time set to %new_time seconds

## AI Assistance
- [ ] No
- [x] Yes (**explain below**)

Copilot CLI - help with migration

## Final Checklist

### DIFF
--- a/src/Ai/Base/Strategy/WaitForAttackStrategy.cpp
+++ b/src/Ai/Base/Strategy/WaitForAttackStrategy.cpp
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU AGPL v3 license, you may redistribute it
+ * and/or modify it under version 3 of the License, or (at your option), any later version.
+ */
+
+#include "WaitForAttackStrategy.h"
+
+#include "Action.h"
+#include "PlayerbotAI.h"
+#include "PlayerbotAIConfig.h"
+#include "Playerbots.h"
+#include "Strategy.h"
+
+void WaitForAttackStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
+{
+    triggers.push_back(new TriggerNode(
+        "wait for attack safe distance",
+        {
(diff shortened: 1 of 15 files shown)


src/Ai/Base/Strategy/WaitForAttackStrategy.cpp
```diff
@@ -0,0 +1,93 @@
...
+    return botAI->GetAiObjectContext()->GetValue<uint8>("wait for attack time")->Get();
+}
+
+float WaitForAttackStrategy::GetSafeDistance()
+{
+    return sPlayerbotAIConfig.spellDistance;
+}
+
+float WaitForAttackMultiplier::GetValue(Action* action)
+{
```
Celandriel, line 77: Will this also disable all healing? Is this the intention?

kadeshar: Yeap it should. Main purpose in cmangos was to allow tank pull mobs and build initial aggro. Healers also interupt that.

Celandriel (approved):

PR #2222, opened 2026-03-18
Fix/talent tree ordered map

### LINKED ISSUE #2050: Generated Level 80 DK only has 25 talent points used
**Bug Description**
Generated Level 80 DKs only have 25 talent points used, untill you use talents spec command to reset their talents points to normal status.

DKs of other levels seem don't have this problem.

**Commit Hash**
fd07e02a8a841a39719c43ff88babb19d34f4b63

**How To Reproduce Bug**
Just grab a generated Level 80 DK and check its talent points.

**Expected Behavior**
Talents point should be fully used.

**Screenshots**
lvl 80 DK No.1

Image
Image

lvl 80 DK No.2
Image
Image

lvl 60 DK 
Image

lvl 70 DK
Image

**Modules**
ah bot plus
ale
character service
individual progression
tansmog

oskov:
Fixes #2050

  InitTalents builds a map of talentRow → [TalentEntry*] and iterates it to teach talents row by row. WoW's talent system requires each row to be filled before unlocking the next, so iteration must happen in ascending row order. Commit
  b474dc4 ("Performance optim") changed the container from std::map to std::unordered_map, which has no guaranteed key ordering. As a result, bots would frequently attempt to learn talents in a row whose prerequisites hadn't been met
  yet, silently skipping them. I belive it's the reason of #2050 issue.

  The fix is a one-character type change: restoring std::map<uint32, ...>, which guarantees ascending key (row) order.

  How to Test the Changes

   1. Make fresh installation
   2. Create new character
   3. Observe talents tree of fresh rnd bots
  

   - [X]  Yes — GitHub Copilot CLI was used to identify the root cause (unordered_map introduced in b474dc4 breaking talent row ordering), stage the one-line fix, and draft this PR description. The code change was reviewed and fully
  understood before submission.

  Root cause commit: b474dc44bb6323430a84fc17c1ec046f9919a101 ("Performance optim") — changed std::map to std::unordered_map in InitTalents, breaking the row-ordering guarantee that WoW's talent prerequisite system depends on.

### DIFF
--- a/src/Bot/Factory/PlayerbotFactory.cpp
+++ b/src/Bot/Factory/PlayerbotFactory.cpp
@@ -762,7 +762,7 @@ void PlayerbotFactory::InitPetTalents()
         // pet_family->petTalentType);
         return;
     }
-    std::unordered_map<uint32, std::vector<TalentEntry const*>> spells;
+    std::map<uint32, std::vector<TalentEntry const*>> spells;
     bool diveTypePet = (1LL << ci->family) & diveMask;
 
     for (uint32 i = 0; i < sTalentStore.GetNumRows(); ++i)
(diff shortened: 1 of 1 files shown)


Celandriel (commented):
I think line 765 of that file should also be changed back to map.

oskov:
> I think line 765 of that file should also be changed back to map.

I didn't notice any issues with pet talents (didn't know that pets have talents too, never played huner), but the logic in InitPetTalents is identical — same unordered_map keyed by row, same row-ordered iteration. So it's likely broken in the same way. 

I've included the revert for that map too.

PR #2227, opened 2026-03-21
Blood dk rotation update

dillyns:
## Pull Request Description

Improve the blood DK rotation, add hysteria use, improve blood pact use. Basic Blood priority should be:
1. Keep diseases up
2. Use rune strike whenever possible
3. Use frost and death runes on Icy Touch (highest threat ability)
4. Use unholy runes on death strike to trigger more death runes
5. use blood runes on heart/blood strike

Hysteria should be used on a physical dps, or a tank if no physical dps is available. Never a caster or healer.
Summon ghoul should be saved for death pact usage. They are honestly a liability that aggros everything in range without the unholy talent that turns them into a pet anyways.

## Feature Evaluation

Removed plague strike from the general rotation. It should only be used for disease application.
Added death strike on "high unholy rune" trigger instead.
Remove raise dead from generic dk. Only cast it right before death pact.
Add hysteria to bloods default actions.
Minimal change

## How to Test the Changes

To test rotation changes:
Use a combat parsing addon such as Details! or watch the combat log of a Blood DK bot.
They should now use death strike instead of using unholy runes on plague strike.

To test hysteria:
Ungrouped blood dk bots can be observed using it on themselves.
Create a group with the blood dk + casters/healers only. Observe that they never use hysteria on the casters.
Create a group with blood dk + physical dps. Observe that they use hysteria on a dps and not themselves.

To test blood pact:
I used GM command .damage to get my Blood DK bot into "critical health" trigger. They will use raise dead then blood pact.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

Used for brainstorming and exploring the codebase to find similar patterns.

## Final Checklist

### DIFF
--- a/src/Ai/Class/Dk/Action/DKActions.cpp
+++ b/src/Ai/Class/Dk/Action/DKActions.cpp
@@ -48,3 +48,55 @@ bool CastRaiseDeadAction::Execute(Event event)
 
     return true;
 }
+
+Unit* CastHysteriaAction::GetTarget()
+{
+    Group* group = bot->GetGroup();
+    if (!group)
+    {
+        if (!bot->HasAura(49016))
+            return bot;
+        return nullptr;
+    }
+
+    Unit* rangedDps = nullptr;
+    Unit* tank = nullptr;
+    
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
+        if (!member || !member->IsAlive())
+            continue;
+
+        if (member->GetMap() != bot->GetMap() || bot->GetDistance(member) > sPlayerbotAIConfig.spellDistance)
+            continue;
+
+        // Skip if already has hysteria
+        if (member->HasAura(49016))
+            continue;
+
+        // Priority 1: Melee DPS
+        if (botAI->IsMelee(member) && botAI->IsDps(member))
+            return member;
+
+        // Priority 2: Ranged DPS (physical, not casters)
+        if (!rangedDps && botAI->IsRanged(member) && botAI->IsDps(member) && !botAI->IsCaster(member))
+            rangedDps = member;
+
+        // Priority 3: Tank
+        if (!tank && botAI->IsTank(member))
+            tank = member;
+    }
+
+    if (rangedDps)
+        return rangedDps;
+
+    if (tank)
+        return tank;
+
+    // Fallback to self if no hysteria
+    if (!bot->HasAura(49016))
+        return bot;
+
+    return nullptr;
+}
--- a/src/Ai/Class/Dk/Action/DKActions.cpp
+++ b/src/Ai/Class/Dk/Action/DKActions.cpp
@@ -61,7 +61,7 @@ Unit* CastHysteriaAction::GetTarget()
 
     Unit* rangedDps = nullptr;
     Unit* tank = nullptr;
-    
+
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
--- a/src/Ai/Class/Dk/Action/DKActions.h
+++ b/src/Ai/Class/Dk/Action/DKActions.h
@@ -340,4 +340,11 @@ public:
     CastBloodTapAction(PlayerbotAI* botAI) : CastMeleeSpellAction(botAI, "blood tap") {}
 };
 
+class CastHysteriaAction : public CastSpellAction
+{
+public:
+    CastHysteriaAction(PlayerbotAI* botAI) : CastSpellAction(botAI, "hysteria") {}
+    Unit* GetTarget() override;
+};
+
 #endif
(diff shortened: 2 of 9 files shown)


Celandriel:
Question from a non expert in DK. 

Should Death Strike should also be a default action? the rotation i see online is IT→PS→HS→HS→DS→DS

Should we add "no rune"  triggers for  "empower rune weapon" to Blood?

Moving "no pet" away from generic means that other specs wont use Raise Dead. I know they cant control them, but should we do that?

dillyns:
> Question from a non expert in DK.
> 
> Should Death Strike should also be a default action? the rotation i see online is IT→PS→HS→HS→DS→DS
> 
> Should we add "no rune" triggers for "empower rune weapon" to Blood?
> 
> Moving "no pet" away from generic means that other specs wont use Raise Dead. I know they cant control them, but should we do that?

The rotation you posted looks more like an opener than a full priority. Both icy-veins and wowhead state that icy touch should be used to consume frost and death runes, and death strike should be used to consume unholy runes.
Because of that you can't put death strike in default actions. It would have to have lower priority than icy touch, but since death strike requires 1 frost and 1 unholy rune, you will never have available frost runes because the default action will spam icy touchs first.
Having it on the high unholy rune lets us queue a death strike at a higher priority than icy touch, so we can guarantee the unholy rune is spent, but then return to icy touch priority once the high unholy trigger is over.

Regarding empower rune weapon, I actually think that it should be added to generic DK on the "no runes" trigger, as all DK specs should be using it, or it can be added to each strategies default actions with low priority like it already is for Frost DK strategy. If there is a performance gain using default actions over triggers I can do that.

Regarding "no pet" I think that blood DK should always be saving the ghoul for death pact use. I could add "no pet" and the other pet triggers to frost if we really want, but the uncontrolled ghoul is honestly a menace. It goes around pulling everything in dungeons and raids so I don't think it is worth the danger. I always add it to my Dks "ss" list

Celandriel:
> Regarding empower rune weapon, I actually think that it should be added to generic DK on the "no runes" trigger, as all DK specs should be using it, or it can be added to each strategies default actions with low priority like it already is for Frost DK strategy. If there is a performance gain using default actions over triggers I can do that.

If its useful for all the specs to have it then a generic one makes more sense for maintainability purposes.

Celandriel (dismissed):

src/Ai/Class/Dk/Action/DKActions.cpp
```diff
@@ -48,3 +48,55 @@ bool CastRaiseDeadAction::Execute(Event event)
 
     return true;
 }
+
+Unit* CastHysteriaAction::GetTarget()
```
Celandriel, line 52: I just realized this is going to run every tick. Group scans every tick are a bad idea. Can we gate this behind a trigger somehow?

Celandriel (changes requested):

src/Ai/Class/Dk/Action/DKActions.cpp
dillyns: added the hysteria no cd trigger

Celandriel (approved):

src/Ai/Class/Dk/Action/DKActions.cpp
```diff
@@ -48,3 +48,55 @@ bool CastRaiseDeadAction::Execute(Event event)
...
+        if (botAI->IsMelee(member) && botAI->IsDps(member))
+            return member;
+
+        // Priority 2: Ranged DPS (physical, not casters)
+        if (!rangedDps && botAI->IsRanged(member) && botAI->IsDps(member) && !botAI->IsCaster(member))
+            rangedDps = member;
+
+        // Priority 3: Tank
+        if (!tank && botAI->IsTank(member))
+            tank = member;
```
SmashingQuasar, line 88: You could return immediately the member here instead of storing it in a variable since you will never reaffect it. This would improve performance and make the code cleaner.

src/Ai/Class/Dk/Action/DKActions.cpp
```diff
@@ -48,3 +48,55 @@ bool CastRaiseDeadAction::Execute(Event event)
...
+
+        if (member->GetMap() != bot->GetMap() || bot->GetDistance(member) > sPlayerbotAIConfig.spellDistance)
+            continue;
+
+        // Skip if already has hysteria
+        if (member->HasAura(49016))
+            continue;
+
+        // Priority 1: Melee DPS
+        if (botAI->IsMelee(member) && botAI->IsDps(member))
```
SmashingQuasar, line 79: I've noticed that the code often does `botAI->IsX` but these methods are static, so you can just do `PlayerbotAI::IsX` which is cleaner and less confusing for new contributors.

PR #2229, opened 2026-03-21
Implement Auchenai Crypts Strategies and TBC Dungeon Contexts

flashtate98:
## Pull Request Description

This PR aims to add bot strategies to the Auchenai Crypts dungeon, specifically for the boss Shirrak the Dead Watcher, as his focus fire mechanic is really annoying to deal with.

Additionally I have added TbcDungeonActionContext.h and TbcDungeonTriggerContext.h for future reference as I add more strategies to TBC dungeons that need it.

A HUGE thank you to @brighton-chi for all his help. This has been a fun learning experience!

## Feature Evaluation

Trigger: A 20 yard radius grid search is performed to locate the focus fire trigger NPC

Action: Bots will flee from the trigger NPC using MoveAway with a 5 yard safety buffer

Multiplier: A for loop that prevents bots from running back into the focus fire mechanic while its active.

Tanking: Tanks have a set coordinate to drag the boss to for the duration of the fight.

Minimal, these scripts only execute when Shirrak has been engaged.

## How to Test the Changes

Go into Auchenai Crypts (Normal or Heroic) and engage Shirrak the Dead Watcher

## Impact Assessment

    - - [ ] No, not at all
    - - [X] Minimal impact (**explain below**)

This logic is only applied to Auchenai Crypts and activates during Shirrak's encounter.

    - - [ ] No
    - - [X] Yes (**explain why**)

Behavior only applies when in the instance and when engaged with Shirrak.

    - - [ ] No

New dungeon contexts were added (TbcDungeonActionContext and TbcDungeonTriggerContext) to provide a clean and dedicated structure for future TBC dungeon strategies that will be developed. This was done to be in alignment with the existing WOTLK contexts that already exist.

## Messages to Translate

    - - [X] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

Gemini was used to help me understand existing code in the module I was referencing and reusing to develop this strategy as I am not a programmer and hardly know anything about C++.

## Final Checklist

## Notes for Reviewers

If there are any better alternatives that could be used to help improve the strategy in any way, please suggest it. I am very new to this and want to learn and improve.


src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,73 @@
...
+                                                   position.GetPositionY());
+        if (distToPosition > 6.0f)
+        {
+            float dX = position.GetPositionX() - bot->GetPositionX();
+            float dY = position.GetPositionY() - bot->GetPositionY();
+            float moveDist = std::min(2.0f, distToPosition);
+            float moveX = bot->GetPositionX() + (dX / distToPosition) * moveDist;
+            float moveY = bot->GetPositionY() + (dY / distToPosition) * moveDist;
+
+            return MoveTo(558, moveX, moveY, bot->GetPositionZ(), false, false,
```
brighton-chi: Since you're only using the map id once, I'd probably use bot->GetMapId(). It costs practically nothing anyway and for a single usage is clearer in meaning.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        if (flare && flare->IsAlive())
+        {
+            float currentDistance = bot->GetDistance2d(flare);
+            constexpr float safeDistance = 20.0f; 
+            constexpr float buffer = 5.0f;
+
+            if (currentDistance < safeDistance)
+            {
+
```
brighton-chi: Extra line

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.h
```diff
@@ -0,0 +1,25 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONS_H
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONS_H
+
+#include "Action.h"
```
brighton-chi: I don't think you need to include Action.h since neither of your actions inherit from it

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.h
```diff
@@ -0,0 +1,25 @@
...
+#include "AttackAction.h"
+#include "MovementActions.h"
+#include "AuchenaiCryptsTriggers.h"
+
+// Shirrak the Dead Watcher
+
+class ShirrakTankPositionBossAction : public AttackAction
+{  
+public:
+    ShirrakTankPositionBossAction(PlayerbotAI* botAI) : AttackAction(botAI, "shirrak tank position") {}
```
brighton-chi: Since the action is called ShirrakTankPositionBossAction, you might want the string to align and be "shirrak tank position boss" (and then change it throughout)

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "FollowActions.h"
+#include "AIObjectContext.h"
+#include "Playerbots.h"
+
+// Shirrak the Dead Watcher
+
+// Flee from Focus Fire and dont run back in
+float FleeFocusFireMultiplier::GetValue(Action* action)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
```
brighton-chi: Since you're not using the boss variable, you don't need it. The check can just be:
```
if (!AI_VALUE2(Unit*, "find target", "shirrak the dead watcher")
    return 1.0f;
```

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,48 @@
...
+        return 1.0f;
+
+   constexpr float searchRadius = 20.0f;
+        std::list<Creature*> creatureList;
+        bot->GetCreatureListWithEntryInGrid(creatureList, NPC_FOCUS_FIRE, 20.0f);
+
+    for (Creature* flare : creatureList)
+    {
+        if (flare && flare->IsAlive())
+
```
brighton-chi: Extra line

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,48 @@
...
+        {
+
```
brighton-chi: Extra line

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,48 @@
...
+            if (dynamic_cast<CastReachTargetSpellAction*>(action))
+            return 0.0f;
```
brighton-chi: Missing indent

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+            float currentDistance = bot->GetDistance2d(flare);
+            constexpr float safeDistance = 20.0f; 
+
+            if (currentDistance < safeDistance)
```
brighton-chi: Gromleq is going to want you to combine the nested if statements =P

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.h
```diff
@@ -0,0 +1,14 @@
...
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSMULTIPLIERS_H
+
+#include "Multiplier.h"
+
+class FleeFocusFireMultiplier : public Multiplier
+{
+public:
+    FleeFocusFireMultiplier(PlayerbotAI* botAI) : Multiplier(botAI, "flee focus fire") {}
+    float GetValue(Action* action) override;
+
```
brighton-chi: Extra line

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "AuchenaiCryptsTriggers.h"
+#include "AiObject.h"
+#include "AiObjectContext.h"
+
+// Shirrak the Dead Watcher
+
+bool ShirrakTankPositionBossTrigger::IsActive()
+{
+    return botAI->IsTank(bot) &&
+            AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
```
brighton-chi: Think you might have one too many spaces here

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.cpp
```diff
@@ -0,0 +1,30 @@
...
+}
+
+bool FleeFocusFireTrigger::IsActive()
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
```
brighton-chi: Same as in the multiplier, no need for a variable for the boss

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.h
```diff
@@ -0,0 +1,34 @@
...
+    SPELL_FOCUS_FIRE_VISUAL         = 32286,
+};
+ 
+class ShirrakTankPositionBossTrigger : public Trigger
+{
+public:
+    ShirrakTankPositionBossTrigger(PlayerbotAI* botAI) : Trigger(botAI, "shirrak tank position") {}
+
+    bool IsActive() override;
+
```
brighton-chi, line 26: Extra line

src/Ai/Dungeon/AuchenaiCrypts/AuchenaiCryptsActionContext.h
```diff
@@ -0,0 +1,28 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONSCONTEXT_H
```
brighton-chi: Should be ACTIONCONTEXT

src/Ai/Dungeon/AuchenaiCrypts/AuchenaiCryptsActionContext.h
```diff
@@ -0,0 +1,28 @@
...
+            creators["flee focus fire"] = 
+                &TbcDungeonAuchenaiCryptsActionContext::flee_focus_fire;
+        }
+    private:
+        
+        static Action* shirrak_tank_position(
+            PlayerbotAI* botAI) { return new ShirrakTankPositionBossAction(botAI); }
+        
+        static Action* flee_focus_fire(
+            PlayerbotAI* botAI) {return new FleeFocusFireAction(botAI); }
```
brighton-chi: Space before the return for consistency

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.h
```diff
@@ -0,0 +1,34 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSTRIGGERS_H
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSTRIGGERS_H
+
+#include "Trigger.h"
+#include "GenericTriggers.h"
+#include "DungeonStrategyUtils.h"
+
+enum AuchenaiCryptsIDs
```
brighton-chi: So Keleborn has decided that apparently we are going to follow AC codestyle (which AC seems not to follow) of using enum class instead of enum. You'll need to cast all of the constants when you use them with enum class because they won't implicitly convert to integers. Like instead of just using NPC_FOCUS_FIRE, it will need to be something like:
```
static_cast<uint32>(AuchenaiCryptsIDs::NPC_FOCUS_FIRE)
```

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    return false;
+}
+
+//Flee from Shirrak's Focus Fire
+
+bool FleeFocusFireAction::Execute(Event /*event*/)
+{
+    constexpr float searchRadius = 20.0f;
```
Celandriel: Not used?

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.cpp
```diff
@@ -0,0 +1,29 @@
...
+    return botAI->IsTank(bot) &&
+        AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
+}
+
+bool FleeFocusFireTrigger::IsActive()
+{
+    if (!AI_VALUE2(Unit*, "find target", "shirrak the dead watcher"))
+        return false;
+
+   constexpr float searchRadius = 20.0f;
```
Celandriel: Not used

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+// Shirrak the Dead Watcher
+
+// Flee from Focus Fire and dont run back in
+float FleeFocusFireMultiplier::GetValue(Action* action)
+{
+    if (!AI_VALUE2(Unit*, "find target", "shirrak the dead watcher"))
+        return 1.0f;
+
+   constexpr float searchRadius = 20.0f;
```
Celandriel: Not used.

src/Ai/Dungeon/AuchenaiCrypts/Strategy/AuchenaiCryptsStrategy.cpp
```diff
@@ -0,0 +1,18 @@
+#include "AuchenaiCryptsTriggers.h"
+#include "AuchenaiCryptsStrategy.h"
+#include "AuchenaiCryptsMultipliers.h"
+
+void TbcDungeonAuchenaiCryptsStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
+{
+  // Shirrak The Dead Watcher
```
Celandriel: Should be 4 spaces even for comment.

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.cpp
```diff
@@ -0,0 +1,29 @@
...
+        std::list<Creature*> creatureList;
```
Celandriel, line 19: Review file tabbing.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,71 @@
...
+        std::list<Creature*> creatureList;
```
Celandriel: Extra tab

src/Ai/Dungeon/AuchenaiCrypts/AuchenaiCryptsTriggerContext.h
```diff
@@ -0,0 +1,27 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSTRIGGERCONTEXT_H
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSTRIGGERCONTEXT_H
+
+#include "AIObjectContext.h"
+#include "TriggerContext.h"
+#include "AuchenaiCryptsTriggers.h"
+
+class TbcDungeonAuchenaiCryptsTriggerContext : public NamedObjectContext<Trigger>
+{
+    public:
```
Celandriel: No tab on public/private, and then 1 tab below.

src/Ai/Dungeon/AuchenaiCrypts/AuchenaiCryptsActionContext.h
```diff
@@ -0,0 +1,28 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONCONTEXT_H
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONCONTEXT_H
+
+#include "AIObjectContext.h"
+#include "Action.h"
+#include "AuchenaiCryptsActions.h"
+
+class TbcDungeonAuchenaiCryptsActionContext : public NamedObjectContext<Action>
+{
+    public:
```
Celandriel: no tab on public, and then start for 1 tab

Celandriel (changes requested):

brighton-chi:
Alrighty, I fought him 5 times and tried to trigger some different scenarios. I think by having a Paladin healer, that gave me some difficulties because they are definitely the worst healer for this fight, which I think is good for pressure testing.

1. Tank movement is good. My Warrior tank would charge into combat so he was already at the top of the stairs when the fight started, so I tried making him wait back so that he had to engage the boss on the stairs. He still had no issues with backing up the stairs into position. If it's possible to do faster than 2 yards/s without fucking up the pathfinding, that would be great, but if not, I think this works well regardless. I do suggest a multiplier to disable TankFaceAction anytime there is a fixed tank waypoint because otherwise the tank will still try to turn the boss once arriving at the waypoint and ends up sometimes running side-to-side.

2. I had some issues with melee running back in after getting out. This shouldn't happen with the multiplier, and I suspect the issue is actually the buffer's interaction with the search radius. Since the search radius is the same distance as the danger zone, but movement calls for the bot to move farther, then when the bot gets farther than 20 yards + hitboxes from the trigger NPC, the NPC is no longer found so the multiplier fails at the flare check (as does the fleeing action), at which point the tank uses Charge, the melee DPS runs back in, etc. That is my guess anyway. I think increasing the search radius to say 30 yards could solve this issue.

3. I may be wrong, but I think 20 yards might be a greater radius than is needed. I think the actual damaging spell from Focus Fire is 32302, which has a radius of 12 yards.

4. My casters frequently insisted on finishing their casts before running. I have no idea why since you have InterruptNonMeleeSpells() there. That doesn't seem to interrupt channeled spells in my experience (I still haven't figured out how to do that so my Shadow Priest is always dying), but I'm talking here about a Warlock finishing Shadow Bolts and a Paladin finishing Holy Lights. The only thing I can think of is to try botAI->Reset() instead of AttackStop() and InterruptNonMeleeSpells(), but this one is puzzling to me.

5. The Attract Magic mechanic actually gave me some trouble. The boss would pull everybody in, and then the casters would get the max effect of Inhibit Magic as well as get hit by Carnivorous Bite, and since my Paladin was then casting 5-second Holy Lights, that would cause the damage to continue to stack, when ideally she'd step away a bit before casting even if everybody needs healing at the moment. My suggestion is to try a method for ranged bots to MoveAway from the boss if within X yards, but have an early return on the trigger if a flare is found since if that is the case they just need to get away from the flare at all costs. According to the boss script, Inhibit Magic increases to 2 stacks at less than 35 yards away, 3 stacks at less than 25 yards, and 4 stacks at less than 15 yards. I'm not sure what distance to use for such an action for the best tradeoff, but that's the context to consider anyway.

At the end of the day though, this is still much better than before. I could still clear this encounter, and without a strategy I used to have to make use of Flee a lot and still hope for some good luck on placement.

Minor comments: I don't think you need to suppress AvoidAoeAction in your multiplier. I don't think there's any mechanic in the fight that would cause it to be called. Also, not sure if you were intending to keep STAIRS_TOP_POSITION for another purpose but otherwise that can be removed.

src/Ai/Dungeon/AuchenaiCrypts/Strategy/AuchenaiCryptsStrategy.cpp
flashtate98: Can you elaborate? Not sure I understand.

Celandriel: The comment is preceded by two spaces instead of 4

flashtate98:
Thank you for all the feedback. This is extremely helpful. I'll work on getting actions for ranged bots to keep their distance.

That STAIRS_TOP_POSITION was originally there for when the tank wouldn't go to the position in the center of the corridor for some reason. I was attempting a bread crumb method at the time. But yeah, that will be removed. I'll also see if I can get the tank to be faster, it used to be quicker but the tank was pretty erratic especially when it got Shirrak to the correct position. But I will see if I can make it better.

I have tested it with warlocks and holy paladins and yeah for some strange reason they are hellbent on completing whatever they are casting before running away from focus fire. Hopefully your suggestion works.

I am noticing warriors are still charging back in, I will follow your suggested solution and hope that resolves the issue,

src/Ai/Dungeon/AuchenaiCrypts/Strategy/AuchenaiCryptsStrategy.cpp
flashtate98: Gotcha, thank you.

Celandriel (dismissed):
@flashtate98 Anymore changes incoming?

flashtate98:
> @flashtate98 Anymore changes incoming?

Yes, I will convert this to draft while I add more actions to this.

flashtate98:
New action was added for ranged bots to keep their distance from Shirrak to minimize the inhibit magic mechanic. They will stand at the top of the stairs while the tank drags the boss further down the room he resides in. 

For some reason, bots selected for focus fire will not interrupt any of their casts to move out of the incoming AOE. However, once they finish, they will then move out of the AOE like they are supposed to. Thankfully, since the ranged bots keep their distance, they are not as affected by inhibit magic, and will almost always escape the AOE even after finishing their casts. At worst, I have seen them take one tick of damage as they were running out.

I've done several tests with various ranged classes and it seems to be good now. But if anyone wants to do further testing themselves, that would be most welcome. Any and all further feedback is much appreciated.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,124 @@
...
+bool ShirrakTankPositionBossAction::Execute(Event /*event*/)
+{
+    Unit* shirrak = AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
+    if (!shirrak)
+        return false;
+
+    if (bot->GetVictim() != shirrak)
+        return Attack(shirrak);
+
+    if (shirrak->GetVictim() == bot && bot->IsWithinMeleeRange(shirrak))
```
brighton-chi: Suggestion here--I would add a check for the bot's HP because the bot will be backing up possibly out of LoS and/or range of the healer that is running up the stairs, and while it is backing up, it is not going to be using abilities. What I've started doing is incorporating HP checks in some cases for tanks positioning bosses if the movement could be dangerous. That way, the bot will stop and use some defensive ability or at least keep healers in range if the bot is getting low on health.

Something like
```
if (shirrak->GetVictim() == bot && bot->IsWithinMeleeRange(shirrak) &&
    bot->GetHealthPct() > 50.0f)
``` 
Or whatever other health threshold you think works best here.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,124 @@
...
+        if (flare && flare->IsAlive())
+        {
+            float currentDistance = bot->GetDistance2d(flare);
+            constexpr float safeDistance = 12.0f; 
+            constexpr float buffer = 5.0f;
+
+            if (currentDistance < safeDistance)
+            {
+                bot->AttackStop();
+                bot->InterruptNonMeleeSpells(true);
```
brighton-chi: As discussed, this can be deleted. I am thinking about a way to interrupt spells in raid strategies, I think the best approach (without modifying AC) is probably adding a condition in UpdateAI that will allow an external script hook to interrupt. But that's not for this PR, just something that we might have for strategies int he future.

Edit: I see you did it already. Disregard.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+    return false;
+}
+
+// Ranged should keep distance from Shirrak, staying at the edge of the stairs
+
+bool ShirrakRangedKeepDistanceAction::Execute(Event /*event*/)
+{
+    Unit* shirrak = AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
```
brighton-chi: You shouldn't need the manual attack, the bots won't have any other target so they should attack Shirrak on their own. So then you don't need the boss check either since that is already gated in the trigger.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
flashtate98: Thats a great idea! I didn't know such a thing was possible.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,124 @@
...
+    if (bot->GetVictim() != shirrak)
+        return Attack(shirrak);
+
+    std::vector<Player*> rangedBots;
+    if (Group* group = bot->GetGroup())
+    {
+        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+        {
+            Player* member = ref->GetSource();
+            if (member && botAI->IsRanged(member) && member->IsAlive())
```
brighton-chi: I would toss the IsAlive() check. You're not doing anything to store position assignments for bots (and you don't need to); what that means though is that if a bot dies, other ranged bots could have their positions reassigned due to the change in ordering of the vector. So they'd play musical chairs unnecessarily. I'd just keep the dead bot in the vector and then you won't have that problem.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,124 @@
...
+    size_t botIndex = (findIt != rangedBots.end()) ? std::distance(rangedBots.begin(), findIt) : 0;
+    size_t count = rangedBots.size();
+
+    constexpr float arcSpan = M_PI / 2.0f; 
+    float arcCenter = M_PI;
+    float arcStart = arcCenter - (arcSpan / 2.0f);
+
+    float angle = (count <= 1) ? arcCenter : (arcStart + (arcSpan * (float)botIndex / (float)(count - 1)));
+    
+    float spreadRadius = 5.0f;
```
brighton-chi: constexpr

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.h
```diff
@@ -0,0 +1,31 @@
...
+};
+
+class ShirrakFleeFocusFireAction : public MovementAction
+{
+public:
+    ShirrakFleeFocusFireAction(PlayerbotAI* botAI) : MovementAction(botAI, "shirrak flee focus fire") {}
+    bool Execute(Event event) override;
+};
+
+class ShirrakRangedKeepDistanceAction : public AttackAction
```
brighton-chi: If you remove the attack call this should be MovementAction

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,46 @@
+#include "AuchenaiCryptsMultipliers.h"
+#include "AuchenaiCryptsActions.h"
+#include "AuchenaiCryptsTriggers.h"
+#include "MovementActions.h"
+#include "AttackAction.h"
```
brighton-chi: I don't think AttackAction is needed here?

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "ReachTargetActions.h"
+#include "FollowActions.h"
+#include "AIObjectContext.h"
```
brighton-chi: Should be AiObjectContext.h. Windows is case-insensitive so it compiled fine for me but I'm not sure about other operating systems and in any case it should just match the file.

src/Ai/Dungeon/AuchenaiCrypts/Trigger/AuchenaiCryptsTriggers.cpp
```diff
@@ -0,0 +1,34 @@
...
+        if (flare && flare->IsAlive())
+            return true;
+    }
+    return false;
+}
+
+bool ShirrakRangedKeepDistanceTrigger::IsActive()
+{
+    return botAI->IsRanged(bot) &&
+        AI_VALUE2(Unit*, "find target", "shirrak the dead watcher");
```
brighton-chi: Need 3 more spaces on this line to align

src/Ai/Dungeon/AuchenaiCrypts/AuchenaiCryptsActionContext.h
```diff
@@ -0,0 +1,34 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONCONTEXT_H
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSACTIONCONTEXT_H
+
+#include "AIObjectContext.h"
```
brighton-chi: AiObjectContext.h

src/Ai/Dungeon/AuchenaiCrypts/AuchenaiCryptsTriggerContext.h
```diff
@@ -0,0 +1,35 @@
+#ifndef _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSTRIGGERCONTEXT_H
+#define _PLAYERBOT_TBCDUNGEONAUCHENAICRYPTSTRIGGERCONTEXT_H
+
+#include "AIObjectContext.h"
```
brighton-chi: AiObjectContext.h

src/Bot/PlayerbotAI.cpp
```diff
@@ -1529,7 +1529,7 @@ void PlayerbotAI::ApplyInstanceStrategies(uint32 mapId, bool tellMaster)
     static const std::vector<std::string> allInstanceStrategies =
     {
         "aq20", "bwl", "karazhan", "gruulslair", "icc", "magtheridon", "moltencore",
-        "naxx", "onyxia", "ssc", "tempestkeep", "ulduar", "voa", "wotlk-an", "wotlk-cos",
+        "naxx", "onyxia", "ssc", "tempestkeep", "tbc-ac", "ulduar", "voa", "wotlk-an", "wotlk-cos",
```
brighton-chi: Alphabetically should be before tempestkeep

brighton-chi (commented):
Tested it and works great. Thanks!

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+        std::list<Creature*> creatureList;
+        bot->GetCreatureListWithEntryInGrid(creatureList, static_cast<uint32>(AuchenaiCryptsIDs::NPC_FOCUS_FIRE), 20.0f);
+
+    for (Creature* flare : creatureList)
+    {
+        if (flare && flare->IsAlive())
+        {
+            if (dynamic_cast<CastReachTargetSpellAction*>(action))
+                return 0.0f;
```
kadeshar, line 26: That always should be true? Even for bots which are far from flare or bots which want to reach boss to cast range spell?

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
```diff
@@ -0,0 +1,45 @@
...
+// Shirrak the Dead Watcher
+
+// Flee from Focus Fire and dont run back in
+float ShirrakFleeFocusFireMultiplier::GetValue(Action* action)
+{
+    if (!AI_VALUE2(Unit*, "find target", "shirrak the dead watcher"))
+        return 1.0f;
+
+        std::list<Creature*> creatureList;
+        bot->GetCreatureListWithEntryInGrid(creatureList, static_cast<uint32>(AuchenaiCryptsIDs::NPC_FOCUS_FIRE), 20.0f);
```
kadeshar: FIx indentation

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,119 @@
...
+    if (!shirrak)
+        return false;
+
+    if (bot->GetVictim() != shirrak)
+        return Attack(shirrak);
+
+    if (shirrak->GetVictim() == bot && bot->IsWithinMeleeRange(shirrak) &&
+        bot->GetHealthPct()>50.0f)
+    
+    {
```
kadeshar: Blank line to delete

kadeshar:
Codestyle check fails

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
```diff
@@ -0,0 +1,119 @@
...
+        float moveDist = std::min(2.0f, distToSpot);
+        float moveX = bot->GetPositionX() + (dX / distToSpot) * moveDist;
+        float moveY = bot->GetPositionY() + (dY / distToSpot) * moveDist;
+
+        return MoveTo(bot->GetMapId(), moveX, moveY, bot->GetPositionZ(), false, false,
+                      false, false, MovementPriority::MOVEMENT_COMBAT, true, false);
+    }
+    return false;
+}
+
```
noisiver: Missing new line at the end of the file

noisiver (dismissed):

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
brighton-chi: CastReachTargetSpellAction is only Charge and similar spells. Moving into spell or attack range is ReachTargetAction. The flare spawns only when the attack starts and despawns immediately so I think this is correct.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
flashtate98: Pushed an update. Let me know if its fixed.

src/Ai/Dungeon/AuchenaiCrypts/Action/AuchenaiCryptsActions.cpp
flashtate98: Pushed an update. Let me know if its fixed.

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
flashtate98: Pushed an update. Let me know if its fixed.

src/Ai/Dungeon/AuchenaiCrypts/Multiplier/AuchenaiCryptsMultipliers.cpp
flashtate98: As what @brighton-chi said, its to prevent warriors from charging into the Focus Fire AOE while they attempt to run away.

kadeshar:
Codestyle stills fail

obraz

flashtate98:
> Codestyle stills fail
> 
> obraz

Should be fixed now. Let me know if there are any further issues.

kadeshar (approved):

Issue #2002, opened 2026-01-10
Unscripted abilities


brighton-chi:
I'm not sure if this is really a bug or a missing feature, but I wanted to create a running list of abilities that are not scripted for bots to use and should be added in the future.

Edit: Actually, changed my mind--this can also keep track of abilities that are referenced in the code but not complete or correct such that the bots will never use them.

- Druid: Challenging Roar
- Paladin: Hand of Freedom
- Racials: Will of the Forsaken
- Rogue: Blind, Deadly Throw, Dismantle, Ghostly Strike, Gouge, Hemorrhage, Preparation, Riposte, Shadow Dance, Shadowstep

Wishmaster117:
Good idea!

Wishmaster117:
You missed : https://github.com/mod-playerbots/mod-playerbots/pull/2074 :)

brighton-chi:
Oh there are a ton that are not on this list. I just haven't gone through all the class strategy code to check.

PR #2231, opened 2026-03-22
Will of the forsaken

kadeshar:
## Pull Request Description

Added Will of the Forsaken support. 
Made structure fix for affected files.

Partially resolves: #2002 

## How to Test the Changes

- create/invite undead bot
- start fight (you can use dummy target)
- use `.aura 6215` to fear bot
- bot should use "Will of the Forsaken" to remove debuff

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Undead bot use now "Will of the Forsaken" to remove charm, fear or sleep.

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

OpenCode to review changes

## Final Checklist

## Notes for Reviewers

obraz

### DIFF
--- a/src/Ai/Base/ActionContext.h
+++ b/src/Ai/Base/ActionContext.h
@@ -164,6 +164,7 @@ public:
         creators["blood fury"] = &ActionContext::blood_fury;
         creators["berserking"] = &ActionContext::berserking;
         creators["every man for himself"] = &ActionContext::every_man_for_himself;
+        creators["will of the forsaken"] = &ActionContext::will_of_the_forsaken;
         creators["use trinket"] = &ActionContext::use_trinket;
         creators["auto talents"] = &ActionContext::auto_talents;
         creators["auto share quest"] = &ActionContext::auto_share_quest;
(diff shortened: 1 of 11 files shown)


brighton-chi:
@Celandriel 

Calling you here because of the issue with respect to skipping spells being housed only in CastSpellAction::isUseful(). I don't think anybody is really going to want a bot to skip using WotF, but I'm wondering about general approach right now with new abilities. Maybe if a spell is overriding isUseful, it should directly call CastSpellAction::isUseful()?

kadeshar:
> @Celandriel
> 
> Calling you here because of the issue with respect to skipping spells being housed only in CastSpellAction::isUseful(). I don't think anybody is really going to want a bot to skip using WotF, but I'm wondering about general approach right now with new abilities. Maybe if a spell is overriding isUseful, it should directly call CastSpellAction::isUseful()?

Sounds reasonable. Will test that.

Celandriel:
What about making that section of code a protected member of CastSpellAction, and then inherited classes can call it while implementing their own isUseful checks?

brighton-chi:
> What about making that section of code a protected member of CastSpellAction, and then inherited classes can call it while implementing their own isUseful checks?

You mean a new member just for the spell skip? But how do you force inherited classes to call it? Or you just mean going forward?

I wonder about all these checks for CastSpellAction, it seems like they are often being skipped with no obvious consequence. I know that ultimately CanCastSpell and CastSpell also have a zillion checks so maybe that ends up covering everything. But basically there are a lot of interwoven parts here, and it's not clear what is necessary or helpful to have.

kadeshar:
You can use interceptor design pattern for that but that was be a huge change to change that in all of actions

EDIT:
We can add check to c++ codestyle to monitor that

Celandriel:
Well its really just this line     "AI_VALUE2(bool, "spell cast useful", spell);" That we need t add back in right?

kadeshar:
Hmmm im testing that (if work then i will refactor that)
obraz

EDIT:
this specific spell (and Every man for himself) cant use base class isPossible. But isUseful, yes
obraz

Celandriel (approved):

PR #2234, opened 2026-03-23
Fix Shaman Weapon Enchants & Cure Toxins/Cleanse Spirit

brighton-chi:
## Pull Request Description

1. I've been having persistent issues with Enhancement Shamans sometimes applying Rockbiter to both weapons instead of MH Windfury and OH Flametongue. Rockbiter is the alternative for Flametongue and, through Flametongue, the alternative for Windfury. But there seemed to be no obvious reason why a Shaman that had all three abilities would ever use Rockbiter, which costs more mana than Windfury and Flametongue. Claude's take on it is that there is instability from ItemForSpellValue related to its poor way of distinguishing handedness, in addition to it having a 1-second cache, which can cause in some scenarios stale caches for the action running on each hand back-to-back. I still can't say I fully understand why the issue exists, but the most straightforward fix that should prevent this from happening is to just have separate mainhand and offhand actions for each enchant. So that's what this PR does. The relevant ActionNodes are now:
 
- The MH-specific chain for Enhancement is WF -> FT -> RB. In practice, Enhancement should never apply RB because all Shamans under level 10 (when FT is learned) are considered Elemental. The FT -> RB node is just for Elemental.
- The MH-specific Resto chain (not that Resto can dual-wield) is EL -> FT -> RB. Againt, FT -> RB is just for Elemental.
- OH for Enhancement is only FT. You cannot be Enhancement before level 10, nor can Enhancement dual-wield before level 40, so no alternative is needed.

3. I commented out Frostbrand Weapon actions/triggers because the ability is not included in any strategy. I didn't delete the code because in the future somebody might want to implement it as I understand it can be useful for Enhancement Shamans in PvP.
4. Shamans are coded to use "cure poison" and "cure disease", which do not exist in WotLK, having been combined into Cure Toxins. Wishmaster has PR #1844  that has been open on this for a long time, but I decided to correct the abilities here anyway as he was going for a more limited approach, and I decided to rename all the actions and redo the structure to rely on ActionNode alternatives, which is pretty much the exact framework that should be used for this type of situation w/r/t bots. Now, Shamans prefer Cleanse Spirit (Resto talent, which costs the same as Cure Toxins and also dispels curses), with an alternative of Cure Toxins (for poisons and disease only). I tested this and it seems to work well.
5. I deleted empty ActionNodes.
6. I did some cleanup of formatting and such, but this is not intended to be a comprehensive refactor.

## Feature Evaluation

I've followed the general intended structure of class strategies with triggers and actions. The same triggers exist, just different actions are called based on the trigger that fires, so I don't think there should be any impact on performance.

## How to Test the Changes

The best way to get a grasp on if things work is probably to just do group play for a while with Shamans and make sure they apply the right enchants and properly cast Cleanse Spirit and Cure Toxins.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Shamans previously did not cure poisons or disease at all, and now they do with the default "cure" strategy applied.

    - - [ ] No

One might say having separate actions per hand for enchants is somewhat more complex, but ultimately I think it is less confusing to keep those paths separate.

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

I had Claude try to diagnose the weapon enchant issue. It proposed and provided the separate MH/OH WF/FT actions.  The other things were easy enough for me to do.

## Final Checklist


src/Ai/Class/Shaman/Action/ShamanActions.cpp
```diff
@@ -105,18 +105,53 @@ bool SetTotemAction::Execute(Event /*event*/)
...
+    return item->GetTemplate()->Class == ITEM_CLASS_WEAPON;
+}
+
+bool CastWindfuryWeaponMainHandAction::isPossible()
+{
+    Item* item = bot->GetItemByPos(INVENTORY_SLOT_BAG_0, EQUIPMENT_SLOT_MAINHAND);
+    if (!item || item->GetEnchantmentId(TEMP_ENCHANTMENT_SLOT))
+        return false;
+
+    return item->GetTemplate()->Class == ITEM_CLASS_WEAPON;
```
kadeshar: 3 times same implementation. Consider making base class

Also CastEnchantItemAction.isPossible shouldnt be call?

src/Ai/Class/Shaman/Strategy/TotemsShamanStrategy.cpp
```diff
@@ -132,7 +124,7 @@ void FireResistanceTotemStrategy::InitTriggers(std::vector<TriggerNode*>& trigge
 {
     GenericShamanStrategy::InitTriggers(triggers);
     triggers.push_back(new TriggerNode("set fire resistance totem", { NextAction("set fire resistance totem", 60.0f) }));
-    triggers.push_back(new TriggerNode("no water totem", { NextAction("fire resistance totem", 55.0f) }));
+    triggers.push_back(new TriggerNode("no fire totem", { NextAction("fire resistance totem", 55.0f) }));
```
kadeshar: Shouldnt stay "no water totem"?
I know nothing about shaman only based on website

Image

brighton-chi: You're right. Will fix.

src/Ai/Class/Shaman/Action/ShamanActions.cpp
brighton-chi: Sounds good. I'll make both changes.

kadeshar: In your changes i dont see CastEnchantItemAction.isPossible call

Celandriel (dismissed):

src/Ai/Class/Shaman/Action/ShamanActions.cpp
brighton-chi: I hardcoded the same checks into the base classes since it was just two simple checks.

kadeshar: Maybe but from maintenance pov is bad pattern. You should get value calling base.
Base class (CastEnchantItemAction) also have same problem but its other story :D

EDIT: also you can pull changes from test-staging to fix github actions compilation problems

brighton-chi: Fineeeeee

I wonder why CastSpellAction::isPossible() is commented out there.

kadeshar: Im think is related with CastCastSpell target. But other vehicle/mount checks are ok.

kadeshar (approved):

PR #2238, opened 2026-03-23
Challenging Roar support

kadeshar:
## Pull Request Description

Added Challenging Roar support to druid taunt spell
Related with: #2002 

## How to Test the Changes

- run dungeon with druid tank and monitor spell usage

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Druid also use Challenging Roar as taunt spell

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [x] No

## Final Checklist


brighton-chi:
WIth Challenging Roar being an AoE taunt, I'm not sure it makes sense to have the same trigger as growl--that seems like not a good usage of a 3-minute cooldown. The Warrior strategies use "high aoe" as a trigger for Challenging Shout--maybe the implementation should be the same here.

kadeshar:
Good idea will change that

src/Ai/Class/Druid/Strategy/BearTankDruidStrategy.cpp
```diff
@@ -212,6 +212,7 @@ void BearTankDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
             }
         )
     );
+    triggers.push_back(new TriggerNode("high aoe", {NextAction("challenging roar", ACTION_HIGH + 8)}));
```
brighton-chi, line 215: Should the format match the other TriggerNodes, which break across multiple lines?

kadeshar: Its not repository convention and personaly for me that is more practically (and often used in repo)

brighton-chi: Every class has its strategies either entirely or mostly in the broken out format. I don't personally have any opinion about which is better. But it seems odd to have one node be different from all the others in a single file (or every other node in the file be different from this one, depending on the perspective).

Celandriel (approved):

PR #2242, opened 2026-03-25
Module update for CollisionChange

Celandriel:
## Pull Request Description

Some code changes due to core update change from MMapFactory.h ->MapCollisionData.h

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

Identifying problematic areas generated by the change

## Final Checklist

### DIFF
--- a/src/Ai/Base/Trigger/StuckTriggers.cpp
+++ b/src/Ai/Base/Trigger/StuckTriggers.cpp
@@ -8,7 +8,7 @@
 #include "CellImpl.h"
 #include "PathGenerator.h"
 #include "Playerbots.h"
-#include "MMapFactory.h"
+#include "MapCollisionData.h"
 
 bool MoveStuckTrigger::IsActive()
 {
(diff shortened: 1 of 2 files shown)


src/Mgr/Travel/TravelMgr.cpp
```diff
@@ -687,10 +687,13 @@ std::vector<WorldPosition> WorldPosition::frommGridCoord(mGridCoord GridCoord)
     return retVec;
 }
 
+// TODO: Cleanup — this function proactively loads mmap tiles for the TravelMgr cross-grid
+// pathfinding system. The old MMapFactory singleton allowed loading from anywhere; the new
+// Dead code (if(false) VMap loading, unreachable MMap branch) was removed during the API migration.
```
kadeshar: if(false) is dead code but else no. Still should be commented?

src/Ai/Base/Trigger/StuckTriggers.cpp
```diff
@@ -90,7 +90,7 @@ bool MoveLongStuckTrigger::IsActive()
     }
 
     if (cell.GridX() > 0 && cell.GridY() > 0 &&
-        !MMAP::MMapFactory::createOrGetMMapMgr()->loadMap(botPos.GetMapId(), cell.GridX(), cell.GridY()))
+        bot->GetMap()->GetMapCollisionData().LoadMMapTile(cell.GridX(), cell.GridY()) == MMAP::MMAP_LOAD_RESULT_ERROR)
```
kadeshar: loadMap only returned false on error? What with ignored?

Celandriel: My thinking is that bots in invalid maps are def stuck, while if the map is already loaded, its not a guarantee that they are stuck?

src/Mgr/Travel/TravelMgr.cpp
Celandriel: I left the only part that actually ran. The entire function is not called at all right now anyway, and I need to fix it.

kadeshar (approved):

PR #2250, opened 2026-03-28
feat(Performance): BotActiveAlone activity interval fixes and default settings for avg player

hermensbas:
## Pull Request Description

- Bugfix the jittering on/off of botAlone activity
- BotActiveAlone activity duration configurable
- Updated the default config values for general user for a smoother experience
- Added offset jittering for the check allowedActivity and check next AI delay to prevent cpu spikes

(disabled WhenIsFriend can cause race conditions)

## Feature Evaluation

## Impact Assessment

    - - [ ] No, not at all

In a positive way, bots in your zone and 150 radius will always be active, meanwhile other bots will be active 40%
of the time with intervals for 60 seconds per bot. With much lower latencies. All configurable without say.

40% and 60 seconds for more balance for those who seek bots create world feel more natural and live vs bots leveling without killing the server performance. Why not 50 due activity of bots itself 40% will result more into 45-50% like behavior and 50% prolly more 55%-60%. This it not something we want incorporate when calculating the value since it depends on various config and situation.  But 40% is good base with default config.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [x] No

## Final Checklist


NoxMax:
25% is a bit low for default, no? I mean in the end its all gonna revolve around how many bots the user has on the server, but your original choice of 50% is probably better.

Second, is `BotActiveAloneDurationSeconds` useful as a config? And onto that point, is 30 seconds a measurable improvement over 60? (which I think was the original intended cycle duration)

hermensbas:
well yes and no, it comes down to the intend of this feature. The feature itself ensures better spreading once you enter an zone at the otherhand 50% does prevent to much of performance difference when all zones are active which rarely happens. 

The intend was not bots continue leveling when nobody to witness it, but that being said. 50% might a good middle ground for both worlds; where some like the bots to level while others just want to world feel as real when moving around.

30 seconds should be enough, since you also have to consider during that timeframe pre checks can be fired for example when in combat or various others which extends that time by design. Which can result in longer times in 30 seconds depending on situation and env.

But lemme run some additional test with 50% and 60 seconds this evening. After this i am gonna look into integrating level bracket mod of Dustin.

hermensbas:
Done, modified slighty for getting active 50% balance for avg user. And added additional offset delay for the allowedActivity checks to prevent event clustering, and slight ms delay on ai checks to prevent cpu stuttering. 

That being said personally i use different settings, just enough to keep the world feel alive, i dont care much for them for level when nobody is around. This PR balanced between bots to lvl when no real player around vs keeping the world feel alive.

My personal settings e..g:
```
AiPlayerbot.BotActiveAloneDurationSeconds = 15
```

This PR:
```
AiPlayerbot.BotActiveAlone = 40 (which in practise leans towards 45-50%)
AiPlayerbot.BotActiveAloneDurationSeconds = 60
```

NoxMax:
Great! Config comments need updating for the 40% though. And I think some more explicit comments would help
```
# The default is 40%, but will be automatically adjusted if botActiveAloneSmartScale
# is enabled. Regardless, this value is only applied to inactive areas where no real players
# are detected. When real players are nearby, the value is always enforced to 100%
AiPlayerbot.BotActiveAlone = 40
AiPlayerbot.BotActiveAloneDurationSeconds = 60      # Minimum time between activity switching.

# Force botActiveAlone to 100% when...
AiPlayerbot.BotActiveAloneForceWhenInRadius = 150   # bot is within this distance from a real player.
AiPlayerbot.BotActiveAloneForceWhenInZone = 1       # bot is in the same zone as a real player.
AiPlayerbot.BotActiveAloneForceWhenInMap = 0        # bot is in the same continent as a real player.
AiPlayerbot.BotActiveAloneForceWhenIsFriend = 0     # bot is a real player's friend.
AiPlayerbot.BotActiveAloneForceWhenInGuild = 1      # bot is in a real player's guild.
```
I mention "bot is in the same continent as a real player." because while GetMap also covers raids, dungeons, BGs, and arenas, they have their own activity enforcement logic. So for all practices, this config is for continents.

hermensbas:
Updated config comments

Celandriel (dismissed):

hermensbas:
There is still some internal discussion for the default value of active bots., so "waiting" for that..

hermensbas:
Update the default config based on internal talks best value and which values would benefit most ppl out of the box.

Celandriel (approved):

hermensbas:
### performance:  ac/test-staging & pb/test-staging: 4/4/2025
```
gobal + default config
AiPlayerbot.Min/MaxRandomBots = 4000
AiPlayerbot.BotActiveAloneDurationSeconds = 30
```

### test 1 (default values, 10%)
AiPlayerbot.botActiveAloneSmartScale = 1
```
Connected players: 1. Characters in world: 4001.
Connection peak: 1.
Server uptime: 17 minute(s) 32 second(s)
Update time diff: 15ms. Last 500 diffs summary:
|- Mean: 29ms
|- Median: 18ms
|- Percentiles (95, 99, max): 66ms, 75ms, 88ms
```

### test 2 (40%)
AiPlayerbot.BotActiveAlone = 40
AiPlayerbot.botActiveAloneSmartScale = 1
```
Connected players: 1. Characters in world: 4001.
Connection peak: 1.
Server uptime: 5 minute(s) 55 second(s)
Update time diff: 19ms. Last 500 diffs summary:
|- Mean: 38ms
|- Median: 27ms
|- Percentiles (95, 99, max): 84ms, 92ms, 100ms
```

### test 3 (40% without smartscale)
AiPlayerbot.BotActiveAlone = 40
AiPlayerbot.botActiveAloneSmartScale = 0
```
Connected players: 1. Characters in world: 4001.
Connection peak: 1.
Server uptime: 5 minute(s) 29 second(s)
Update time diff: 84ms. Last 500 diffs summary:
|- Mean: 48ms
|- Median: 35ms
|- Percentiles (95, 99, max): 110ms, 116ms, 123ms
```

### test 4 (10% without smartscale)
AiPlayerbot.botActiveAloneSmartScale = 0
```
Connected players: 1. Characters in world: 4001.
Connection peak: 1.
Server uptime: 3 minute(s) 27 second(s)
Update time diff: 14ms. Last 500 diffs summary:
|- Mean: 30ms
|- Median: 18ms
|- Percentiles (95, 99, max): 74ms, 78ms, 93ms
```

### test 5
AiPlayerbot.BotActiveAlone = 100
AiPlayerbot.botActiveAloneSmartScale = 0
```
Connected players: 1. Characters in world: 4001.
Connection peak: 1.
Server uptime: 4 minute(s) 9 second(s)
Update time diff: 39ms. Last 500 diffs summary:
|- Mean: 73ms
|- Median: 64ms
|- Percentiles (95, 99, max): 138ms, 151ms, 176ms
```

PR #2258, opened 2026-03-28
Implement Battle for Mount Hyjal Strategies

brighton-chi:
## Pull Request Description

This PR implements raid strategies for all bosses in everybody's favorite TBC instance, the Battle for Mount Hyjal. As before, I have designed these all to work with IP with 50% damage and healing. I also did not merge the 1.88x buff to Vanilla & TBC healing items that IP recently implemented.

The next post will outline all implemented strategies.

Note: Set to draft for now as I may tweak Archimonde some more, but I generally consider Hyjal complete, subject to comments.

## Feature Evaluation

As with previous strategies, I've worked within the existing context of actions/multipliers/triggers. This strategy does implement new spell hooks, but I don't think that is problematic for performance, and I'll explain why they are necessary in the next post.

## Impact Assessment

    - - [ ] No, not at all

There are many new triggers, multipliers, and actions, but they will be evaluated only if the "hyjal" strategy is added. Additionally, I've attempted to order and implement checks in a manner to limit performance impact and have tested with .pmon active. In general, I consider performance to be highly important so am always working on ways to limit the impact (e.g., trying to use the most targeted grid searches available when needed).

    - - [ ] No
    - - [x] Yes (**explain why**)

Only in the Hyjal instance, for obvious reasons.

    - - [ ] No

New decision branches apply only in the new Hyjal instance, for obvious reasons. Maintenance complexity should not be increased as code outside of the new Hyjal files is not impacted, except to the extent needed to register and implement the strategy in the same manner as all existing strategies. Exception: As noted, I did add new spell hooks, but if they become problematic, they can easily be removed.

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

Claude Sonnet 4.6 was used for more complex implementation and occasionally GPT-5 mini was used for simple questions. I do not use AI for brainstorming or developing strategies, only for implementation and review of code. Most of this was written by me directly, but most notably I needed AI support to implement the spell hooks and triggers/actions that relied on them. Everything was reviewed and tested many times.

## Final Checklist

          Caveat: The full impact of implementing the spell hooks on a broad scale is beyond my knowledge to evaluate.


brighton-chi:
### Implemented Strategies

One note: Hyjal Summit's terrain is highly uneven so the effectiveness of the strategies is especially dependent on the quality of the maps. I first started working on these strategies after the previous mmaps update and overhaul of the movement system, and at the time, proper strategies were practically impossible due to repeated failure of movement checks. That's simply a risk factor with this instance.

### Global

- The use of Bloodlust/Heroism is suppressed by a multiplier during the entirety of the instance except for boss fights (in each case, after HP reaches a threshold to allow positioning before the buff is invoked). Due to the way that values are computed for using cooldowns, Bloodlust/Heroism will be blown on cooldown on almost every trash wave without a global multiplier. It is too valuable to waste in such a manner.
- Maps used for timers, positional trackers, and dynamic object trackers are reset for each boss when bots are out of combat and do not have the applicable boss on their threat list.

### Rage Winterchill

- Hunters will Misdirect the boss to the main tank upon pull.
- The main tank will position the boss a bit back toward the base to give bots the opportunity to spread around the boss.
- Ranged bots will begin the fight by moving to designated positions in a 360-degree spread around the boss. They will move freely once the positions have been reached; the method is just to achieve an initial spread to limit the impact of Death and Decay.
- Note: Melee bots perform poorly when Death and Decay is cast on top of them. The default AvoidAoeAction() uses FleePosition(), which moves bots to positions that allow them to continue to attack. Needless to say, the action is not effective in cases when the boss simply cannot be attacked by melee (notable, with respect to large AoEs like Death and Decay that, depending on placement, can cover the full 360-degree radius around the boss. This will cause running in and out of the AoE repeatedly. There is no doubt to me that the issue can be solved, but it would require another spell hook to get the dynamic object that generates the Death and Decay. Since this is literally the easiest boss in TBC 25-man content, I did not view it as worthwhile to implement an avoidance method, but I'm not opposed to it.

### Anetheron

- Hunters will Misdirect the boss to the main tank upon pull. They will also Misdirect spawned Infernals to the first assist tank.
- The main tank will position the boss a bit back toward the base to allow a centralized spread.
- There are two positions, one East and one West, that are designated for Infernal tanking. When a bot is targeted by Inferno (the Infernal-spawning spell), unless it is the main tank, it will run to the nearest Infernal tanking spot. The first assist tank will move to the targeted bot to pick up the Infernal when it spawns and drag it to the nearest Infernal tanking spot.
- Nearby ranged DPS will kill the Infernal.
- The initial spread of ranged bots is 360degrees around the boss, with an inner ring for ranged DPS and outer ring for healers.
- After the initial spread, ranged bots will not return to their originally assigned positions. They will instead generally attempt to spread from other bots (as well as Infernals).

### Kaz'rogal

**General Setup**
- Hunters will Misdirect the boss to the main tank upon pull.
- The main tank will tank the boss near the entrance to the base, turning him toward the entrance away from the raid. This is intended to engage the boss as soon as possible since it is a DPS race.
- Assist tanks stack on top of the main tank to spread cleave damage.
- Ranged DPS fan out away from the boss, in the direction of the Horde base, in a small 45-degree arc, just outside of the boss's War Stomp range.

**Mana Management**
- When a bot drops under 3200 mana, the following actions take place:
  - Hunters switch to Aspect of the Viper
  - Warlocks use Life Tap
  - All casters (including Warlocks if they are unable to Life Tap) run away from the group, at least 16 yards from the nearest player. When they get 16 yards away from the group, they then will attempt to spread in any direction from the nearest player. Due to MoveFromGroup()'s implementation (move away from group until bot is X yards from the nearest player), if two bots are near each other and each gets low on mana, they will keep moving clear across the base before they get 16 yards apart.
  - If the bot gets back to 4000 mana, it will rejoin the group to get back into spell range of the boss (and Hunters will switch off Viper)
- In the event that a Priest's Shadow Resistance is not applied, a Paladin will switch to Shadow Resistance Aura.
- Warlocks with less than 3000 mana that are Marked will cast Shadow Ward.
- Mages and Paladins with less than 1200 mana will cast Ice Block and Divine Shield, respectively.
- Note: Melee with mana will also run from the group under same conditions as ranged. This should take them toward the entrance of the base. In practice, they do not have enough mana to be of any use in this fight.

### Azgalor

**General**
- Hunters will Misdirect the boss to the main tank upon pull.
- The main tank will drag the boss back toward Thrall's keep and then turn and move back toward the base entrance a bit to face Azgalor away from the raid.
- Melee (other than the main tank) will wait back near the base for the main tank to pull the boss into position.
- Ranged will be active right away but will maintain a greater distance from the boss while the boss is being positioned (as compared to after the boss is positioned).

**Doom**
- Bots with Doom will run to a designated position a bit aways behind Azgalor. The first assist tank will go there also to pick up the Doomguard when it spawns. If the first assist tank gets Doom, the second assist tank will go to pick up the Doomguard.
- Ranged bots within 65 yards of the Doomguard will DPS it down (this keeps any ranged bots that are near the front of Azgalor from trying to attack and potentially getting cleaved.
- Ranged will maintain a small spread from each other (except while DPSing Doomguards). They will always maintain distance from Doomguards and the boss.

**Rain of Fire Avoidance**
- Tanks will not avoid Rain of Fire.
- Ranged will rely on the standard AvoidAoeAction().
- Melee requires a custom strategy. The standard AvoidAoeAction() is not effective because it relies on FleePosition(), which will attempt to move in a manner that allows the bot to still attack. This means that depending on RoF placement, melee bots may run in front of Azgalor to avoid it and get one-shot by a cleave. Because RoF is generated by a dynamic object and AC does not expose a dynamic object getter (as far as I can tell), a spell hook is needed.
  - All active RoFs are recorded (10-second persistence)
  - Melee dps that are within the RoF radius will escape it by running directly away from Azgalor. There are situations in which they could move to the side and theoretically it would be better, but the implementation would be extraordinarily complex
  - Melee dps will wait out the 10s duration and then move back in. If further RoFs spawn during this time, they'll keep moving farther out.

### Archimonde

- Hunters will Misdirect the boss to the main tank upon pull.
- Priests will cast Fear Ward on the main tank, and Shamans will switch to the Tremor Totem strategy.
- The main tank will move the boss up the hill a little bit at the beginning of the fight to try to place a bit of distance between the boss and the World Tree. I've had issues with wipes due to him getting too close to the tree (it's an autowipe).
- Ranged will also spread a bit at the beginning of the fight so the first Air Burst doesn't blow away the entire ranged portion of the raid.
- Bots will attempt to get away from the Air Burst target. I'm not sure if this one is working at all. Air Burst is a very fast cast.
- Bots suffering damage from Doomfire and low on HP will remove the DoT with Ice Block (Mage), Cloak of Shadows (Rogue), or Divine Shield (Paladin).

**Doomfire**
- Doomfire's mechanic is interesting. Archimonde summons two NPCs, a Doomfire NPC and a Doomfire Spirit. The Spirit teleports randomly (very short distances), and the Doomfire moves to the Spirit. That creates the winding trail. The NPC will persist for 27 seconds. However, the trail is generated by a dynamic object and persists for only 18 seconds. Basically, as the trail moves, the beginning of it starts to dissipate while the end of it continues to move.
- A spell hook is used to get all active trails. Bots read the dynamic objects and from that attempt to position in a manner that is safe from all Doomfires. 
- I've tried a few different things , such as moving only from the closest point on a Doomfire from the bot, or different types of movement calls (currently it's a MoveTo() that is FORCED for non-healers but COMBAT for healers to allow them to cross Doomfire to reach a healing target if necessary). It's unclear what works best; the fight is just so random. So I may continue to tweak this avoidance.
- An amusing note here is that if I don't specifically block the trigger for the avoidance when a bot is Feared, they actually will refuse to be Feared and will avoid Doomfires anyway. Just another quirk of not having a client, I suppose.

brighton-chi:
**Edit: Disregard below--covered in separate PR**

I've added to PlayerbotAI::UpdateAI a boolean variable, pendingCastInterrupt, and a public function, RequestCastInterrupt(), that toggles the variable. This lets an external script hook into UpdateAI to interrupt a spell that is in the process of being cast. This should allow raid/dungeon strategies to actually interrupt spells, such as for avoiding hazards, as you cannot do so by calling InterruptSpells() or Reset() or anything else with a raid/dungeon action method. This method is very simple, but I did have Claude come up with it for me after weighing a few different options.

I've made use of the method for Archimonde's Doomfire--bots that are near Doomfire will call RequestCastInterrupt() and thus will actually interrupt their spells to run away from the hazard. I tested and it works well. I also tested this with the open Auchenai Crypts PR because that one is very easy to test whether this method works or not, and it did work.

Ultimately, I think this interrupt method has no additional cost or disadvantage and just provides a framework to give more capability to future strategies.

noisiver:
Considering that we typically enforce the rule of not adding more than necessary in a single PR, shouldn't the interrupt spell feature be a separate PR?

brighton-chi:
> Considering that we typically enforce the rule of not adding more than necessary in a single PR, shouldn't the interrupt spell feature be a separate PR?

I'm fine with that. It just doesn't actually do anything by itself so I wasn't sure if that made any sense in this case. But I can definitely just split that off to address first and then PR Hyjal later. I can take care of that later.

brighton-chi:
Decided to redo a bunch of stuff. I added more movement validation, which seems to be working better in the rough Hyjal terrain.

Other changes:
- Added Death & Decay avoidance for Rage Winterchill because I was sick of bots looking like idiots
- Changed melee RoF avoidance for Azgalor so that they'll swap to a Doomguard if RoF is dropped onto Azgalor instead of doing nothing and waiting it out
- Added a spell script + spell interrupt request to try to prevent all of the melee from being Air Burst at once (which leads to Archimonde pointing his finger at players and one-shotting them)

src/Ai/Raid/HyjalSummit/Strategy/RaidHyjalSummitStrategy.cpp
```diff
@@ -0,0 +1,137 @@
...
+ */
+
+#include "RaidHyjalSummitStrategy.h"
+#include "RaidHyjalSummitMultipliers.h"
+
+void RaidHyjalSummitStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
+{
+    // General
+    triggers.push_back(new TriggerNode("hyjal summit bot is not in combat", {
+        NextAction("hyjal summit erase trackers", ACTION_EMERGENCY + 11) }));
```
kadeshar, line 13: Death of one bot dont erase trackers?

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+bool AzgalorBossCastsRainOfFireOnRangedTrigger::IsActive()
+{
+    if (botAI->IsMelee(bot))
+        return false;
+
+    Unit* azgalor = AI_VALUE2(Unit*, "find target", "azgalor");
+    return azgalor && azgalor->GetVictim() != bot &&
+           !bot->HasAura(static_cast<uint32>(HyjalSummitSpells::SPELL_DOOM));
+}
```
kadeshar: Trigger name doesnt fit for content for me. If ranged bot dont have doom aura that means boss cast rain of fire on ranged?

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
```diff
@@ -0,0 +1,268 @@
...
+            return nullptr;
+
+        return &instanceIt->second;
+    }
+
+    TankPositionState GetAzgalorTankPositionState(PlayerbotAI* botAI, Player* bot)
+    {
+        Player* mainTank = GetGroupMainTank(botAI, bot);
+        if (!mainTank)
+            return TankPositionState::Unknown;
```
kadeshar, line 211: That means when main tank dies this method return unknown and that is desired state?

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
```diff
@@ -0,0 +1,268 @@
...
+            return nullptr;
+        }
+
+        if (elapsed >= RAIN_OF_FIRE_DURATION)
+            return nullptr;
+
+        return &instanceIt->second;
+    }
+
+    TankPositionState GetAzgalorTankPositionState(PlayerbotAI* botAI, Player* bot)
```
kadeshar, line 207: I assume that if player is main tank then must know this strategy and realize that himself. Right?

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if (bot->getClass() == CLASS_DRUID && tab == DRUID_TAB_FERAL)
+        return false;
+
+    if (!AI_VALUE2(Unit*, "find target", "kaz'rogal"))
+        return false;
+
+    if (bot->getClass() == CLASS_HUNTER)
+    {
+        return true;
+    }
```
kadeshar, line 148: That is ok that hunter always is consider and need escape path?

src/Ai/Raid/HyjalSummit/Strategy/RaidHyjalSummitStrategy.cpp
brighton-chi: I don't think they can run the strategy when they are dead. The trigger for being not in combat is a little wonky for bots though so that's why I do a separate gate with find target in the action. I guess maybe if they released and revived somewhere else they might erase? I'm not sure about that, but I can just add a map ID gate to the trigger then.

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
brighton-chi: Yeah, the action is just to spread in a position that allows for an escape path. There's a distance gate in the action (so if the bot is close enough to their position, they won't move, and the action just returns false). The reason Hunters just return true here is that I don't make them move away from other bots even if they get low on mana, they just put on Viper Sting, whereas casters need to get away if they are low on mana so they return false for this trigger if they are low on mana.

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
brighton-chi: So the associated action is to keep distance from nearby bots because Azgalor casts Rain of Fire. The reason Doom is excluded from everything is if bots have Doom they have to go to a particular place so they can't be worrying about other movements.

I could name it something generic like "AzgalorBossEngagedByRangedTrigger" if you prefer.

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
brighton-chi: The trigger has an early return for if the main tank is a player and Azgalor is under 95% health. So what should happen with a player main tank is that melee runs away to allow for positioning, but ranged is still going to attack so when ranged takes Azgalor to 95%, melee should attack.

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
brighton-chi: If you use the main tank flag, the main tank will still return true for IsMainTank() if it dies. If you don't use the flag, then the next living tank just becomes the main tank. So there should always be a main tank unless you don't use the flag and all of your tanks die, in which case you're wiping anyway.

kadeshar: GetGroupMainTank have alive check
obraz

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
kadeshar: But in this trigger nothing checking what boss casting. Its ok? If yes then something like AzgalorBossRangedUnsafePositionTrigger

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
brighton-chi: Mm, that doesn't really make sense, I think it should be removed from GetGroupMainTank and let IsMainTank decide whether alive or not matters.

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
brighton-chi: Yeah, it's just a general spread. RoF is from a dynamic object so no preemptive avoidance is possible without a spell listener, which I have for melee but is not really worth it for ranged here. So the actual avoidance is just through AvoidAoeAction().

This action is really just a generic positioning action for ranged based on the mechanics of the fight, which I do for many fights and just call something like "BossEngagedByRanged" because it's just standard positioning the bots maintain over the fight. I can rename it.

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
kadeshar: Maybe, but that is breaking change and require checking all usage

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
kadeshar: Ok, rename to much better fit real content

src/Ai/Raid/HyjalSummit/Util/RaidHyjalSummitHelpers.cpp
brighton-chi: Ok, I am going to add an early return in the multiplier that relates to going to the waiting position. Everywhere else there is already a check for GetGroupMainTank. So now if the main tank dies and isn't revived, the whole concept of going to the waiting position will just not happen. I think that's fine.

kadeshar (dismissed):

brighton-chi:
Decided to make the action for melee to wait out of the way just return false for a human main tank. Too much trouble, and players can manually handle the pull in a way to avoid melee getting cleaved if they really need to (and honestly the chance isn't that high to happen in the first place).

I'm sick of this one so I'm not making any more unsolicited changes unless I realize something is broken =P

src/Ai/Raid/HyjalSummit/Trigger/RaidHyjalSummitTriggers.cpp
```diff
@@ -0,0 +1,360 @@
...
+    return GetInfernoTarget(anetheron) != bot;
+}
+
+bool AnetheronBotIsTargetedByInfernalTrigger::IsActive()
+{
+    Unit* anetheron = AI_VALUE2(Unit*, "find target", "anetheron");
+    if (!anetheron || botAI->IsMainTank(bot))
+        return false;
+
+    if (botAI->IsMainTank(bot))
```
kadeshar: Duplicate conditions

kadeshar (changes requested):

kadeshar (approved):

PR #2259, opened 2026-03-29
Fix Warrior Battle Shout Spam

brighton-chi:
## Pull Request Description

There is a known problem that Warriors will repeatedly spam Battle Shout (BS) if there is a stronger Blessing of Might (BoM) present that they cannot override. It has been long known that the issue arises from the lack of a BS trigger. It was not a problem in the past because BS and BoM used to (erroneously) stack, and that was eliminated when AC ported TC's stacking rules.

This PR adds a trigger for BS that prevents the bot from attempting to cast it if a stronger BoM (or Greater BoM) is present. To do this, it compares the AP bonus of any BoM/GBoM applied to the Warrior bot, and then will attempt BS only if the Warrior has a stronger one.

I also did some minor reformatting of the other Warrior triggers, mostly just deleting comments and extra braces.

## Feature Evaluation

The trigger is fairly complicated because it uses an AP comparison and takes into account spell ranks and the talent Commanding Presence (which increases the strength of BS). The bug could be solved by a very simple aura check for BoM/GBoM. That's suboptimal though because BS will be stronger if the Paladin has a lower rank of BoM or the Warrior has Commanding Presence and the Paladin does not have Improved BoM. 

At the end of the day, this is a new trigger that runs every tick, but it's not really more expensive than generic BuffTriggers, which all do this. And it is needed to solve a very annoying bug that essentially renders Warriors useless in any optimized raid unless BS is disabled.

I don't believe this more robust trigger (as opposed to just calling BuffTrigger and additionally checking for any BoM/GBoM aura) has any further impact on performance--any added logic is practically free AFAIK (e.g., checking if the bot has a particular talent is just an integer lookup).

## How to Test the Changes

Form a group with a Warrior and Paladin bot, with the Warrior bot having a weaker BS than the Paladin's BoM (this is naturally the case at max level, unless the Warrior has Commanding Presence and the Paladin does not have Improved BoM). Have the Paladin cast BoM on the Warrior if it has not done so automatically. Enter combat and see if the Warrior uses BS. Try again with the Warrior having a stronger BS.

## Impact Assessment

    - - [ ] No, not at all

As noted above, this is a new BuffTrigger that runs every tick.

    - - [ ] No
    - - [x] Yes (**explain why**)

Default behavior was broken.

    - - [x] No

This is, with respect to the code, a pretty complex trigger. But that should not impact other logic, and if need be, there is a very simple version that could be implemented that would at least fix the bug, as noted above.

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

I used Claude Sonnet 4.6. This is essentially the first PR that I entirely vibe coded. I knew what the problem was and how to fix it, including the conditions I wanted in the trigger, and the LLM did the rest. I can tell you what each block does and why that's needed, but I can't tell you how they all work technically.

I tested this ingame in scenarios where the available BS was stronger and not as strong, and in each case, the correct behavior was observed. I've since run several raids without BS disabled on my Warriors, and they are correctly not using it (as my Ret Paladin is applying a stronger BoM). The trigger does not show up as an expensive one with pmon (unlike, say, VIgilance, which is probably the most expensive class trigger in the mod (excluding any DK triggers just because I've never used one so I have no idea)).

## Final Checklist


brighton-chi:
Weird, somebody left a comment and then deleted it. But to respond, the linked PR is a broader PR with respect to Paladin greater blessings, and it did include a change to address this issue, but such change just disables Battle Shout use entirely when BoM/GBoM is active. It fixes the spamming, but in some cases, Battle Shout is stronger so it would be better if the bot were to use it.

src/Ai/Class/Warrior/Trigger/WarriorTriggers.cpp
```diff
@@ -117,5 +84,74 @@ bool ShatteringThrowTrigger::IsActive()
...
+        for (uint8 eff = 0; eff < MAX_SPELL_EFFECTS; ++eff)
+        {
+            if (bomInfo->Effects[eff].ApplyAuraName == SPELL_AURA_MOD_ATTACK_POWER)
+            {
+                int32 bomApValue = bomInfo->Effects[eff].BasePoints + 1;
+                if (bomApValue >= effectiveBsAp)
+                    return false;
+                break;
+            }
+        }
```
kadeshar, line 153: That work with Improved Blessing of Might?

brighton-chi: It does, although it's being wonky with Commanding Presence right now it seems (Warrior with Rank 7 Battle Shout and Commanding Presence (300 AP) is not using it when buffed by Rank 7 Improved Blessing of Might (290 AP)). Lemme see if I can figure that out.

brighton-chi: Actually I bet I know the issue. I think the talent check should use HasAura() instead of HasSpell(). I'll take a closer look tomorrow, but I think that's it.

brighton-chi: Yep, just needed to change to HasAura(). Good to go.

kadeshar (approved):

PR #2269, opened 2026-03-31
Bug fix. Edge case where bots would get stuck in cities. 

Celandriel:
## Pull Request Description

When I refactored flight destinations, I wanted to make where bots go more intentional. so I made it dependent on the allianceHubsPerLevelCache and hodeHubsPerLevelCache. This system relied on there being an innkeeper in each area that the bots would fly to. However, not every zone has an innkeeper, and so there was an odd situation where bots had nowhere to fly to. (Most notably at level 53.) This solves that by hardcoding the flightmasters in those areas into the cache. 

I also put back in the city teleport probability check which was forcing every bot to teleport to a city on level up. 

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [ ] No

Debugging and comments.

## Final Checklist

### DIFF
--- a/src/Mgr/Travel/TravelMgr.cpp
+++ b/src/Mgr/Travel/TravelMgr.cpp
@@ -4419,6 +4419,7 @@ std::vector<std::vector<uint32>> TravelMgr::GetOptimalFlightDestinations(Player*
                                             bot->GetTeamId());
     if (!fromNode)
         return validDestinations;
+
     std::vector<WorldLocation> candidateLocations;
     if (bot->GetLevel() >= 10 && urand(0, 100) < sPlayerbotAIConfig.probTeleToBankers * 100)
         candidateLocations = GetCityLocations(bot);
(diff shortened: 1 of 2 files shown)


src/Mgr/Travel/TravelMgr.cpp
```diff
@@ -4673,6 +4674,31 @@ void TravelMgr::PrepareDestinationCache()
                 if (forAlliance)
                     allianceFlightMasterCache[guid] = pos;
                 flightMastersCount++;
+
+                // Zones that have flight masters but no innkeepers — use flight master as hub
+                static const std::set<uint32> zonesWithoutInnkeeper = {
+                    4,    // Blasted Lands (52-57)
```
brighton-chi, line 4680: I thought Blasted Lands had an innkeeper (Alliance only, but so is the flightpath).

Edit: Ignore me, guess I'm crazy, there isn't one.

hermensbas (approved):

hermensbas:
Still a few build errors though. hence not approved label yet

Celandriel:
> Still a few build errors though. hence not approved label yet

Build check is broken because test-staging has breaking core changes merged into it. Should be fine, but will confirm on update

PR #2270, opened 2026-03-31
Remove fire totem override from enhancement AOE strategy

dillyns:
## Pull Request Description

Currently in AOE enchancement shamans are dropping magma totems. They should not be overriding the set fire totem strategy.

## Feature Evaluation

## How to Test the Changes

Bring an enhancement shaman into medium aoe scenario (3 enemies)
They should no longer replace their fire totem with magma totem. They should only drop the fire totem that is set in their strategies.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## Messages to Translate

    - - [x] No

|			 |			      |
|			 |			      |

## AI Assistance

    - - [x] No

## Final Checklist

### DIFF
--- a/src/Ai/Class/Shaman/Strategy/GenericShamanStrategy.cpp
+++ b/src/Ai/Class/Shaman/Strategy/GenericShamanStrategy.cpp
@@ -156,9 +156,7 @@ void ShamanAoeStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
     }
     else if (tab == 1)  // Enhancement
     {
-        triggers.push_back(new TriggerNode("medium aoe",{
-                                                    NextAction("magma totem", 24.0f),
-                                                    NextAction("fire nova", 23.0f), }));
+        triggers.push_back(new TriggerNode("medium aoe",{ NextAction("fire nova", 23.0f), }));
 
         triggers.push_back(new TriggerNode("maelstrom weapon 5 and medium aoe", { NextAction("chain lightning", 22.0f), }));
         triggers.push_back(new TriggerNode("maelstrom weapon 4 and medium aoe", { NextAction("chain lightning", 21.0f), }));


kadeshar:
Currently there is totem strategy which automatic switch between Totem of Wrath/Searing Totem (depends of Demonic Pact) in single target fight to Magma Totem in AOE fight and back?

kadeshar:
Conflict

dillyns:
> Currently there is totem strategy which automatic switch between Totem of Wrath/Searing Totem (depends of Demonic Pact) in single target fight to Magma Totem in AOE fight and back?

You may need to clarify what you are trying to say here, I don't think I quite understand. But the issue here is that this enhancement aoe strategy is overwriting whatever totem strategy is set in order to drop magma/searing totem if there are 3+ enemies. If I tell my shaman I want flametongue totem for example, it should only drop flametongue totem regardless of aoe or not.

Celandriel:
> Currently there is totem strategy which automatic switch between Totem of Wrath/Searing Totem (depends of Demonic Pact) in single target fight to Magma Totem in AOE fight and back?

@kadeshar Im not following here either.

kadeshar:
Worst is that me neither xD

kadeshar (approved):

YunataSavior:
On a side note, we should probably better document `enh` combat strat (among other things) in the Playerbots Commands page. Last week, I started a lvl 1 playthrough with an altbot shaman who I wanted to be enhancement. The bot was casting lightning bolt and not meleeing, and I was confused why until I discovered that the bot had `ele` strategy. Things like these aren't really documented.

ThePenguinMan96:
Bots acquire their strategies based on their talent points. Prior to level 10, the default spec for shamans is elemental. Each class has a default spec prior to level 10, typically the one that performs the best at that level.

PR #2276, opened 2026-04-01
Add default case to mount initialization for bots.

heyitsbench:
## Pull Request Description

This PR adds a default case for the mount initialization function for player bots, allowing custom race additions to not crash when added to an AzerothCore server (such as [`mod-worgoblin`](https://github.com/heyitsbench/mod-worgoblin)).

## Feature Evaluation
This feature ideally does not get touched in standard `mod-playerbots` usage, as it only adds a case to a switch statement, and would not be getting taken with Blizzlike data (which the vast majority of users likely use).

## How to Test the Changes
1. Add custom races to your AzerothCore instance (most easily accomplished with the above linked module).
2. Start the server and configure it to create random player bots (that fall under the added custom races).
3. Observe no crash.

## Impact Assessment
    - - [X] No, not at all

    - - [X] No

This does add another path in a switch statement. It also creates another set of data for that case. This could have been avoided by consolidating the human/orc cases into the default case, but I elected not to do that in case players make their own changes. If preferred, I can definitely consolidate the cases to not have those redundant data points.

## Messages to Translate
    - - [X] No

## AI Assistance
    - - [X] No

## Final Checklist

## Notes for Reviewers
Pretty please? 🥹

### DIFF
--- a/src/Bot/Factory/PlayerbotFactory.cpp
+++ b/src/Bot/Factory/PlayerbotFactory.cpp
@@ -3039,8 +3039,10 @@ void PlayerbotFactory::InitMounts()
                 fast = {23228, 23227, 23229};
             }
             else // Human mounts
-            slow = {6654, 6653, 580};
-            fast = {23250, 23252, 23251};
+            {
+                slow = {6654, 6653, 580};
+                fast = {23250, 23252, 23251};
+            }
     }
 
     switch (bot->GetTeamId())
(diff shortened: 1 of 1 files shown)


Celandriel (approved):

heyitsbench:
Thanks for catching that, I'm apparently a noob. 😆

Celandriel:
Happens to the best of us :)

hermensbas (approved):

PR #2288, opened 2026-04-06
Performance(Core): Some activity sec to ms init fixes, global activity loop check and some additional minor fixes

hermensbas:
## Pull Request Description

1. Corrected the init activity times; Since ive changed (previous) the calc from seconds to ms to increase more scope for the offset execute jitter (removed timer(), 0 is correct way todo now with MS) (also broke self-bot)
2. Global loop checks in activityAllowed instead vs multiple loops, this function is called very often so we better optimize it.
3 Fixed the broken 'HasManyPlayersNearby' function and then deleted it :O To fragile due various edge cases and rather expensive call, besides lets activeAlone deal with this situation which is way more controlled.
4. Some additional small fixes that where unnoticed overtime.
5. Added/Changed inline comments which makes more sense and explains what it does.
6. Removed dead code freeze bots during init.
7. self-bot fix

## Impact Assessment

    - - [ ] No, not at all

When playing with larger amount of real players makes the allowed activity perform better.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -4485,81 +4485,16 @@ bool PlayerbotAI::HasPlayerNearby(float range)
     return HasPlayerNearby(&botPos, range);
 };
 
-bool PlayerbotAI::HasManyPlayersNearby(uint32 trigerrValue, float range)
-{
-    uint32 found = 0;
-
-    for (auto& player : sRandomPlayerbotMgr.GetPlayers())
-    {
-        if ((!player->IsGameMaster() || player->isGMVisible()) &&
-            ServerFacade::instance().GetDistance2d(player, bot) < range)
-        {
-            found++;
-
-            if (found >= trigerrValue)
-                return true;
-        }
-    }
-
-    return false;
-}
-
-inline bool HasRealPlayers(Map* map)
-{
-    Map::PlayerList const& players = map->GetPlayers();
-    if (players.IsEmpty())
-    {
-        return false;
-    }
-
-    for (auto const& itr : players)
-    {
-        Player* player = itr.GetSource();
-        if (!player || !player->IsVisible())
-        {
-            continue;
-        }
-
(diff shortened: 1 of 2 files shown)


src/Bot/PlayerbotAI.cpp
kadeshar, line 4566: Why modify method which is not used. Only usage of this method is deleted in this PR

hermensbas: Aah wrong copy/pasta, fixed but back to draft since i just realized something;  if (HasManyPlayersNearby(10, 40))
doesnt seem to trigger. I wanna investigate that first, otherwise we will have halfass version.

kadeshar: I also saw that but didnt commented that. Now instead using squared distance use normal which means a lot of difference

hermensbas: getPlayers instead of getAllBots :O was the main issue, the rule was mostly interested in other bots. But this entire rule with new RPG makes no sense. So after fixing it i kinda realize we dont want this logic in first place.
Lets activeAlone deal with this way more stabile way, there are to many edge cases which can cause keeping bots active while they should not. 

Still ended up combining some checks to prevent those multi loops and removing the inline functions. Less readable, but since this function is called so often we rather see it optimized.

hermensbas:
The only measure left is to prioritize the actions, but the difference between performance is minimal to none.  I may add at some point. For now i dont see enough reason to add more complexity.

hermensbas:
Wanna run some additional test tommorow, before putting it up for review again

hermensbas:
.playerbot bot self, /w maintenance, /w autogear, /w nc +new rpg,+grind+loot working again or default in BG

hermensbas:
Done testing

kadeshar (approved):

PR #2289, opened 2026-04-06
Feat: Reintroduce timed logouts

NoxMax:
## Pull Request Description

Timed logouts was enabled in the past, but was disabled due to a misdiagnosis on an issue that caused crashes. That issue was better understood and resolved in #2131. Now timed logouts are reintroduced for alt-bots and addclass-bots. As before, random-bots do not and should not accept logout commands.

Note that if the bot's master has instant logout privileges according to `InstantLogout` in worldserver.conf, so would the bot. If not, neither would the bot.

## Feature Evaluation

Feature at minimum logic needed to implement. No measurable processing cost.

## How to Test the Changes

1. Have your account gmlevel set to 2 in acore_auth>account_access just to test how it works with security parameters.
2. Set `InstantLogout = 3` in worldserver.conf.
3. Invite an addclass or alt-bot.
4. Bot should logout if whispered "logout" to or if you logout.
5. If not in a resting place, neither you nor the bot should be able to instant logout according to the security setting.
6. After the bot logout, log it back in and whisper "logout" again, but right after whisper "cancel logout" or "logout cancel". That should cancel the logout.
7. Have your account gmlevel set to 3 in acore_auth>account_access. (when you change your gmlevel, you need to log out of your account for the change to take effect)
8. You are now at admin gmlevel, and with  `InstantLogout = 3`, you and any bot under your command should be able to logout instantly.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)
Bots now trigger instant or timed logout under the same circumstances that would apply to their master.

    - - [ ] No
Checks if bot is eligible for instant logout or not. If not, timed logout applies to them.

## AI Assistance

- - [x] No

## Final Checklist

## Notes for Reviewers

While testing this I was having some odd issues with the command `account set gmlevel`. Not sure what's going on there, but for purposes of testing this and changing your account security level, doing it directly in the DB at "acore_auth>account_access" is going to be most reliable.

### DIFF
--- a/src/Bot/PlayerbotMgr.cpp
+++ b/src/Bot/PlayerbotMgr.cpp
@@ -293,6 +293,11 @@ void PlayerbotHolder::LogoutAllBots()
         if (!botAI || botAI->IsRealPlayer())
             continue;
 
+        // If bot is mid-countdown, cancel the timer so LogoutPlayerBot proceeds immediately.
+        WorldSession* session = bot->GetSession();
+        if (session && session->isLogingOut())
+            session->SetLogoutStartTime(0);
+
         LogoutPlayerBot(bot->GetGUID());
     }
 }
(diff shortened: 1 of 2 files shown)


hermensbas:
Still not huge fan of functionality; but if i remember correctly same ppl wanted it including Revision

src/Bot/PlayerbotMgr.cpp
```diff
@@ -374,17 +388,30 @@ void PlayerbotHolder::LogoutPlayerBot(ObjectGuid guid)
         if (masterWorldSessionPtr && !masterWorldSessionPtr->GetPlayer())
             logout = true;
 
-        if (bot->HasFlag(PLAYER_FLAGS, PLAYER_FLAGS_RESTING) || bot->HasUnitState(UNIT_STATE_IN_FLIGHT) ||
-            botWorldSessionPtr->GetSecurity() >= (AccountTypes)sWorld->getIntConfig(CONFIG_INSTANT_LOGOUT))
+        // Master's socket is already gone (EXIT GAME -> EXIT NOW is the most typical cause).
+        // Force instant logout. Without this, the bot restarts its 20-second countdown and fires LogoutPlayer() 20 seconds
+        // after the master's Player object has been deleted, causing the bot's logout to crash on the now deleted master.
+        if (masterWorldSessionPtr && masterWorldSessionPtr->IsSocketClosed())
+            logout = true;
```
NoxMax: Pushed this exception for EXIT NOW.
EXIT NOW doesn't actually EXIT NOW in core. Instead your character hangs around for another 60 seconds after the client shutdown. The fix is simple and I will post it to AC, but the fix can also cause a crash triggered Playerbots-side. This exception prevents that.

kadeshar: I know its not you but when logout == true then rest conditions until line 407 should be skipped

NoxMax:
Made the entire logout check into a single ||, with a more or less cheapest checks first. Format wise it looks okay-ish, but regardless it will exit as true on the first true condition. Also removed `botWorldSessionPtr->ShouldLogOut` after a closer look. If it's true, the block would never be reached anyway because of the `if (botWorldSessionPtr->ShouldLogOut(time(nullptr)))` from the block above

kadeshar (approved):

PR #2292, opened 2026-04-06
Druid - lifebloom

kadeshar:
## Pull Request Description

Restoration Druid now use lifebloom when get Clearcasting proc.

## How to Test the Changes

1. Invite to group restoration druid and tank
2. Attack target (for example dummy)
3. When druid heals and get clearcasting should cast lifebloom on tank

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Restoration druid by default use lifebloom on tank when got clearcasting proc.

    - - [x] No

## AI Assistance

- - [ ] No

Copilot CLI to review.

## Final Checklist

## Notes for Reviewers

obraz
obraz

### DIFF
--- a/src/Ai/Class/Druid/Action/DruidActions.cpp
+++ b/src/Ai/Class/Druid/Action/DruidActions.cpp
@@ -23,6 +23,16 @@ std::vector<NextAction> CastAbolishPoisonOnPartyAction::getAlternatives()
                              CastSpellAction::getPrerequisites());
 }
 
+bool CastLifebloomOnMainTankAction::isUseful()
+{
+    Unit* target = GetTarget();
+    if (!target || !target->IsAlive() || !CastSpellAction::isUseful())
+        return false;
+
+    Aura* lifebloom = botAI->GetAura("lifebloom", target, true, true);
+    return !lifebloom || lifebloom->GetStackAmount() < 3 || lifebloom->GetDuration() < 2000;
+}
+
 Value<Unit*>* CastEntanglingRootsCcAction::GetTargetValue()
 {
     return context->GetValue<Unit*>("cc target", "entangling roots");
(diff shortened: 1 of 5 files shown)


src/Ai/Class/Druid/Action/DruidActions.cpp
```diff
@@ -23,6 +23,16 @@ std::vector<NextAction> CastAbolishPoisonOnPartyAction::getAlternatives()
...
 }
 
+bool CastLifebloomOnMainTankAction::isUseful()
+{
+    Unit* target = GetTarget();
+    if (!target || !target->IsAlive() || !CastSpellAction::isUseful())
+        return false;
+
+    Aura* lifebloom = botAI->GetAura("lifebloom", target, true, true);
+    return !lifebloom || lifebloom->GetStackAmount() < 3 || lifebloom->GetDuration() < 2000;
```
brighton-chi, line 33: So it seems the intent here is to keep 3 stacks rolling. Is that how you'd normally use Lifebloom in WotLK? Since it refunds half of the mana cost when it completes, it could also be used to regen mana, but that approach would better have the condition to be just no lifebloom on the target at all (and not consider stacks or duration). I never played WotLK so I don't know what is typical.

kadeshar: I researched that area long time ago but yes. Clearcasting from trigger = free mana then half cost back is redundant
obraz

brighton-chi: I think the cost is "refunded" even if not paid. Basically you just get half the base cost added as a net positive.

kadeshar: Better keep stacks up instead of getting mana.

Celandriel (approved):

PR #2293, opened 2026-04-07
Fix: Compatibility patch for #2250 BotActiveAlone update

NoxMax:
## Pull Request Description

Possible fix to self-bot breaking. Issue was at this evaluation:
```
    if (!checkNow && getMSTime() < (allowActiveCheckTimer[activityIndex] + 4500 + offset))
        return allowActive[activityIndex];
```
allowActiveCheckTimer was set to current time, this made sense with the old evaluation:
```
    if (!checkNow && time(nullptr) < (allowActiveCheckTimer[activityIndex] + 5))
        return allowActive[activityIndex];
```
but with the new evaluation, it will always return true. Or at least return true until the server is on for +1.77 billion ms (about 20 days)

Technically the problem should've affected all bots, but regular bots get
```
    // check activity
    botAI->AllowActivity(ALL_ACTIVITY, true)
```
at OnBotLogin. And because it passes checkNow = true, that evaluation fails for them and allowActiveCheckTimer gets properly set to getMSTime()

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -119,7 +119,7 @@ PlayerbotAI::PlayerbotAI()
 
     for (uint8 i = 0; i < MAX_ACTIVITY_TYPE; i++)
     {
-        allowActiveCheckTimer[i] = time(nullptr);
+        allowActiveCheckTimer[i] = 0;
         allowActive[i] = false;
     }
 }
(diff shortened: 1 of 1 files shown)


NoxMax:
already fixed with #2288

PR #2295, opened 2026-04-07
Modify UpdateAI to Allow Future Methods to Interrupt Bot Spells

brighton-chi:
Note: Resubmitted because the prior PR had master for the source

## Pull Request Description

This PR adds to PlayerbotAI::UpdateAI a boolean variable, pendingCastInterrupt, and a public function, RequestCastInterrupt(), that toggles the variable and would then call InterruptSpells in UpdateAI. This lets an external script hook into UpdateAI to interrupt a spell that is in the process of being cast. This should allow raid/dungeon strategies to actually interrupt spells, such as for avoiding hazards, as you cannot do so by calling InterruptSpells() or Reset() or anything else with a raid/dungeon action method.

## Feature Evaluation

There is no processing cost from this PR itself since it just adds a simple boolean check that will always return false unless other methods are implemented to call RequestCastInterrupt().

## How to Test the Changes

This PR doesn't do anything by itself, but confirmation of no performance impact can be tested by just playing with the PR merged. I tested this by introducing scripts that called RequestCastInterrupt() for Archimonde (included in the Hyjal PR now) and for Auchenai Crypts (built on the strategy from flashtate's PR just for test purposes), and spells were interrupted as intended.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [ ] No

I'm not sure if it counts as "yes" here, but this PR opens up the ability for scripts to be added that would add real checks to UpdateAI. The impact of such scripts will depend on how they are implemented, however. It is possible (and intended) for the external calls to be highly limited in scope (e.g., only bots in a particular circumstance in a particular boss fight).

## AI Assistance

- - [ ] No

Claude Sonnet 4.6 was used to brainstorm different possibilities to allow external scripts to interrupt spells. I considered different options before settling on this one due to it not requiring core changes, being easily usable in boss strategies, and not having any performance impact on its own.

## Final Checklist


src/Bot/PlayerbotAI.cpp
```diff
@@ -331,6 +331,15 @@ void PlayerbotAI::UpdateAI(uint32 elapsed, bool minimal)
                 ServerFacade::instance().SetFacingTo(bot, spellTarget);
             }
 
+            // Allow external scripts to interrupt a cast in progress
+            if (pendingCastInterrupt)
+            {
+                pendingCastInterrupt = false;
```
kadeshar: What if 
pendingCastInterrupt == true 
and dont meet one of this condition
if (currentSpell)
if (spellInfo && getState() == SPELL_STATE_PREPARING)
Shouldnt be changed on false?
If not that not interupt next spell?

brighton-chi: Good point. How about this?

```
if (pendingCastInterrupt)
{
    pendingCastInterrupt = false;
    if (currentSpell && spellInfo &&
        currentSpell->getState() == SPELL_STATE_PREPARING)
    {
        InterruptSpell();
        YieldThread(bot, GetReactDelay());
        return;
    }
}

kadeshar: This spellInfo is required, right?

brighton-chi: Well, actually thinking about it again, everything is already nested inside currentSpell && spellInfo && SPELL_STATE_PREPARING checks. So I'm not sure we do need to change anything.

kadeshar: We should clear flag when dont meet existing conditions otherwise that interupt next spell

brighton-chi: Oh I see what you mean. I don't think there is a way to clear the flag unless we had some sort of expiration to it, or at least nothing comes to mind to me. What if the spell cast conditions were built into RequestCastInterrupt() so the flag is not set to true unless there is a spell casting? That seems more straightforward to me.

Actually your point about SPELL_STATE_PREPARING made me realize this is not currently interrupting channels. That's easy enough to modify I think.

ThePenguinMan96: There is a cancel channel action I have previously used in the mage, boomkin, hunter, and warlock PRs. I believe the file is cancel channel action.h and .cpp. It stops the bot from channeling - not sure if it works for casting, but might be worth a try. I linked it to custom triggers I made, such as "volley channel check" where it checks the number of enemies nearby, and if that is less than 3, it stops channeling the spell and moves back into the single target rotation. 

Might help this instance :)

brighton-chi: Ah I see that now. Interesting that it works, as calling InterruptNonMeleeSpells does not seem to have ever worked for me to cancel channeled spells, and you are using AC's InterruptSpell so I think they should function the same (but maybe passing a specific spell ID changes the outcome, idk). There is a PlayerbotAI::InterruptSpell() also that seems more robust and sends packets, but I haven't tried it before.

Regardless though, a spell in the process of casting should be currently impossible to interrupt because UpdateAI just early returns when there's a spell casting. For simplicity (I don't see why somebody would want to interrupt a pending cast but not a channel), I think we can use the subject of this PR for both pending and channeling casts since going forward it means that externally you just make one call, and that call is just a void within your avoidance action (no need to do a separate action).

And I think spellInfo is not actually required here. It shouldn't matter. So I can reorder things.

kadeshar: obraz

Dont understand, this is delete on purpose? If yes, why.

brighton-chi: Was messing with channeled spells and forgot to put that back for channels being prepared, thanks

kadeshar (approved):

NoxMax:
> Note: Resubmitted because the prior PR had master for the source

Quick sidenote: You can click the edit pencil next to the PR title, and that would allow you to change the base branch.

brighton-chi:
> > Note: Resubmitted because the prior PR had master for the source
> 
> 
> 
> Quick sidenote: You can click the edit pencil next to the PR title, and that would allow you to change the base branch.

Oh, great, thanks. When test-staging was first set up, I was trying to figure out how to change it with Keleborn, and at the time, he was able to but I didn't have the option. But I guess we all can now so that's convenient.

PR #2298, opened 2026-04-08
Fix Undead/Draenei bots stuck in starting zones

ghost:
## Pull Request Description

  Undead and Draenei bots get stuck in an idle/rest loop in their starting zones because the default NPC scan range (150f) and quest giver filter (80f) are too small and not enough. NPCs fall within range for `WANDER_NPC` to activate (requires >= 3).

  This adds a configurable area-based override that increases both ranges to 200f only in affected areas. All other zones remain at default values.

  ## Feature Evaluation

  - **Minimum logic**: One `std::set::count()` lookup per `Calculate()` call to check if the bot's current area is in the override list. If yes, scan range is 200f instead of 150f.
  - **Processing cost**: `GetAreaId()` is a cached uint32 read (~1ns). `std::set::count()` on a 2-element set is O(log n) ≈ 1 comparison. Negligible compared to the grid scan itself  (~33,000-53,000ns).

  ## How to Test the Changes

  1. Create Undead or Draenei bots (level 1)
  2. Observe that they pick up quests and start moving in Deathknell/Ammen Vale
  3. Without this fix, they sit in REST status indefinitely
  4. Optionally add/remove area IDs via `AiPlayerbot.RpgScanRangeOverrideAreaIds` in playerbots.conf

  ## Impact Assessment

      - [x] Minimal impact
      - pmon data (500 bots, extended run):
        - Default 150f: 0.033ms avg
        - Global 200f: 0.053ms avg
        - Area check (this PR): 0.042ms avg
      - The 0.009ms increase over default is caused by bots currently in starting zones scanning at 200f. Bots outside override areas are unaffected. 

      - [x] No

      - [x] No
      - Uses the existing `LoadSet`/`std::set` config pattern already used throughout the codebase.

  ## AI Assistance

  - [x] Yes
    - Used AI to speed up understanding the codebase, locate relevant functions, and compare with the cmangos playerbots implementation. All code was reviewed and tested manually.

  ## Final Checklist

  ## Notes for Reviewers

  The root cause is that `WANDER_NPC` requires `possibleTargets.size() >= 3`, but sparse starting zones have fewer than 3 NPC-flagged units within 150f. Increasing the scan range to 200f brings enough NPCs into range for the status check to pass. The override is configurable via `AiPlayerbot.RpgScanRangeOverrideAreaIds` so server admins can add more areas without code changes.

### DIFF
--- a/src/Ai/Base/Value/PossibleRpgTargetsValue.cpp
+++ b/src/Ai/Base/Value/PossibleRpgTargetsValue.cpp
@@ -13,6 +13,7 @@
 #include "ServerFacade.h"
 #include "SharedDefines.h"
 #include "NearestGameObjects.h"
+#include <unordered_set>
 
 std::vector<uint32> PossibleRpgTargetsValue::allowedNpcFlags;
 
@@ -88,8 +89,11 @@ bool PossibleRpgTargetsValue::AcceptUnit(Unit* unit)
 
 std::vector<uint32> PossibleNewRpgTargetsValue::allowedNpcFlags;
 
+// Sparse starting zones where the default scan range is insufficient for WANDER_NPC (requires >= 3 NPCs)
+static const std::unordered_set<uint32> rpgRangeOverrideAreaIds = { 3526 /* Ammen Vale */, 2117 /* Deathknell */ };
+
 PossibleNewRpgTargetsValue::PossibleNewRpgTargetsValue(PlayerbotAI* botAI, float range)
-    : NearestUnitsValue(botAI, "possible new rpg targets", range, true)
+    : NearestUnitsValue(botAI, "possible new rpg targets", range, true), defaultRange(range)
 {
     if (allowedNpcFlags.empty())
     {
@@ -119,6 +123,11 @@ PossibleNewRpgTargetsValue::PossibleNewRpgTargetsValue(PlayerbotAI* botAI, float
 
 GuidVector PossibleNewRpgTargetsValue::Calculate()
 {
+    if (rpgRangeOverrideAreaIds.count(bot->GetAreaId()) && defaultRange < 200.0f)
+        range = 200.0f;
+    else
+        range = defaultRange;
+
     std::list<Unit*> targets;
     FindUnits(targets);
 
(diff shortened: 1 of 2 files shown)


NoxMax:
Given that this is a fix to a niche situation, I don't see a need for it to be a config. Keeping it as list internally defined in PossibleRpgTargetsValue.cpp should be fine.

Also, this condition is not best practice
```
    if (sPlayerbotAIConfig.rpgScanRangeOverrideAreaIds.count(bot->GetAreaId()))
        range = 200.0f;
    else
        range = 150.0f;
```
A better approach is to capture the default range at the header as a private member, and override it if need be.
```
    if (rpgRangeOverrideAreaIds.count(bot->GetAreaId()) && defaultRange < 200.0f)
        range = 200.0f;
    else
        range = defaultRange;
```
This way in the future if for whatever reason the range was changed to +200, then this wouldn't override it with the lesser value of 200. And the default is not hardcoded to 150 here; instead just use `defaultRange` that was captured from the header.

ghost:
Refactored as suggested and defaultRange now captures the header default in the constructor, override only applies if defaultRange < 200.0f. Removed config, area IDs are defined internally. Diff is down to 2 files.

NoxMax:
Two more things. `unordered_set` is gonna be a better option than `set`. Makes the lookup of area ID ever so slightly faster. Just make sure to add `#include <unordered_set>`. And leave a short comment above the list explaining why this override was needed in these areas.

ghost:
switched to unordered_set and added a comment. Thanks for the feedback, much cleaner this way :)

ghost:
Verified via DB: the 3rd nearest NPC-flagged unit is at 168f (Deathknell) and 191f (Ammen Vale) from spawn. 200f is close to the minimum needed.

kadeshar:
How that looks in performence numbers? For me looks ok, only asking from performance side.

ghost:
pmon enabled:

default: ```1.320%      0.280s |     0.0 ..     0.3 (     0.033 of       8513) - Value     : PossibleNewRpgTargetsValue```


area check: ```1.488%      0.579s |     0.0 ..     0.3 (     0.041 of      14248) - Value     : PossibleNewRpgTargetsValue```

kadeshar (approved):

PR #2312, opened 2026-04-17
Enable travel node system for RPG pathfinding

Celandriel:
## Pull Request Description

Enable the TravelNode system for RPG pathfinding. This allows bots to pathfind beyond the distances of the path generating system, can handle anywhere where a node currently exists. Some notable limitations it resolves is bots getting stuck in Darnassus. 

I made a couple of changes to the system overall. I removed the use of hearthstones or mage portals, because it mutated the node system and I couldnt think about how to make it more safe. 

I also modified the system for generating additional nodes, but havent included that in this PR just yet. There are some new ones that are still needed. 

I created a shared teleport command in BotAI that reduced duplication in code and gave a single point to maintain the code. 

I did something similar with path generators since its being used in many places. 

I moved the namespace functions used for transport out of follow to movementactions since it inherits it anyway. 

## Feature Evaluation

It can get expensive, but I did a number of things to try and simplify that. The first is pre-compute a connection node to identify if the route between nodes exists. Then I made made a zone based node lookup to try to reduce the number of data points that it needs to scan before finding the nearest node. 

Then when developing the spline it only runs up to 100 points along the path. 

The cost isnt *huge* from what I have tested in the past. WIll test again with this isolated version. 

## How to Test the Changes

Just enable the option AiPlayerbot.EnableTravelNodes  to try, and bots will start using it. 

## Impact Assessment

    - - [ ] No, not at all
    - - [X] Minimal impact (**explain below**)

    - - [ ] No
    - - [x] Yes (**explain why**)

    - - [ ] No

Basically, there are now two main paths when bots travel that are available to them. Currently its based on a simple distance of 300f. 

The added complexity is configurable right now, with extra logging messages in place to identify hot spots where we need to add additional points. To be removed when new points are added. 

## AI Assistance

- - [ ] No

Claude Opus 4.6
A lot of brainstorming and implementation. but I spent quite a while reviewing everything. 

## Final Checklist

- - [ ] Stability is not compromised.
- - [X] Any new bot dialogue lines are translated.
- - [X] Documentation updated if needed (Conf comments, WiKi commands).

## Notes for Reviewers

It is unclear if this is the source of my crashes or not right now. There is something what I have been experiencing that causes frequent crashes.


src/Mgr/Travel/TravelNode.cpp
```diff
@@ -1965,25 +1784,40 @@ void TravelNodeMap::removeLowNodes()
         std::vector<TravelNode*> nodes = node->getNodeMap(true);
 
         if (nodes.size() < 5)
+        {
+            std::vector<TravelNode*> allInCluster = node->getNodeMap();
+            LOG_INFO("playerbots", "[RemoveLow] Cluster starting at '{}' (map {}) has only {} important nodes, {} total — marking for removal:",
+                     node->getName(), node->GetMapId(), nodes.size(), allInCluster.size());
+            for (auto& rn : allInCluster)
+                LOG_INFO("playerbots", "[RemoveLow]   - '{}' (map {}, {:.0f},{:.0f},{:.0f}) important={}",
+                         rn->getName(), rn->GetMapId(), rn->getX(), rn->getY(), rn->getZ(), rn->isImportant());
```
kadeshar: Im not a fan of putting everywhere LOG_xxx especial to loop. I dont know how many record allInCluster can have but in this situation if possible i would check shouldLog before that.

kadeshar:
Crashes sounds bad :D I didnt had crashes on current test staging.
Could you somehow review that is related with PR?

Also what about performance diff when this feature is enabled? I see a lot loops/nested loops etc

Celandriel:
## Profile results — 1.06M log lines##
Function | Calls
-- | --
MoveFarTo (lines) | 956,520
ExecuteTravelPlan | 77,820
GetFullPath | 17,586
GetTravelPlan | 5,943

# 
MoveFarTo EXIT reasons (cumulative ms = avg × count)
Reason | Count | Avg | Max | Total
-- | -- | -- | -- | --
empty-dest | 327,801 | 42µs | 730ms | 13,825 ms
let-spline-finish | 75,601 | 50µs | 275ms | 3,819 ms
no-path | 6,890 | 503µs | 71ms | 3,469 ms
travel-plan | 25,991 | 96µs | 63ms | 2,509 ms
waiting-last-move | 6,909 | 70µs | 123ms | 488 ms
primary-path | 1,390 | 358µs | 1.3ms | 498 ms
cone-fallback | 825 | 515µs | 1.6ms | 425 ms
short-range | 601 | 116µs | 0.7ms | 70 ms
stuck-teleport | 733 | 77µs | 0.6ms | 56 ms
-- | -- | -- | -- | --

MoveFarTo distribution: 91.6% finish in 10–50µs; only 0.1% take &gt;1ms.

# MoveFarTo internal sections
Section | Count | Avg | Max | Total
-- | -- | -- | -- | --
primary GeneratePath | 9,105 | 245µs | 774µs | 2,230 ms
cone GeneratePath #0 | 7,715 | 85µs | 647µs | 657 ms
cone GeneratePath #1 | 7,715 | 82µs | 1.2ms | 635 ms
cone-fallback TOTAL | 7,715 | 184µs | 1.3ms | 1,422 ms
StartTravelPlan | 1,981 | 130µs | 1.1ms | 259 ms
UpdateTravelPlan | 25,991 | 46µs | 0.8ms | 1,192 ms


# GetFullPath sub-steps (avg / max / total)
Step | Avg | Max | Total
-- | -- | -- | --
shared_lock acquire | 0µs | 4µs | ~0 ms
startNode lookup | 2.4µs | 51µs | 4.7 ms
GetZoneId(dest) | 3.5µs | 29µs | 6.8 ms
endNode lookup | 0.5µs | 23µs | 1.1 ms
hasRouteTo | 0.5µs | 21µs | 1.0 ms
GetNodeRoute (A*) | 25µs | 300µs | 49 ms
BuildPath | 28µs | 154µs | 54 ms
GetFullPath EXIT total | 107µs | 1.1ms | 207 ms

# ExecuteTravelPlan
EXIT total: 40µs avg / 784µs max / 1,049 ms cumulativeover 26K calls
CheckSplineProgress: &lt;1µs avg / 157µs max — essentially free

hermensbas:
Could you explain this travelnode system abit more, how does vary against def pathfinding navmesh? I kinda expected some custom waypoints as in the db, but thats doesnt seem to be the case?

Celandriel:
>Could you explain this travelnode system abit more, how does vary against def pathfinding navmesh? I kinda expected some custom waypoints as in the db, but thats doesnt seem to be the case?

Sure.
So the system is using all the information that is defined in the database — specifically three tables: playerbots_travelnode, playerbots_travelnode_link, and playerbots_travelnode_path.

- Nodes: Points of interest, flight masters, portals, elevators, transports, inns, world bosses, subzone centers, etc.
- Links: What is connected to what, how long it takes to get there, are you swimming, max mob level along the way, and the link type (walk, portal, transport, flightPath, teleportSpell, staticPortal, flyingMount(Not implemented yet, but added for future).
- Paths: The exact series of waypoints used to connect between the linked nodes. 

So then the bot asks, I need to go from point A -> Z
Then the system will do the following.
1. Find nearest node from start to finish. This is indexed by zone to make it faster, fallback to map wide if it fails. 
2. Check if they are connected in our map. (I used the same BFS approach for this, and it helps avoid instances of trying to search for a path between two unconnected nodes that takes forever)
3. Determine the path from A -> Z which will travel through all the nodes B -> Y. This I capped to 500 nodes. The system also does a 'cost' factor, like swimming etc. All stuff that was already built in. 
4. Then I expect that our destination is not near a node, we then use the pathfinding to start/finish the travel .

Then I hooked this all into the rpg strategy, where the path gets stored and processed depending on the path type that we encounter. If for some reason a part of it fails, we cheat and teleport there. 

So, using the nodes (And actually i have a follow up PR that will add new nodes to close some notable travel gaps) we eliminate the need to do pathfinding across long distances (Long > 296f) and can enable bots to go further. 

One notable case is Darnassus. Bots always get stuck there because the distance to the flight master is > 500f and without this they dont really know how to use the portal. This will solve that.

Celandriel:
image
BTW, this image comes from this pathfinding system. Bots are able to fully populate the world.

hermensbas:
interesting, will try and test, this weekend.

hermensbas:
> image BTW, this image comes from this pathfinding system. Bots are able to fully populate the world.

I kinda expect this is mostly due the teleport "hack"?

Celandriel:
Actually not so much, There are just a handful of areas that actually end up requiring teleports. 

Ive had Claude summarize the logs for me as they run, 

```

### Unreachable Destinations

#### Destination: map=1 (6774,-4677,724) — 190 failures
- Location: Winterspring/Hyjal area on Kalimdor
- Reason: A* returns empty (500 node cap reached or no connected route)
- Source bots: scattered across Kalimdor (Desolace, Feralas, Azshara, Stonetalon, Thousand Needles, Dustwallow)
- Root cause: destination node is likely in a disconnected component or very far from A* start with high node count between them

#### Destination: map=530 (9683,-7521,19) / (9694,-7526,19) etc — Silvermoon City area — 15 failures
- Location: Silvermoon City on map 530 (Outland map includes Blood Elf starting area)
- Reason: "no route and no flight" or "route through instance"
- Root cause: route from Outland mainland to Silvermoon passes through instance portals or lacks direct node connectivity

#### Destination: map=530 (-4021,-11738,-151) etc — Exodar area — 21 failures
- Location: The Exodar on map 530
- Reason: "no route and no flight" or "route through instance"
- Root cause: similar to Silvermoon — Draenei starting area on Outland map, disconnected from mainland Outland nodes

#### Destination: map=571 (5928,729,644) / (5630,693,655) — Dalaran area — 12 failures
- Location: Dalaran in Northrend
- Reason: "no route and no flight" or "no map connection"
- Root cause: map 530→571 connection not found by FindMapConnection
```

hermensbas:
I am still running some tests in starting lvl shadowglen and use self-bot. U will see a lot of teleporting going simply in situation grinding quest mob, quest done, and teleport to npc instead of walking.

How does this teleport work? whats the indicator when its failing? based not reaching target in certain timeframe? 

Meanwhile lemme test abit more.

Celandriel:
So I added a crap ton of logging locally, and from what I can see the reason bots is because all of are path generation calls set the forceDestination to True, this in some areas this causes the pathfinding to fail and after a certain number of fails it will teleport. 

The bots still do a dumb dance when they are trying to do the quest, because once they are at the target location the do a moverandomnear, but that is a different issue than a pathfinding one.

hermensbas:
True but there is some regression now in the most core features of the bots. lets talk our with this. This is solvable and lets ensure we have no regression within the zone movements.

hermensbas:
Added for in the world debugging:
Can only be executed as GM2+ just like the BG command.

.playerbots debug zone showpath=all
.playerbots debug zone showpath=node
.playerbots debug zone showpath=path

image

Celandriel:
Closing. Will introduce with a different PR with more travel changes.

PR #2318, opened 2026-04-20
Fix Deep Breath issues during Onyxia encounter

HennyWilly:
## Pull Request Description

The current strategy for Onyxia causes bots to get hit by her breath attack relatively consistently during phase 2. 
The problem was that the safe zone coordinates always use the bot's z-coordinate. If the bots are standing at the lower altitude part of the arena, `SearchForBestPath` inside `MoveTo` causes `INVALID_HEIGHT`, resulting in the bot not moving at all and getting hit by the breath attack.

This PR fixes this behavior by using the actual terrain z-coordinates for the predefined safe zones instead of always using the bot's z-coordinate. These values are chosen to ensure valid pathfinding regardless of the bot's current elevation.
Additionally, bots now interrupt their spells if they are not inside a safe zone during the breath. This causes the bots to immediately start running instead of finishing their casts first.

## Feature Evaluation

Replaced the use of `bot->GetPositionZ()` in `GetSafeZonesForBreath` with predefined safe zone z-coordinates to ensure valid pathfinding.
Added `AttackStop` and `InterruptNonMeleeSpells` to guarantee immediate movement when outside safe zones.
No additional condition checks or branching logic were introduced.

Minimal. The logic only runs within the Onyxia encounter script and calling `AttackStop` and `InterruptNonMeleeSpells` should be negligible.

## How to Test the Changes

Enter Onyxia's Lair (10, 25 or 40 (mod-individual-progression)) and engage Onyxia with the appropriate number of bots.
During phase 2 (Onyxia takes off), check if the bots move to the safe zones during the breath attack.
Tip: Mark Onyxia as moon (RTI), so that phase 2 doesn't end too quickly.

## Impact Assessment

    - - [ ] No, not at all

The calls of `AttackStop` and `InterruptNonMeleeSpells` cause minimal overhead compared to the original strategy. This should be negligible.

    - - [ ] No
    - - [x] Yes (**explain why**)

Yes (encounter-specific). Bots will now interrupt casts earlier during Onyxia phase 2 to prioritize movement to safe zones.

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

## Notes for Reviewers

The strategy for Onyxia might need additional work:
For example, the Onyxian Lair Guards are completely ignored while whelps are alive and their Blast Nova doesn't get handled at all.
This PR focuses on fixing the Deep Breath behavior. Handling of Onyxian Lair Guards is not included and should be implemented in a separate PR.

### DIFF
--- a/src/Ai/Raid/Onyxia/Action/RaidOnyxiaActions.cpp
+++ b/src/Ai/Raid/Onyxia/Action/RaidOnyxiaActions.cpp
@@ -99,8 +99,12 @@ bool RaidOnyxiaMoveToSafeZoneAction::Execute(Event /*event*/)
     if (bot->IsWithinDist2d(bestZone->pos.GetPositionX(), bestZone->pos.GetPositionY(), bestZone->radius))
         return false;  // Already safe
 
+    // Stop current spell first
+    bot->AttackStop();
+    bot->InterruptNonMeleeSpells(false);
+
     // bot->Yell("Moving to Safe Zone!", LANG_UNIVERSAL);
-    return MoveTo(bot->GetMapId(), bestZone->pos.GetPositionX(), bestZone->pos.GetPositionY(), bot->GetPositionZ(),
+    return MoveTo(bot->GetMapId(), bestZone->pos.GetPositionX(), bestZone->pos.GetPositionY(), bestZone->pos.GetPositionZ(),
                   false, false, false, false, MovementPriority::MOVEMENT_COMBAT);
 }
 
--- a/src/Ai/Raid/Onyxia/Action/RaidOnyxiaActions.h
+++ b/src/Ai/Raid/Onyxia/Action/RaidOnyxiaActions.h
@@ -2,7 +2,6 @@
 #ifndef _PLAYERBOT_RAIDONYXIAACTIONS_H_
 #define _PLAYERBOT_RAIDONYXIAACTIONS_H_
 
-#include "Action.h"
 #include "AttackAction.h"
 #include "GenericSpellActions.h"
 #include "MovementActions.h"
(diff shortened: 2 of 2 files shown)


brighton-chi:
Interesting. I was thinking failed movement calls was the most likely cause. Switching to the target's z position is totally reliable in this case? In my experience, the most effective method is to use the bot's z and make movement incremental. Clamping to GetMapWaterOrGroundLevel is even better and I needed that to make Hyjal workable. But if this already works, then no need to try anything else.

Re: AttackStop(), are you finding that that helps? I haven't noticed any difference in speed of avoidance--I thought that call was basically like whether you wanted to disengage the bots before moving (basically like /stopattack for a player) and that it doesn't really matter much.

Are you sure InterruptNonMeleeSpells works? UpdateAI returns early if you have a pending spell casting, with exceptions only for if the target unit dies, target gameobject despawns, or target for healing is at full HP (which is actually a pretty bad exception that maybe should be removed). I haven't been able to get spell interrupts called from actions to actually do anything, hence the PR on test-staging that adds a hook into UpdateAI (which does work very well for me). I also was never able to get in-progress channels to cancel, though I actually don't know why that didn't work because UpdateAI didn't return early for channels that were already cast (but in any case, RequestSpellInterrupt() from the PR does work for cancelling pending channels also).

HennyWilly:
I don't know if switching to the target's z position is **totally reliable**, but after running phase 2 of the encounter for many minutes, no bot got hit by the breath attack.
I remember that I used the incremental movement approach too on Golemagg in Molten Core. In that case, it was necessary because of the relatively steep slope and there were no fixed positions for the tank bots to move to. But if the target positions are fixed, it works fine to set the z position and moving in one go.

I found that `AttackStop` and `InterruptNonMeleeSpells` on their own didn't really work that great. When used together, the bots actually start running almost immediately. But to be honest, I don't know, if there are more efficient ways.
In case of `InterruptNonMeleeSpells`, I noticed that at least hunters keep channeling their `Volley` attack, if `InterruptNonMeleeSpells` is **not** called.
I wasn't aware of `RequestSpellInterrupt`. I'll try that one when I have time.

brighton-chi:
Even with a fixed endpoint, I think incremental movement can have value. I think of it as a sort of a crude linear pathfinding approximation, and you're more likely to succeed with shorter local moves when the terrain is uneven. It shortens the movement lock window. I know it's mainly useful to allow bots to reconsider conditions during a move, but it helps with rough terrain quite a lot too. This doesn't mean it's needed for Onyxia, I just don't want to diminish the value of the technique.

There isn't any reason I am aware of why you shouldn't be able to interrupt an active channel (like volley) with a separate action that calls a spell interrupt. For whatever reason, I've had trouble with interrupting Arcane Missiles and Mind Flay in particular with that approach, maybe those spells have different mechanics in some way. What you should not be able to do with an outside action is interrupt a pending cast because UpdateAI will ignore everything you attempt to do while a cast is pending, no matter the priority. That's probably fine enough for Onyxia as I don't think the timing is that unforgiving. But there's other stuff like for a couple of the Sunwell bosses where you have probably like 2 seconds to run before you wipe the raid, and then it becomes more relevant.

HennyWilly:
@brighton-chi If it's okay with you, I would keep the movement as is for now. While incremental movement has its merits, I don't think it's necessary in this case. After extensive testing, there were no issues when bots moved the full distance at once.

I also tried `RequestSpellInterrupt` as you suggested, and it works very well. I've replaced `InterruptNonMeleeSpells` with it and removed `AttackStop`, as it didn't seem to have any noticeable effect.
That said, it feels a bit strange to be the only one to use `RequestSpellInterrupt`...

brighton-chi:
It shouldn't work for a pending cast in the middle of an action. The thing is once you are casting, UpdateAI returns early and will reject any other action so you need a way to interrupt the cast outside of the action system. RequestSpellInterrupt() just sets a flag inside UpdateAI that when activated will allow interruption of the pending cast. I've used a spell listener script so basically when the cast is detected, RequestSpellInterrupt() is called outside of the action system. I use it in the Hyjal PR.

It can work the way you have it to interrupt an active channel like volley, but I think InterruptSpell() should work for that anyway.

HennyWilly:
Thanks for taking the time to explain this, I appreciate it.
I understand what you are saying. I looked into your Hyjal PR and I think an AllCreatureScript may be overkill for this case.
I reverted to the `AttackStop` and `InterruptNonMeleeSpells` combination, mainly because other strategies use the same approach (e.g. Gruul's Lair, Karazhan, and Tempest Keep).

brighton-chi:
> Thanks for taking the time to explain this, I appreciate it. I understand what you are saying. I looked into your Hyjal PR and I think an AllCreatureScript may be overkill for this case. I reverted to the `AttackStop` and `InterruptNonMeleeSpells` combination, mainly because other strategies use the same approach (e.g. Gruul's Lair, Karazhan, and Tempest Keep).

Yeah, I agree it's not needed here. I brought it up originally not to push for the use but just to note that calling the mid-action interrupt doesn't actually stop any pending casts. There's plenty of time to finish a Shadow Bolt or whatever and then run for Onyxia.

As far as Karazhan and Gruul, I didn't realize the issue for pending spell casts until @flashtate98 and I couldn't figure out why bots would finish their casts before moving for his Shirrak strategy, notwithstanding calling InterruptSpell() or even Reset(). I had no idea back when I was working on Karazhan through TK; I was following a pattern that goes back to all the old WotLK strategies from before my time. It became very obvious the method didn't work with Shirrak because he pulses an aura that slows down spell casts and then his main attack is an almost instant nuke on a random targets position that kills you in seconds.

I've kicked around the idea of using AC hooks as part of boss strategies for a while, but it was @trolloks who actually did the original implementation for Void Reaver's Arcane Orbs because the mechanics of the spell are such that it is impossible for bots to detect it before impact by any traditional means. It's a really powerful tool--for example, it's the only way I've found to preemptively avoid dynamic object AoEs (AC doesn't expose any public method that I have been able to find to get dynamic objects so bots normally can observe them only after having an aura applied, and they immediately "forget" the object as soon as the aura falls off). I think it's indispensable for the more complicated raids if you do decide to work on more (I'm using the method constantly in Sunwell, and I suspect Eredar Twins are unkillable with IP's TBC-tuning without it). But most of the time I agree it is overkill.

brighton-chi:
> @brighton-chi If it's okay with you, I would keep the movement as is for now. While incremental movement has its merits, I don't think it's necessary in this case. After extensive testing, there were no issues when bots moved the full distance at once.
> 
>

Heh, of course, you don't have to listen to a single thing I suggest, and I wasn't necessarily advocating for it anyway. I just still don't have a great grasp on how to best execute movement sometimes and what may work better between different approaches so moreso was just curious that a move from bot z to destination z here made all the difference, given that the strategy was utterly broken before. Onyxia worked _most_ of the time when I was fighting her way back when that strategy was first implemented, but I have heard it practically never works now.

Celandriel (approved):

PR #2326, opened 2026-04-24
clean up for DropQuestAction

Celandriel:
## Pull Request Description

I was getting annoyed by the constant "No event owner detected" message after I enabled bot debugging messaging. 
And then I figured that there is no reason that maintenance should trigger this action every 5 seconds. So I swapped it to seldom, so it runs every 5 mins. More Performance!

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist


hermensbas (dismissed):

hermensbas:
That 5 second could be mitigation, should be tested well though. Seems like an ok change.

PR #2333, opened 2026-04-25
(Reapply) Feat: Reintroduce timed logouts

NoxMax:
## Pull Request Description

Reapplying the reverted #2289. Now with a fixed patch for possible crashes:
1. Deleted the last `delete target` reference in PlayerbotMgr as a double delete was happening with `delete botAI` right after it.
2. Changed 0 to 1 in `session->SetLogoutStartTime(1);`, as it would possibly cause a problem with the > 0 evaluation in WorldSession.h

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

    - - [ ] No

## AI Assistance

- - [x] No

## Final Checklist

## Notes for Reviewers

No other functional changes were made beyond the original commit.

### DIFF
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -244,9 +244,21 @@ void PlayerbotAI::UpdateAI(uint32 elapsed, bool minimal)
         nextAICheckDelay = 0;
 
     // Early return if bot is in invalid state
-    if (!bot || !bot->GetSession() || !bot->IsInWorld() || bot->IsBeingTeleported() ||
-        bot->GetSession()->isLogingOut() || bot->IsDuringRemoveFromWorld())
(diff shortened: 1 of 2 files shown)


hermensbas:
Still against this overload of a feature for such a small use case.

NoxMax:
I don't know about calling it an overload. I mean anything remotely measurable resource-wise won't run outside actually logging out. And as far as code complexity, there isn't really anything far-out here that can confuse future devs.

So really I don't see it being expensive at all, even in the context of something small like not wanting bots to go out like this right away no matter what:

Celandriel:
Sorry about the delay here. I intended to run this locally but have had negative time to do anything specific lately.

NoxMax:
No worries. Honestly someone else should test it, the travel system is huge enough on your plate already.

PR #2336, opened 2026-04-25
Boomkin Update - Typhoon, Cyclone, fixed Eclipse Proc and DoTs

ThePenguinMan96:
## Pull Request Description

Hello everyone! I wanted to get back into working on strategies for classes, and a few people recommended the druid needs work. After building a 25m raid team, I noticed that both cat and boomkin weren't doing great. I decided to work on the boomkin first, as it has a special place in my heart. Below are a list of changes. **I used the AI to generate a list of changes in bold,** _with my comments below them in italics._

**Strategy Rename: CasterDruid → BalanceDruid
Renamed all CasterDruid strategy classes, files, and references to BalanceDruid to match the correct specialization name used everywhere else in the project. This includes the strategy class itself, the action node factory, the AOE sub-strategy, the AI factory registration, and all associated headers. No behavior was changed by the rename — it is a pure refactor for consistency.**

_I am trying to slowly make all strategies uniform so they adhere to their spec's name. The in-game strategy changed from "caster" and "caster aoe" to "balance" and "aoe", as seen in the aifactory changes. When this commit goes live, people will need to reset strategies for their boomkin alts and randombots._

**New: Cyclone (Crowd Control)
Cyclone is newly implemented for Balance Druid. CastCycloneCcAction is registered in the AI object context with a proper CC target value binding, and the Balance strategy registers a cyclone on cc action node with moonkin form as a prerequisite so the bot shifts if needed before casting. The trigger uses the existing HasCcTargetTrigger framework.**

_This follows the hibernate and entangling roots code, just implements cyclone as an option as well._

**New: Typhoon (Defensive Knockback)
Typhoon is newly implemented for Balance Druid within the AOE strategy. Rather than treating it as a pure AoE damage spell, it is triggered by enemy within melee — firing when enemies close into melee range as a defensive knockback. A facing and range check in CastTyphoonAction::isUseful() ensures it only casts when the bot is facing the target and the target is within 15 yards.**

_Basically, it's nested in the aoe strategy, but only triggers if "enemy within melee". I originally tested it with the trigger "enemy too close for spell", but part of that trigger requires the target to not attack the boomkin. After I made the trigger switch, it worked. Keep in mind that the only spec that has typhoon is the boomkin pvp one! You will not see a pve boomkin use this._

**Eclipse Rotation
Implemented a proper Eclipse spell rotation using time_t timers stored directly on CastWrathAction and CastStarfireAction. When Lunar Eclipse procs (aura 48518), Wrath is suppressed so Starfire can fish for Solar Eclipse — and vice versa. The timer clears early if the other Eclipse procs (fishing succeeded) or after 30 seconds (cooldown window over). This replaces a broken approach that called HasSpellCooldown() on spell IDs that never actually enter the cooldown system. For druids without the Eclipse talent, the timers never fire and both spells remain fully available.**

**Starfire added as a default action at priority 5.2 (just above Wrath at 5.1), so it gets preference as the filler spell
Eclipse trigger nodes (eclipse (solar) / eclipse (lunar)) kept to boost priority to 20.0 when the proc is active**

_I was noticing that the boomkin wasn't tracking the eclipse cooldowns correctly. It would:
Spam wrath until lunar eclipse
Spam starfire until lunar eclipse is over
Spam wrath again? When it should be spamming starfire to get a solar eclipse proc._ 

_It was not adhering to the internal cooldown of 30 seconds, causing a pretty significant dps loss.
I looked into the coding behind it:_

relies on cooldown

_The previous checked if the spell had a cooldown. But upon further investigation:_

no cooldown

_This function was pointing to a non-existent cooldown. I originally tried to create a trigger with a timer, but I had no consistent luck. Then I pivoted and modified the isuseful() portion of wrath and starfire. After I made this change, the druids were respecting the eclipse cooldown! I used timing code consistent with that of spirit walk (enhancement shaman), and it worked quite well._

**UPDATE - 5/3/2026: Implemented a proper Eclipse spell rotation using time_t timers. When Lunar Eclipse procs (aura 48518), Wrath is suppressed so Starfire can fish for Solar Eclipse — and vice versa. The timer clears early if the other Eclipse procs (fishing succeeded) or after 30 seconds (cooldown window over). Timer state is stored in file-scope static maps keyed by bot GUID rather than instance variables on the action objects — this ensures the state survives action object recreation (bot death, strategy reset, zone change) and that each Eclipse phase is recorded by whichever action is winning the priority queue at that moment. Previously, _solarProcTime was never written during Solar Eclipse because CastStarfireAction::isUseful() was never evaluated while the Eclipse trigger was winning — the static map approach fixes this by having CastWrathAction (which IS called during Solar Eclipse) update the shared Solar timer, and vice versa for Lunar. For druids without the Eclipse talent, the timers never fire and both spells remain fully available.**

**Starfire added as a default action at priority 5.4 (just above Wrath at 5.3), so it gets preference as the filler spell.**

_I went back to test the eclipse proc, but it was inconsistent - again - it would work fine for the first few eclipses, but something wasn't tracking right. After the changes, it is tracking accurately._

**DoT Application Fixes
CastMoonfireAction, CastInsectSwarmAction, CastMoonfireOnAttackerAction, CastInsectSwarmOnAttackerAction, and CastFaerieFireAction all had their isUseful() overridden to call CastAuraSpellAction::isUseful() instead of the default CastDebuffSpellAction path. This bypasses the time-to-live (TTL) check that was silently blocking DoT application on targets the bot considered low-health or short-lived.**

**FaerieFireTrigger was also updated to match the InsectSwarmTrigger pattern — it now calls BuffTrigger::IsActive() (debuff missing check) rather than DebuffTrigger::IsActive() (which includes its own TTL gate), plus requires bot->HasAura(33602) so it only fires when the bot has Improved Faerie Fire talented.**

_These changes were similar to the warlock's ttl (time to life) check. I believe there are calculations that are wrong with the estimated group dps and time to life checks, so I had the DoTs + faerie fire bypass these checks. Now balance druids will actually cast insect swarm and moonfire at lower levels, in instances, against dummy targets, etc. I also added logic in faerie fire to not cast it until the 3% spell hit buff talent was obtained. It felt quite useless at level 15 using faerie fire as part of their rotation._ 

**UPDATE 5-3-2026: FaerieFireTrigger was updated to call BuffTrigger::IsActive() (debuff missing check) rather than DebuffTrigger::IsActive() (which includes its own TTL gate). The talent aura check (33602) has been replaced with a world/raid boss rank check (CREATURE_ELITE_WORLDBOSS) — Faerie Fire now only fires on raid bosses and world bosses. Since all of these targets are level 60+, the druid will always have the talent at that point, and 3% spell hit actually matters on these targets.**

_In testing, the boomkin using faerie fire on every single mob was actually bad. Now it only is used on raid bosses/world bosses. _

**Eclipse Cooldown Trigger Removal
Removed EclipseSolarCooldownTrigger and EclipseLunarCooldownTrigger entirely — both from the header, the .cpp, and the DruidAiObjectContext factory. The cooldown logic now lives entirely inside the action isUseful() methods where it can actually work.**

_With the modification of isuseful tracking the eclipse timers, these were removed._

**AoE Strategy (DruidAoeStrategy)
Added a Balance Druid-specific AoE strategy block inside DruidAoeStrategy::InitTriggers in GenericDruidStrategy.cpp. The strategy is spec-gated — it only registers triggers when the bot is in the Balance tab — keeping it clean for Feral and Restoration specs.
The AoE priority ladder, from highest to lowest:
Typhoon (40.0) — defensive knockback on enemy within melee (see above)
Starfall (28.5) — on cooldown-ready trigger, with a CC-safety check to avoid breaking crowd control
Hurricane (23.0) — sustained channel on medium aoe (3+ enemies), with a hurricane channel check trigger that cancels the channel if the number of enemies within the Hurricane's actual AoE radius (read from the spell's DynamicObject) drops below threshold mid-cast — rather than checking global attacker count, which was imprecise in multi-bot scenarios.

Insect Swarm on Attacker / Moonfire on Attacker (19.5 / 19.0) — spread DoTs onto additional attackers on light aoe (2+ enemies)**

_I implemented an aoe strategy in genericdruidstrategy, in an attempt to make things more uniform across the board - eventually all class strategies should have their aoe spells controlled by "co +/- aoe". Now, boomkins will only cast starfall if it's off of cooldown and the aoe strategy is added._

**UPDATE 5-3-2026: Insect Swarm on Attacker / Moonfire on Attacker (5.2 / 5.1) — spread DoTs onto additional attackers via dedicated on-attacker triggers, sitting just below Wrath and Starfire so single-target filler always takes precedence**

_After testing this at lower levels, before hurricane and starfall, I realized that multi-dotting with these DoTs isn't worth it, and is actually really bad. The boomkin would just go oom to do suboptimal dps. Nothing would live the entire duration of the DoTs. I reduced the priority of the on-attacker DoTs to lower than the default actions of starfire and wrath (5.4 and 5.3). This way, they will still fire while moving - but if stationary, they will just use starfire/wrath._

**Boost Strategy (DruidBoostStrategy)
Updated DruidBoostStrategy::InitTriggers to be spec-aware.  Force of Nature (Treants) is now conditionally registered — it only appears in the boost strategy when the bot is specced Balance.** 

_Moved force of nature from balance to boost. It was originally using it as part of it's default actions._

## Feature Evaluation

## How to Test the Changes

Make a druid, turn them into balance pve after level 10, and see the changes for yourself. 

## Impact Assessment

    - - [x] No, not at all

It functions similar to how boomkins did before, with slight changes to the triggers and actions. 

    - - [ ] No
    - - [x] Yes (**explain why**)

Rotation is slightly more accurate to what guides and players consider optimal.

    - - [x] No

## AI Assistance

- - [ ] No

AI (Claude Code with Sonnet 4,6 high) was used to review and write most of the code, and gemini reviewed it. Claude also wrote the descriptions in the PR. 

## Final Checklist

## Notes for Reviewers

I saw a noticeable increase in boomkin raid dps. I tested these changes at level 15, 30, 50, 70, and 80, in rdf/raid group settings.
Changing the filler from wrath to starfire was huge at lower levels - wrath is so incredibly mana inefficient compared to the king that is starfire. balance druids shouldn't have nearly as many mana problems at early levels with this switch.

**UPDATE 5-3-2026:**

Here are some dps tests. All 5 minutes in duration, against a single-target boss, with varying gear levels.

Let's start with heroic gear (blue ilevel 200):

**Master Branch:**
Heroic Gear Main Branch 5 min

**Boomkin Update:**
Heroic Gear Boomkin Update 5 min

And Naxx Gear:

**Master Branch:**
NAXX Gear Main Branch 5 min

**Boomkin Update:**
Naxx Gear Boomkin Update 5 min

Ulduar Gear:

**Master Branch:**
ULD Gear Main Branch 5 min

**Boomkin Update:**
ULD Gear Boomkin Update 5 min

TOC Gear:

**Master Branch:**
TOC Gear Main Branch 5 min

**Boomkin Update:**
TOC Gear Boomkin Update 5 min

ICC Gear:
**Master Branch:**
ICC Gear Main Branch 5 min

**Boomkin Update:**
ICC Gear Boomkin Update 5 min

**Summary: DoT Uptime is lower, because eclipse uptime is higher, resulting in higher dps across all gear levels.**


kadeshar:
Github have issues and cant review right now:
- PR doesnt compile right now
- consider not using magic numbers
- CastFaerieFireAction now use isOwner parameter. That debuff can overlap?

ThePenguinMan96:
This boomkin PR needs a bit more work - it was originally branched off of master, and requested to be merged into test staging - my github is a disaster right now. I'm going to work on those 3 things you recommended, and reopen a new PR later down the line as part of a druid update.

PR #2358, opened 2026-05-03
Overhaul party buff/greater blessing system

brighton-chi:
## Pull Request Description

These changes I originally made for myself because as a person who really likes to raid with bots, I felt like the current group buff system is fundamentally broken, and I needed something more consistent and optimal. I debated a lot whether to PR this because it's such an extensive overhaul that was almost entirely reliant on AI, and I know that wishmaster still has a PR open regarding the greater blessings. I decided to after a couple of conversations so at least people can look at it and see if it's something that they want.

The tl;dr version is that this PR overhauls buff handling in two related areas:
1. It adds a dedicated greater blessing assignment system.
2. It generalizes party/raid reagent-buff handling for Paladins, Druids, Mages, and Priests.

Under this PR, greater blessings are determined by assignments for the current group, and those assignments are determined based on:

1. a hardcoded priority list of blessings for each spec;
2. the number of Paladins in the group; and 
3. whether any Paladins have talents for Blessing of Sanctuary, Improved Blessing of Might, or Improved Blessing of Wisdom.

Assignment determinations are cached in a value to avoid constant reevaluation. 

The exact priority list is:

- All casters: Kings, Wisdom, Sanctuary, Might
- Physical-only DPS (Rogues, Warriors, DKs): Might, Kings, Sanctuary, N/A
- Hybrid DPS (Enh, Ret, Hunters, Cats): Might, Kings, Wisdom, Sanctuary
- Druid tanks: Kings, Might, Sanctuary, Wisdom
- Warrior and DK tanks: Kings, Might, Sanctuary, N/A
- Paladin tank: Sanctuary, Might, Wisdom, Kings

Note that Sanctuary is preferred over Kings for Paladin tanks because of the mana regen component but deprioritized for other tanks because Kings provides Agility. The extra 3% damage reduction from Sanctuary does not stack with Disc Priests’ Renewed Hope, which will have 100% uptime.

For group buffs, logic is centralized so that class triggers use the same gating and upgrade rules for Gift of the Wild, Arcane Brilliance, Prayer of Fortitude, Prayer of Spirit, and Prayer of Shadow Protection. Also, Shadow Protection is now a default strategy for Priests (rshadow, which existed before but wasn’t added by default).

I’ve added a config setting for the greater blessing system and adjusted the current config setting for group buffs. In each case, you can pick whether to disable the feature entirely, use it in all groups, or use it only in raid groups. The default is raid only for greater blessings and all groups for group buffs. Note that for group buffs, even if the config is enabled, they will be used only if at least 3 group/raid members on the same map are missing the buff family. This is mainly to stop group buff spamming during wipe recovery as bots are revived one-by-one.

I renamed the Paladin buff strategies to align them with the actual blessing names:
- `bhealth` -> `bsanc`
- `bmana` -> `bwisdom`
- `bdps` -> `bmight`
- `bstats` -> `bkings`

This is an intentional breaking change for saved strategy strings. Bots will need a one-time strategy reset after update.

I removed bots telling you when they are out of reagents for greater blessings. If people like that though, I can add it back.

A small cleanup is also included in TankPaladinStrategy: Holy Shield was subject to three overlapping health triggers with the same priority; I removed the two lower health thresholds which have no purpose.

## Feature Evaluation

I’m going to let the AI answer this one.
> The minimum logic is:
> - a shared config-gated check for whether group/raid buff variants are allowed
> - a shared way to treat single and group variants as equivalent aura families
> - a shared upgrade path from single-target buff to group buff when the group variant is appropriate
> - a Paladin-only cached assignment model that decides which blessing family each Paladin should cover for the current group
> - trigger/action wiring that only attempts casts when a group member is actually missing the assigned buff
>
> This avoids scattering separate per-class heuristics across many triggers and actions.

Processing cost should be minimal but non-zero. The general party buff changes are limited to existing buff trigger paths and mostly replace duplicated checks with shared helpers. They do not add expensive default per-tick behavior outside those existing trigger evaluations.

The Paladin greater blessing logic does add extra decision-making, but it is limited to Paladins, gated by config and group eligibility, subject to a delayed trigger evaluation of only once per 4s, and cached per group assignment set instead of recomputing the full assignment model on every action attempt.

This PR also increases the throttle duration for group buff triggers to limit performance impact; I’m open to adjustments to these durations:
- Mark of the Wild triggers were increased from 4s to 8s
- Arcane Intellect triggers were increased from 4s to 8s
- Priest buff triggers were increased to 8s (previously, Fortitude was 6s, Spirit was 4s, and Shadow Protection had no throttle)
- There is now a 5s delay on buffing (greater blessings and group buffs) after bots log in—I was getting bots spamming buffs as soon as they logged in even when it was not necessary

I’ve tested with pmon, and the impact is minimal—these are very cheap triggers even compared to standard bot rotational ability triggers.

## How to Test the Changes

1. Try different config settings to confirm that they work to enable/disable greater blessings/group buffs in the configured scenarios

2. For greater blessing changes:
   - test with one Paladin in a party/raid
   - test with multiple Paladins in a party/raid
   - confirm the Paladins divide blessing coverage instead of repeatedly overwriting each other
   - include at least one Paladin with Improved Blessing of Might and make sure it casts Might over Paladins without the talent; check the same with a Paladin with Improved Blessing of Wisdom
   - do not include a Paladin that knows Sanctuary, confirm any Paladin tank receives Kings instead (you’ll need a low-level Paladin for this since Sanctuary is a prot talent)
   - confirm bots cast blessings only when a member is actually missing the relevant blessing family
   - confirm there is a 5s delay on buffing when bots log in

3. For group buff changes:
   - confirm there is a 5s delay on buffing when bots log in
   - confirm that single buffs are used when there aren’t at least three unbuffed members in the same map, even if group buffs are enabled in the config

4. For all buffs, test with reagents missing to confirm fallback to single-target buffs and single blessings

5. Confirm the Paladin buff strategy names are changed after resetting AI

## Impact Assessment

    - - [ ] No, not at all

Discussed above in processing costs.

    - - [ ] No
    - - [x] Yes (**explain why**)

Yes—that is the purpose of this PR, to change default buffing behavior.

    - - [ ] No

Yes, but I think it’s inevitable to add complexity to get greater blessings to function consistently, given the challenges brought by their mechanic of applying across each class.

## AI Assistance

- - [ ] No

I used GPT-5.4 extensively for this overhaul. It’s much more complicated than I could handle on my own. I’ve done a lot of testing and have reviewed the code and provided plenty of revisions, but I cannot say I can perfectly explain each addition and how it works, not even close.

## Final Checklist


ThePenguinMan96:
@brighton-chi 
@kadeshar 

After reviewing this for the last couple of days, I am very happy with it. This makes paladin buffing excellent, and it was quite responsive. I did notice a couple of bugs/errors perhaps, and I will explain. Below is a list of screenshots of my experience in testing. I am testing with a group of 4 paladins, so as they level up they should get all of the greater blessings.

**Level 10 looks good - only knows might.**

test level 10

**Level 20 - 3 blessings! So nice to see that... It was bugging me that I would run multiple paladins, and not have max blessings. Even warrior doesn't get wisdom, as the mod intends:**

test level 20 1
test level 20 2

**Level 30 - about the same, 3 blessings**

test level 30

**Level 40 - 4 blessings, sanc learned, and improved talents got priority on the blessings they should have:**

test level 40
test level 40 3

**Level 60, I ran into some issues. They continuously spammed greater blessings over and over, even though the bots had the correct blessings. After they ran out of reagents, they were fine. This could be because they are all the same class, and they might have different priorities (caster vs hybrid dps vs prot pally tank) and greater blessings apply groupwide to all classes of the same type. I was curious because they also learn greater blessing of kings at 60, so maybe I didn't have that learned yet.**

test level 60
test level 60 2
test level 60 3

**Level 70 - leveled them up, did /maintenance, and things were fine again.**

test level 70
test level 70 2

**Level 80 - I wanted to do the raid testing now. Super happy with the results:**

test level 80
test level 80 2
test level 80 3
test level 80 4

**Great. Now to the archetype testing. **

**Prot paladin:**
Prot pally 1 buff
Prot pally 2 buff
Prot pally 3 buff
Prot pally 4 buff

**Druid Tank:**

druid tank 1 buff
druid tank 2 buff
druid tank 3 buff
druid tank 4 buff

**Warrior Tank:**

warrior tank 1 buff kings
warrior tank 2 buff
warrior tank 3 buff
warrior tank 4 buff

**Phys only dps:**

phys dps 1 buff
phys dps 2 buff
phys dps 3 buff
phys dps 4 buff

**Hybrid dps:**

hybrid dps 1 buff kings
hybrid dps 2 buff
hybrid dps 3 buff
hybrid dps 4 buff

**Caster:**

(Sorry, 1 buff image got ruined, but there weren't issues at 1 buff)

caster 2 buff
caster 3 buff
caster 4 buff

So, overall, archetypes look good. The only minor thing I saw was that the druid tank should have had wisdom as the 3rd buff since the prot paladin wasn't present, and the 3rd buff should have been sanc, but that is REALLY splitting hairs. Looks miles better than what we have now. 

Okay, now for the code review. I read through everything you did, and it looks like this replaces entirely the old system that was being worked on. I am fine with that. I would not comment it out, but remove it, but that's just me. I know that this PR makes buffs work, so I don't see why we would need to preserve the old stuff, unless it's almost done.

I did have some concerns about a few things. I am not very good at typing about it, so I used AI to convey my message about modifying the base actions instead of making intermediate actions. I don't think we should be modifying CastBuffSpellAction, but rather making an intermediate action based on it (one that inherits it as it's base). Otherwise you're pushing that logic on every buff action in the game, regardless of if it's a greater blessing or group buff. This is a more technical explanation, written by claude code (looking at the local files) sonnet 4.6 (high)

## Overview

The core idea behind both systems is sound. The coordinated slot-based assignment for Paladins is meaningfully better than the old free-for-all approach. However, the implementation makes a structural mistake that the rest of the codebase consistently avoids: **it places class-specific logic inside generic base classes**, rather than in class-specific subclasses.

---

## Critical Architectural Issue: Modifying Generic Base Classes

### What the PR changes

The PR adds overrides for `isUseful()` and `Execute()` directly to `CastBuffSpellAction` in `GenericSpellActions.cpp/.h`:

```cpp
// PR adds these to CastBuffSpellAction — a generic base used by all classes
bool CastBuffSpellAction::isUseful()
{
    if (ai::buff::IsGroupVariantEnabled(bot, spell))
    {
        std::string const groupVariant = ai::buff::GroupVariantFor(spell);
        if (!groupVariant.empty() && botAI->HasAura(groupVariant, target, ...))
            return false;
    }
    // ...
}

bool CastBuffSpellAction::Execute(Event event)
{
    std::string const castName = ai::buff::UpgradeToGroupIfAppropriate(bot, botAI, spell);
    return botAI->CastSpell(castName, GetTarget());
}
```

It also modifies `BuffTrigger::IsActive()` in `GenericTriggers.cpp` to run a group-variant lookup on every check:

```cpp
// PR adds this to BuffTrigger::IsActive() — used by all classes
if (ai::buff::IsGroupVariantEnabled(bot, spell))
{
    std::string const groupVariant = ai::buff::GroupVariantFor(spell);
    if (!groupVariant.empty() && botAI->HasAura(groupVariant, target, ...))
        return false;
}
```

And adds a new `BuffOnPartyTrigger::IsActive()` override that iterates all group members on every tick:

```cpp
bool BuffOnPartyTrigger::IsActive()
{
    Unit* target = GetTarget();  // full value lookup
    if (ai::buff::ShouldDeferPartyBuffEvaluationForRecentLogin(bot, target, spell))
        return false;
    return BuffTrigger::IsActive();  // calls GetTarget() again internally
}
```

### Why this is a problem

`CastBuffSpellAction` is the **base class for every self-buff action across every class** in the module. This includes:

- `CastArcaneTorrentAction` — Blood Elf racial
- `CastBloodFuryAction` — Orc racial
- `CastBerserkingAction` — Troll racial
- `CastCrowdControlSpellAction` — CC across all classes
- `CastTotemAction` and all its subclasses — every Shaman totem
- `CastStrengthOfEarthTotemAction`, `CastStoneskinTotemAction`, `CastSearingTotemAction`, etc.
- `BuffOnMainTankAction` — main tank buffs

Every one of these now runs `IsGroupVariantEnabled` → `GroupVariantFor` on every `isUseful()` check, and `UpgradeToGroupIfAppropriate` on every cast. For spells with no group variant — which is the vast majority — `GroupVariantFor` walks its full comparison chain and returns `""` every time. This is Druid/Mage/Priest logic running inside Shaman totem placement, Blood Elf racial usage, and Orc cooldown management.

Similarly, `BuffTrigger::IsActive()` is used by **every buff trigger across every class**. The new group-variant check runs for Warrior battle shout triggers, Rogue poison triggers, Death Knight presence triggers — anything backed by `BuffTrigger`. The same is true of the `BuffOnPartyTrigger::IsActive()` addition: `ShouldDeferPartyBuffEvaluationForRecentLogin` iterates all group members on every tick, for every bot, for every party buff trigger regardless of class.

---

## How the Codebase Actually Handles This

Every class in the module adds custom behavior through **subclassing**, not by modifying generic base classes. The clearest example is the Shaman totem system.

### The Totem Pattern

```
CastSpellAction
  └── CastAuraSpellAction
        └── CastBuffSpellAction          ← generic, remains untouched
              └── CastTotemAction        ← Shaman-specific intermediate
                    ├── CastStrengthOfEarthTotemAction
                    ├── CastSearingTotemAction
                    ├── CastMagmaTotemAction   (further overrides isUseful() for melee range)
                    └── CastStoneclawTotemAction (further overrides isUseful() for solo-only)
```

`CastTotemAction` adds non-trivial Shaman-specific logic — totem slot checking, aura checking — without touching `CastBuffSpellAction`:

```cpp
// ShamanActions.cpp
bool CastTotemAction::isUseful()
{
    return CastBuffSpellAction::isUseful()
        && !AI_VALUE2(bool, "has totem", name)
        && !botAI->HasAura(buff, bot);
}
```

When `CastMagmaTotemAction` needs an additional melee-range constraint, it overrides `isUseful()` again in its own subclass — not in `CastTotemAction` or `CastBuffSpellAction`:

```cpp
bool CastMagmaTotemAction::isUseful()
{
    Unit* target = AI_VALUE(Unit*, "current target");
    if (!target || !bot->IsWithinMeleeRange(target))
        return false;
    return CastTotemAction::isUseful();
}
```

### The Party Buff Pattern

Druid, Mage, and Priest all follow the same clean split:

```cpp
// Self-buff: extends CastBuffSpellAction, no custom logic needed
class CastMarkOfTheWildAction : public CastBuffSpellAction
{
public:
    CastMarkOfTheWildAction(PlayerbotAI* botAI) : CastBuffSpellAction(botAI, "mark of the wild") {}
};

// Party buff: extends BuffOnPartyAction, no custom logic needed
class CastMarkOfTheWildOnPartyAction : public BuffOnPartyAction
{
public:
    CastMarkOfTheWildOnPartyAction(PlayerbotAI* botAI) : BuffOnPartyAction(botAI, "mark of the wild") {}
};
```

```cpp
// Mage — identical structure
class CastArcaneIntellectAction : public CastBuffSpellAction { ... };
class CastArcaneIntellectOnPartyAction : public BuffOnPartyAction { ... };
```

When custom logic is required elsewhere — Rejuvenation target prioritization, Focus Magic healer preference, Starfall CC-avoidance, Lifebloom stack checking — it is always overridden in the specific subclass. **No generic base class has ever been modified to accommodate a single class's needs.**

---

## Recommended Fix

Following the existing totem pattern, the group buff upgrade logic should live in an **intermediate class** between `BuffOnPartyAction` and the individual Druid/Mage/Priest party buff actions. The individual spell actions stay as one-liners, and `CastBuffSpellAction` remains untouched.

### Proposed: `CastGroupBuffOnPartyAction`

```cpp
// In GenericSpellActions.h — or a new GroupBuffActions.h

class CastGroupBuffOnPartyAction : public BuffOnPartyAction
{
public:
    CastGroupBuffOnPartyAction(PlayerbotAI* botAI, std::string const spell)
        : BuffOnPartyAction(botAI, spell) {}

    bool isUseful() override;
    bool Execute(Event event) override;
};
```

```cpp
// In GenericSpellActions.cpp

bool CastGroupBuffOnPartyAction::isUseful()
{
    Unit* target = GetTarget();
    if (!target)
        return false;

    // Suppress single-target cast if the group variant is already up on this target
    if (ai::buff::IsGroupVariantEnabled(bot, spell))
    {
        std::string const groupVariant = ai::buff::GroupVariantFor(spell);
        if (!groupVariant.empty() && botAI->HasAura(groupVariant, target))
            return false;
    }

    return BuffOnPartyAction::isUseful();
}

bool CastGroupBuffOnPartyAction::Execute(Event event)
{
    std::string const castName = ai::buff::UpgradeToGroupIfAppropriate(bot, botAI, spell);
    return botAI->CastSpell(castName, GetTarget());
}
```

### Updated Inheritance Chain

```
CastSpellAction
  └── CastAuraSpellAction
        └── CastBuffSpellAction              ← generic, untouched
              └── BuffOnPartyAction           ← party targeting, untouched
                    └── CastGroupBuffOnPartyAction  ← group-variant logic lives here
                          ├── CastMarkOfTheWildOnPartyAction
                          ├── CastArcaneIntellectOnPartyAction
                          ├── CastPrayerOfFortitudeOnPartyAction
                          ├── CastPrayerOfSpiritOnPartyAction
                          └── CastShadowProtectionOnPartyAction
```

This is structurally identical to how `CastTotemAction` sits between `CastBuffSpellAction` and the individual totem actions. The group-variant logic is contained to the classes that actually need it, and the individual spell action classes remain one-liners.

### Trigger Fix

The same principle applies to `BuffTrigger` and `BuffOnPartyTrigger`. The group-variant check and post-login deferral should not be in the generic trigger base. The specific triggers for Druid, Mage, and Priest party buffs (`MarkOfTheWildOnPartyTrigger`, `ArcaneIntellectOnPartyTrigger`, etc.) should override `IsActive()` individually, the same way `CastMagmaTotemAction` overrides `isUseful()` for its melee-range check.

---

### Double Target Lookup in `BuffOnPartyTrigger::IsActive()`

The new override calls `GetTarget()` before delegating to `BuffTrigger::IsActive()`, which also calls `GetTarget()` internally:

```cpp
bool BuffOnPartyTrigger::IsActive()
{
    Unit* target = GetTarget();                                         // lookup #1
    if (ai::buff::ShouldDeferPartyBuffEvaluationForRecentLogin(...))
        return false;
    return BuffTrigger::IsActive();  // calls GetTarget() again        // lookup #2
}
```

Even if the value is cached, the call is redundant. If the intent is to defer only on the bot's own login grace (not the target's), `GetTarget()` need not be called at all before the deferred check.

---

**back to human being**

While the AI said some silly stuff, some of it makes sense. I would consider changing it from modifying those base classes to making a subclass, and putting the logic there. And I'd look over the double target lookup, not that I know exactly what it means. 

Overall, super happy with the PR. I am very eager for this to launch. It will help my precious, precious dps numbers.

brighton-chi:
This is fabulous, thank you. I'll need a bit of time to go through it but it's so helpful. 

Sanctuary is not handled in any sophisticated way right now--the logic basically is if a bot is supposed to get Sanctuary but there isn't a Paladin that can cast it, the bot gets Kings in that slot. So that's how it is running into 3 Paladins/2 buffs since the Druid is supposed to get Kings 3rd I think so the allocation becomes Kings, Might, Kings (i.e., only 2 buffs). I'll think about if there's a better way to handle that.

The issue with the spamming is not immediately evident to me off the top of my head so I'll need to look into it.

ThePenguinMan96:
> This is fabulous, thank you. I'll need a bit of time to go through it but it's so helpful. 
> 
> Sanctuary is not handled in any sophisticated way right now--the logic basically is if a bot is supposed to get Sanctuary but there isn't a Paladin that can cast it, the bot gets Kings in that slot. So that's how it is running into 3 Paladins/2 buffs since the Druid is supposed to get Kings 3rd I think so the allocation becomes Kings, Might, Kings (i.e., only 2 buffs). I'll think about if there's a better way to handle that.
> 
> The issue with the spamming is not immediately evident to me off the top of my head so I'll need to look into it.

I mean, it is such a small/superficial issue that functionally, they don't perform any different - MP5 won't be the decision maker in a bear tanks success. I'm fine with leaving that alone. 

All I did for the paladin glitch was hit .character level 60 on all of them, and they were level 40 before. I didn't maintenance on them yet. That could be the issue, something about them not knowing the spell

brighton-chi:
> > This is fabulous, thank you. I'll need a bit of time to go through it but it's so helpful.
> > Sanctuary is not handled in any sophisticated way right now--the logic basically is if a bot is supposed to get Sanctuary but there isn't a Paladin that can cast it, the bot gets Kings in that slot. So that's how it is running into 3 Paladins/2 buffs since the Druid is supposed to get Kings 3rd I think so the allocation becomes Kings, Might, Kings (i.e., only 2 buffs). I'll think about if there's a better way to handle that.
> > The issue with the spamming is not immediately evident to me off the top of my head so I'll need to look into it.
> 
> I mean, it is such a small/superficial issue that functionally, they don't perform any different - MP5 won't be the decision maker in a bear tanks success. I'm fine with leaving that alone.
> 
> All I did for the paladin glitch was hit .character level 60 on all of them, and they were level 40 before. I didn't maintenance on them yet. That could be the issue, something about them not knowing the spell

In practice I think the Sanctuary bug exists only for Bears (who debatably might want Wisdom over Sanctuary anyway, and thus I could just swap the priority of the two) and casters (who would lose out on Might as a result, whoop de doo). So I'd be tempted to take the cheap route there, except I think the spam buffing issue is indicative of a greater systemic issue that may need significant reworking (in a manner that would make the Sanctuary matter sensible to resolve since it would just be a tiny piece of a much bigger puzzle). Let's see how much time my daughter allows me this weekend =P

brighton-chi:
So regarding the code review, I think it is definitely a good point that we shouldn't shred up CastBuffSpellAction, and a subclass is warranted. And BuffTrigger should stay generic. But I'm not sure about creating a subclass under BuffOnPartyTrigger. That would I guess distinguish between single buffs cast on party members and group buffs cast on party members, but I'm not sure there is really enough there to warrant another inheritance level when this module is already so many layers deep in a way that is often hard to follow and results in some bad overrides that already exist.

ThePenguinMan96:
Just reviewed the latest commits, super happy with the blessingonparty trigger you made, as well as the groupbuffspellaction and groupbuffonparty addition. Perfect use of subclassing. And I see the extended logic to prevent blessings from looping. When I get time, I will test again.

brighton-chi:
This has been the biggest PITA PR I've ever dealt with. There's always some edge case that seems not to work. Whatever, marking for review.

src/Ai/Base/Actions/GenericSpellActions.cpp
```diff
@@ -175,14 +158,58 @@ bool CastAuraSpellAction::isUseful()
...
+{
+    Unit* target = GetTarget();
+    if (!target || !CastSpellAction::isUseful())
+        return false;
+
+    Aura* aura = botAI->GetAura(spell, target, isOwner, checkDuration);
+    if (!aura || (beforeDuration && aura->GetDuration() < beforeDuration))
         return true;
+
     return false;
```
kadeshar: :D
`return !aura || (beforeDuration && aura->GetDuration() < beforeDuration)`

src/Ai/Base/Trigger/GenericTriggers.cpp
```diff
@@ -36,52 +37,50 @@ bool MediumManaTrigger::IsActive()
...
-    return AI_VALUE2(bool, "has mana", "self target") && AI_VALUE2(uint8, "mana", "self target") > 85;
+    return AI_VALUE2(bool, "has mana", "self target") &&
+           AI_VALUE2(uint8, "mana", "self target") > 85;
 }
 
 bool EnoughManaTrigger::IsActive()
 {
-    return AI_VALUE2(bool, "has mana", "self target") && AI_VALUE2(uint8, "mana", "self target") > sPlayerbotAIConfig.highMana;
+    return AI_VALUE2(bool, "has mana", "self target") &&
+           AI_VALUE2(uint8, "mana", "self target") > sPlayerbotAIConfig.highMana;
```
kadeshar: Refactoring shouldnt be done with feature PR only in separate one

src/Ai/Base/Util/GenericBuffUtils.cpp
```diff
@@ -72,75 +200,33 @@ namespace ai::buff
...
+        if (groupName.empty())
+            return baseName;
 
-            // We check usefulness on the **basic** buff (not the greater version),
-            // because "spell cast useful" may return false for the greater variant.
-            bool const usefulBase = botAI->GetAiObjectContext()
-            ->GetValue<bool>("spell cast useful", baseName)->Get();
+        // Prefer singles when no more than two same-map members are missing this buff family;
+        // use the group spell once enough nearby members still need it.
+        if (!HasEnoughSameMapMissingPlayersForGroupVariant(bot, botAI, baseName, groupName))
```
kadeshar, line 220: Im not sure that comment states true.
"still need it" means for HasEnoughSameMapMissingPlayersForGroupVariant that dont have aura. 
What about bot which aura almost cancelled?

src/Ai/Base/Util/GenericBuffUtils.cpp
```diff
@@ -72,75 +200,33 @@ namespace ai::buff
...
-                    announce(announceText);
-                    last = now;
-                }
-            }
-        }
-        return castName;
+        if (groupSpellId && HasRequiredReagents(bot, groupSpellId))
+            return groupName;
+
+        return baseName;
```
kadeshar, line 229: There was issue with notification about missing reagents?
Now there is silent switch on single target.

kadeshar:
Please dont make PR with both feature and refactoring especial such huge one.

brighton-chi:
> Please dont make PR with both feature and refactoring especial such huge one.

Want me to split it up? Tempting to both when looking at the same file =P

src/Ai/Base/Util/GenericBuffUtils.cpp
brighton-chi: I deleted it because it was noted by some people in Discord, such as Revision, that they did not like it (and I didn't really care for it, either).

src/Ai/Base/Util/GenericBuffUtils.cpp
brighton-chi: I can rephrase the comment.

Ideally I think bots should refresh buffs a little bit before they expire. And there should also be a chat command to rebuff. Those are things I have in mind to maybe do in the future.

kadeshar:
> > Please dont make PR with both feature and refactoring especial such huge one.
> 
> Want me to split it up? Tempting to both when looking at the same file =P

Now is too late to split. I know is tempting but in such big PR its problem :P

src/Ai/Base/Util/GenericBuffUtils.cpp
kadeshar: Its personal PoV. This regeants are refill automatic?

src/Ai/Base/Util/GenericBuffUtils.cpp
kadeshar: Ok, if that is something TODO i believe in you :D

src/Ai/Base/Util/GenericBuffUtils.cpp
brighton-chi: I didn't change any behavior regarding reagents. So bots get them with the maintenance command, unless the InitReagents part of the command is disabled in the config. It is possible bots would run out and only be able to use singles. That is the case with group buffs too like Prayer of Fortitude and so on. Unlike greater blessings, those never had any text about being out of reagents (the text was originally added by wishmaster for the greater blessings only).

kadeshar (approved):

PR #2366, opened 2026-05-06
update to level 80 pve specs

ScoobyPwnsOnU:
## Pull Request Description

Some of the specs needed improvement.

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist


brighton-chi:
Having Glyph of SS is fine, but EA still needs to be a major. Bots are coded to use EA with 3 or fewer combo points, and without the glyph, that is not a workable approach (would need to EA at 4 or 5). Although it really be 2 or fewer with the glyph, just another thing for the Rogue refactor...

Anyway, maybe replace KS with EA? KS is I think generally the highest DPS glyph for Combat, but it's not huge, and people can always set their own glyphs at max level.

What did you change with Assassination?

ScoobyPwnsOnU:
I have reverted the combat changes as idc to argue about it anymore.

kadeshar (approved):

PR #2381, opened 2026-05-11
Implement Black Temple Strategies

brighton-chi:
## Pull Request Description

This PR implements strategies for all bosses in the Black Temple. As always, I’ve written these with the intent that all bosses be completable with appropriate gear and 50/50 IP nerfs and boss HP returned to TBC levels. Illidan is difficult at those parameters, but he is certainly doable. You just won’t be able to roll a bunch of meme specs or shitty compositions. Probably.

## Feature Evaluation

My goal with raid strategies is the same—I’m not just trying to make encounters doable but am trying to allow people to experience them in a way that feels like they are part of a coordinated raid of real players. To me, being able to achieve that is the minimum, and that requires getting bots to respond to all major mechanics, even if they can technically just be powered through.

## How to Test the Changes

Run the Black Temple raid and see if my descriptions of strategies in the next post check out.

## Impact Assessment

    - - [ ] No, not at all
 
The performance impact exists only when the “blacktemple” strategy is on. I’ve tested with pmon and done my best to properly gate checks to limit the performance impact even during the instance and boss encounters.

    - - [ ] No
    - - [x] Yes (**explain why**)

The triggers and multipliers will be evaluated as long as the "blacktemple" strategy is active, and it will be applied automatically in the instance (and removed automatically when changing maps outside the instance).

    - - [x] No

## AI Assistance

- - [ ] No

GPT-5.4, mainly for calculations and things that are more intermediate concepts that I’m not proficient with like lambdas.

## Final Checklist

### DIFF
--- a/src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
+++ b/src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
@@ -1485,8 +1485,7 @@ bool IllidariCouncilMainTankReflectJudgementOfCommandAction::Execute(Event /*eve
     return false;
 }
 
-// Rename to something like "focus" Malande if this works
-bool IllidariCouncilFirstAssistTankPositionMalandeAction::Execute(Event /*event*/)
+bool IllidariCouncilFirstAssistTankFocusMalandeAction::Execute(Event /*event*/)
 {
     Unit* malande = AI_VALUE2(Unit*, "find target", "lady malande");
     if (!malande)
@@ -1504,25 +1503,6 @@ bool IllidariCouncilFirstAssistTankPositionMalandeAction::Execute(Event /*event*
 
     if (AI_VALUE(Unit*, "current target") != malande)
         return Attack(malande);
-
-    /* if (malande->GetVictim() == bot)
-    {
-        const Position& position = MALANDE_TANK_POSITION;
-        const float distToPosition = bot->GetExactDist2d(position.GetPositionX(),
-                                                         position.GetPositionY());
-        if (distToPosition > 5.0f)
-        {
-            const float dX = position.GetPositionX() - bot->GetPositionX();
-            const float dY = position.GetPositionY() - bot->GetPositionY();
-            const float moveDist = std::min(5.0f, distToPosition);
-            const float moveX = bot->GetPositionX() + (dX / distToPosition) * moveDist;
-            const float moveY = bot->GetPositionY() + (dY / distToPosition) * moveDist;
-
-            return MoveTo(BLACK_TEMPLE_MAP_ID, moveX, moveY, position.GetPositionZ(),
-                          false, false, false, false, MovementPriority::MOVEMENT_COMBAT,
-                          true, true);
-        }
-    } */
 
     return false;
 }
@@ -2360,6 +2340,9 @@ bool IllidanStormrageAssistTanksHandleFlamesOfAzzinothAction::RepositionToAvoidB
         }
         waypoints = &W_GLAIVE_TANK_POSITIONS;
     }
+
+    if (!waypoints)
+        return false;
 
     size_t& waypointIndex = flameTankWaypointIndex[bot->GetGUID()];
     const Position& target = (*waypoints)[waypointIndex];
--- a/src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
+++ b/src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
@@ -1,0 +1,3058 @@
+/*
+ * Copyright (C) 2016+ AzerothCore <www.azerothcore.org>, released under GNU AGPL v3 license, you may redistribute it
+ * and/or modify it under version 3 of the License, or (at your option), any later version.
+ */
+
+#include "RaidBlackTempleActions.h"
+
+#include <vector>
+
+#include "CreatureAI.h"
+#include "Playerbots.h"
+#include "RaidBlackTempleHelpers.h"
+#include "RaidBossHelpers.h"
+
+using namespace BlackTempleHelpers;
+
+// General
+
+bool BlackTempleEraseTimersAndTrackersAction::Execute(Event /*event*/)
+{
+    const ObjectGuid guid = bot->GetGUID();
+    const uint32 instanceId = bot->GetMap()->GetInstanceId();
+
+    if (botAI->IsTank(bot))
+    {
+        bool erased = false;
+        if (!AI_VALUE2(Unit*, "find target", "illidan stormrage") &&
+            !AI_VALUE2(Unit*, "find target", "flame of azzinoth"))
+        {
+            if (illidanBossDpsWaitTimer.erase(instanceId) > 0)
+                erased = true;
+            if (illidanFlameDpsWaitTimer.erase(instanceId) > 0)
+                erased = true;
+            if (illidanLastPhase.erase(instanceId) > 0)
+                erased = true;
+            if (illidanShadowTrapGuid.erase(guid) > 0)
+                erased = true;
+            if (illidanShadowTrapDestination.erase(guid) > 0)
+                erased = true;
+            if (flameTankWaypointIndex.erase(guid) > 0)
+                erased = true;
+            if (westFlameGuid.erase(instanceId) > 0)
+                erased = true;
+            if (eastFlameGuid.erase(instanceId) > 0)
+                erased = true;
+        }
+        if (!AI_VALUE2(Unit*, "find target", "gathios the shatterer"))
+        {
+            if (councilDpsWaitTimer.erase(instanceId) > 0)
+                erased = true;
+            if (gathiosTankStep.erase(guid) > 0)
+                erased = true;
+        }
+        if (!AI_VALUE2(Unit*, "find target", "mother shahraz") &&
+            shahrazTankStep.erase(guid) > 0)
+        {
+            erased = true;
+        }
+        return erased;
+    }
+    else if (botAI->IsHeal(bot))
+    {
+        if (zerevorHealStep.erase(guid) > 0)
+            return true;
+        else
+            return false;
+    }
+    else
+    {
+        bool erased = false;
+        if (!AI_VALUE2(Unit*, "find target", "supremus") &&
+            supremusPhaseTimer.erase(instanceId) > 0)
+        {
+            erased = true;
+        }
+        if (!AI_VALUE2(Unit*, "find target", "ashtongue channeler") &&
+            hasReachedAkamaChannelerPosition.erase(guid) > 0)
+        {
+            erased = true;
+        }
+        if (!AI_VALUE2(Unit*, "find target", "gurtogg bloodboil") &&
+            gurtoggPhaseTimer.erase(instanceId) > 0)
+        {
+            erased = true;
+        }
+        return erased;
+    }
+}
+
+// High Warlord Naj'entus
+
+bool HighWarlordNajentusMisdirectBossToMainTankAction::Execute(Event /*event*/)
+{
+    Unit* najentus = AI_VALUE2(Unit*, "find target", "high warlord naj'entus");
+    if (!najentus)
+        return false;
+
+    Player* mainTank = GetGroupMainTank(botAI, bot);
+    if (!mainTank)
+        return false;
+
+    if (botAI->CanCastSpell("misdirection", mainTank))
+        return botAI->CastSpell("misdirection", mainTank);
+
+    if (bot->HasAura(static_cast<uint32>(BlackTempleSpells::SPELL_MISDIRECTION)) &&
+        botAI->CanCastSpell("steady shot", najentus))
+    {
+        return botAI->CastSpell("steady shot", najentus);
+    }
+
+    return false;
+}
+
+bool HighWarlordNajentusTanksPositionBossAction::Execute(Event /*event*/)
+{
+    Unit* najentus = AI_VALUE2(Unit*, "find target", "high warlord naj'entus");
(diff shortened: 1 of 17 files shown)


brighton-chi:
### Strategies

Note: Reliquary has a tendency to be pulled through the floor on the way to Gorefiend, so it is heavily recommended that you kill Reliquary (and then, for efficiency, Bloodboil) before Gorefiend.

I'm not going to go over Misdirection actions except where they are particularly relevant, as I use them constantly and mostly they are for pulls.

**High Warlord Naj'entus**
- The active tank has a designated spot not too far away from the boss's spawn point.
- Ranged attempts to maintain at least 10 yards from the boss and 7 yards from other players to avoid chaining Needle Spines.
- When a bot is impaled, the nearest non-tank bot will remove the Impaling Spine; there is a slight built-in delay to try to make this seem more realistic.
- When Tidal Shield goes up, after a short built-in delay for realism, bots will throw any Impaling Spines in their inventory to break the shield.

**Supremus**
- During the tanking phase, ranged will attempt to maintain at least 8 yards from other players to mitigate the disruptive impact of the Molten Flame trails (which actually chase players).
- During the kite phase, the fixated bot kites Supremus just by running straight away from him. It's not a fabulous implementation, but due to the size of the room, this generally works OK.
- Bots will avoid Volcanos during the kite phase.
- Rogues are prevented from using Killing Spree because Supremus's hitbox is buggy.

**Shade of Akama**
- Melee dps will move onto the platform with the Channelers at the beginning of the encounter and then will mark the Channelers with skull and prioritize them in turn.

**Teron Gorefiend**
- Tanks will position the boss near where he spawns.
- Ranged line up in an arc with their backs up against the balcony.
- Bots targeted with Shadow of Death will try to escape the spell during the cast if they have an ability that can do so (Vanish, Ice Block, Divine Shield, Feign Death).
- Bots with Shadow of Death will, when the aura is about to expire, run to the corner of the room left of the entrance to die.
- Bots killed by Shadow of Death will play the Vengeful Spirit game by using the Spirit's abilities to slow and kill the Shadowy Constructs; they should not have any problem getting all the Constructs down before they reach the raid.
- Once a bot's Constructs are dead, it will go cast Spirit Strike on the boss and also help any subsequent bot kill Constructs.

**Gurtogg Bloodboil**
- Tanks will attempt to tank the boss up against the statues behind where the boss spawns.
- Ranged bots split into 3 groups of 5 bots each. Every 10 seconds, a different group moves to be the farthest from the boss to take the next Bloodboil.
- When a player gets Fel Rage, bots will move 20 yards away from the targeted player.

**Reliquary of Souls**

Essence of Suffering
- Tanks attempt to be the closest to the boss since it targets the closest player; meanwhile, ranged and melee dps try to maintain distance from the boss.
- Normal healing is disabled by multiplier since it does not work; instead, healers are directed to dps the boss. Priests are still allowed to use Power Word: Shield.

Essence of Desire
- Mages are directed to Spellsteal the boss's Rune Shield; in practice this may not happen because another bot may dispel the buff first, but I decided not to disable dispelling because the benefit of ensuring a Mage gets the buff is far outweighed by the downside of the buff not being removed if there are no Mages or all Mages are dead.
- If the tank is a Warrior, it will attempt to use Spell Reflection on Deaden; in practice, it is very likely that Deaden will be interrupted, but I decided not to prevent interrupting it since there's no guarantee that a player would be using a Warrior tank, and while I could have had a condition that the tank be a Warrior for suppressing interrupts, I thought the downside of Deaden hitting the tank was too high to risk.

Essence of Anger
- There's nothing, just burn it.

**Mother Shahraz**
- Tanks will pull the boss to about under the statue to the left of her starting position (when facing her).
- Ranged will position under the statue to the right of her starting position; melee dps will also position there until tanks get the boss into position, after which melee dps will attack.
- The positioning is technically to stop fall damage from Sinister Beam by putting a vertical barrier in the way; this doesn't actually work in practice due to the collision being wonky on AC, and bots not really obeying regular collision rules anyway. Plus bots don't take fall damage, so it doesn't really matter. But I just wanted the fight to look more realistic in execution.
- Bots with Fatal Attraction will run away from each other to break their chains. Rarely, they may get stuck in a position that prevents them from breaking their chains; I put in a call for some random movement in case that happens, but this getting stuck has happened to me only twice ever, and it hasn't happened since I put in the failsafe movement method, so I don't know how well the fallback works.

**Illidari Council**

You will need three traditional tanks and a Mage tank (no special gear needed, though more stam is good). I find it easiest to pull by having the Mage attack High Nethermancer Zerevor instead of sending your traditional tanks in.
- Misdirects on pull go in order of Mage tank/Zerevor, first assist tank/Malande, second assist tank/Darkshadow, main tank/Gathios.
- When the Mage tank grabs aggro on Zerevor, it will drag Zerevor over to near the ramp on the left side of the room (facing into the room from the entry). One healer will go over to be the designated healer for the Mage tank.
- Gathios will be kited by the main tank in a rectangle-ish pattern with four points. Whenever an AoE (Gathios's Consecration or Zerevor's Blizzard or Flamestrike) hits the main tank, it will pull Gathios to the next point. This will continue through the fight.
- Malande is a caster who does not properly chase, so she is just tanked by the first assist tank wherever she ends up, which in practice is near where she starts.
- The second assist tank picks up Darkshadow whenever he emerges from Vanish. Darkshadow is moved to Gathios to stack them up for cleave damage.
- DPS generally targets Darkshadow when he is up; otherwise they target Gathios. Exception: Warriors and Rogues will attack Malande when she doesn't have physical immunity, and Shamans will attack Malande when she doesn't have magic immunity. This is to ensure somebody is available to interrupt her Circle of Healing.

**Illidan Stormrage**

You need 3 traditional tanks, with the first and second assist tanks having considerable fire resistance. This is probably needed even without IP nerfs. With IP nerfs, you should try to get close to max while still maintaining crit immunity. The reason is because of a really nasty bug—the Demon Fires left after Eye Blast stack. They shouldn’t stack, it should be 2k per second unresisted, but they do when coming from different trigger NPCs in AC, and thus depending on where the tank is standing, it is either 4k or 6k per second unresisted, which is insane when piled on top of all the other damage.
You also need a Warlock tank with a lot of shadow resistance if you are playing with IP nerfs; without nerfs, you'll probably never see a demon phase, so a Warlock tank is not needed. If you do see a demon phase, try not to take Illidan into Phase 5 while he is in demon phase because he WILL NOT wipe threat during that scene and therefore will basically just proceed to kill your Warlock tank and all your ranged dps as soon as Phase 5 starts.

Phase 1
- Major cooldowns, including Bloodlust/Heroism, are saved until Phase 3
- Non-tanks briefly hold DPS at the start of the phase to give the tank time to establish aggro.
- The main tank will step back to get out of Flame Crashes.
- If a bot gets Parasitic Shadowfiend, it will move away from nearby players, and ranged dps will prioritize Parasitic Shadowfiends.

Phase 2
- Hunters will Misdirect the Flames of Azzinoth to the flame tanks. There is a built-in DPS wait at the start of the flame phase to give the assist tanks time to get aggro.
- The two assist tanks handle the Flames of Azzinoth. Before the Flames spawn, the assist tanks move to waiting positions by the glaives. While tanking the Flames, the assist tanks reposition to avoid Blaze patches (moving through set kite paths) and will also sidestep Eye Blast if it is about to pass through them.
- Everyone who is not actively tanking a Flame moves to one of 3 spread positions on the grate, which are set up to be far enough apart not to chain Fireball and also are positioned in a way not to be in range of any (1) Blaze dropped on the standard tank kiting paths or (2) Demon Fire left from an Eye Blast.
- Melee will stay at the grate positions so mostly they will be useless during this phase (though they will attack Flames that come into range). It’s too much of a shitshow to try to figure out a way to get them to attack safely (without chaining Fireball or dying in fire).
- Pets are forced passive during this phase so they do not instantly pull aggro or die in the fire.
- Mages, Paladins, and Rogues will try to remove Dark Barrage with Ice Block, Divine Shield, or Cloak of Shadows.

Transition to Phase 3
- As Illidan prepares to land, bots move away from the landing point.

Phase 3
- Non-tanks briefly hold DPS at the start of the phase.
- Ranged fan out behind Illidan.
- If a bot is infected with Parasite, it will run behind the ranged group and a designated Hunter will go there to drop a Frost Trap to slow the spawned Shadowfiend.
- Ranged dps will prioritize Shadowfiends over Illidan.
- The main tank will continue to move Illidan away from Flame Crashes.

Phase 4
- Hunters will Misdirect Illidan to the Warlock tank
- The Warlock tank stays at range, uses Shadow Ward, and spams Searing Pain to hold Illidan.
- Non-Warlock DPS briefly waits at the start of demon phase to give the Warlock tank time to establish aggro.
- Ranged forms a wide circle around Illidan, leaving the Warlock tank’s sector clear.
- Melee back away and wait at a farther distance than ranged, but they will move in if it is safe to attack a Shadow Demon or low-health Shadowfiend.
- Ranged dps prioritizes Shadow Demons first, then Shadowfiends, then Illidan.

Phase 5
- This is mostly the same as phase 3, except the main tank will move Illidan toward the Shadow Trap if it is not too far away (otherwise, the trap will be ignored). Then, the closest bot to the trap will trigger it when Illidan is in range.

Cheats
- Bots will clear residual hazards between phases (Flame Crashes, Blazes, Demon Fires). This is not gated behind the cheat config option; the strategy simply cannot work without this being done.
- Bots will autokill residual Shadowfiends in phase 2. This is gated behind the cheat config.
- Shadow Demons will have their HP nerfed to 25% of max in phase 4. This is gated behind the cheat config.

src/Ai/Raid/BlackTemple/Util/RaidBlackTempleHelpers.cpp
```diff
@@ -0,0 +1,471 @@
...
+            {
+                PlayerbotAI* memberAI = GET_PLAYERBOT_AI(member);
+                if (memberAI && memberAI->IsRanged(member))
+                    rangedMembers.push_back(member);
+            }
+        }
+
+        for (size_t i = 0; i < rangedMembers.size(); ++i)
+        {
+            groups[i / 5].push_back(rangedMembers[i]);
```
kadeshar, line 74: In theory 16+ ranged can cause crash?

src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
```diff
@@ -0,0 +1,3058 @@
...
+            const float dY = position.GetPositionY() - bot->GetPositionY();
+            const float moveDist = std::min(5.0f, distToPosition);
+            const float moveX = bot->GetPositionX() + (dX / distToPosition) * moveDist;
+            const float moveY = bot->GetPositionY() + (dY / distToPosition) * moveDist;
+
+            return MoveTo(BLACK_TEMPLE_MAP_ID, moveX, moveY, position.GetPositionZ(),
+                          false, false, false, false, MovementPriority::MOVEMENT_COMBAT,
+                          true, true);
+        }
+    } */
```
kadeshar: This commented code is for something?

src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
```diff
@@ -0,0 +1,3058 @@
...
+        if (!westFlame || westFlame->GetVictim() != bot ||
+            !bot->IsWithinMeleeRange(westFlame))
+        {
+            return false;
+        }
+        waypoints = &W_GLAIVE_TANK_POSITIONS;
+    }
+
+    size_t& waypointIndex = flameTankWaypointIndex[bot->GetGUID()];
+    const Position& target = (*waypoints)[waypointIndex];
```
kadeshar, line 2348: There is no possibility that waypoints are here still nullptr?

src/Ai/Raid/BlackTemple/Util/RaidBlackTempleHelpers.cpp
```diff
@@ -0,0 +1,471 @@
...
+
+namespace BlackTempleHelpers
+{
+    // High Warlord Naj'entus
+    const Position NAJENTUS_TANK_POSITION = { 438.515f, 772.436f, 11.931f };
+
+    // Supremus
+    const Position SUPREMUS_TANK_POSITION = { 704.651f, 684.401f, 72.608f };
+
+    std::unordered_map<ObjectGuid, Position> supremusRangedPositions;
```
kadeshar: I dont see usage

src/Ai/Raid/BlackTemple/Util/RaidBlackTempleHelpers.cpp
brighton-chi: It shouldn't. The function breaks after 15 so the remainder just aren't assigned. I almost always have more than 15 ranged so would have experienced an issue if there was one.

src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
brighton-chi: No, I previously had a fixed position then decided to try without one, and it works better without it so I just need to delete the commented code.

src/Ai/Raid/BlackTemple/Util/RaidBlackTempleHelpers.cpp
brighton-chi: I'll delete it. That's from wayyy back, I never really seriously pursued that path.

src/Ai/Raid/BlackTemple/Action/RaidBlackTempleActions.cpp
brighton-chi: Practically I don't think so, but no harm in adding a check so I'll do that.

kadeshar (approved):

PR #2383, opened 2026-05-12
Updates the Windows CI workflow to build AzerothCore and mod-playerbots reliably with Ninja instead of the Visual Studio/MSBuild generator

Wishmaster117:
Added a Windows CI fix to build with Ninja and avoid Windows command-line length failures.

## Pull Request Description
This PR updates the Windows CI workflow to build AzerothCore + mod-playerbots reliably with Ninja instead of the Visual Studio/MSBuild generator.

The previous Windows build could fail during the final `worldserver` build steps because MSBuild/RC generated command lines that exceeded Windows command-line limits. Enabling long-path support was not enough, because the failure was caused by tool command length rather than only filesystem path length.

This workflow now:
- installs and uses Ninja as the CMake generator;
- creates an AzerothCore `conf/config.sh` for CI;
- forces the compiler to MSVC `cl`;
- enables CMake/Ninja response files to reduce command-line length;
- moves the source tree to a short path (`C:\ac`) before configuring and building.

This keeps the Windows build on MSVC while avoiding the MSBuild/RC command-line length issue.

**Pros:**
- Fixes Windows CI command-line length failures;
- avoids unreliable `rc.exe` / `MSB6003` errors from the MSBuild generator;
- keeps the build on MSVC;
- makes the Windows CI build more predictable;
- does not affect runtime code or bot behavior.

**Cons:**
- Visual Studio project files (`.vcxproj`) are no longer generated in CI;
- the workflow now copies the source tree to a short path before building.

## Feature Evaluation
- Minimum logic: CI-only workflow changes to use Ninja, force MSVC, enable response files, and build from a short path.
- Processing cost: None at runtime. This only affects GitHub Actions build tooling.

## How to Test the Changes
- Run the Windows CI workflow on GitHub Actions.
- Confirm that CMake reports `-- Building for: Ninja`.
- Confirm that CMake detects MSVC as the C and C++ compiler.
- Confirm that the Windows build completes successfully.
- Confirm that no MSBuild `MSB6003`, `rc.exe`, or Ninja `CreateProcess` command-line length errors occur.

## Impact Assessment
    - [x] No, not at all

    - [x] No

    - [x] No

## AI Assistance
- [x] No
Maybe if should so i'd made less commits...

## Final Checklist

## Notes for Reviewers
This change is CI-only and does not affect gameplay, bot logic, database logic, or runtime performance.

The Windows workflow was failing because generated build commands became too long for Windows process creation. Switching the CI build to Ninja, forcing MSVC, enabling response files, and building from `C:\ac` fixes the issue while keeping the compiler/toolchain consistent with the Windows target.


Celandriel (approved):

Celandriel (changes requested):
AI Review
```
Two real bugs in the change:

Wrong registry value name. It writes LongPathsEnable — the documented Windows value is LongPathsEnabled (with trailing d). Microsoft's docs and every standard reference use LongPathsEnabled. As written, it creates a useless value and the kernel never sees the long-path opt-in.

Missing type, data, and force flags. reg add ... /v LongPathsEnable with no /t//d//f creates the value as REG_SZ with empty data (not REG_DWORD = 1), and if the value already exists it prompts Overwrite (Yes/No/All)?. In non-interactive PowerShell that prompt makes the step hang or fail. The correct form is:


reg add HKLM\SYSTEM\CurrentControlSet\Control\FileSystem /v LongPathsEnabled /t REG_DWORD /d 1 /f
A simpler, fully equivalent PowerShell-native version would be:


      - name: Enable long paths
        shell: powershell
        run: |
          Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name 'LongPathsEnabled' -Value 1 -Type DWord
Also worth noting: enabling the registry flag isn't sufficient for every tool. Git in particular needs git config --system core.longpaths true to honour long paths — if the issue this PR addresses is Git checkout failures, the registry flag alone won't fix it.

My recommendation: request changes — fix the name (LongPathsEnabled) and add /t REG_DWORD /d 1 /f, and consider whether git config core.longpaths true is also needed for the original failure.
```

Wishmaster117 (commented):
Changed to Ninja

Celandriel (approved):

PR #2390, opened 2026-05-14
Fix Mages' Armor Strategies & Light Refactor

brighton-chi:
## Pull Request Description

Mages have a "bdps" strategy to use Molten Armor (default for Fire and Arcane) and "bmana" strategy to use Mage Armor (default for Frost). The existing code uses a series of alternatives for armor (Molten->Mage->Ice->Frost), which is needed for Mages that have not learned Molten or Mage. However, I was noticing that sometimes my Fire and Arcane Mages would end up with Mage Armor, presumably because there could be a situation in which the casting of Molten Armor failed and the fallback to Mage Armor kicked in (for example, due to there being not enough mana to cast Molten Armor). This PR makes bdps always mean Molten Armor, if it is learned, and bmana always mean Mage Armor, if it is learned, by gating through triggers.

Other changes:

- Added bdps and bmana to default Mage combat strategies (still bdps for Fire and Arcane and bmana for Frost) so that Mages will reapply armor if it expires in combat.
- Deleted Arcane Explosion strategy--it was not fully implemented because there was no associated action (such as through a CastArcaneExplosionAction class). I debated implementing it, but there isn't a suitable targeting mechanism that exists in the code from what I can tell. Arcane Mages generally just use Blizzard for AoE (and Flamestrike with PoM); Arcane Explosion is useful to use if (1) a player is moving or (2) mobs are almost dead in an AoE situation. Scenario (1) is irrelevant for bots since they cannot cast while moving. With respect to scenario (2), the existing AoE triggers in fact look for highest HP mobs so to implement Arcane Explosion in a useful manner would probably require a new Value, and that is not worth it for what would be miniscule benefit anyway.
- General cleanups of Mage code (e.g., deleted empty ActionNodes). These were based on a quick review; I did not do any sort of detailed or comprehensive review and have no desire to with this PR. Note: I know that FrostMageStrategy.cpp had a Fireball alternative for Frostfire Bolt, but I deleted it anyway because the same ActionNode is already in GenericMageStrategy.cpp.
- General cleanups of AiFactory default combat/noncombat strategies (e.g., removal of deprecated bdps and bmana strategies for Shamans).

## Feature Evaluation

Added getAlternatives for the armor strategies; this approach already exists for Druids. Added one new trigger for Molten Armor that is throttled by 10s like the existing Mage Armor trigger.

## How to Test the Changes

Give a Mage the "bdps" strategy. They should cast Molten Armor. Make them cast a different armor. After the trigger throttle period (10s), they should reapply Molten Armor. Same goes for bmana and Mage Armor. Try this in combat, and it should work too.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

The point of this PR is to fix the default Mage armor buffing behavior.

    - - [x] No

## AI Assistance

- - [ ] No

I had GPT-5.4 present a couple of possibilities to fix the issue of Mage Armor being cast with bdps, and from there I settled on the getAlternatives approach. I did everything else.

## Final Checklist

## Notes for Reviewers

@Dreathean Perhaps the wiki can be updated to document the bdps and bmana strategies for Mages?

### DIFF
--- a/src/Ai/Class/Mage/Trigger/MageTriggers.cpp
+++ b/src/Ai/Class/Mage/Trigger/MageTriggers.cpp
@@ -4,7 +4,6 @@
  */
 
 #include "MageTriggers.h"
-#include "MageActions.h"
 #include "Playerbots.h"
 #include "Player.h"
 #include "Spell.h"
@@ -23,7 +22,7 @@ bool NoManaGemTrigger::IsActive()
         5513,   // Mana Jade
         5514    // Mana Agate
     };
-    Player* bot = botAI->GetBot();
+
     for (uint32 gemId : gemIds)
     {
         if (bot->GetItemCount(gemId, false) > 0)  // false = only in bags
@@ -45,17 +44,35 @@ bool ArcaneIntellectTrigger::IsActive()
 bool MageArmorTrigger::IsActive()
 {
     Unit* target = GetTarget();
+    if (botAI->HasAura("mage armor", target))
+        return false;
+
+    if (AI_VALUE2(uint32, "spell id", "mage armor"))
+        return true;
+
     return !botAI->HasAura("ice armor", target) && !botAI->HasAura("frost armor", target) &&
-           !botAI->HasAura("molten armor", target) && !botAI->HasAura("mage armor", target);
+           !botAI->HasAura("molten armor", target);
+}
+
+bool MoltenArmorTrigger::IsActive()
+{
+    Unit* target = GetTarget();
+    if (botAI->HasAura("molten armor", target))
+        return false;
+
+    if (AI_VALUE2(uint32, "spell id", "molten armor"))
+        return true;
+
+    return !botAI->HasAura("ice armor", target) && !botAI->HasAura("frost armor", target) &&
+           !botAI->HasAura("mage armor", target);
 }
 
 bool FrostNovaOnTargetTrigger::IsActive()
 {
     Unit* target = GetTarget();
     if (!target || !target->IsAlive() || !target->IsInWorld())
-    {
         return false;
-    }
+
     return botAI->HasAura(spell, target);
 }
 
@@ -63,15 +80,14 @@ bool FrostbiteOnTargetTrigger::IsActive()
 {
     Unit* target = GetTarget();
     if (!target || !target->IsAlive() || !target->IsInWorld())
-    {
         return false;
-    }
+
     return botAI->HasAura(spell, target);
 }
 
 bool NoFocusMagicTrigger::IsActive()
 {
-    if (!bot->HasSpell(54646))
+    if (!bot->HasSpell(54646)) // Focus Magic
         return false;
 
     Group* group = bot->GetGroup();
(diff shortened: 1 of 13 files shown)


src/Ai/Class/Mage/Trigger/MageTriggers.cpp
```diff
@@ -45,33 +44,50 @@ bool ArcaneIntellectTrigger::IsActive()
...
+    Unit* target = GetTarget();
+    if (botAI->HasAura("molten armor", target))
+        return false;
+
+    if (AI_VALUE2(uint32, "spell id", "molten armor"))
+        return true;
+
     return !botAI->HasAura("ice armor", target) && !botAI->HasAura("frost armor", target) &&
-           !botAI->HasAura("molten armor", target) && !botAI->HasAura("mage armor", target);
+           !botAI->HasAura("mage armor", target);
```
kadeshar, line 67: That not block bdps via casting manual armor?

brighton-chi: No, the trigger returns true above the last return if the bot knows Molten Armor. So if one of the other armors is manually cast, the next time the trigger evaluates, the bot will override with Molten Armor.

```
if (AI_VALUE2(uint32, "spell id", "molten armor"))
    return true;

kadeshar: Nice finding. Can we make HasSpell(string spellName) to replace all such things? Like 
obraz

brighton-chi: Yeah, so we could have PlayerbotAI::HasSpell() that takes a string? Do we want to use the same name as the core function? I guess HasAura() does that.

kadeshar: For me is ok such solution :)

brighton-chi: I think no point in making a wrapper for HasSpell() that takes spellId, right? There is PlayerbotAI::HasAura(uint32 spellId, Unit const* unit), but I don't know why that exists. It's hardly even used, almost every time a spellId is used the code just uses bot->HasAura().

What if we just get rid of the spell id HasAura wrapper and change the few botAI->HasAura(spellId) calls to bot? Then the format for HasAura and HasSpell are the same. They are both called on bot for the spellId or botAI for the string.

That can be a different PR of course. Like maybe we just merge this PR to solve the problem, and I'll make a separate one to implement PlayerbotAI::HasSpell(string) and get rid of PlayerbotAI::HasAura(spellId), and I can replace the calls in the trigger with the new function.

kadeshar: Most probably migration leftover. Cleanup always sounds good

kadeshar (approved):

kadeshar (changes requested):
Conflict to resolve

brighton-chi:
> Conflict to resolve

From deletion of Value<Unit*>* CastPolymorphAction::GetTargetValue() in your PR. Resolved.

Celandriel (approved):

PR #2392, opened 2026-05-15
Druid Overhaul

ThePenguinMan96:
druids

Hello playerbots community!! After my pvp gear update, I was itching to get back to class strategies. I was testing raids, and noticed that druids... Well, they kinda sucked. 

They had glaring issues, would randomly die from bugs, and overall felt bad. Looking at issues discussed on github/discord, I decided to start working on a druid update. I started with the boomkin, and made a boomkin PR, but there were some technical issues with the storing of the eclipse mapping not clearing, and it was not good...

I closed that PR, and went back to the drawing board, with one goal in mind: Make the druid class function as best as possible WHILST keeping the code consistent with what already exists. There is a TON of _yoink and twist_ (copy and paste with or without slight edits), and anything that is custom/new is discussed in the section below. I am very proud and excited to release this though - after 45 days of coding and testing, the druid finally feels good to have in the group.

Disclaimer - this PR aims to address bugs, utility, and overall performance of Druids. It will not magically make them top DPS. I have done hours of testing, and druids can occasionally top the charts - but inconsistently. Boomkins are inconsistent due to Eclipse (until later gear phases), and Feral Cats are incredibly dependent on positioning, timing, combo points, energy, clearcasting procs... The stars have to align for things to go right for the Cat (It doesn't help that all movement, targeting, and actions are ran through the same engine, so cats really suffer in boss fights with scripted movement) But the changes I have made help those situations occur a fair bit more frequently. Let's dive in! 

## Pull Request Description

A druid overhaul across all four specs — new Cat stealth and CC systems, Eclipse tracking and rotation fixes for Balance, a Bear threat rotation rework, a Resto healing priority overhaul, and a restructured CC and AoE strategy architecture.

---

## Talent & Glyph Config

- **Balance:** Moved points out of Improved Moonfire and into Nature's Reach for threat reduction and extended cast range.
- **Bear:** Moved points out of King of the Jungle (15% enrage damage) and into Feral Instinct for increased Swipe damage and AoE threat.
- **Cat:** Swapped the Glyph of Typhoon (useless for feral) for Glyph of Dash, which benefits the cat a ton in prowl.
- **Resto:** Swapped Glyph of Rejuvenation for Glyph of Nourish (better HPS for difficult content), and Glyph of Typhoon (useless for resto) for Glyph of Dash. Moved one point from Nature's Bounty to Empowered Touch.

---

## Balance

- **New skill - Typhoon:** Added Typhoon, triggered by "enemy within melee," using cone targeting logic ported from Cone of Cold. Typhoon is from the "balance pvp" spec only, and is housed in the shared AoE strategy.

- **New skill - Cyclone:** Added Cyclone targeting the RTI CC-marked target (default moon). Incapacitates for 6 seconds with full damage immunity — cannot be broken by AoE. Priority 24.0f > Hibernate (23.0f) > Entangling Roots (22.0f). *(See CC Implementation in the code notes below.)*

- **Bug fix - Eclipse cooldown tracking:** Eclipse procs referenced a cooldown not registered in the database, so boomkins immediately reverted to the wrong filler after an Eclipse buff fell off. Fixed with manual timestamp tracking. *(See Eclipse Cooldown Tracking in the code notes below.)*

- **Bug fix — Starfall no longer pulls out-of-combat hostile enemies:** Previously fired on cooldown regardless of surroundings — its 36-yard radius (the largest AoE in the game) could silently pull entire unengaged packs. Now suppressed if any non-combat hostile NPC is within 40 yards. *(See Starfall Pull Safety in the code notes below.)*

- **Filler changed to Starfire:** Starfire is now the filler (priority 5.4) over Wrath (5.3). Starfire has a 100% spellpower coefficient vs. Wrath's 12%, is more mana-efficient per point of damage, and has a 100% eclipse proc chance on crit vs. Wrath's 40%. In practice this significantly reduced mana consumption, especially before level 40 when Moonkin Form's mana-on-crit passive isn't available.

- **Moonfire / Insect Swarm on Attacker rework:** Previously tied to the light AoE trigger (2+ enemies, 13.0f), causing low-level boomkins to multi-dot instead of casting fillers — mana-inefficient at low levels where targets rarely live long enough to tick the full DoT. Re-added as dedicated on-attacker triggers at lower priority than the fillers, so they fire only as a movement fallback. The triggers have been changed to override the TTL check (time to life), similar to the warlocks DoTs.

- **Hurricane channel check rework:** Previously cancelled based on distance from the bot — enemies could leave the AoE but remain within 30 yards, keeping the channel alive. Now reads the Hurricane DynamicObject's actual radius and counts only attackers physically inside it. *(See Hurricane Channel Cancel in the code notes below.)*

---

## Cat

- **Prowl (Stealth):** Implemented using the same logic as the existing Rogue stealth system. The bot enters Prowl when out of combat and a target is within range. Engagement distances are:
  - 30 yards baseline
  - −10 yards if the target already has a victim (engaged in combat)
  - −10 more yards if the target is also moving (minimum 10 yards)
  - +15 yards in Battlegrounds or Arenas
  - Enemy player targets take priority over grind/DPS targets when evaluating distance.

- **Prowl openers:** The bot approaches the target in Prowl and opens based on approach angle and level:
  - **From behind:** Ravage (learned at level 32). Before Ravage is learned, Shred is used as the opener.
  - **From the front:** Pounce (stun + bleed, learned at level 36). Before Pounce is learned, Claw is used as the opener.

- **New skill - Maim:** Added Maim as a 5 second stun-finisher at 5 combo points, but only against player targets. It will only fire when Rip and Savage Roar are already active.

- **Innervate on healer:** Cats now cast Innervate when a healer drops below the low mana threshold (`AiPlayerbot.LowMana`, default 15%). *(See Healer Low Mana Framework in the code notes below for the shared value/trigger infrastructure backing this.)*

- **Predator's Swiftness with CC spells:** Added a twotrigger pairing Predator's Swiftness (the instant-cast proc from finishing moves) with the existing CC triggers (Cyclone, Hibernate, Entangling Roots). Feral cats can now instant-cast CC the RTI CC-marked target (default moon) after a finisher using the Predator's Swiftness proc.

- **Predator's Swiftness with Rebirth:** Added a twotrigger pairing Predator's Swiftness with the combat resurrection trigger. Cats can now use a Predator's Swiftness proc to instantly cast Rebirth on a dead party member.

- **Bug fix - Autoattack no longer breaks prowl:** `MeleeAction::isUseful()` now returns false while the bot has the Prowl aura, preventing autoattack from breaking stealth before an opener spell fires. The code comment notes this pattern should be reused for a future Rogue autoattack in stealth fix.

- **Bug fix - Non-prowl skills no longer break prowl:** `isUseful()` overrides were added to Feral Charge (Cat), Mangle (Cat), Swipe (Cat), Rake to return false while Prowl is active, preventing accidental prowl breaks before the opener fires.

- **Bug fix - Clearcasting proc with energy spells:** Added dedicated `ClearcastingTrigger` / action pairings to ensure Clearcasting procs are consumed immediately. On single target, Shred is used; on AoE, Swipe (Cat) is used. This prevents the cat from using it's valuable clearcasting proc on a low energy spell (rake, feral charge - cat, cower, etc). This also fixes a bug where Clearcasting would sometimes linger for 4+ seconds without being used, as energy-based spells do not recognize the free cast and would not fire until their energy condition was met. 

- **Tiger's Fury rework:** TF previously fired on cooldown as a default action, meaning the bot would use it at full or near-full energy and gain no benefit from the energy it generates. The trigger now requires energy to be below 30 before firing, ensuring the bot recovers the full 60 energy granted by the King of the Jungle talent.

- **Faerie Fire (Feral) rework for cat:** With Omen of Clarity, spams on cooldown to fish for Clearcasting procs. Without it, applies as a normal debuff and does not reapply while active. *(See Faerie Fire (Feral) Trigger in the code notes below.)*

- **Feral Charge toggleable strategy:** Feral Charge (Cat) is now housed in a dedicated `feral charge` strategy, enabled by default for Cat druids. It can be disabled with `co -feral charge` and re-enabled with `co +feral charge`. Disabling this strategy can be useful for encounters where charging in would be unfavorable.

- **Rake on Melee Attackers removed:** The Rake on Melee Attackers action was removed from the Cat AoE strategy. Applying Rake to multiple attackers and spreading out combo points produced far lower AoE DPS than simply continuing the single-target rotation. 

- **Antiquated Omen of Clarity framework removed:** `OmenOfClarityTrigger` and `CastOmenOfClarityAction` were not functional and have been removed. The `ClearcastingTrigger` appropriately tracks Omen of Clarity procs. I believe this was from TBC when Omen of Clarity was a spell.

---

## Bear

- **Berserk tracking for Mangle:** A `berserk active` trigger fires Mangle (Bear) at priority 25.0f while Berserk is up. Previously, Mangle sat at 5.5f in the default actions while Swipe (Bear) sat at 25.0f on the light AoE trigger — meaning in any 2+ enemy encounter, Swipe would always win regardless of Berserk. Now Mangle (25.0f) sits above the AoE triggers (24.5f), so it takes priority during Berserk. It's pretty cool to see a Bear's dps on pull!

- **Faerie Fire (Feral) rework:** Previously applied once and stopped. Now spams on cooldown for continuous threat generation (~3.5k threat per cast at level 80), and serves as a ranged soft-taunt fallback on the `lose aggro` trigger when Growl is on cooldown. *(See Faerie Fire (Feral) Trigger in the code notes below.)*

- **Lacerate rework:**  Previously, lacerate was a low priority default action and bears had no duration awareness. They would occasionally let 5 stacks of Lacerate fall off, resulting in pretty significant threat loss. Now, `LacerateTrigger` fires when the target has no Lacerate debuff, the stack count is below 5, or the remaining duration is ≤ 6 seconds. 

- **Demoralizing Roar on single target:** Previously, Demoralizing Roar only fired on the medium AoE trigger (3+ enemies, skipped on bosses). A dedicated trigger now applies it in any encounter, provided Vindication, Demoralizing Shout, or Curse of Weakness are not already present, as they don't stack with Demoralizing Roar.

- **Bug fix - Rebirth on bears:** Bears no longer attempt to cast Rebirth in combat. The generic combat resurrection trigger was firing for all druid specs, causing bears to shift out of Dire Bear Form mid-fight to cast Rebirth — dropping their armor and HP while still holding aggro. Lots of sudden tank deaths...

- **Feral Charge toggleable strategy:** Feral Charge (Bear) is now housed in a dedicated `feral charge` strategy, enabled by default for Bear druids. It can be disabled with `co -feral charge` and re-enabled with `co +feral charge`. Disabling this strategy can be useful for encounters where charging in would be unfavorable.

---

## Resto

- **Blanketing strategy:** Added a `blanketing` strategy (enabled by default, `co -blanketing` to disable) that pre-HoTs group members with Wild Growth and Rejuvenation regardless of current health, prioritizing tanks → melee → ranged to maximize Revitalize uptime. *(See Blanketing Strategy in the code notes below.)*

- **Nature's Swiftness → instant Healing Touch combo:** Nature's Swiftness was previously included in the boost strategy, where it would fire proactively on cooldown regardless of context. This wasted the proc on situations where it provided no benefit. It is now exclusively reactive — triggered at high priority (56.0f) when a party member hits critical health. A paired `nature's swiftness active` trigger then immediately fires Healing Touch (55.0f) on the lowest-health party member, consuming the proc as an instant-cast emergency heal.

- **Lifebloom priority lowered (29.0f → 13.0f):** Lifebloom on the main tank is cast on Omen of Clarity procs. At 29.0f it previously outprioritised all low health reactive healing (21.4f), meaning a Clearcasting proc while a party member was at 25–44% HP would cause the bot to cast Lifebloom on the tank instead of Swiftmend or Nourish on the injured target. Lowered to 13.0f so it fires only when no reactive healing is queued.

- **Healing spell priority order reworked:** All three reactive categories (critical, low, medium) now follow the same sequence: Swiftmend → Wild Growth → Nourish → Regrowth → Healing Touch. 

- **Tranquility toggleable strategy:** Tranquility is now housed in a dedicated `tranquility` strategy, enabled by default for Resto druids. It can be disabled with `co -tranquility`. In raids, Tranquility only heals the druid's own group — not the full raid — making it situationally poor during raid-wide damage or heavy movement phases. Disabling this strategy lets players suppress the cast on those encounters without affecting the rest of the healing rotation.

---

## CC & Strategy

- **Boost strategy:** The boost strategy is assigned to all druid specs by default. It is spec-gated internally — Balance druids use it for Force of Nature (treants on cooldown), and Feral druids (both cat and bear) use it for Berserk.

- **CC strategy enabled by default:** The CC strategy was previously not assigned to druids by default. It is now enabled for Balance and Feral Cat, with behavior gated by spec:
  - Balance receives the full RTI CC trigger set (Cyclone, Hibernate, Entangling Roots).
  - Feral cats only receive RTI CC triggers when Predator's Swiftness is active (see above).

- **AoE strategy reorganized:** All AoE spells — Hurricane, Starfall, Typhoon, Swipe (Cat), and DoTs on attackers (Moonfire, Insect Swarm) — are now handled by the shared AoE strategy, consistent with how the rest of the playerbot project structures AoE logic. This does not affect Bear druids.

- **Aquatic Form while submerged:** The non-combat strategy now shifts into Aquatic Form when the bot is fully submerged out of combat (`LIQUID_MAP_UNDER_WATER`). If the bot is in another shapeshift (Bear, Cat, Moonkin, Tree), it first shifts to caster form as a prerequisite before entering Aquatic Form. The trigger intentionally does not fire at the water surface (`LIQUID_MAP_IN_WATER`), or use the "swimming" trigger, because it caused the druid to loop caster form and aquatic form endlessly while surfaced.

---

## Code Consolidation & Refactoring

- The Bear, Cat, Heal, and Caster strategy files have been renamed to Bear, Cat, Balance, and Resto respectively, to match the naming conventions used elsewhere in the project.
- The Melee and Offheal strategy files have been deleted. The offheal healing logic is preserved in full — it now lives as an optional strategy (`CatOffhealStrategy`) inside `CatDruidStrategy`, sharing the same action node factories as the base cat strategy rather than duplicating them.
- All action relevance values across the druid strategies have been converted from named constants (e.g. `ACTION_NORMAL`, `ACTION_HIGH + 4`) to explicit numerical floats (e.g. `10.0f`, `24.0f`). This makes priority ordering immediately visible in the source without needing to cross-reference the constant definitions.

---

### Eclipse Cooldown Tracking (`DruidActions.cpp`)

The previous implementation tracked the Eclipse cooldown using `EclipseSolarCooldownTrigger` and `EclipseLunarCooldownTrigger`, both of which extended `SpellCooldownTrigger` and called `bot->HasSpellCooldown(48517/48518)`. `SpellCooldownTrigger` is the standard project pattern for this — it works correctly for spells whose cooldowns are registered in the database. However, Eclipse (Solar) and Eclipse (Lunar) both have **Cooldown: n/a** in the DB. `HasSpellCooldown` always returned false, so boomkins never respected the cooldown and would revert to the wrong filler immediately after an Eclipse buff fell off.

Since the cooldown can't be read from the DB, the fix tracks it manually. When `CastWrathAction::isUseful()` or `CastStarfireAction::isUseful()` detects that the corresponding Eclipse aura has become active, it records the current timestamp and suppresses the opposing filler for 30 seconds — the actual in-game cooldown duration.

The timestamps are stored using `ManualSetValue<time_t>`, the same pattern as `LastSpellCastTimeValue` (`src/Ai/Base/Value/LastSpellCastTimeValue.h`), which is already used throughout the project to record when spells were last cast. Two new value classes — `EclipseSolarProcTimeValue` and `EclipseLunarProcTimeValue` — are registered in a new `DruidValueContextInternal` factory inside `DruidAiObjectContext.cpp`, following the same factory pattern as `DruidTriggerFactoryInternal` and `DruidAiObjectContextInternal` in the same file. Because these values live in the per-bot `AiObjectContext`, they are automatically destroyed when the bot logs out — no manual cleanup needed, and no shared state between bots.

---

### Healer Low Mana Framework (`PartyMemberToHeal.h/.cpp`, `HealthTriggers.h/.cpp`, `ValueContext.h`, `TriggerContext.h`)

`HealerLowMana` and `HealerLowManaTrigger` are added to the shared base framework rather than the druid-specific code. Currently used by the Cat Innervate trigger; designed so Mana Tide Totem, Hymn of Hope, and similar spells from other classes can hook into the same trigger without duplicating the group-scanning logic.

The pair follows the same pattern as the existing `PartyMemberToHeal` / `PartyMemberLowHealthTrigger` — the project's standard design for "scan the group for the most in-need member, then trigger when that member crosses a threshold."

**Value** (`HealerLowMana : PartyMemberValue`): `Calculate()` walks the group reference list, skips non-healers via the existing `IsHeal()` check, and uses `MinValueCalculator` to return the lowest-mana healer as a `Unit*`. Registered in `ValueContext.h` under the key `"healer low mana"`.

**Trigger** (`HealerLowManaTrigger : Trigger`): `GetTargetName()` returns `"healer low mana"`, which the base `Trigger::GetTarget()` resolves against the value context — exactly how `PartyMemberLowHealthTrigger::GetTargetName()` returns `"party member to heal"`. `IsActive()` calls `GetTarget()` and checks `GetPowerPct(POWER_MANA) < sPlayerbotAIConfig.lowMana`.

The trigger doesn't extend `HealthInRangeTrigger` because that class is specifically for health (it reads the `"health"` value). Mana requires a direct `GetPowerPct(POWER_MANA)` call, so a plain `Trigger` with a custom `IsActive()` is used instead.

Both are registered in the global `ValueContext.h` and `TriggerContext.h` rather than a class-specific factory, consistent with how all other `PartyMemberValue` subclasses are registered in the project.

---

### Blanketing Strategy (`RestoDruidStrategy.h/.cpp`, `DruidActions.h/.cpp`, `DruidAiObjectContext.cpp`)

**Strategy structure:** `DruidBlanketStrategy` is a standalone `Strategy` overlay, not embedded inside `RestoDruidStrategy`. This is the same pattern as `DruidTranquilityStrategy`, `DruidBoostStrategy`, and `DruidCcStrategy` — additive overlays that layer behavior on top of the base strategy and can be toggled independently via `co +/-blanketing`.

**Triggers:** Both `"wild growth blanket"` and `"rejuvenation blanket"` are instantiated as `BuffOnPartyTrigger(ai, spellName)` — the project's existing class from `GenericTriggers.h` for party-wide buff maintenance, used throughout the codebase for things like Blessings and Mark of the Wild. `BuffOnPartyTrigger` extends `BuffTrigger` and fires when any party member is missing the named aura. No custom trigger class was needed.

**Actions:** Both actions inherit from a shared `CastBlanketHotAction` base that itself extends `CastSpellAction`. Inheriting from `CastSpellAction` means `isPossible()` is handled for free — spell known, off cooldown, target reachable, resources available. The constructor sets `range = botAI->GetRange("heal")` to use the standard healing range.

**`GetBlanketTarget(auraName)`:** The custom part of the implementation. Walks the group in three prioritized passes — tanks first, then melee non-tanks, then ranged — returning the first eligible member found. Eligible is defined as: alive, not a GM, within `spellDistance`, and `!botAI->HasAura(auraName, member, false, true)` (not already carrying the HoT). Returns nullptr if every member is already covered.

**`isUseful()`:** On both actions simply returns `GetTarget() != nullptr` — fires as long as `GetBlanketTarget` finds someone without the HoT, and suppresses itself the moment all targets are covered.

---

### CC Implementation — Cyclone, Hibernate, Entangling Roots (`DruidTriggers.h/.cpp`, `DruidActions.h/.cpp`, `GenericDruidStrategy.cpp`)

The druid CC spells use the project's strict RTI-only pattern rather than the fallback "best candidate" pattern used by other classes (e.g., Mage Polymorph).

**`"rti cc target"` value:** A direct raid icon lookup. Reads the `"rti cc"` string value (default `"moon"`, configurable per-bot via `rti cc <icon>`), converts it to a raid icon index, and returns the live `Unit*` for that GUID. If no CC icon is set, it returns `nullptr`. There is no fallback to a best-candidate scan.

**Triggers** (`CycloneTrigger`, `HibernateTrigger`, `EntanglingRootsTrigger`): All three extend `HasCcTargetTrigger` and override `IsActive()`. The first check is always `"rti cc target"` — if it returns `nullptr`, the trigger is immediately silent. If an icon is set, it checks `"cc target"` (with the spell name as a qualifier) to verify the RTI target matches and delegates to `HasCcTargetTrigger::IsActive()`, which handles the "don't re-cast while already CC'd" check.

**Actions** (`CastCycloneCcAction`, `CastHibernateCcAction`, `CastEntanglingRootsCcAction`): All three extend `CastCrowdControlSpellAction` rather than plain `CastSpellAction`. The action names are `"cyclone on cc"`, `"hibernate on cc"`, `"entangling roots on cc"` — not the raw spell names. This matters because `CastSpellAction` stores its constructor argument as both the action name and the spell name, and `isPossible()` calls `CanCastSpell(spell, target)` using that string. Passing `"cyclone on cc"` to `CastSpellAction` would resolve to spell ID 0 and silently return false forever. `CastCrowdControlSpellAction` keeps the spell name separate from the action name, avoiding this. `GetTargetValue()` on all three returns `context->GetValue<Unit*>("rti cc target")` directly.

**Form prerequisite:** The action nodes for `"cyclone on cc"` and `"hibernate on cc"` have `NextAction("caster form")` as a prerequisite, so the bot automatically shifts out of Bear, Cat, or Moonkin form before casting. Entangling Roots has the same prerequisite.

**Priority order:** Cyclone (24.0f) > Hibernate (23.0f) > Entangling Roots (22.0f). Cyclone is preferred because it works on any target type and the target is immune to all damage and healing while cycloned — it cannot be broken by AoE. Hibernate is beast/dragonkin only. Entangling Roots can be broken by damage.

**Feral Cat CC:** Wired through `TwoTrigger` pairings with `"predator's swiftness"` (see Cat section above). Because the Predator's Swiftness proc makes the spell instant-cast, no form shift is needed — the cat casts directly from Cat Form after a finisher.

---

### Ferocious Bite Execute (`DruidTriggers.h`, `DruidCatActions.h`, `CatDruidStrategy.cpp`)

Two separate triggers fire the same `CastFerociousBiteAction`, which is a plain `CastMeleeSpellAction` with no custom logic — all the intelligence lives in the triggers.

**`FerociousBiteTimeTrigger`** ("ferocious bite time", 22.5f) — the normal rotation path. Requires 5 combo points, Savage Roar active with >10 seconds remaining, and Rip active on the target with >10 seconds remaining. The duration checks prevent spending combo points on Ferocious Bite when either buff is about to fall off and needs to be refreshed first.

**`FerociousBiteExecuteTrigger`** ("ferocious bite execute", 24.0f) — the execute window, higher priority than the time trigger. Requires only 1 combo point, and fires when the target is below **both** 25% HP and 20,000 absolute HP. The dual condition is the key design detail: the 25% threshold alone would trigger on a raid boss at 25% health — which could still be millions of HP remaining. The 20,000 HP cap ensures the execute behavior only activates when the target is genuinely close to death, at which point dumping even a partial combo point buildup into Ferocious Bite is better than continuing a normal builder-spender cycle.

---

### Faerie Fire (Feral) Trigger (`DruidTriggers.h`)

A single `FaerieFireFeralTrigger` class handles both Bear and Cat with spec-branched behavior inside `IsActive()`. It extends `DebuffTrigger` — the project's standard class for debuff maintenance on the current target — but overrides `IsActive()` to produce three distinct behaviors depending on form and talent state:

**Bear:** Bypasses `DebuffTrigger::IsActive()` entirely. Returns true whenever the target is alive and in world, regardless of whether the debuff is already present. Every cast generates immediate threat and damage, so there is no reason to wait for it to fall off before recasting.

**Cat with Omen of Clarity (talent aura 16864):** Same bypass — spams on cooldown to fish for Clearcasting procs. Faerie Fire (Feral) has no energy cost, making it a free input that can proc Omen of Clarity on any hit.

**Cat without Omen of Clarity:** Falls through to `DebuffTrigger::IsActive()` — the standard base class behavior, which checks: target alive and in world, debuff not already present (`!botAI->HasAura("faerie fire (feral)", target)`), and estimated remaining lifetime of the target is at least `needLifeTime` seconds (default 8.0f — no point applying a 30-second debuff to something about to die). Applied as a normal debuff; does not reapply while active.

Both spam paths additionally guard against Prowl — `IsActive()` returns false while the bot has the Prowl aura to prevent casting from breaking stealth.

**Strategy wiring:**
- Bear: standard rotation slot at 17.0f, plus wired into the `"lose aggro"` trigger at 25.5f as a soft-taunt fallback when Growl is on cooldown.
- Cat: low-priority filler at 5.0f.

---

### Starfall Pull Safety (`DruidActions.cpp`)

Starfall's 36-yard AoE radius is the largest in the game. A single cast near an unengaged patrol or mob pack would silently pull everything in that area. The previous implementation fired on cooldown with no awareness of the surrounding area.

`CastStarfallAction::isUseful()` now applies two guards before allowing the cast:

**CC safety check** (standard project pattern): reads `"current cc target"` and `"aoe position"`; suppresses the cast if the CC'd target is within `aoeRadius` of the bot's AoE position.

**Unengaged hostile NPC scan (custom)**: reads `"nearest hostile npcs"` (`NearestHostileNpcsValue`), which uses the project's standard `Acore::AnyUnitInObjectRangeCheck` + `Cell::VisitObjects` grid searcher at `sightDistance` (~50 yards). The value pre-filters via `AcceptUnit()`: non-players only, and `unit->IsHostileTo(bot)` must be true — this excludes neutral-faction trigger creatures, dummies, and invisible spawns that would otherwise appear in a raw range scan. The loop then applies four additional filters:

- Skip null / dead / out-of-world units (standard guard).
- Skip the current target — it is the reason we're in combat; its in-combat flag is already covered.
- Skip `!bot->IsValidAttackTarget(unit)` — safety net for hostile-faction trigger creatures carrying `UNIT_FLAG_NON_ATTACKABLE` that `IsHostileTo` alone doesn't filter.
- Skip units beyond 40 yards — Starfall's listed radius is 36; 40 adds a small buffer for patrols about to enter range.

If any remaining unit is `!unit->IsInCombat()`, the cast is suppressed — that mob is unengaged and would be pulled.

**Why `"nearest hostile npcs"` and not `"attackers"`:** `attackers` only contains units currently targeting the bot. We need to scan all hostile units in the area, not just those already aggro'd.

---

### Hurricane Channel Cancel (`DruidTriggers.h/.cpp`, `GenericDruidStrategy.cpp`)

The previous cancel condition checked whether fewer than 3 enemies were within 30 yards of the bot. This is a poor proxy — enemies could scatter laterally but still sit within that radius, keeping the channel alive while none of them were taking damage. 

The replacement is `HurricaneChannelCheckTrigger`, which locates the actual Hurricane `DynamicObject` on the field and measures from it directly.

**`IsActive()` logic:**

1. Checks `bot->GetCurrentSpell(CURRENT_CHANNELED_SPELL)` — if the bot isn't channeling at all, returns false immediately. If it is channeling but the spell isn't a Hurricane rank, also returns false. This check is necessary because `CURRENT_CHANNELED_SPELL` is a slot, not a specific spell — the same cancel action is reused for other channeled spells in the codebase, so the trigger must verify it's specifically Hurricane before acting.

2. Iterates through `HURRICANE_SPELL_IDS` (all five ranks: 16914, 17401, 17402, 27012, 48467) calling `bot->GetDynObject(spellId)` until a non-null result is found. Hurricane places a `DynamicObject` on the field that the server uses as the actual AoE cylinder — each damage tick queries which units are inside it. The DynamicObject is keyed by spell ID, so the trigger must try each rank to find whichever one the bot currently has learned and placed.

3. Reads `dynObj->GetRadius()` — the actual radius stored on the DynamicObject itself rather than a hardcoded constant. This matches exactly what the server uses to calculate damage, so the trigger's cancel condition is spatially identical to the server's hit detection.

4. Walks the `"attackers"` GuidVector and counts how many live attackers are within `dynObj->GetRadius()` of the DynamicObject's position using `unit->GetDistance(dynObj->GetPosition()) <= radius`.

5. Returns `count < minEnemies` (default 3). The trigger fires — cancelling the channel — when fewer than 3 attackers are physically inside the Hurricane AoE.

**Why `"attackers"` and not a full area scan:** Hurricane only deals damage to units that are attacking the bot (or in its threat list). Scanning all nearby hostile units would cause premature cancellation if non-aggro'd enemies happened to be standing outside the AoE. Attackers is the right scope.

**Strategy wiring:** The trigger is paired with `NextAction("cancel channel", 22.0f)` in the AoE strategy for both Balance and Resto druids. The cancel priority (22.0f) sits below the Hurricane cast priority (23.0f), so if the medium AoE trigger re-activates on the same tick the cancel fires — meaning enemies came back into range — the new cast wins over the cancel.

---

## Feature Evaluation

Most new triggers are simple aura or cooldown checks. The heavier ones are the group scans (for the blanketing HoTs and the healer mana check), but these are identical in cost to group scans already running throughout the project (all of the party member health checks). The Starfall safety check is the only genuinely new scan — it looks for nearby hostile NPCs before allowing a cast, using the same grid search the project already uses elsewhere. That being said, it's loaded on the end of the trigger/action pairing - so in the StarfallNoCDTrigger, the bot has to already have learned starfall, already be in combat, and have Starfall off of cooldown and ready to use. The Hurricane cancel check only runs while the bot is actively channeling, so it's tightly gated.

Negligible for almost everything in this PR. The vast majority of new logic is aura/buff/debuff lookups and cooldown checks that cost nothing at scale. The group scans for blanketing and healer mana follow the same pattern as existing party scans that already run on every healer bot every tick. No new unbounded operations, no shared state between bots.

## How to Test the Changes

All druids perform a bit better now - I'd say test the branch out with the druids y'all currently use. JUST REMEMBER TO DO reset botAI or talents spec "x" again, since there have been some strategies changed!! The big one being the blanketing strategy for resto druids. They heal so much better now. Also being able to control when they pre-hot is really great.

## Impact Assessment

    - - [x] No, not at all

Tested before an after with the same performance logs. I tested it with a 25 man group of only druids versus my normal 25 man group on several raid bosses - no difference in pmon.

    - - [ ] No
    - - [x] Yes (**explain why**)

Druid bots currently have several bugs/issues with them. This doesn't exactly change the skills they were already using - just refines the scenarios in which they should be used. For example, a boomkin won't use starfall when there is a pack within range but not aggro'd. You can turn off feral charge for cat druids now, so they don't fly into a bosses aoe (locust swarm on anub, overload on iron council). Bear druids don't battle rez anymore. They just feel less clunky and heal/hold aggro better. 

    - - [x] No

There are only 2 changes to files outside of the druid strategy, which is the healer low mana framework and the modification to autoattack not being used while in prowl.

## AI Assistance

- - [ ] No

AI was used heavily in the process to make this PR. First in the research necessary into how systems work, to the initial code implementation, to the testing results (explaining the outcome/why it sucks), to the fix, and then to the review of the code at the end. I will say that after I started researching how to use AI, use .md files for context, clearing sessions, I got a lot better results. 

I'll be the first to admit that it is 10 times easier to introduce a bug with AI than it is to solve one or implement something new. That is why every time it proposed a change, I asked it if the code was consistent with the project (Already present somewhere else) and if it wasn't, it was heavily scrutinized. 

It was written with Claude Code with Sonnet 4.6 (high), and peer reviewed by Github copilot. AI also made the description part of the PR, in which I modified myself. 

## Final Checklist

**Wiki commands**

@Dreathean 

While this PR does add strategies, they are all enabled by default:

co +feral charge (feral druids, both cat and bear) - enabled by default, allows/prevents the use of feral charge
co +tranquility (resto druids) - enabled by default, allows/prevents the use of tranquility
co +blanketing (resto druids) - enabled by default, allows the druid to pre-hot with wild growth and rejuvenation

But it would be worth a mention on the wiki - there are scenarios where having these strategies disabled would be beneficial.

## Notes for Reviewers

@kadeshar 
@Celandriel 
@brighton-chi 

Thank you for taking the time to look this over. There is a lot of copied code, a bit of new code (which is explained in the code explanation part of it, but please still ask questions), and a lot of refactoring. Please remember to reset the bot strategies before/after you test this branch, due to the several changes (blanketing, feral charge strategies). Reset with reset botAI or "talents spec balance pve" for any testers out there that didn't know. If/When this PR goes to the master branch, it will need to be noted to the people this same thing about resetting strategies.


src/Ai/Class/Druid/Action/DruidActions.cpp
```diff
@@ -89,12 +165,27 @@ bool CastStarfallAction::isUseful()
             return false;
     }
 
-    // Avoid single-target usage on initial pull
-    uint8 aoeCount = *context->GetValue<uint8>("aoe count");
-    if (aoeCount < 2)
+    // Suppress if any unengaged hostile NPC is within 40 yards — Starfall's 36-yard radius would pull them.
+    // "nearest hostile npcs" pre-filters to IsHostileTo, which excludes neutral-faction trigger/dummy/invisible creatures.
+    Unit* currentTarget = AI_VALUE(Unit*, "current target");
+    GuidVector const& nearbyNpcs = AI_VALUE(GuidVector, "nearest hostile npcs");
```
brighton-chi: Can this work with possible targets no los? Nearest hostile npcs is kind of misleading because its grid search is AnyUnitInObjectRangeCheck and then filters after for hostility; it's basically the most expensive grid search built into the module.

src/Ai/Class/Druid/Action/DruidActions.cpp
```diff
@@ -149,3 +258,69 @@ bool CastRejuvenationOnNotFullAction::isUseful()
...
+
+    auto eligible = [&](Player* member) -> bool
+    {
+        return member && member->IsAlive() &&
+               !member->IsGameMaster() &&
+               bot->GetDistance2d(member) <= sPlayerbotAIConfig.spellDistance &&
+               !botAI->HasAura(auraName, member, false, true);
+    };
+
+    // Pass 1: tanks
```
brighton-chi: I think instead of 3 passes you can do just 1 and remember the eligible members.

src/Ai/Class/Druid/Strategy/BearDruidStrategy.cpp
```diff
@@ -32,7 +32,7 @@ class BearTankDruidStrategyActionNodeFactory : public NamedObjectFactory<ActionN
     {
         return new ActionNode(
             "melee",
```
brighton-chi: If an ActionNode is empty you can remove it. The engine will create a default empty one.

src/Ai/Class/Druid/Strategy/BearDruidStrategy.cpp
```diff
@@ -159,99 +159,81 @@ class BearTankDruidStrategyActionNodeFactory : public NamedObjectFactory<ActionN
...
+        }
+    ));
+    triggers.push_back(new TriggerNode("berserk active", { NextAction("mangle (bear)", 25.0f) }));
     triggers.push_back(
         new TriggerNode(
             "medium aoe",
             {
-                NextAction("demoralizing roar", ACTION_HIGH + 6),
-                NextAction("swipe (bear)", ACTION_HIGH + 6)
+                NextAction("demoralizing roar", 24.5f),
```
brighton-chi, line 194: Is the intent to queue demo roar and swipe? If so, should demo roar be queued in high aoe or light aoe situations too? I don't have any clue how bears are supposed to function.

src/Ai/Class/Druid/Strategy/CatDruidStrategy.cpp
```diff
@@ -0,0 +1,446 @@
...
+    triggers.push_back(
+        new TriggerNode(
+            "cat form", {
+                NextAction("cat form", 28.0f)
+            }
+        )
+    );
+    triggers.push_back(
+        new TriggerNode(
+            "tiger's fury", {
```
brighton-chi: Would you mind making a generic low energy trigger instead? I think it would be useful for other actions too so would be good to centralize. Like AR for Rogues.

src/Ai/Class/Druid/Strategy/GenericDruidNonCombatStrategy.cpp
```diff
@@ -23,6 +23,8 @@ class GenericDruidNonCombatStrategyActionNodeFactory : public NamedObjectFactory
         creators["remove curse on party"] = &remove_curse_on_party;
```
brighton-chi, line 22: I think the commented out innervate can be removed. Not sure why it existed originally, but I can't imagine it ever being a good idea to blow that out of combat.

src/Ai/Class/Druid/Strategy/GenericDruidNonCombatStrategy.cpp
```diff
@@ -165,11 +184,16 @@ void GenericDruidNonCombatStrategy::InitTriggers(std::vector<TriggerNode*>& trig
                        NextAction("remove curse on party", ACTION_DISPEL + 7),
                        }));
 
+    triggers.push_back(new TriggerNode("aquatic form", { NextAction("aquatic form", 10.0f) }));
+
     int specTab = AiFactory::GetPlayerSpecTab(botAI->GetBot());
     if (specTab == 0 || specTab == 2) // Balance or Restoration
```
brighton-chi: Can we use constants for all tabs?

src/Ai/Class/Druid/Strategy/GenericDruidStrategy.cpp
```diff
@@ -107,52 +126,123 @@ void GenericDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
...
     triggers.push_back(
-        new TriggerNode("low health", { NextAction("barkskin", ACTION_HIGH + 7) }));
+        new TriggerNode("almost full health", { NextAction("barkskin", 40.0f) }));
+
+    Player* bot = botAI->GetBot();
+    int tab = AiFactory::GetPlayerSpecTab(bot);
+
+    if (tab == DRUID_TAB_FERAL)
+    {
+        if (!bot->HasAura(16931) /*thick hide — bear spec*/)
```
brighton-chi, line 137: I wonder if there is some way to allow bears maybe to still use rebirth in the event that they aren't actually tanking. I'm not sure if there is a way to generically check for if a player is being targeted by any hostile unit. If nothing works, then fine, just thinking that it could still be valuable in some PvE situations (and definitely PvP).

Alternatively rebirth could be a strategy that is added by default, except only for arenas/BGs for bears.

src/Ai/Class/Druid/Strategy/BalanceDruidStrategy.cpp
```diff
@@ -125,130 +129,76 @@ class CasterDruidStrategyActionNodeFactory : public NamedObjectFactory<ActionNod
...
-    triggers.push_back(
-        new TriggerNode(
-            "faerie fire",
-            {
-                NextAction("faerie fire", ACTION_HIGH)
-            }
-        )
-    );
+    // Utility/Defensive
+    triggers.push_back(new TriggerNode("medium mana", { NextAction("innervate", 29.0f) }));
```
brighton-chi, line 202: Being able to innervate oneself is useful, but I'm wondering if it would be preferable to have the boomkin innervate a healer in a group setting?

Or a class that does actual DPS. JKJK

src/Ai/Class/Druid/Strategy/RestoDruidStrategy.cpp
```diff
@@ -0,0 +1,120 @@
...
+    triggers.push_back(
+        new TriggerNode("party member almost full health",
+                        {
+                            NextAction("wild growth on party",  10.3f),
+                            NextAction("rejuvenation on party", 10.2f),
+                            NextAction("regrowth on party",     10.1f),
+                        }));
+
+    triggers.push_back(
+        new TriggerNode("medium mana", { NextAction("innervate", 25.0f) }));
```
brighton-chi, line 99: Same thought here as with boomkins re: innervate

src/Ai/Class/Druid/Trigger/DruidTriggers.cpp
```diff
@@ -47,17 +134,38 @@ bool HurricaneChannelCheckTrigger::IsActive()
 {
     Player* bot = botAI->GetBot();
```
brighton-chi: bot is already available for triggers

src/Ai/Class/Druid/Strategy/GenericDruidStrategy.cpp
```diff
@@ -107,52 +126,123 @@ void GenericDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
...
+    if (tab == DRUID_TAB_BALANCE || tab == DRUID_TAB_RESTORATION)
+    {
+        triggers.push_back(new TriggerNode(
+            "cyclone", { NextAction("cyclone on cc", 24.0f) }));
+        triggers.push_back(new TriggerNode(
+            "hibernate", { NextAction("hibernate on cc", 23.0f) }));
+        triggers.push_back(new TriggerNode(
+            "entangling roots", { NextAction("entangling roots on cc", 22.0f) }));
+    }
+    if (tab == DRUID_TAB_FERAL)
```
brighton-chi, line 198: Are the cc strategies still available to bears, just not enabled by default in AiFactory?

brighton-chi (commented):
Ok, I am absurdly excited about Aquatic Form of all things.

But this is awesome. I can't wait to test it out.

Wishmaster117:
Amazing work, and even if druid are not on top dps after this PR don't worry, some real players sucks too :D
Dis you introduce new nc + and co + strategies, if yes can you write them here so i can update MultiBot addon easyly.

ThePenguinMan96:
> Amazing work, and even if druid are not on top dps after this PR don't worry, some real players sucks too :D
> Dis you introduce new nc + and co + strategies, if yes can you write them here so i can update MultiBot addon easyly.

Thanks for the kind words! It was a blast to work on. 

The new strategies for the addon are all enabled by default:

For feral (bear and cat)
co -feral charge 

For resto 
co -blanketing
co -tranquility

src/Ai/Class/Druid/Trigger/DruidTriggers.h
```diff
@@ -55,22 +55,30 @@ class ThornsTrigger : public BuffTrigger
...
+public:
+    NaturesSwiftnessActiveTrigger(PlayerbotAI* botAI) : HasAuraTrigger(botAI, "nature's swiftness") {}
+    bool IsActive() override { return botAI->HasAura("nature's swiftness", bot); }
 };
 
 class RakeTrigger : public DebuffTrigger
 {
 public:
     RakeTrigger(PlayerbotAI* botAI) : DebuffTrigger(botAI, "rake", 1, true) {}
+    bool IsActive() override { return !botAI->HasAura("prowl", bot) && BuffTrigger::IsActive(); }
```
kadeshar: Here shouldnt be DebuffTrigger::IsActive?

src/Ai/Class/Druid/DruidAiObjectContext.cpp
```diff
@@ -62,26 +61,24 @@ class DruidDruidStrategyFactoryInternal : public NamedObjectContext<Strategy>
         creators["bear"] = &DruidDruidStrategyFactoryInternal::bear;
         creators["tank"] = &DruidDruidStrategyFactoryInternal::bear;
         creators["cat"] = &DruidDruidStrategyFactoryInternal::cat;
-        creators["caster"] = &DruidDruidStrategyFactoryInternal::caster;
+        creators["balance"] = &DruidDruidStrategyFactoryInternal::balance;
+        creators["caster"] = &DruidDruidStrategyFactoryInternal::balance;
```
kadeshar: Its correct? Because caster is default strategy for resto druid

src/Ai/Class/Druid/Strategy/RestoDruidStrategy.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include "RestoDruidStrategy.h"
+
+#include "Playerbots.h"
+
+class RestoDruidStrategyActionNodeFactory : public NamedObjectFactory<ActionNode>
+{
+public:
+    RestoDruidStrategyActionNodeFactory() {
+        creators["nourish on party"] = &nourtish_on_party;
```
kadeshar: nourish_on_party

src/Ai/Base/Actions/GenericActions.cpp
```diff
@@ -49,6 +49,12 @@ bool MeleeAction::isUseful()
     if (botAI->IsInVehicle() && !botAI->IsInVehicle(false, false, true))
         return false;
 
+    // Do not start autoattack while prowled — let opener spells break stealth intentionally.
+    // Future rogue stealth implementation should use this instead:
+    // if (botAI->HasAura("stealth", bot) || botAI->HasAura("prowl", bot))
+    if (botAI->HasAura("prowl", bot))
+        return false;
+
     return true;
```
kadeshar: return !botAI->HasAura("prowl", bot);

src/Ai/Class/Druid/Strategy/GenericDruidStrategy.cpp
```diff
@@ -107,52 +126,123 @@ void GenericDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
...
+        triggers.push_back(new TriggerNode("hurricane channel check", { NextAction("cancel channel", 22.0f) }));
+        triggers.push_back(new TriggerNode("medium aoe", { NextAction("hurricane", 23.0f) }));
+        triggers.push_back(new TriggerNode("insect swarm on attacker", { NextAction("insect swarm on attacker", 5.2f) }));
+        triggers.push_back(new TriggerNode("moonfire on attacker", { NextAction("moonfire on attacker", 5.1f) }));
+    }
+
+    if (tab == DRUID_TAB_FERAL)
+    {
+        triggers.push_back(new TriggerNode("clearcasting and medium aoe", { NextAction("swipe (cat)", 25.5f) }));
+        triggers.push_back(new TriggerNode("medium aoe", { NextAction("swipe (cat)", 25.0f) }));
```
kadeshar, line 259: That shouldnt check druid form?

src/Ai/Class/Druid/Action/DruidActions.h
```diff
@@ -193,12 +198,11 @@ class CastEntanglingRootsAction : public CastSpellAction
     CastEntanglingRootsAction(PlayerbotAI* botAI) : CastSpellAction(botAI, "entangling roots") {}
 };
 
-class CastEntanglingRootsCcAction : public CastSpellAction
+class CastEntanglingRootsCcAction : public CastCrowdControlSpellAction
 {
 public:
-    CastEntanglingRootsCcAction(PlayerbotAI* botAI) : CastSpellAction(botAI, "entangling roots on cc") {}
+    CastEntanglingRootsCcAction(PlayerbotAI* botAI) : CastCrowdControlSpellAction(botAI, "entangling roots") {}
     Value<Unit*>* GetTargetValue() override;
```
kadeshar, line 205: Not better use default implementation from CastCrowdControlSpellAction? I think we should have that unified to all cc spell behavior in same way in term of target selection

src/Ai/Class/Druid/Action/DruidCatActions.h
```diff
@@ -48,28 +56,50 @@ class CastRakeAction : public CastDebuffSpellAction
...
 class CastClawAction : public CastMeleeSpellAction
 {
 public:
     CastClawAction(PlayerbotAI* botAI) : CastMeleeSpellAction(botAI, "claw") {}
+
+    bool isUseful() override
+    {
+        if (botAI->HasAura("prowl", bot))
+        {
+            if (bot->HasSpell(9005)) // Pounce Rank 1 — learned at level 36
```
kadeshar: merge if's

src/Ai/Class/Druid/Strategy/FeralDruidStrategy.cpp
```diff
@@ -99,15 +82,23 @@ void FeralDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
...
-    triggers.push_back(
-        new TriggerNode("berserk", { NextAction("berserk", ACTION_HIGH + 6) }));
+                                       { NextAction("dash", 92.0f) }));
+}
+
+void FeralChargeDruidStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
+{
+    Player* bot = botAI->GetBot();
+
+    if (bot->HasSpell(768) /*cat form*/ && !bot->HasAura(16931) /*thick hide*/)
```
kadeshar: Magic numbers we mostly resolve via const not comment

src/Ai/Class/Druid/Trigger/DruidTriggers.cpp
```diff
@@ -22,6 +24,47 @@ bool ThornsOnPartyTrigger::IsActive()
...
+    Unit* rtiCcTarget = context->GetValue<Unit*>("rti cc target")->Get();
+    if (!rtiCcTarget)
+        return false;
+
+    Unit* ccTarget = context->GetValue<Unit*>("cc target", "cyclone")->Get();
+    if (ccTarget && ccTarget == rtiCcTarget)
+        return HasCcTargetTrigger::IsActive();
+
+    return botAI->CanCastSpell("cyclone", rtiCcTarget);
+}
```
kadeshar: 3 times almost the same. Can we create base action or helper on that?

src/Ai/Class/Druid/Trigger/DruidTriggers.h
ThePenguinMan96: Good catch. Testing with the new debuff trigger instead of buff trigger, it respects time-to-life checks now. preventing it from using rake on a 1%hp mob. Seems good to me! It will be in the next commit.

src/Ai/Class/Druid/Strategy/RestoDruidStrategy.cpp
ThePenguinMan96: typo fixed in next commit. Thank you

src/Ai/Base/Actions/GenericActions.cpp
ThePenguinMan96: This will be what it is in the next commit:

return !botAI->HasAura("prowl", bot);

thank you

src/Ai/Class/Druid/Strategy/FeralDruidStrategy.cpp
ThePenguinMan96: Okay, in next commit these will be defined:

FeralDruidStrategy.h — SPELL_CAT_FORM = 768, AURA_THICK_HIDE = 16931
DruidTriggers.h — AURA_OMEN_OF_CLARITY = 16864
DruidCatActions.h — SPELL_POUNCE_RANK_1 = 9005, SPELL_RAVAGE_RANK_1 = 6785

I just copied the logic from aifactory, but thank you for the suggestion. I will try and do this going forward

brighton-chi: Gromleq is speaking aspirationally and not about what this codebase actually says =P

src/Ai/Class/Druid/Strategy/GenericDruidStrategy.cpp
ThePenguinMan96: Are you suggesting a rebirth strategy, and the only spec that doesn't have it by default is bears? And cats get the nature's swiftness logic? I'd be down to make that.

src/Ai/Class/Druid/Action/DruidCatActions.h
ThePenguinMan96: newest commit will have merged if's:

    // Block Claw once Pounce is learned; Claw remains available as the stealth opener before then.
    if (botAI->HasAura("prowl", bot) && bot->HasSpell(SPELL_POUNCE_RANK_1))
        return false;

src/Ai/Class/Druid/Strategy/BearDruidStrategy.cpp
ThePenguinMan96: Nice. newest commit will have removed the action nodes for Melee and Feral Charge (bear). Any others you'd recommend?

src/Ai/Class/Druid/Strategy/BalanceDruidStrategy.cpp
ThePenguinMan96: I mean, it could be useful. Boomkins need the innervate until about 40% crit chance, and then they rarely run out of mana. That being said, I am not sure if I would have it as default behavior... For most of the game, might as well call them oomkin.

src/Ai/Class/Druid/Strategy/RestoDruidStrategy.cpp
ThePenguinMan96: Talking about using innervate on other healers? That could be an idea. Innervate could also be a strategy, like how warlocks have soulstone strategies - innervate on healer, innervate on self, innervate on dps (with mana). But that is diving into the rabbit hole. I feel like most times they want to use it on themselves.

src/Ai/Class/Druid/Action/DruidActions.cpp
ThePenguinMan96: next commit will have 1 pass, and remember the members by role. nice optimization!

src/Ai/Class/Druid/Trigger/DruidTriggers.cpp
ThePenguinMan96: this is the current logic for fear and banish, so I will make a base trigger (RTICCTrigger) all 5 will inherit from in rtitriggers:

RtiTriggers.h/.cpp — RtiCcTrigger base class added here for any spec to reuse
DruidTriggers.h — 3 triggers swapped to RtiCcTrigger, IsActive() declarations removed
DruidTriggers.cpp — 3 IsActive() implementations deleted (~40 lines)
WarlockTriggers.h — 2 triggers swapped to RtiCcTrigger, IsActive() declarations removed
WarlockTriggers.cpp — 2 IsActive() implementations deleted (~14 lines)

Newest commit will have these changes. Thanks for the input

src/Ai/Class/Druid/Strategy/BearDruidStrategy.cpp
ThePenguinMan96: Basically, it's pretty low priority against 1 target, and a little bit better with 2, but the attack power reduction is noticable at 3 or 4 mobs (higher the better). It reduces AP roughly by 9%, and it's noticeable when fighting a lot of units at the same time. Currently it starts kicking in at medium aoe - if there is a mob nearby without the debuff and there are 3+ targets, it will use that before spamming swipe. I tried this at 2 mobs, but in practice as the bot pulled more consecutive mobs I felt like 2 was too early. 3 felt right, 4 wouldn't trigger on packs of 3, which this game has a lot of.

src/Ai/Class/Druid/Strategy/CatDruidStrategy.cpp
ThePenguinMan96: Your will is my command. next commit has this:

- Added LowEnergyTrigger(PlayerbotAI* botAI, uint8 threshold = 30) to GenericTriggers.h/.cpp as a generic reusable trigger, following the same pattern as LowManaTrigger
- Registered as "low energy" in TriggerContext.h
- TigersFuryTrigger deleted entirely — the energy check is self-sufficient since Tiger's Fury instantly restores 60 energy, making the old BuffTrigger::IsActive() check redundant
- Strategy trigger node updated from "tiger's fury" → "low energy", with tiger's fury remaining as the action

src/Ai/Class/Druid/Strategy/GenericDruidStrategy.cpp
ThePenguinMan96: Not yet, but next commit will allow them at ~34.0f priority. It will not be enabled by default, but if enabled, they will shift out of form as needed to CC the RTI target.

src/Ai/Class/Druid/Strategy/GenericDruidStrategy.cpp
ThePenguinMan96: Yes, and the newest commit will reflect this. Originally, AoE is handled for bears in their own strategy. But I forgot that people use co +aoe macros, and if they did that raidwide, the bear would adopt this strategy, switching out of cat form. Good catch, thank you.

src/Ai/Class/Druid/Action/DruidActions.h
ThePenguinMan96: Originally it was using CastSpellAction and the value "cc target" but CC target is not the rti target - it is a calculated value. I think that CC spells should use the CastCrowdControlSpellAction as the base class, and if the intent is to CC the RTI target, they should use the "rti cc target" value as seen below in druidactions.cpp:

Value<Unit*>* CastCycloneCcAction::GetTargetValue() { return context->GetValue<Unit*>("rti cc target"); }

I have found this to be the most consistent way for bots to CC the rti target ONLY - and not a secondary target that isn't marked. See explanation by my AI:

---

- The GetTargetValue() override returning "rti cc target" is intentional — we only want the druid to CC the RTI moon-marked target, not opportunistically CC other targets
- The default "cc target" value has a fallback that selects non-RTI targets based on health, AoE position, and tank distance — that fallback behavior is undesirable here
- Since the triggers are already RtiCcTrigger-based, the override keeps the action and trigger consistent: both only fire when there is an explicit RTI CC target

Look at how "cc target" is calculated:

```
Unit* CcTargetValue::Calculate()
{
    FindTargetForCcStrategy strategy(botAI, qualifier);
    return FindTarget(&strategy);
}
```
Which runs FindTargetForCcStrategy::CheckAttacker() against all attackers:

```
void CheckAttacker(Unit* creature, ThreatManager* /*threatMgr*/) override
{
    if (!botAI->CanCastSpell(spell, creature))
        return;

    // RTI target gets immediate priority
    if (*botAI->GetAiObjectContext()->GetValue<Unit*>("rti cc target") == creature)
    {
        result = creature;
        return;
    }

    // Skip current target
    if (*botAI->GetAiObjectContext()->GetValue<Unit*>("current target") == creature)
        return;

    // Skip low health targets
    if (health < sPlayerbotAIConfig.mediumHealth)
        return;

    // Skip targets in AoE cluster
    // Skip targets being tanked
    // Otherwise select by max distance from tanks
}
```

"cc target" does prioritize the RTI moon target first — **but if no RTI target is in the attacker list, it falls through to opportunistically select another target.** That fallback is what you don't want. The "rti cc target" override correctly restricts it to moon-only. 

...versus how "rti cc target" is calculated:

src/Ai/Base/Value/RtiValue.cpp:

`RtiCcValue::RtiCcValue(PlayerbotAI* botAI) : ManualSetValue<std::string>(botAI, "moon", "rti cc") {}`

"rti cc" defaults to "moon". That feeds into RtiTargetValue::Calculate():

```
Unit* RtiTargetValue::Calculate()
{
    std::string const rti = AI_VALUE(std::string, type); // "moon"
    int32 index = GetRtiIndex(rti);                      // index 4

    ObjectGuid guid = group->GetTargetIcon(index);       // moon-marked unit's GUID
    if (!guid)
        return nullptr;

    Unit* unit = botAI->GetUnit(guid);
    if (!unit || unit->isDead() || !bot->IsWithinLOSInMap(unit) || ...)
        return nullptr;

    return unit;
}
```

"rti cc target" is a direct GUID lookup of whatever unit has the moon icon — completely different from "cc target" which iterates all attackers. 



---

I eventually do want to create separate CC options - one that is strictly for "cc rti" that will only cc the rti marked target, and one that is "cc auto" that will use the "cc target" value that it defaults as.  This is the reason that the mage's polymorph was hitting targets that weren't marked - I believe it was using the cc target value instead of the "rti cc target" value. 

But, as it stands, I do want the druid to ONLY cc the RTI target (default moon). **So I don't think these should be changed.**

src/Ai/Class/Druid/Strategy/GenericDruidNonCombatStrategy.cpp
ThePenguinMan96: next commit will reflect the removed comment:

// creators["innervate"] = &innervate;

will be gone.

src/Ai/Class/Druid/Strategy/GenericDruidNonCombatStrategy.cpp
ThePenguinMan96: Next commit will reflect:

// Before:
if (specTab == 0 || specTab == 2) // Balance or Restoration
if (specTab == 1) // Feral

// After:
if (specTab == DRUID_TAB_BALANCE || specTab == DRUID_TAB_RESTORATION)
if (specTab == DRUID_TAB_FERAL)

src/Ai/Class/Druid/Trigger/DruidTriggers.cpp
ThePenguinMan96: removing Player* bot = botAI->GetBot(); in the next commit

src/Ai/Class/Druid/DruidAiObjectContext.cpp
ThePenguinMan96: Newest commit will remove caster entirely. It was an artefact from before. explanation:

"caster" was a leftover alias from before the Balance rename. Adding it to solo Resto druids was silently ejecting the "heal" strategy via the sibling-exclusion mechanism in DruidDruidStrategyFactoryInternal — only one strategy from that factory can be active at a time. Fixed by removing "caster" from the solo block (and the alias itself); healer dps already covers solo DPS, and aoe is safe to add since it lives in the non-exclusive factory.

ThePenguinMan96:
@kadeshar 
@brighton-chi 

Worked on all of the suggestions in the latest commit. here are the changes:

1 — RakeTrigger: BuffTrigger::IsActive() → DebuffTrigger::IsActive() (DruidTriggers.h)
2 — Typo: nourtish_on_party → nourish_on_party (RestoDruidStrategy.cpp)
3 — MeleeAction prowl: simplified to return !botAI->HasAura("prowl", bot) (GenericActions.cpp)
5 — Prowl opener nested-if flattened to && (DruidCatActions.h)
6 — Magic numbers → AURA_OMEN_OF_CLARITY, SPELL_POUNCE_RANK_1, SPELL_RAVAGE_RANK_1, SPELL_CAT_FORM, AURA_THICK_HIDE (DruidTriggers.h, DruidCatActions.h, FeralDruidStrategy.h)
7 — RtiCcTrigger base class added; Druid CC triggers + Warlock Banish/Fear inherit it — duplicate IsActive() overrides removed (RtiTriggers.h/.cpp, DruidTriggers.h/.cpp, WarlockTriggers.h/.cpp)
8 — Empty melee and faerie fire (feral) ActionNodes removed (BearDruidStrategy.cpp)
10 — TigersFuryTrigger deleted; generic LowEnergyTrigger(threshold = 30) added, registered as "low energy" (GenericTriggers.h/.cpp, TriggerContext.h, CatDruidStrategy.cpp, DruidTriggers.h/.cpp, DruidAiObjectContext.cpp)
12 — Blanketing: three group passes → single pass with tank/melee/ranged priority buckets (DruidActions.cpp)
13 — Starfall: "nearest hostile npcs" → "possible targets" (cheaper grid search, LoS-aware) (DruidActions.cpp)
16 — Swipe (Cat) AoE guard: requires SPELL_CAT_FORM and not AURA_THICK_HIDE (GenericDruidStrategy.cpp)
17 — Solo Resto "caster" removed — was silently ejecting "heal" via sibling-exclusion mechanism; alias removed (AiFactory.cpp, DruidAiObjectContext.cpp)
18 — Bear CC split from cat CC: standard Cyclone/Hibernate/Entangling Roots at 42/41/40 when co +cc enabled (GenericDruidStrategy.cpp)
19 — Redundant Player* bot = botAI->GetBot() removed in HurricaneChannelCheckTrigger::IsActive() (DruidTriggers.cpp)
20 — Commented-out innervate removed from non-combat strategy (GenericDruidNonCombatStrategy.cpp)
21 — Spec tab magic numbers → DRUID_TAB_BALANCE/FERAL/RESTORATION (GenericDruidNonCombatStrategy.cpp)
CC priority raised to 42/41/40 across all specs (was 24/23/22 for Balance/Resto, 34/33/32 for Feral) (GenericDruidStrategy.cpp)

src/Ai/Class/Druid/Action/DruidActions.cpp
```diff
@@ -275,34 +274,30 @@ Unit* CastBlanketHotAction::GetBlanketTarget(std::string const& auraName)
...
 
-    // Pass 3: ranged
-    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
-    {
-        Player* member = ref->GetSource();
-        if (!eligible(member) || !PlayerbotAI::IsRanged(member))
-            continue;
-        return member;
+        if (!firstTank && PlayerbotAI::IsTank(member))
+            firstTank = member;
```
kadeshar: Instead of assign that should simply return firstTank?

ThePenguinMan96: good point - modified it for GetBlanketTarget to return immediately on finding an eligible tank, removing the firstTank bucket on the newest commit I just made now.

src/Ai/Class/Druid/Strategy/BearDruidStrategy.cpp
brighton-chi: I must have put the comment on a bad spot. What I mean is if an ActionNode has no prerequisites, alternatives, or continuers, then you don't need to have an ActionNode for it at all. For example, swipe (bear) can be removed (the ActionNode itself and the registration under the ActionNodeFactory). But melee and faerie fire (feral) should actually not be removed. Sorry for the confusion.

ThePenguinMan96: Well, melee and faerie fire (feral) used to be linked to Feral Charge - Bear, but now that has it's own strategy. Both melee and faerie fire (feral) had no P/A/C's. Do you think they are good to stay removed? Also, do you want me to remove all others without P/A/C?

ThePenguinMan96: That is good to know though - I had a suspicion, since other classes had so few nodes. I was wondering how the hell they were working. Now I know!

Celandriel:
Reviewer tool flagged the following. They dont need to be resolved, just acknowledged and if you find it meaningful implement something. 

```
Warlock CC regression from the shared RtiCcTrigger (RtiTriggers.cpp:25-36). The new base added a CanCastSpell(rtiCcTarget) fallback the old Banish/Fear triggers didn't have, and Warlock's actions still target "cc target" while the trigger now keys on "rti cc target". Trigger and action can disagree. Druid actions were updated to "rti cc target" and are self-consistent; Warlock was left behind. Either keep Warlock's strict cc target == rti cc target semantics (drop the fallback) or move Warlock actions to "rti cc target" too (a behavior change needing retest).

Barkskin retargeted to a healthy band at interrupt priority (GenericDruidStrategy.cpp:130) — affects all druids. It now fires on almost full health (~50–85% HP) at 40.0f, so a defensive cooldown is wasted while healthy and preempts real interrupts (bash at 41–42). Should trigger on low health and live in the high/emergency band.

Resto Nature's Swiftness combo at 55–56 (RestoDruidStrategy.cpp) sits above dispel (50) and mount (54), so a single-target clutch heal preempts cleanses/CC. Wiring is correct; the band is wrong — put it in the critical-heal band or the emergency band intentionally.

Typhoon queued at 40.0f (GenericDruidStrategy.cpp:243) collides with the interrupt band but it's a proximity AoE, not an interrupt. Drop to the 20–29 high band.

Starfall scans at ~100yd sight range then discards beyond 40yd (DruidActions.cpp:170-184) — ~6× more grid work than needed, and it reads "possible targets" (LOS-filtered, level-gated) while the comment/reviewer both say "nearest hostile npcs". Fetch a 40yd value (or the cheaper "nearest hostile npcs") and align comment with code. Also: LOS-blocked mobs within 40yd currently over-suppress Starfall — a deliberate decision, not a bug.
```

Celandriel:
Merge conflict as well.

src/Ai/Class/Druid/Strategy/BearDruidStrategy.cpp
brighton-chi: Ah ok, I was just looking at what existed before this PR. But yeah, Keleborn and I talked about this a while back, think we should remove all the empty ones. Saves a good amount of space I think. The engine just creates an empty node if you don't have anything so there's no need.

ThePenguinMan96: Roger that. I'll go through the druid strategy files and remove all action node factories without P/A/C that I see. If I miss any, feel free and correct or let me know. Next commit will reflect changes.

ThePenguinMan96:
> Reviewer tool flagged the following. They dont need to be resolved, just acknowledged and if you find it meaningful implement something.
> 
> ```
> Warlock CC regression from the shared RtiCcTrigger (RtiTriggers.cpp:25-36). The new base added a CanCastSpell(rtiCcTarget) fallback the old Banish/Fear triggers didn't have, and Warlock's actions still target "cc target" while the trigger now keys on "rti cc target". Trigger and action can disagree. Druid actions were updated to "rti cc target" and are self-consistent; Warlock was left behind. Either keep Warlock's strict cc target == rti cc target semantics (drop the fallback) or move Warlock actions to "rti cc target" too (a behavior change needing retest).
> 
> Barkskin retargeted to a healthy band at interrupt priority (GenericDruidStrategy.cpp:130) — affects all druids. It now fires on almost full health (~50–85% HP) at 40.0f, so a defensive cooldown is wasted while healthy and preempts real interrupts (bash at 41–42). Should trigger on low health and live in the high/emergency band.
> 
> Resto Nature's Swiftness combo at 55–56 (RestoDruidStrategy.cpp) sits above dispel (50) and mount (54), so a single-target clutch heal preempts cleanses/CC. Wiring is correct; the band is wrong — put it in the critical-heal band or the emergency band intentionally.
> 
> Typhoon queued at 40.0f (GenericDruidStrategy.cpp:243) collides with the interrupt band but it's a proximity AoE, not an interrupt. Drop to the 20–29 high band.
> 
> Starfall scans at ~100yd sight range then discards beyond 40yd (DruidActions.cpp:170-184) — ~6× more grid work than needed, and it reads "possible targets" (LOS-filtered, level-gated) while the comment/reviewer both say "nearest hostile npcs". Fetch a 40yd value (or the cheaper "nearest hostile npcs") and align comment with code. Also: LOS-blocked mobs within 40yd currently over-suppress Starfall — a deliberate decision, not a bug.
> ```

1. Good catch - I tested it, and it worked fine, but "cc target" is not what we want. We want "rti cc target". I should have tested longer than 15 minutes on the warlock change. Newest commit will reflect changes that those warlockactions will target rti cc target.

2. Barkskin's priority was too low for casters - it has 20% damage reduction, but the 2nd part where it prevents spellcasting pushback was the main reason I moved it to high hp - so the caster would use it if it was getting hurt, and protect their spellcasting, resulting in better dps/hps. 

3. Good catch. I literally just picked 55/56 because it seemed high, but it should be above removing curse/poison. saving someone from dying with a instant cast heal is nearly always more important. newest commit will move it to 58, above remove curse (57). 

4. While typhoon is in the aoe strategy, that is simply because it hits multiple targets - it is a get-off-me tool. it was originally tested with lower priority (25) but I found that the druid would not cast it above other certain spells and enemies would keep getting in melee range and attacking a few times before it casted typhoon. With the higher priority, it casts it nearly immediately when they walk in melee range.

5. I think the AI is suggesting to go back to "nearest hostile mobs". But I believe you said that was more expensive than "possible targets no los", which is "possible targets" cousin. Do you recommend I change it back?

Thank you for the review and feedback :)

Celandriel:
>     5. I think the AI is suggesting to go back to "nearest hostile mobs". But I believe you said that was more expensive than "possible targets no los", which is "possible targets" cousin. Do you recommend I change it back?

I think Crow is the expert on that. I dont have a good sense of which is best.

ThePenguinMan96:
> > ```
> > 5. I think the AI is suggesting to go back to "nearest hostile mobs". But I believe you said that was more expensive than "possible targets no los", which is "possible targets" cousin. Do you recommend I change it back?
> > ```
> 
> I think Crow is the expert on that. I dont have a good sense of which is best.

Oh, gotcha. I am fine with either - possible targets or nearest hostile npc. He recommended against nearest hostile NPC because it was an expensive scan. I am going to test possible targets some more.

Also, just resolved merge conflicts.

brighton-chi:
Point the AI to the grid search files in AC and ask it to reevaluate what is the more expensive check. It is probably assuming based on the names alone what is more or less expensive. Nearest hostile npcs is not a pre-search filter. It does nearest npcs (which looks at everything, other players, shopkeepers, even critters) and then does a faction filter afterward. It's like the most taxing search you can do.

Possible targets does a more limited search so you don't pick up shit like critters or triggers. Possible targets no los is a lot cheaper than just possible targets because determining whether you have los is expensive. I don't know what is best here but I know that nearest npcs and nearest hostile npcs are overused in this codebase and they have very high performance cost.

ThePenguinMan96:
> Point the AI to the grid search files in AC and ask it to reevaluate what is the more expensive check. It is probably assuming based on the names alone what is more or less expensive. Nearest hostile npcs is not a pre-search filter. It does nearest npcs (which looks at everything, other players, shopkeepers, even critters) and then does a faction filter afterward. It's like the most taxing search you can do.
> 
> Possible targets does a more limited search so you don't pick up shit like critters or triggers. Possible targets no los is a lot cheaper than just possible targets because determining whether you have los is expensive. I don't know what is best here but I know that nearest npcs and nearest hostile npcs are overused in this codebase and they have very high performance cost.

I recall trying possible targets no los, but I noticed that in a dungeon the boomkin fired off starfall around a corner, 2 seconds before it made LoS contact. So I believe the extra cost justifies the means.

brighton-chi:
Yeah I imagine for targeting a particular spell that is probably like the exact use case.

Celandriel (approved):

Wishmaster117:
@Celandriel  Wiki & Multibot Need update, can you tag please?

PR #2405, opened 2026-05-19
Use Stones for Prot Paladins & Reduce Oils

brighton-chi:
## Pull Request Description

Back when I did the Stones/Oils PR, I gave Wizard Oil to Prot Paladins. I was operating based on their mechanics in TBC and have since realized that their damage is much more physical-based in WotLK as opposed to spell-based in TBC. So I've changed them to now get stones upon maintenance and apply stones.

I also reduced weapon oils given during maintenance from 4 to 2 stacks. I did 4 originally because oils stack to 5 so it would align with the amount given of stones and poisons (which stack to 20). But 4 inventory slots taken up by oils is excessive, and 2 still last far longer than buffing reagents in a raid context. So this is just to reduce inventory clutter.

Other stuff is just little code stuff like reformatting or reordering things to be more efficient.

## Feature Evaluation

There's nothing new--essentially with respect to consumables, prot paladins now are lumped with ret.

## How to Test the Changes

Create a prot paladin, turn on selfbot. Use maintenance and the paladin should be given stones and no oils. The paladin should then apply the stone to its weapon (if it is equipping a level-eligible weapon).

Create a caster, turn on selfbot. Use maintenance and they should be given 2 stacks of oils.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Prot paladins will now use stones in their inventory and will not use oils in their inventories. This is better for their mechanics.

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/src/Bot/Factory/PlayerbotFactory.cpp
+++ b/src/Bot/Factory/PlayerbotFactory.cpp
@@ -850,36 +850,36 @@ void PlayerbotFactory::Refresh()
 
 void PlayerbotFactory::InitConsumables()
 {
-    int specTab = AiFactory::GetPlayerSpecTab(bot);
+    uint8 specTab = AiFactory::GetPlayerSpecTab(bot);
     std::vector<std::pair<uint32, uint32>> items;
 
     switch (bot->getClass())
     {
         case CLASS_PRIEST:
         {
-            // Discipline or Holy: Mana Oil
-            if (specTab == 0 || specTab == 1)
+            if (specTab == PRIEST_TAB_SHADOW)
             {
-                std::vector<uint32> mana_oils = {BRILLIANT_MANA_OIL, SUPERIOR_MANA_OIL, LESSER_MANA_OIL, MINOR_MANA_OIL};
-                for (uint32 itemId : mana_oils)
-                {
-                    ItemTemplate const* proto = sObjectMgr->GetItemTemplate(itemId);
-                    if (proto->RequiredLevel > level || level > 75)
-                        continue;
-                    items.push_back({itemId, 4});
-                    break;
-                }
-            }
-            // Shadow: Wizard Oil
-            if (specTab == 2)
-            {
-                std::vector<uint32> wizard_oils = {BRILLIANT_WIZARD_OIL, SUPERIOR_WIZARD_OIL, WIZARD_OIL, LESSER_WIZARD_OIL, MINOR_WIZARD_OIL};
+                std::vector<uint32> wizard_oils = {
+                    BRILLIANT_WIZARD_OIL, SUPERIOR_WIZARD_OIL, WIZARD_OIL, LESSER_WIZARD_OIL, MINOR_WIZARD_OIL };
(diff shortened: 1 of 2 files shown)


dillyns:
Where are you getting the information that a stone would be better for a prot paladin vs. Wizard Oil. According to stat weights on https://sixtyupgrades.com/wotlk (which may or may not be correct):
Brilliant Wizard Oil > Ele Sharpening Stone > Superior Wizard Oil > Wizard oil = Adamantite Stones

brighton-chi:
> Where are you getting the information that a stone would be better for a prot paladin vs. Wizard Oil. According to stat weights on https://sixtyupgrades.com/wotlk (which may or may not be correct): Brilliant Wizard Oil > Ele Sharpening Stone > Superior Wizard Oil > Wizard oil = Adamantite Stones

https://www.wowhead.com/wotlk/guide/lots-of-knowledge-wotlk-pve-protection-paladin-and-tanking-guide-17306

"The best weapons now are standard tanking weapons, since HotR scales strongly with weapon dps and these weapons also have mitigation/avoidance stats on them (especially handy for reaching and maintaining crit immunity). While a spellpower weapon would boost damage from a number of different abilities, the overall effect would still be weaker than the scaling of a melee weapon on HotR damage. (If there were weapons that had sacrificed dps converted into AP, those would be the best weapons, but those don't exist.)"

brighton-chi:
(That being said, I am going off a statement, which I will not give greater weight to than actual math, and noted that Stat A is more important than Stat B also doesn't mean that Item A that provides Stat A is better than Item B that provides Stat B)

src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -850,75 +850,78 @@ void PlayerbotFactory::Refresh()
...
-                for (uint32 itemId : mana_oils)
+                std::vector<uint32> wizard_oils = {
+                    BRILLIANT_WIZARD_OIL, SUPERIOR_WIZARD_OIL, WIZARD_OIL, LESSER_WIZARD_OIL, MINOR_WIZARD_OIL };
+                for (uint32 itemId : wizard_oils)
                 {
                     ItemTemplate const* proto = sObjectMgr->GetItemTemplate(itemId);
                     if (proto->RequiredLevel > level || level > 75)
                         continue;
-                    items.push_back({itemId, 4});
+                    items.push_back({itemId, 2});
```
kadeshar, line 869: That is not pieces? You mentioned about 2 stacks

brighton-chi: Sorry, it is two items. I was thinking in terms of uses--each item has 5 charges, but the item itself doesn't stack. Basically right now oils eat 4 inventory slots, and that gives them 20 uses like one stack of stones, but I feel like that is too much space so it's better to take up 2 inventory slots for 10 charges, which still is a lot of usage since they do persist through death.

kadeshar (approved):

PR #2409, opened 2026-05-21
Rewrite Equipment-Randomization-Related Configs

brighton-chi:
## Pull Request Description

The main focus on this PR is to clarify and eliminate overlap between three very confusingly described yet important config options: AutoUpgradeEquip, EquipmentPersistence, and IncrementalGearInit.

For context, after initialization, randombots generally get randomized each time after logging in and periodically between MinRandomBotRandomizeTime (configurable, default 2 hours) to MaxRandomBotRandomizeTime (configurable, default 14 days). This randomization happens only when the bot is idle.

Now let’s look at what those three config options currently say.
-	AutoUpgradeEquip: Randombots automatically upgrade their equipment on levelup.
-	EquipmentPersistence: Enable/Disable bot equipment persistence (stop random initialization) after certain level (EquipmentPersistenceLevel).
-	IncrementalGearInit: If disabled, random bots can only upgrade equipment through looting and quest

None of those descriptions are accurate.
-	AutoUpgradeEquip determines if randombots, upon leveling up, refresh ammo, reagents, food, consumables, potions, and, ONLY IF IncrementalGearInit is also enabled and EquipmentPersistence is disabled, upgrade equipment (yes, three config options required for this one thing).
-	EquipmentPersistence affects both equipment and talents. 
-	Disabling IncrementalGearInit does not prevent randombots from changing their equipment through the login/periodic randomization process unless EquipmentPersistence is enabled.

These config options shouldn’t overlap with or be dependent on each other, and their names and descriptions should reflect what they actually do. Thus, this PR does the following:
-	AutoUpgradeEquip solely controls whether or not randombots automatically upgrade their gear upon level up. No other config option is involved for this purpose, and AutoUpgradeEquip no longer impacts inventory items. This does mean that it is no longer possible to stop randombots from being given ammo, potions, etc. when they level up. I tend to think that randombots as they currently are cannot fully function otherwise so I have no issue with the loss of this ability, but if there is disagreement, then we need to introduce a new config option. AutoUpgradeEquip is also now set to true in PlayerbotAIConfig to reflect that the default is true in the .dist.
-	I originally wanted to combine EquipmentPersistence and EquipmentPersistenceLevel into a single config option that also included talents in the name, but EquipmentPersistence is used in the level brackets mod so I don’t want to make a breaking change there. I settled for making EquipmentPersistence enabled by default and also reducing EquipmentPersistenceLevel to 1 by default, in effect entirely disabling periodic/login randomization of randombot talents and gear by default. I only see people complain about these features so I think unless there is some compelling performance or structural reason to the contrary, the default should be that randombots do not continuously randomize their talents and equipment.
- IncrementalGearInit is eliminated. There are two real functional impacts: (1) as noted above, randombots cannot be blocked from receiving standard inventory items upon level-up (though the way it is currently being gated does not make sense anyway), and (2) you can no longer have equipment persistence for equipment only but not talents (I cannot imagine anybody would ever want to do that, and I don't think it was even intended given the strange interaction between config options that was needed to even accomplish that before).
-	For all these settings, the config descriptions are updated to try to be clear about what the player is actually configuring.

Beyond that, I made some clean-up changes to the config to fix some typos and try to shorten it in places, and I also deleted a few config options that do not appear in operative code anymore (and any corresponding references in PlayerbotAIConfig). I also deleted some comments and made some minor style changes in AutoMaintenanceOnLevelupAction.cpp.

## Feature Evaluation

The main substance of the PR consists of changes to checks in PlayerbotFactory and AutoMaintenanceOnLevelupAction.

## How to Test the Changes

- To test AutoUpgradeEquip, use the gm command .level 1 while targeting an rndbot.
- To test EquipmentPersistence, the only way I could find to force an incremental randomization was to use .playerbots rndbot level BotName. This will do an incremental randomization and a level up. However, this level does not go through the AutoMaintenanceOnLevelupAction path so just disregard the +1 level aspect and consider it a pure test of EquipmentPersistence.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

I had GPT-5.4 review my changes and come up with the testing method.

## Final Checklist


conf/playerbots.conf.dist
```diff
@@ -774,19 +741,31 @@ AiPlayerbot.RandomBotFixedLevel = 0
...
+
+# Randombots automatically receive upgraded equipment on levelup
+# Default: 1 (enabled)
+AiPlayerbot.AutoUpgradeEquip = 1
+
+# Enable/disable which (if any) randombots preserve their current equipment and talents during login
+# and periodic randomization (i.e., persistence).
+# Default: 1 (enabled)
+AiPlayerbot.EquipmentPersistence = 1
+
```
ThePenguinMan96, line 792: Big fan of this. I feel like the community has been asking for a way to make their randombots feel a bit more permanent. Even just having the option is nice.

src/Ai/Base/Actions/AutoMaintenanceOnLevelupAction.cpp
```diff
@@ -89,21 +87,17 @@ void AutoMaintenanceOnLevelupAction::LearnQuestSpells(std::ostringstream* out)
...
         Quest const* quest = i->second;
 
-        // only process class-specific quests to learn class-related spells, cuz
-        // we don't want all these bunch of entries to be handled!
-        if (!quest->GetRequiredClasses())
-            continue;
-
-        // skip quests that are repeatable, too low level, or above bots' level
-        if (quest->IsRepeatable() || quest->GetMinLevel() < 10 || quest->GetMinLevel() > bot->GetLevel())
+        if (!quest->GetRequiredClasses() || quest->IsRepeatable() || quest->GetMinLevel() < 10 ||
```
ThePenguinMan96, line 90: Interesting - a GetRequiredClasses check prevents wrong class quests?

Celandriel (dismissed):

brighton-chi:
After discussion with Keleborn, I am changing equipmentPersistence to equipAndSpecPersistence to eliminate confusion going forward. This is going to cause level brackets to fail to compile. I can submit a PR for that mod.

Celandriel (dismissed):

brighton-chi:
https://github.com/DustinHendrickson/mod-player-bot-level-brackets/pull/83

Celandriel (approved):

PR #2413, opened 2026-05-23
Fix: Dismount cleanup 

NoxMax:
## Pull Request Description

Cleans some stale flight flags, and gets rid of the parachute given in forced flight dismount. Refactors some dismount code, and get rid of the unused mountData parameter in CalculateMasterMountSpeed as well

`MOVEMENTFLAG_WATERWALKING` was removed because it might've been added to protect MOVEMENTFLAG_CAN_FLY, but if the bot remounts over water, MOVEMENTFLAG_CAN_FLY would be reapplied anyway. However if it has a function that I missed, we can re-add it.

## Feature Evaluation

Feature is currently at minimum cost.

## How to Test the Changes

1. As GM hover high above Dalaran, but not high enough that you are outside the resting zone.
2. .summon some lvl +70  bots until you get some that are on a flying mount.
3. Watch as they lose their mount after 10 seconds, and be given a parachute as they fall.
4. When they reach the ground, they should be stuck with a parachute forever, and they shouldn't be stuck hovering just above the ground forever.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [ ] No
Checks if bot has `HasFeatherFallAura` to remove any stale flight flags.

## AI Assistance

- - [x] No

## Final Checklist

## Notes for Reviewers

Comment paragraph might be verbose, but worth it for whomever is looking at the code years from now trying to figure out why bots need special treatment with dismounting anyway.

### DIFF
--- a/src/Ai/Base/Actions/CheckMountStateAction.cpp
+++ b/src/Ai/Base/Actions/CheckMountStateAction.cpp
@@ -17,6 +17,7 @@
 #include "SpellAuraEffects.h"
 
 static constexpr uint32 SPELL_COLD_WEATHER_FLYING = 54197;
+static constexpr float PARACHUTE_LAND_THRESHOLD = 15.0f;
 
 // Define the static map / init bool for caching bot preferred mount data globally
 std::unordered_map<uint32, PreferredMountCache> CheckMountStateAction::mountCache;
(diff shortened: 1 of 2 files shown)


NoxMax:
Marking this as ready for review for now. It can be implemented in the code to patch the current issue with parachutes, and later on if there's more extensive work to the flying mechanics then the code here can be modified however needed.

Celandriel (approved):

Vortikai:
I know that waterwalking and such was a big focus on the post movement refactor https://github.com/mod-playerbots/mod-playerbots/pull/2134

NoxMax:
I did check if it has an effect there. Now as before, they can't mount up in water without water walking, they can mount up with it. 

Players can ground mount in water without water walking (doesn't affect their swim speed), bots can't. This hasn't changed.

They have an odd relationship with water surface, if they dismount above water they fall through it at first as if falling through air, then remember to swim back to the surface. This likely depends on on client data they have no access to. Regardless, nothing changed as far as this either.

PR #2415, opened 2026-05-23
Autogear suffixes

### LINKED ISSUE #2370: Autogearing should utilize items with suffixes
Currently, autogear for bots doesn't fully utilize random gear suffixes "e.g. of the Monkey"

Many years back, I wrote a custom command for TrinityCore that auto-gives the player the "best possible gear at current level". A lot of gear while leveling up utilizes random suffixes, and I had to account for:
1. Their stats
2. Assigning the best suffix based on class+spec

Here's what I did, see autobis_misc.cpp:

https://github.com/YunataSavior/TC_autobis

n/a. When leveling myself numerous times on my own pserver with autobis, I'd always be using a gear piece with a random suffix

**Additional context**
Feel free to reach out to me on Discord if you have any questions.

Noscopezz:
## Pull Request Description
Suffix scoring was already wired up in playerbots, bots could evaluate suffix items they owned, but factory init scored candidates with randomPropertyId = 0 and equipped via EquipNewItem which doesn't roll suffixes, so suffix items always looked like junk and never got picked.

Added a cached item_enchantment_template pool in RandomItemMgr, a PickBestRandomPropertyId helper that picks the best suffix per bot class/spec, and added it on the equipped item via SetItemRandomProperties. 

Tested levels 20-80, suffixes match spec. Closes #2370.

## Feature Evaluation

For items with RandomProperty != 0 or RandomSuffix != 0, run the item_enchantment_template pool once, score each candidate suffix against the bot's existing class/spec stat weights, keep the highest, and stamp that id on the item right after EquipNewItem. Items without a suffix pool skip the helper entirely. Without this, every suffix template scores as base stats only (often near zero) and never gets picked during factory init.

Startup: one SELECT entry, ench FROM item_enchantment_template (few hundred rows), parsed into an unordered_map<uint32, vector<uint32>> once. No SQL afterwards.
Per bot during init: PickBestRandomPropertyId only runs on candidates that actually have a suffix pool. Each call is a hash lookup plus a loop over 5-15 enchantment ids, each doing DBC lookups already used by the scoring code. Cost is mostly by the existing CalculateRandomProperty work, not new logic.
Scales linearly with bot count, same as the rest of factory init. No new per-tick work.

## How to Test the Changes

1. Generate a fresh random bot at level 20-79 and use autogear (cloth, leather, mail, plate, lots of suffix gear at that range).
2. Inspect the bot's gear. Items with names like "of the Eagle", "of the Monkey", "of Healing", etc... should be equipped, with the suffix stats visible on the tooltip.
3. Generate a level 80 bot at high gear-score limit. Confirm it still equips raid epics normally (epics have no suffix pool, so this path is untouched).

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)
Bots now consider random-suffix items during factory init and equip them with the best suffix for their class/spec. Before, suffix items were effectively invisible to init autogear because they scored as base stats only. This is the intended fix for #2370.

    - - [ ] No
Added one new public method on StatsWeightCalculator (PickBestRandomPropertyId) and one new cache in RandomItemMgr (LoadEnchantmentPool + GetEnchantmentPool). The candidate list in InitEquipment changed from vector<uint32> to vector<pair<uint32, int32>> to carry the chosen suffix id alongside the item id. Logic mirrors existing patterns in the same files (DBC lookups, SQL-backed caches, signed randomPropertyId encoding), no new abstractions, no new wrappers.

## AI Assistance

- - [ ] No
Used AI to help me find the bug. It mapped the existing scoring pipeline and pointed out that PlayerbotFactory::InitEquipment was scoring candidates with randomPropertyId = 0 and equipping via EquipNewItem (which doesn't roll suffixes), so the scoring side was already built, just never fed real suffix ids during init. I reviewed and tested all the code in-game across levels 20-80 and multiple classes/specs.

## Final Checklist


src/Mgr/Item/StatsWeightCalculator.cpp
```diff
@@ -190,6 +191,52 @@ void StatsWeightCalculator::CalculateRandomProperty(int32 randomPropertyId, uint
...
+        isSuffix = true;
+    }
+    if (!poolEntry)
+        return 0;
+
+    std::vector<uint32> const& pool = sRandomItemMgr.GetEnchantmentPool(poolEntry);
+    if (pool.empty())
+        return 0;
+
+    GenerateWeights(player_);
```
Celandriel, line 215: Claude says that you need to add a Reset() before Generate weights.

src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -2304,7 +2311,11 @@ void PlayerbotFactory::InitEquipment(bool incremental, bool second_chance)
         if (oldItem)
             continue;
 
-        bot->EquipNewItem(dest, bestItemForSlot, true);
+        if (Item* equipped = bot->EquipNewItem(dest, bestItemForSlot, true))
```
Celandriel, line 2314: from AI review
SetItemRandomProperties sets ITEM_FIELD_RANDOM_PROPERTIES_ID / UpdateItemSuffixFactor() and writes the enchantment slots via SetEnchantment(...) — but SetEnchantment only updates the item's data fields and marks it ITEM_CHANGED; it does not call Player::ApplyEnchantment. Since this runs after EquipNewItem (which already ran _ApplyItemMods with randomPropertyId = 0), and InitEquipment does no re-apply afterward, the suffix bonuses are written to the item but never applied to the bot's live stats.

Result: the item shows the suffix on the tooltip (client reads the item fields) but the bot's combat stats don't include it until a full item-mods reapplication — i.e. next relog/reload. Random bots self-heal on their next login cycle; persistent addclass/alt bots won't until relog. Worth re-verifying by inspecting the bot's actual stats rather than the tooltip.

Re-applying the item's mods after stamping fixes it:

```
if (Item* equipped = bot->EquipNewItem(dest, bestItemForSlot, true))
{
    if (bestRandomPropForSlot != 0)
    {
        bot->_ApplyItemMods(equipped, slot, false);
        equipped->SetItemRandomProperties(bestRandomPropForSlot);
        bot->_ApplyItemMods(equipped, slot, true);
    }
}
```

Noscopezz: Tested it and it works, it makes sense since suffix bonuses are now applied to live stats at init, not only after relog

src/Mgr/Item/StatsWeightCalculator.cpp
Noscopezz: Reset() zeroes stats_weights_[], this is correct

Celandriel (approved):

PR #2419, opened 2026-05-25
Make .playerbots bot commands case-insensitive

Lichborne-AC:
## Pull Request Description

\`.playerbots bot add jared\` fails with "Character 'jared' not found" even when \`Jared\` exists, because the lookup is case-sensitive. Same issue hits \`remove\`, \`init=<...>\`, \`refresh\`, \`addaccount\`, and any other subcommand that takes a character name.

Fix at the parser layer in \`HandlePlayerbotCommand\`: every subcommand that takes a charname flows through one shared name-lookup step. Normalize the typed name to canonical form (first letter upper, rest lower) at that step so \`jared\`, \`JARED\`, \`JaReD\` all resolve to \`Jared\`. The \`addaccount\` path still tries the raw token as an account name first (account names are case-insensitive on the auth side), then falls back to a normalized character-name lookup.

No subcommand's behavior changes — only the name lookup that precedes them. Subcommands with no charname (\`initself\`, \`list\`, \`reload\`, \`tweak\`, \`self\`, \`lookup\`, \`addclass\`) are unaffected.

## Feature Evaluation

  One \`normalizePlayerName(s)\` call on each comma-separated token before the existing \`GetCharacterGuidByName(s)\` lookup. Uses the helper this module already calls in four other places.

  Runs only when a player types a \`.playerbots bot ...\` command, on the typed tokens. Not in any bot-tick or per-bot path. Cost does not scale with bot count.

## How to Test the Changes

\`\`\`
.playerbots bot add Jared           → adds Jared
.playerbots bot add jared           → adds Jared
.playerbots bot add JARED           → adds Jared
.playerbots bot add JaReD           → adds Jared
.playerbots bot remove Jared        → removes Jared
.playerbots bot remove jared        → removes Jared
.playerbots bot init=auto jared     → re-rolls Jared's gear
.playerbots bot levelup jared       → re-rolls Jared at his current level
.playerbots bot refresh jared       → refreshes Jared (HP/mana restored)
.playerbots bot refresh=raid jared  → unbinds Jared from saved raid IDs
.playerbots bot add jAred,saLLy,BOB → adds all three
.playerbots bot add nonexistent     → still reports not found
.playerbots bot initself            → still works (untouched path)
\`\`\`

All cases verified in-game.

## Impact Assessment

    - [x] No, not at all

    - [x] No
    - [ ] Yes (**explain why**)

    - [x] No
    - [ ] Yes (**explain below**)

## AI Assistance

- [ ] No
- [x] Yes (**explain below**)

Used Claude to help trace the call graph (confirming which subcommands are affected and which are untouched) and to draft the patch. Reviewed and tested in-game before pushing.

## Final Checklist

## Notes for Reviewers

Diff is small (~25 lines in one file). One minor cosmetic side effect: "Character 'X' not found" errors now echo the normalized form (\`Bogus\` instead of \`bogus\`).

### DIFF
--- a/src/Bot/PlayerbotMgr.cpp
+++ b/src/Bot/PlayerbotMgr.cpp
@@ -1251,16 +1251,11 @@ std::vector<std::string> PlayerbotHolder::HandlePlayerbotCommand(char const* arg
 
         if (!strcmp(cmd, "addaccount"))
         {
-            // When using addaccount, first try to get account ID directly.
-            // Account names are case-insensitive on the auth side, so try
-            // the raw token before falling back to a normalized character
-            // name lookup.
+            // When using addaccount, first try to get account ID directly
             uint32 accountId = GetAccountId(s);
             if (!accountId)
             {
-                // If not found, try to get account ID from character name.
-                // Character names are stored canonically (first letter upper,
-                // rest lower), so normalize the user input before lookup.
+                // If not found, try to get account ID from character name
                 std::string charName = s;
                 if (!normalizePlayerName(charName))
                 {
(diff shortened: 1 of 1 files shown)


brighton-chi:
This is awesome as the case sensitivity of these commands (and other commands too) annoys the hell out of me. One thing imo is we shouldn't have all the Claude exposition in comments; I'd either go back to the prior comments or just remove the comments entirely but ofc will leave that up to the maintainers.

Lichborne-AC:
Agreed, i tested it and told claude to upload it.  Didnt realize those comments were in the code.  Reverted to old comments, code didnt change - only the comments.

Celandriel (approved):

PR #2421, opened 2026-05-25
init cmd spam fix

Noscopezz:
## Pull Request Description

This change will add cooldown for all init commands to prevent init spam which will lag/crash server.
Cooldown is per bot (guid) and it will not allow command execution if cooldown is active.
Cooldown value is by default 30s per bot (not set in stone) and it can be changed via cfg (AiPlayerbot.AutoInitCooldown = 30). 

## Feature Evaluation

1. New config key AiPlayerbot.AutoInitCooldown (uint32, default 30, 0 disables).
2. BotInitGuard gains a static map<ObjectGuid, time_t> lastInitTime plus two helpers:
   - SecondsUntilReady(guid): returns remaining cooldown seconds (0 if ready or cooldown disabled).
   - StampInit(guid): writes time(nullptr) to the map.
3. In ProcessBotCommand, after the existing IsLocked() check and only when cmd starts with "init=":
   - If SecondsUntilReady(bot guid) > 0, return "Init cooldown, wait Ns." and do nothing.
   - Otherwise, on each successful init=* branch, call StampInit immediately before returning "ok".

Effectively free vs. the seconds to minutes of stalled inits it prevents.

## How to Test the Changes
1. Add bot
2. Use init command (any variation)
3. Use again within 30s
4. You should get cooldown msg and how many seconds left

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)
 Second init=... on the same bot within 30 seconds is now rejected with "Init cooldown, wait Ns." instead of running again.

    - - [ ] No
Minimal. One new branch in ProcessBotCommand: "if cmd starts with init= and SecondsUntilReady > 0, return cooldown message". Two new static methods on BotInitGuard (SecondsUntilReady, StampInit) and one new static map. Seven existing init=* branches each gain a single StampInit(...) call before their return. No new classes, no new files, no new external API. Maintenance is one config key and one map.

## AI Assistance

- - [x] No

## Final Checklist


src/Bot/PlayerbotMgr.cpp
kadeshar, line 758: resetInstanceIdForAltBots is deleted in this PR and still in code.
Same for property which cause compilation error.
Im not sure about whole PR. There are a lot of changes unrelated with PR description

kadeshar (changes requested):

src/Bot/PlayerbotMgr.cpp
Noscopezz: I think i made this before reset instance id. I will close pull and remake later, when I will have time.

PR #2422, opened 2026-05-25
Reset instance ID via existing cmd refresh=raid for alt bots

Noscopezz:
## Pull Request Description

Alt bots can now use refresh=raid cmd if cfg is set to 1.
`AiPlayerbot.ResetInstanceIdForAltBots = 1`

## Feature Evaluation

## How to Test the Changes

Add alt bot, use .playerbots bot refresh=raid and it should say ok after reseting instance.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)
Alt bots will now reset instance ID if cfg is set to 1.

    - - [ ] No
It will check if cfg is set to 1 to enable alt bots to reset instance id.

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/src/Bot/PlayerbotMgr.cpp
+++ b/src/Bot/PlayerbotMgr.cpp
@@ -737,7 +737,10 @@ std::string const PlayerbotHolder::ProcessBotCommand(std::string const cmd, Obje
     bool addClassBot = sRandomPlayerbotMgr.IsAddclassBot(guid.GetCounter());
 
     if (!addClassBot)
-        return "ERROR: You can not use this command on non-addclass bot.";
+    {
+        if (!(cmd == "refresh=raid" && sPlayerbotAIConfig.resetInstanceIdForAltBots))
+            return "ERROR: You can not use this command on non-addclass bot.";
+    }
 
     if (!admin)
     {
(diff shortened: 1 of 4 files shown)


src/Bot/PlayerbotMgr.cpp
```diff
@@ -737,7 +737,10 @@ std::string const PlayerbotHolder::ProcessBotCommand(std::string const cmd, Obje
     bool addClassBot = sRandomPlayerbotMgr.IsAddclassBot(guid.GetCounter());
 
     if (!addClassBot)
-        return "ERROR: You can not use this command on non-addclass bot.";
+    {
+        if (!(cmd == "refresh=raid" && sPlayerbotAIConfig.resetInstanceIdForAltBots))
+            return "ERROR: You can not use this command on non-addclass bot.";
```
Celandriel: Should this be translatable?

brighton-chi: Also, just to be an ass here, the word is "cannot"

I think this error message is not accurate though.

Noscopezz: It is the same as it was, only added new if before exisiting error, if you insist I can translate, but I think it would be better that translations come in logical order, this way, they will be scattered and hard to maintain. e.g. gear related 1-100, text cmds 101-300, etc...

Celandriel (approved):

Issue #715, opened 2024-11-15
When Threads>1 or ThreadPool>2 playbots are very prone to crashing  when there is more than one real player in the server


zbhcn:
When configuring the option MapUpdate in the worldserver. conf file in the Azerothcore core When Threads>1 or ThreadPool>2, playbots are very prone to crashing when there is more than one real player in the server. I think it is most likely because playbots do not support multithreading

zbhcn:
and When MapUpdate.Threads=1 or ThreadPool=2, no more crashes occur

PaoloBenigni:
How many bots are you running? What are your system specs for the server? What is your memory usage? What about diff?

zbhcn:
> How many bots are you running? What are your system specs for the server? What is your memory usage? What about diff?
> 
> Many people routinely run 3-4 mapupdate threads.

![5929c8f0c7a0e2aed6cf0968e02db207](https://github.com/user-attachments/assets/4b1535c0-fa8b-4615-821e-637235dbc7bc)

zbhcn:
*** Hardware ***
Processor: AMD EPYC 74F3 24-Core Processor                
Number Of Processors: 16
Physical Memory: 16776756 KB (Available: 9117480 KB)
Commit Charge Limit: 19791412 KB

*** Operation System ***
Windows Server 2016 Datacenter (Version 10.0, Build 20348)

//=====================================================
Exception code: C0000420 出现声明故障。

Assertion message: 
#----------------------------------------------------------------------#
 
>> ABORTED

# Location 'G:\azerothcore-wotlk-Playerbot\src\server\apps\worldserver\Main.cpp:650'
# Function 'FreezeDetector::Handler'
# Message 'World Thread hangs for 240161 ms, forcing a crash!'

PaoloBenigni:
Did you change `MaxCoreStuckTime = 0` in the worldserver.conf?

zbhcn:
> Did you change `MaxCoreStuckTime = 0` in the worldserver.conf?

yes, i set MaxCoreStuckTime = 240

PaoloBenigni:
Was there a specific reason? If the diff is typically less than 100 ms, and only sometimes you get a spike then I would just leave it off.

PaoloBenigni:
I would turn that setting off then unless you see issues. It should eliminate the error.

noisiver:
[...]

PR #2425, opened 2026-05-28
Improve bot trinket usage and fix related bugs

brighton-chi:
## Pull Request Description

This PR makes three changes to UseTrinketAction:
1. It adds health and mana gating (based on existing config thresholds) for bots to activate mana recovery, mana efficiency, and defensive trinkets. The thresholds are mediumMana (default 40%) for mana recovery trinkets, highMana (default 65%) for mana efficiency trinkets, and lowHealth (default 45%) for defensive trinkets.
2. It removes the old overinclusive procflag/specproc gate introduced by PR 1385, which prevents bots from using dozens of valid trinkets, including some extremely powerful ones (such as Skull of Gul’dan, the iconic TBC expansion BiS trinket for all casters), and replaces it with a narrower exclusion that still addresses the original issue.

- Regarding PR 1385, focusing on liyunfan’s post specifically, I interpet the issue to be relating to trinkets where the item data is screwed up such that a passive effect was implemented as an on-use spell. I had AI do a scan, and it seems that issue impacts only 2 trinkets in the game, Oracle Talisman of Ablution and Frenzyheart Insignia of Fury, and specifically only the versions of those items that are not legitimately obtainable (unclear why they exist at all). I’ve excluded those trinkets from bots via the config now, and this PR maintains the guards against those trinkets regardless but does so in a narrower fashion. PR 1385's approach of using ProcFlags != 0 (i.e., excluding all on-use trinkets with non-zero ProcFlags) works only if proc metadata can be used to distinguish between active/passive trinkets, and that’s not even close to being the case. AI came up with 44 false positives, including many significant trinkets beyond Skull of Gul’dan such as the ZG Hakkar quest trinkets, Badge of the Swarmguard, Petrified Scarab, Scarab Brooch, Eye of the Dead, Essence of the Martyr, Abacus of Violent Odds, Ribbon of Sacrifice, and Pendant of the Violet Eye (and I’m not mentioning WotLK trinkets only because I don’t know anything about what is relevant for that expansion).

3. It fixes an issue where bots were not respecting trinket cooldowns in some cases. This resulted because trinkets with shared cooldown categories (i.e., those that are not stackable) would substitute the individual trinket cooldowns with shared category cooldowns, which let bots spam usages of trinkets, by tracking per-item and per-category trinket cooldowns locally. The root of this is based in AC; I don't know if it should be considered a bug or not, but regardless it doesn't impact players, presumably because cooldowns are enforced on the client side.

- Here’s an illustration to explain the issue in practice. Skull of Gul’dan has a 240s cooldown and Shifting Naaru Sliver has a 180s cooldown, and they share a cooldown category so their usages cannot be stacked. The shared cooldown matches the length of the on-use effect (so 20s for Skull and 15s for Sliver). The below is what can happen without this PR (and is what I observed in testing).

  - t = 0s: Shifting Naaru Sliver used, writes its 90s personal cooldown and 15s shared category cooldown.
  - t = 15s: Skull of Gul’dan used, writes its 120s personal cooldown and 20s shared category cooldown. Skull’s shared category cooldown overwrites Sliver’s spell-cooldown entry, giving Sliver 20s left on its personal cooldown instead of 75s.
  - t = 35s: Sliver incorrectly appears ready and can be used again (55s sooner than should be possible). Then Sliver’s shared category cooldown in turn overwrites Skull’s longer personal cooldown.
  - t = 50s: Skull incorrectly appears ready and can be used again (85s sooner than should be possible).
  - Repeat.

## Feature Evaluation

The logic runs through the existing trinket use code. Specifically:

- Helpers are used to classify on-use trinket effects into mana restoration, mana efficiency, and defensive/tank categories using spell-effect checks
- Existing configured mana and health thresholds are applied to those trinkets
- The old procflag gate is replaced with a one-time cached set of mixed ON_USE/ON_EQUIP trinket spell ids
- Two small cooldown maps per bot are tracked in UseTrinketAction: one for item cooldowns and one for shared category cooldowns

## How to Test the Changes

I did all of these things, but verification is always good. Overall, I think it is worth running with this PR merged into test-staging if/when that happens and keeping an eye on overall performance.
1.	Equip a bot with a trinket referenced above, such as Skull of Gul'dan, and confirm it is now used in combat.
2.	Test mana-based classes with mana recovery and mana-efficiency trinkets and confirm they are used only when below the applicable configured mana threshold. An easy one to check is Glimmering Naaru Sliver because it is a channel.
3.	Test defensive on-use trinkets and confirm they are used only when health is below the configured low-health threshold. Something like Shadowmoon Insignia, which increases maximum health, is pretty obvious.
4.	Confirm that the error versions of Oracle/Frenzyheart don’t stack auras on bots (i.e., the bug addressed in PR 1385 has not returned). You can do this by having a bot kill mobs and check .listauras, though I checked through logging in the code because auras are noisy as hell.
5.	Equip a bot with two trinkets that have shared cooldowns. I used Skull of Gul’dan and Shifting Naaru Sliver. Go fight a mostly tank-and-spank boss, such as Gruul. Use an add-on like Skada that tracks buffs. You should see that before this PR, bots will use the trinkets multiple times in one cooldown period, and after, they observe the actual cooldowns.

## Impact Assessment

    - - [ ] No, not at all

Any additional impact is confined to UseTrinketAction. The exclusion of the busted trinkets uses a cache that is built once per server process followed by constant-time lookups. The per-bot trinket cooldown maps also use constant-time lookups and store only a few timestamp entries per bot.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [ ] No

Sort of--trinkets previously didn't have any consideration for effects with respect to usage so that is new. I think it is necessary though to have half-decent bot trinket usage, and there could be further refinement for how bots decide to use trinkets based on this structure.

## AI Assistance

- - [ ] No

GPT-5.4 was used for investigation and root-cause analysis and assistance with drafting. All resulting code and the PR rationale was validated through in-game testing.

## Final Checklist

## Notes for Reviewers

Bots also suck at using DPS trinkets properly, but I don't think there's a simple way to address that unlike with mana recovery or defensive trinkets. So that's to consider another day.

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -862,8 +862,9 @@ AiPlayerbot.LimitGearExpansion = 1
 AiPlayerbot.RandomGearLoweringChance = 0
 
 # Unobtainable or unusable items (comma-separated list of item IDs)
-# Default: Chilton Wand (12468), Totem of the Earthen Ring (46978)
-AiPlayerbot.UnobtainableItems = 12468,46978
+# Defaults: Chilton Wand (12468), Frenzyheart Insignia of Fury test/on-use row (44869),
+#           Oracle Talisman of Ablution test/on-use row (44870), Totem of the Earthen Ring (46978)
+AiPlayerbot.UnobtainableItems = 12468,44869,44870,46978
 
 # Randombots check player's gearscore level and deny the group invitation if it's too low
 # Default: 0 (disabled)
--- a/src/Ai/Base/Actions/GenericSpellActions.cpp
+++ b/src/Ai/Base/Actions/GenericSpellActions.cpp
@@ -6,7 +6,6 @@
 #include "GenericSpellActions.h"
 
 #include <ctime>
-#include <unordered_map>
 #include <unordered_set>
 
 #include "Event.h"
@@ -553,10 +552,6 @@ bool UseTrinketAction::UseTrinket(Item* item)
     int32 itemSpellCooldown = 0;
     uint32 itemSpellCategory = 0;
     int32 itemSpellCategoryCooldown = 0;
-    static std::unordered_map<ObjectGuid::LowType, std::unordered_map<uint64, uint32>>
-        trinketItemCooldownExpiries;
-    static std::unordered_map<ObjectGuid::LowType, std::unordered_map<uint32, uint32>>
-        trinketCategoryCooldownExpiries;
 
     for (uint8 i = 0; i < MAX_ITEM_PROTO_SPELLS; ++i)
     {
@@ -567,38 +562,30 @@ bool UseTrinketAction::UseTrinket(Item* item)
             itemSpellCooldown = item->GetTemplate()->Spells[i].SpellCooldown;
             itemSpellCategory = item->GetTemplate()->Spells[i].SpellCategory;
             itemSpellCategoryCooldown = item->GetTemplate()->Spells[i].SpellCategoryCooldown;
-            ObjectGuid::LowType const botGuid = bot->GetGUID().GetCounter();
             uint64 const itemCooldownKey = (static_cast<uint64>(item->GetEntry()) << 32) | spellId;
             uint32 const now = getMSTime();
 
-            auto itemCooldownsItr = trinketItemCooldownExpiries.find(botGuid);
-            if (itemSpellCooldown > 0 && itemCooldownsItr != trinketItemCooldownExpiries.end())
+            if (itemSpellCooldown > 0)
             {
-                auto const itemCooldownItr = itemCooldownsItr->second.find(itemCooldownKey);
-                if (itemCooldownItr != itemCooldownsItr->second.end())
+                auto const itemCooldownItr = trinketItemCooldownExpiries.find(itemCooldownKey);
+                if (itemCooldownItr != trinketItemCooldownExpiries.end())
                 {
                     if (itemCooldownItr->second > now)
                         return false;
 
-                    itemCooldownsItr->second.erase(itemCooldownItr);
-                    if (itemCooldownsItr->second.empty())
-                        trinketItemCooldownExpiries.erase(itemCooldownsItr);
+                    trinketItemCooldownExpiries.erase(itemCooldownItr);
                 }
             }
 
-            auto categoryCooldownsItr = trinketCategoryCooldownExpiries.find(botGuid);
-            if (itemSpellCategory && itemSpellCategoryCooldown > 0 &&
-                categoryCooldownsItr != trinketCategoryCooldownExpiries.end())
+            if (itemSpellCategory && itemSpellCategoryCooldown > 0)
             {
-                auto const categoryCooldownItr = categoryCooldownsItr->second.find(itemSpellCategory);
-                if (categoryCooldownItr != categoryCooldownsItr->second.end())
+                auto const categoryCooldownItr = trinketCategoryCooldownExpiries.find(itemSpellCategory);
+                if (categoryCooldownItr != trinketCategoryCooldownExpiries.end())
                 {
                     if (categoryCooldownItr->second > now)
                         return false;
 
-                    categoryCooldownsItr->second.erase(categoryCooldownItr);
-                    if (categoryCooldownsItr->second.empty())
-                        trinketCategoryCooldownExpiries.erase(categoryCooldownsItr);
+                    trinketCategoryCooldownExpiries.erase(categoryCooldownItr);
                 }
             }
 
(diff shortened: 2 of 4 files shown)


Celandriel (commented):
First Pass at AI review. As always take with a grain of salt. 

**Main issue — the two cooldown maps are static (process-global), not per-bot — Request changes**
GenericSpellActions.cpp:154-157:


```
static std::unordered_map<ObjectGuid::LowType, std::unordered_map<uint64, uint32>> trinketItemCooldownExpiries;
static std::unordered_map<ObjectGuid::LowType, std::unordered_map<uint32, uint32>> trinketCategoryCooldownExpiries;
```
The PR body calls these "two small cooldown maps per bot." They're not — static inside UseTrinket() means one shared instance for the whole process, with the bot GUID as the outer key. Every bot's Execute does find/erase/operator[] on that shared outer map (:173, :182-184, :188, :198-200, :293, :298). Bot AI runs on map threads, so this is two problems at once:

- B11 data race: with MapUpdate.Threads > 1, concurrent mutation of the shared unordered_map → rehash/iterator corruption → crash (the #715 failure mode). It's latent today because the project forces MapUpdate.Threads = 1, so it won't crash under the default — but the protocol's rule is explicitly "new code must not make #715 worse," and this does, while permanently poisoning this path for any future multi-map-thread work.
- Unbounded growth: the outer map gains one entry per bot GUID that ever uses a trinket, and there's no logout cleanup — inner entries are only erased on the read path when found-and-expired. Slow but real with random-bot churn.
- Both vanish with the same one-line-structural fix. UseTrinketAction is a per-bot, persistent action instance (GenericSpellActions.h:321 — created and cached per bot by the context factory), and a given bot only updates on its own map thread. So make them non-static members and drop the GUID layer entirely:
- 

```
// UseTrinketAction, private:
std::unordered_map<uint64, uint32> _itemCdExpiry;      // key (entry<<32)|spellId
std::unordered_map<uint32, uint32> _categoryCdExpiry;  // key category
```
That's per-bot by construction, needs no synchronization, and dies with the bot (no leak). It also simplifies the code (no botGuid lookups). This is the correct tool — static was a mis-reach for "persist across Execute calls," which a member already does.

**Low — getMSTime() comparison isn't wrap-safe**
:179, :195 compare expiry > now directly on getMSTime() values, which wrap every ~49.7 days. The idiom here is getMSTimeDiff(now, expiry) (Timer.h:110). Tiny window, low impact, but worth switching while restructuring.

**Validated — no other defects**

- Build: the new CanCastSpell(spellId, bot, false, nullptr, item) and GetSpellCooldownDelay both exist with matching signatures — compiles.
- A6 (no new bypass): still the pre-existing CMSG_USE_ITEM → HandleUseItemOpcode path, unchanged. Priority untouched (ACTION_NORMAL+4, won't preempt heals).
- Mana/health gating logic is correct: precedence is (restoresMana && pct>=mediumMana) || (pct>=highMana) — restore trinkets fire <40%, efficiency <65%; verified by trace. Allowing restoresMana without an aura (:227) correctly lets instant ENERGIZE trinkets through — a real fix vs the old if (!applyAura) return false.
- Mixed-trigger set: the GetMixedTriggerTrinketSpellIds() lazy static is thread-safe to initialize (C++11 guarantee) and read-only afterward — safe. The narrower exclusion is a genuine improvement over the old ProcFlags != 0 gate (which the author shows nukes ~44 legit trinkets incl. Skull of Gul'dan).
- Cooldown read-after-use is valid — bot item use is synchronous, so GetSpellCooldownDelay reflects the just-applied cooldown.

**Design notes (non-blocking)**

- Defensive trinkets gate on health <= lowHealth (45%), i.e. reactive only — a tank won't proactively pop a mitigation trinket before a big hit. Reasonable heuristic, just a behavior choice worth noting.
- The cooldown gate lives in UseTrinket()/Execute, not isUseful(), so an on-cooldown trinket stays queued and is rejected at execution each tick — cheap, acceptable.

**Verdict**
Request changes — one real issue (the static maps; fix = make them per-instance members, dropping the GUID key), plus the low wrap-safety nit. Everything else — the gating logic, the ENERGIZE fix, and the narrower mixed-trigger exclusion — is correct and a solid improvement. The fix is small and localized.

brighton-chi:
Thanks, I thought I fixed that by not using operator[] but maybe I forgot to bring it over from my main branch. I'll take a look later today.

brighton-chi:
Ok I think this should be fixed. I guess this relates to what you first raised months ago about action persistence. I should probably use private members instead of global containers with instanceId or guid keys for some boss strategy storage too, like per-bot positioning or waypoint sequences. But whatever, for another day.

Celandriel (approved):

PR #2426, opened 2026-05-28
fix bot leader handling

### LINKED ISSUE #2420: Random dungeon leader bot loses movement/AI after leadership is passed to player
**Bug Description**
When a random dungeon group is formed and one of the playerbots is initially assigned as group leader, the bot automatically transfers leadership to the real player upon entering the dungeon. After transferring leadership, that specific bot appears to lose its steering/movement state entirely and begins running around erratically, ignoring normal follow behavior and commands.

If the bugged bot is kicked from the group and a new replacement bot joins, the newly joined bot behaves normally and follows commands correctly.

The issue appears tied specifically to the bot that previously held leadership before passing it to the player.
`follow` command doesn't do anything, `flee` gets the bot to the player, but that's about it.

I do believe I read something similar, someone mentioning "altbots", but I can't for the life of me search it up now.

**Commit Hash**
`a02e4ca5c17e636ac0e36277108480293ab8358c`
Was also an issue prior to this commit, but I didn't properly see the pattern until recently.

**How To Reproduce Bug**
Join a dungeon through RDF.
- If the player is already leader in the beginning, all is okay it appears.
- If the player is NOT leader in the beginning and a bot goes "I don't know this dungeon, please lead the way" and passes the torch, that bot starts running around and becomes rather uncontrollable.

**Expected Behavior**
Pass leader to the player - bot should still be a functioning bot like the rest.

**Modules**
- mod-ah-bot-plus
- mod-pvp-titles
- mod-quest-loot-party
- mod-transmog
- mod-player-bot-level-brackets

**System**
OS: Linux

### LINKED ISSUE #2424: Party Leader randomly changes and give leader command not working
**Bug Description**
Party leader sometimes randomly changes for no reason with no input and also resets loot filter back to group loot. When using the give leader command there is no no response 90% of the time, when the give leader command is rarely detected and the bot whispers back "Lead the way" they don't actually give party lead. So naturally I try to login in to the bot that has lead but then it auto passes to the char I was just on, so I relogin back to my main character and it goes back to my altbot again (basically never-ending cycle). I'm forced to reform group or get stuck in party lead loop.

**Commit Hash**
#2399 

**How To Reproduce Bug**
Seems to happen randomly I've tried my best debugging but anytime I think i found the cause it happens again while doing something different. 

**Expected Behavior**
Party leader should never change randomly without any input. Give leader should force current party lead bot to give back party lead to requested char

**Screenshots**
Image
Crown on frame shows party leader never passed when command is sent. Tried without addons and fresh client same result. Doesn't work for friend either

**Modules**
Mod-Multibot-Bridge
Mod-Ale
Mod-AH-Bot-Plus 
Mod-Dynamic-XP
Mod-Fly-Anywhere 
Mod-Starter-Guild (Disabled 
Mod-LLM-Chatter (Disabled) 
AoE-Loot (Disabled)

**Playerbot Settings**
Tested with 1 ALTbot and also with 2 players 2 altbots (1 altbot from each player)

**System**
Windows 11

**Additional Context**
There has also been this really annoying issue (Not sure if related might make another issue later) it doesn't seem to be addon related. Where after a while I am unable to talk in party chat anymore. I can type something and press enter but nothing comes through, but /say and other channels work fine. My friend is also getting the same issue. Relogging seems to fix it but reloading UI doesn't.
This is a completely fresh build with very few mods, most of which I have disabled to try and find the cause.

Noscopezz:
## Pull Request Description

This pr fixes bots "freaking out" after leader change and bots will promote player to party leader after give leader command
closes #2420 #2424 

## Feature Evaluation

## How to Test the Changes

1. Join rdf or make group
2. Promote bot to leader (rnd or alt), if bot was not leader
3. Type /p give leader
4. Bot will promote you to leader and it will follow you instead of freaking out

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

1. Bots now reset their AI on every SMSG_GROUP_SET_LEADER packet, so old leader strategies get cleared after a leader change.
2. Random bots auto-bind their master to a real-player group member during reset botAI, so commands like give leader and follow logic that depend on HasActivePlayerMaster() start working for them.
3. On OnBotLogin, bots no longer steal leadership from a real player, they only force the leader change if the current leader is a bot or offline.

    - - [ ] No

Adds two small guards: an "is current leader a real player" check on login, and a "find first real-player member" loop inside ResetAiAction. 
Both reuse existing patterns (IsRealPlayer(), the OnPlayerLogin master-assign loop).

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/src/Ai/Base/Actions/ResetAiAction.cpp
+++ b/src/Ai/Base/Actions/ResetAiAction.cpp
@@ -44,6 +44,24 @@ bool ResetAiAction::Execute(Event event)
             }
         }
     }
+    if (sRandomPlayerbotMgr.IsRandomBot(bot) && !bot->InBattleground())
+    {
+        if (Group* group = bot->GetGroup())
+        {
+            if (!botAI->GetMaster() || GET_PLAYERBOT_AI(botAI->GetMaster()))
+            {
+                for (GroupReference* gref = group->GetFirstMember(); gref; gref = gref->next())
+                {
+                    Player* member = gref->GetSource();
+                    if (member && member != bot && !GET_PLAYERBOT_AI(member))
+                    {
+                        botAI->SetMaster(member);
+                        break;
+                    }
+                }
+            }
+        }
+    }
     PlayerbotRepository::instance().Reset(botAI);
     botAI->ResetStrategies(false);
     botAI->TellMaster("AI was reset to defaults");
--- a/src/Ai/Base/Actions/ResetAiAction.cpp
+++ b/src/Ai/Base/Actions/ResetAiAction.cpp
@@ -44,6 +44,15 @@ bool ResetAiAction::Execute(Event event)
             }
         }
     }
+    if (Player* master = botAI->GetMaster())
+    {
+        Group* botGroup = bot->GetGroup();
+        Group* masterGroup = master->GetGroup();
+        if (botGroup && (!masterGroup || masterGroup != botGroup))
+        {
+            botAI->SetMaster(nullptr);
+        }
+    }
     if (sRandomPlayerbotMgr.IsRandomBot(bot) && !bot->InBattleground())
     {
         if (Group* group = bot->GetGroup())
--- a/src/Ai/Base/Actions/ResetAiAction.cpp
+++ b/src/Ai/Base/Actions/ResetAiAction.cpp
@@ -49,26 +49,14 @@ bool ResetAiAction::Execute(Event event)
         Group* botGroup = bot->GetGroup();
         Group* masterGroup = master->GetGroup();
         if (botGroup && (!masterGroup || masterGroup != botGroup))
-        {
             botAI->SetMaster(nullptr);
-        }
     }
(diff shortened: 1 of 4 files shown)


src/Ai/Base/Actions/ResetAiAction.cpp
```diff
@@ -44,6 +44,33 @@ bool ResetAiAction::Execute(Event event)
...
+                    Player* member = gref->GetSource();
+                    if (member && member != bot && !GET_PLAYERBOT_AI(member))
+                    {
+                        botAI->SetMaster(member);
+                        break;
+                    }
+                }
+            }
+        }
+    }
```
Celandriel, line 61: Here I would call botAI->FindNewMaster() which unifies the logic you are calling here.

Noscopezz: Replaced with `botAI->FindNewMaster()` idk how I missed it, tested and works.

Celandriel (approved):

PR #2427, opened 2026-05-29
Sunder Armor fixes

kadeshar:
## Pull Request Description

Arms and Fury using Sunder Armor when there is not Protection Warrior in group

## How to Test the Changes

Invite warrior/warriors to group and check how they apply Sunder Armor on dummy target

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Yes, expained in Pull Request Description

    - - [x] No

## AI Assistance

- - [ ] No

Analyze orginal behavior

## Final Checklist

### DIFF
--- a/src/Ai/Class/Warrior/Action/WarriorActions.cpp
+++ b/src/Ai/Class/Warrior/Action/WarriorActions.cpp
@@ -41,7 +41,7 @@ bool CastSunderArmorAction::isUseful()
     if (!group)
         return false;
 
-    if (AiFactory::GetPlayerSpecTab(bot) != WARRIOR_TAB_PROTECTION)
+    if (!botAI->IsTank(bot, false))
     {
         for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
         {
@@ -53,7 +53,7 @@ bool CastSunderArmorAction::isUseful()
             }
 
             if (member->getClass() == CLASS_WARRIOR &&
-                AiFactory::GetPlayerSpecTab(member) == WARRIOR_TAB_PROTECTION)
+                botAI->IsTank(member, false))
                 return false;
         }
     }
--- a/src/Ai/Class/Warrior/Action/WarriorActions.cpp
+++ b/src/Ai/Class/Warrior/Action/WarriorActions.cpp
@@ -5,6 +5,7 @@
 
 #include "WarriorActions.h"
 
+#include "AiFactory.h"
 #include "Playerbots.h"
 
 bool CastBerserkerRageAction::isPossible()
@@ -36,6 +37,23 @@ bool CastBerserkerRageAction::isUseful()
 
 bool CastSunderArmorAction::isUseful()
 {
+    if (AiFactory::GetPlayerSpecTab(bot) != WARRIOR_TAB_PROTECTION)
+    {
+        Group* group = bot->GetGroup();
+        if (group)
+        {
+            for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+            {
+                Player* member = ref->GetSource();
+                if (!member || member == bot || !member->IsAlive() || !member->IsInWorld())
+                    continue;
+
+                if (member->getClass() == CLASS_WARRIOR && AiFactory::GetPlayerSpecTab(member) == WARRIOR_TAB_PROTECTION)
+                    return false;
+            }
+        }
+    }
+
     Aura* aura = botAI->GetAura("sunder armor", GetTarget(), false, true);
     return !aura || aura->GetStackAmount() < 5 || aura->GetDuration() <= 6000;
 }
--- a/src/Ai/Class/Warrior/Action/WarriorActions.cpp
+++ b/src/Ai/Class/Warrior/Action/WarriorActions.cpp
@@ -45,10 +45,14 @@ bool CastSunderArmorAction::isUseful()
             for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
             {
                 Player* member = ref->GetSource();
-                if (!member || member == bot || !member->IsAlive() || !member->IsInWorld())
+                if (!member || member == bot || !member->IsAlive() || !member->IsInWorld() ||
+                    member->GetMapId() != bot->GetMapId())
+                {
                     continue;
+                }
 
-                if (member->getClass() == CLASS_WARRIOR && AiFactory::GetPlayerSpecTab(member) == WARRIOR_TAB_PROTECTION)
+                if (member->getClass() == CLASS_WARRIOR &&
+                    AiFactory::GetPlayerSpecTab(member) == WARRIOR_TAB_PROTECTION)
                     return false;
             }
         }
--- a/src/Ai/Class/Warrior/Action/WarriorActions.cpp
+++ b/src/Ai/Class/Warrior/Action/WarriorActions.cpp
@@ -37,24 +37,24 @@ bool CastBerserkerRageAction::isUseful()
 
 bool CastSunderArmorAction::isUseful()
 {
+    Group* group = bot->GetGroup();
+    if (!group)
+        return false;
+
     if (AiFactory::GetPlayerSpecTab(bot) != WARRIOR_TAB_PROTECTION)
     {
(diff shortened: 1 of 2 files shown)


src/Ai/Class/Warrior/Action/WarriorActions.cpp
```diff
@@ -36,6 +37,23 @@ bool CastBerserkerRageAction::isUseful()
 
 bool CastSunderArmorAction::isUseful()
 {
+    if (AiFactory::GetPlayerSpecTab(bot) != WARRIOR_TAB_PROTECTION)
```
brighton-chi: Is it better to make a variable for tab since it's checked twice?

src/Ai/Class/Warrior/Action/WarriorActions.cpp
```diff
@@ -36,6 +37,23 @@ bool CastBerserkerRageAction::isUseful()
...
+    {
+        Group* group = bot->GetGroup();
+        if (group)
```
brighton-chi: I think the function should immediately return false if there is no group; no need to further check for an aura.

src/Ai/Class/Warrior/Action/WarriorActions.cpp
```diff
@@ -36,6 +37,23 @@ bool CastBerserkerRageAction::isUseful()
...
+        {
+            for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+            {
+                Player* member = ref->GetSource();
+                if (!member || member == bot || !member->IsAlive() || !member->IsInWorld())
```
brighton-chi: Should there be a check for the member's mapId not being the same as the bot's?

src/Ai/Class/Warrior/Action/WarriorActions.cpp
kadeshar: Its checked only once for bot. Second one is for member

src/Ai/Class/Warrior/Action/WarriorActions.cpp
kadeshar: Why? Arms should not apply this if is not in group?

src/Ai/Class/Warrior/Action/WarriorActions.cpp
kadeshar: Good idea, added

src/Ai/Class/Warrior/Action/WarriorActions.cpp
brighton-chi: Hmm I figured it was not worth using if solo since nobody else is benefitting and fights are short. But now I'm reading about it, and there seems to be lots of different views about whether to use it. Some people say don't use Heroic Strike and use Sunder instead. But then there are leveling guides that don't talk about Sunder at all.

src/Ai/Class/Warrior/Strategy/ArmsWarriorStrategy.cpp
```diff
@@ -112,6 +112,7 @@ std::vector<NextAction> ArmsWarriorStrategy::getDefaultActions()
     return {
         NextAction("bladestorm", ACTION_DEFAULT + 0.2f),
         NextAction("mortal strike", ACTION_DEFAULT + 0.1f),
+        NextAction("sunder armor", ACTION_DEFAULT + 0.05f),
```
ThePenguinMan96: Sunder armor should not be in default actions, atleast not for anything besides prot, and they use devastate. It's bad in pvp as arms. If it stays in default actions,  you could add a player check in the isuseful for sunder armor. I added one to maim in the cat druid, where it only fires on players.

Celandriel:
## Review — "Sunder Armor fixes"

The intent is sound and the Arms default-action addition is correct (`sunder armor` is registered globally for warriors in `WarriorAiObjectContext.cpp:217`, and priority `ACTION_DEFAULT + 0.05f` sits in the 5.0–5.9 default band). But the `isUseful()` change has a logic bug and a real per-tick cost, and it touches a **shared** action used by all three warrior specs.

### 🔴 Blocker — `if (!group) return false` breaks solo warriors, including Prot tanks

`CastSunderArmorAction` is shared by **Tank, Fury, and Arms** (the tank strategy uses it directly and as the `devastate` alternative in `TankWarriorStrategy.cpp`). The new early return sits **above** the spec check, so it applies to every spec:

```cpp
if (!group)
    return false;          // <-- hits Prot tanks too
```

A **Protection warrior tanking solo** (leveling, questing, world grinding) will now never apply Sunder Armor — a core threat/armor ability for tanks. That's a default-behavior regression not covered by the PR description.

It's also logically backwards even by the PR's own stated intent ("use Sunder when there is *not* a Protection warrior in group"): a warrior with **no group has no Prot warrior to defer to**, so it *should* sunder, not skip. The correct behavior for `!group` is to fall through to the existing aura check.

Suggested shape:

```cpp
if (Group* group = bot->GetGroup();
    group && AiFactory::GetPlayerSpecTab(bot) != WARRIOR_TAB_PROTECTION)
{
    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
    {
        // ... if a Prot warrior is found, return false
    }
}
// fall through to the existing aura check (covers solo + "no prot in group")
```

### 🟠 High — `GetPlayerSpecTab()` in a per-tick `isUseful()`

`isUseful()` runs every tick for every queued action, and `sunder armor` is now a **default action** (always queued in combat). The new code calls `AiFactory::GetPlayerSpecTab()` once for the bot plus once per warrior group member. That function is not lightweight (`AiFactory.cpp:100`): it **heap-allocates a `std::map`**, iterates the entire talent map, and does `sTalentStore` / `sSpellMgr` lookups per talent. In a raid with multiple warriors this is meaningful per-tick cost, and it contradicts the template's "No, not at all" impact answer.

Mitigations: cache via a `CalculatedValue<bool>` ("group has prot warrior") with `checkInterval` ~1000ms instead of recomputing every tick; or gate on a cheaper role check before paying for `GetPlayerSpecTab` (noting that role-based detection would also match feral/prot-pala tanks, which changes the semantics).

### 🟡 Minor

- **Parent contract:** `CastSunderArmorAction::isUseful()` still doesn't chain `CastSpellAction::isUseful()` (range / spell-known checks). Pre-existing, but since you're editing this method, adding `&& CastSpellAction::isUseful()` (as `CastBerserkerRageAction` does) is a cheap correctness win.
- **Consistency:** Fury queues sunder at `ACTION_DEFAULT + 0.3f` but Arms at `+0.05f`. Both in-band; just confirming this ordering (below Mortal Strike/Bladestorm) is intended.

### Verdict

Request changes. The Arms default-action addition is fine, but the `isUseful()` rewrite (a) regresses solo Prot tanks via the `!group` early-return, which is logically inverted vs. the PR's own goal, and (b) adds non-trivial per-tick cost in a hot path marked as zero-impact. Fix the `!group` fall-through and cache/cheapen the spec lookup and this is good to merge.

brighton-chi:
A Protection warrior tanking solo (leveling, questing, world grinding) will now never apply Sunder Armor — a core threat/armor ability for tanks. That's a default-behavior regression not covered by the PR description.

Wtf does this mean. What is tanking solo. LLM oxymoron

Celandriel:
Yea, as always take these with a grain of salt. 
But the question is whether this should only be cast while in a group or not is valid, as is the point of expensive lookups in isUseful

brighton-chi:
Yeah, maybe prot should still use Sunder when soloing even if Arms and Fury don't.

kadeshar:
Protection use devastate and i changed that like was discussed on discord. I dont see sense to apply sunder armor in leveling.
Second point is not true because isgroup filter most ticks.
Third is preexisted and dont plan analyze that.

Celandriel:
Any idea of the performance hit?

kadeshar:
Edited comment above

brighton-chi:
Not necessarily a problem here, but I remember figuring out a while back that AiFactory::GetPlayerSpecTab() is pretty expensive. It's noticeable in pmon.

kadeshar:
Changed on strategy check

PR #2430, opened 2026-05-30
Remove hardcoded "mana potion" actions from Druid, Mage, and Spriest

dillyns:
## Pull Request Description

Currently, a mage druid or shadow priest bot would use a mana potion without the "potions" strategy if Innervate, Evocation, or Disperse were unavailable.
Removes hardcoded calls to "mana potion". Mana potions should be used with the potions strategy only.

## Feature Evaluation

## How to Test the Changes

Create a druid, mage, or shadow priest bot and make sure they do not have the potions strategy (co -potions).
Make sure the bot has a mana potion in their inventory.
Get them to go out of mana. They should not use the potion.
Add the potions strategy (co +potions). They should now use the potion on the "low mana" trigger like everyone else.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Claude was used to search for any instances of mana potion usage outside of the potions strategy.

## Final Checklist


Celandriel:
Empty action nodes should be deleted.

ThePenguinMan96:
I disagree with this, I added the potion action / medium mana trigger for mage so they use a mana potion earlier in the fight so the 3min CD starts and ends sooner. It helped the endurance of the mage quite a bit in 4-6+ minute fights, especially arcane (mana hungry). These were added after tons of testing on long raid fights.

Celandriel:
I think the concern is whether they should be triggering it as an alternative to a spell instead of an independent strategy. I see how potion use should remain a separate from a spell alternative, and at least on first pass agree with it.

ThePenguinMan96:
Evocation triggers at low mana, so I agree with removing it as an alternative action from evocation, since there is a medium mana trigger for mana potion. 

When making the Mage PR, they were the most mana hungry class I've dealt with. In longer fights, they would end up losing a lot on DPS because they would get to a point where evocation and potion were both on cooldown. Then I added in the mana gems, at high mana threshold, so they start their cooldown earlier. Then, I realized evocation and mana potion were both triggering at the same time at low mana threshold - evocation was triggering from the mage strategy, the mana potion was triggering from the potion strategy. Then I added a separate medium mana trigger -> mana potion trigger directly into the generic Mage strategy. So then the mage had 3 mana triggers:

High mana - mana gem
Medium mana - potion
Low mana - evocation 

Three separate cooldowns, each triggered at different points in the fight. 

For the owner of this PR, I recommend you test a fresh arcane made at level 80 with blue gear. Go to a Target dummy and see how long they last before going out of mana with these proposed changes. Then test the current branch without these changes. You'll see that the arcane Mage goes out of mana quite a bit faster. 

And, if we truly do want to have all potion usage in the potion strategy itself, the strategy needs to understand that some classes need a higher threshold to trigger a mana potion. Not work I've got the time to do, but someone else could do it.

ThePenguinMan96:
I do agree with removing mana potion as an alternative action to dispersion. Dispersion is on the low mana trigger, as well as the mana potion from the potion strategy. Having it as an alternative is redundant.

brighton-chi:
The concern I have is if hardcoding the potion use will cause problems in arena. I know that does happen with the potion strategy; when the bot gets low enough on mana it will lock up trying to use a potion repeatedly and do nothing. I solved that by removing the potion strategy from default arena strategies, but I'm not sure if these strategies would still cause the same as I don't use these classes in arena.

Edit: Of course even if this is the case, it could be solved by an independent potion use trigger for these classes with an arena check.

dillyns:
In my opinion, if there is a potions strategy then bots shouldn't be using potions outside of it, full stop.
 
Perhaps the mana potion usage in the potions strategy should be bumped up to medium mana, or some other more robust solution should be considered for configuring potions on a per class basis. That probably warrants a more indepth discussion and most likely a separate PR.

ThePenguinMan96:
I'm fine with the mana potion being bumped to medium, and removing all potions from class strategies at that point

Celandriel (dismissed):

Celandriel:
Merge conflict

Celandriel (approved):

PR #2435, opened 2026-06-02
Refactor HasSpell/HasAura and convert spellIds to constants

brighton-chi:
## Pull Request Description

This is a PR to follow-up on a discussion with @kadeshar  during the recent mage armor PR. Main changes:
- Add new PlayerbotAI::HasSpell() overload that accepts a string for the spell name as an argument, in order to replace repeated bot->HasSpell(spellId) checks for places where multiple ranks of a spell are checked, and to replace a few calls to AI_VALUE(uint32, "spell id", ...) where it existed only to check for the spell being known.
- Removed PlayerbotAI::HasAura() overload that takes a spell Id as an argument. It is rarely used, and the few instances of its usage are easily replaced with AC's Unit::HasAura(), which is the predominant usage for spell Id aura checks already anyway.
- When modifying DruidPullStrategy to use the new HasSpell() method, I went ahead and also simplified the file overall to pare down duplication (for example, the first check in CanCastSpell is for the spell Id so we don't need to separately check it before calling CanCastSpell()). I then did the same for other pull strategies so they can be consistent.
- Changed many magic numbers for spell Ids to be defined as compile-time constants. I didn't do this throughout the code, but I did in the files where I was making changes anyway due to the HasSpell() and/or HasAura() changes.

## Feature Evaluation

  - Minimum logic is described above.
  - Processing cost shouldn't change much. See Impact Assessment below.

## Impact Assessment

    - - [ ] No, not at all

  - Passing a string for the spell to check for it being known or the aura being present is more taxing than looking up a spell Id, although the usefulness of the string-based overload is for situations with spells that have multiple ranks.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

  - I think these changes should simplify maintenance complexity.

## AI Assistance

- - [ ] No

  - I had AI make many of the changes and then review since they were pretty mundane (for example, I'd tell it to just take all the magic numbers from a file and define them in an anonymous namespace, then replace the magic numbers with the spell constants). There isn't anything complicated in this PR; it was mostly busywork.

## Final Checklist

### DIFF
--- a/src/Ai/Class/Rogue/RogueTriggers.cpp
+++ b/src/Ai/Class/Rogue/RogueTriggers.cpp
@@ -9,6 +9,12 @@
 #include "Playerbots.h"
 #include "ServerFacade.h"
 
+namespace
+{
+constexpr uint32 SPELL_STEALTH = 1784;
+constexpr uint32 SPELL_SPRINT_RANK_1 = 2983;
+}
+
 // bool AdrenalineRushTrigger::isPossible()
 // {
 //     return !botAI->HasAura("stealth", bot);
@@ -29,7 +35,7 @@ bool UnstealthTrigger::IsActive()
 
 bool StealthTrigger::IsActive()
 {
-    if (botAI->HasAura("stealth", bot) || bot->IsInCombat() || bot->HasSpellCooldown(1784))
+    if (bot->HasAura(SPELL_STEALTH) || bot->IsInCombat() || bot->HasSpellCooldown(SPELL_STEALTH))
         return false;
 
     float distance = 30.f;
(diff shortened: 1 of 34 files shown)


src/Ai/Class/Rogue/RogueTriggers.cpp
```diff
@@ -29,7 +35,7 @@ bool UnstealthTrigger::IsActive()
 
 bool StealthTrigger::IsActive()
 {
-    if (botAI->HasAura("stealth", bot) || bot->IsInCombat() || bot->HasSpellCooldown(1784))
+    if (bot->HasAura(SPELL_STEALTH) || bot->IsInCombat() || bot->HasSpellCooldown(SPELL_STEALTH))
```
kadeshar, line 38: Stealth dont have ranks?

brighton-chi: Not in WotLK

kadeshar (approved):

PR #2438, opened 2026-06-02
Fix: Prevent user error when using lower than needed value for RandomBotAccountCount

NoxMax:
## Pull Request Description

The required number of accounts is now always calculated on server start. The manual value of RandomBotAccountCount is only used when it's >= the calculated value. This prevents user error of using lower than needed value, during initial account creation or after.

Moved `AddClassAccountPoolSize` and `RandomBotAccountPrefix` up. They belong with the other account creation configs. 

Minor comment edits in `CalculateAvailableCharsPerAccount`; no code change there.

## Feature Evaluation

Logic is at minimum to prevent this user error. Required number of accounts is now always calculated, but the calculation only happens on server start.

## How to Test the Changes

- Make sure to have a snapshot of your server as this testing involves account deletions, so that it's easier to revert things.
- The functional change is only in RandomPlayerbotFactory.cpp, so you can quickly copy/paste the whole of `CalculateTotalAccountCount` to run this test
1. First delete your accounts `DeleteRandomBotAccounts = 1`
2. Then reboot with 
```
AiPlayerbot.MaxRandomBots = 500
RandomBotAccountCount = 0
DeleteRandomBotAccounts = 0
AddClassAccountPoolSize = 10
```
server should create the correct number of accounts needed: (500/10) +10 = 60
3. Reboot with a `RandomBotAccountCount` value equal to 100, server should create for you 50 new accounts, for a total of 110.
4. Delete your accounts again.
5. Reboot with a `RandomBotAccountCount = 30` (which is 20 less than required), server should give you a terminal warning and proceed to create 50+10 accounts. The 500 randombots should login and behave normally.

## Impact Assessment

    - - [ ] No, not at all
Account calculations will always be done on boot. No effect on the rest of the runtime.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -89,10 +89,20 @@ AiPlayerbot.MinRandomBots = 500
 AiPlayerbot.MaxRandomBots = 500
 
 # Randombot accounts
-# If you are not using any expansion at all, you may have to set this manually, in which case please
-# ensure that RandomBotAccountCount is at least greater than (MaxRandomBots / 10 + AddClassAccountPoolSize)
+# Set to 0 to automatically calculate needed accounts based on MaxRandomBots, EnablePeriodicOnlineOffline
+# and its ratio, and AddClassAccountPoolSize. Set manually if you want to create more accounts than needed.
+# If the manual value is less than the required, the system will override it with the required value.
 # Default: 0 (automatic)
 AiPlayerbot.RandomBotAccountCount = 0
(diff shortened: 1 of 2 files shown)


NoxMax:
Got around to actually test it. It works as expected.

Celandriel (approved):

PR #2441, opened 2026-06-06
Explicit bot types

Celandriel:
## Pull Request Description

This changes how bot type definitions are determined. Instead of lookups, or potentially weak logic, we instead set a flag within the botAI. This eliminates the lookup for players (bots) with botAI. All else fall back to the lookup. 

Some performance details for the current approach.

Metric	Time
min	87 ns
median	7.6 µs
mean	12.5 µs
p90	33 µs
p99	69 µs
max	144 µs

THe min values were when first starting the server, as we loaded more bots this took longer and longer. 

## Feature Evaluation

## How to Test the Changes

Alts, addclass, and randombots still perform as expected. 

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

All bots who are loaded by a player are either alts or addclass bots. This is regardless of whether the bot is from an addclass labled account or not. 

    - - [x] No

## AI Assistance

- - [ ] No

Claude Opus. Implementation. USer reviewed. 

## Final Checklist

## Notes for Reviewers

In case you missed it before this defines all user summoned bots as either alts, or addclass.


src/Bot/PlayerbotAI.h
```diff
@@ -230,6 +230,15 @@ enum class BotTypeNumber : uint8
...
+enum class BotAccountType : uint8
+{
+    UNDEFINED = 0,
+    ALT,
+    RANDOMBOT,
+    ADDCLASS,
+    REALPLAYER  // self-bot
+};
+
 enum class GrouperType : uint8
```
NoxMax, line 242: Using the term "account" is a bit confusing with the pre-existing system that handles account creations and logins. Not to mention that a single account can have alts, self-bots, and regular player characters. Drop the term entirely, particularly that this is a per-bot related matter.

current | proposed
-- | --
BotAccountType | BotType
_accountType | _botType
GetAccountType() | GetBotType()
SetAccountType() | SetBotType()
BotAccountType::UNDEFINED | BotType::UNDEFINED

Also you accidentally posted to master. You also accidentally created the branch out of master so switching the target to test-staging would bring the merge commits from master 😅 but it should be fine.

src/Bot/PlayerbotMgr.cpp
```diff
@@ -1632,6 +1633,13 @@ void PlayerbotMgr::OnBotLoginInternal(Player* const bot)
     botAI->SetMaster(master);
     botAI->ResetStrategies();
 
+    // Any bot summoned off a randombot account (random pool or addclass) is
+    // treated as an addclass bot; only a player's own characters are alts.
+    uint32 botAccountId = bot->GetSession()->GetAccountId();
+    botAI->SetAccountType(sPlayerbotAIConfig.IsInRandomAccountList(botAccountId)
+                              ? BotAccountType::ADDCLASS
+                              : BotAccountType::ALT);
```
kadeshar: You shouldnt use IsAddclassBot?

src/Bot/PlayerbotAI.h
```diff
@@ -533,15 +542,18 @@ class PlayerbotAI : public PlayerbotAIBase
...
+    bool IsRandomBot() { return GetAccountType() == BotAccountType::RANDOMBOT; }
+    bool IsAddclass() { return GetAccountType() == BotAccountType::ADDCLASS; }
     // Bot has a master that is a player.
     bool HasRealPlayerMaster();
     // Bot has a master that is activly playing.
     bool HasActivePlayerMaster();
     // Get the group leader or the master of the bot.
     // Checks if the bot is summoned as alt of a player
-    bool IsAlt();
+    bool IsAlt() { return GetAccountType() == BotAccountType::ALT; }
```
kadeshar: There is possibility that bot will loose alt status in session? Like unlink linked account? That is covered?

src/Bot/PlayerbotAI.h
```diff
@@ -650,6 +662,7 @@ class PlayerbotAI : public PlayerbotAIBase
     Position jumpDestination = Position();
     uint32 nextTransportCheck = 0;
     bool spellInterruptRequested = false;
+    BotAccountType _accountType = BotAccountType::UNDEFINED;
```
kadeshar: This property initialization for REALPLAYER is confusing. Why in HandlePlayerbotCommand?

kadeshar: All methods Is... assume that _accountType is initialized. But can be called before that. Shouldnt be error or something like this?

src/Bot/PlayerbotAI.h
Celandriel: Not really. It's only set during bot login, so I don't think there is a way that it would change mid session

src/Bot/PlayerbotAI.h
Celandriel: So in this case it's because it's the best way to identify if a botAI is for a real player. The only way real players have AI is through self bot command.

Celandriel: It seemed like good practice to have a starting value. Maybe an error can be set in the getter.

src/Bot/PlayerbotAI.h
kadeshar: That not means that will resolve wrong because of that?

kadeshar:
Also that should be PR to test-staging right?

NoxMax (changes requested):

Celandriel:
Did some clean up and extra testing on this. I cant think of any weird edge cases right now.

src/Bot/PlayerbotAI.h
Celandriel: Right now, I dont think that an account that is unlinked would cause a loss of permissions over an altbot you already control, because they would already have you as master, and that behavior isnt changed by this. Really, its just intended to make the typing much explicit.

src/Bot/PlayerbotAI.h
kadeshar: Maybe GetBotType should raise error when called before proper initalization

kadeshar:
Conflict to resolve

Celandriel (commented):
> Maybe GetBotType should raise error when called before proper initalization 

Resolved.

kadeshar (approved):

Issue #2428, opened 2026-05-30
Restoration Druid Randombots issues grinding/killing mobs in open world


HoodOG1:
**Bug Description**
Restoration druid bots seem to have issues killing mobs while grinding/questing etc. The bots are constantly switching between tree form and human form, rooting their targets, moving a few steps back and then continue form switching. I noticed a few rare occasions where the bot started doing some damage casting wrath, but thats to like 1% the case. Noticed it in level ranges 52-62, cant tell if its different for other ranges.

**Commit Hash**
AzerothCore rev. e98e7a97e3f2+ 2026-05-29 13:10:13 -0700 (Playerbot branch)
Playerbots: [1041b0b0705e3347f766ce22f5799ca03836ccae](https://github.com/mod-playerbots/mod-playerbots/commit/1041b0b0705e3347f766ce22f5799ca03836ccae)

**How To Reproduce Bug**
Just find a restoration bot in the open world where i should grind/quest/level.

**Expected Behavior**
Should not be form switching constantly and actually try to kill the mobs.

**Modules**
|- mod-ah-bot
|- mod-groupquests
|- mod-player-bot-level-brackets
|- mod-playerbots
|- mod-quest-loot-party
|- mod-server-auto-shutdown
|- mod-skip-dk-starting-area
|- mod-transmog

**Additional Context**
I dont know if it could relate to [PR 2392](https://github.com/mod-playerbots/mod-playerbots/pull/2392), since i haven't updated the server/bots for a long while. But im sure that this wasn't the case in earlier versions. 

@ThePenguinMan96, could you validate that? :)

Already tried to completly reset bots/generate new bots, didnt solve the issue.

YunataSavior:
If you don't have a scheme to constantly pull the latest playerbots revision and rebuild, device a scheme.

kadeshar:
Pull latest master branch and reset randombot because there was druid overhaul. Give feedback after that

HoodOG1:
<img width="2412" height="1356" alt="Image" src="https://github.com/user-attachments/assets/71f88621-76ef-41ce-b47e-468dca2d274c" />

@kadeshar, the core the issue was built on already included the druid overhaul, thats why i linked [PR 2392](https://github.com/mod-playerbots/mod-playerbots/pull/2392) for further investigation.

I have updated the core and playerbots, resetted the randombots once more just to make sure, the issue still prevails.

A restoration bot still just shapeshifts continuously without even attacking the target.

[...]

PR #2453, opened 2026-06-08
Druid update hotfixes

ThePenguinMan96:
## Pull Request Description

Hello playerbots community! Through testing and feedback, two issues have been brought to my attention with the druid update:

1. **The blanketing strategy is messing with the healer dps/healer should attack function**, as outlined in issue #2428 . It's also clunky in small groups/solo play. I have decided to remove the blanketing strategy from the list of default resto strategies, and have it be applied with the ApplyInstanceStrategy, which currently governs these instances:

aq20, blacktemple, bwl, gruulslair, hyjal, icc, karazhan, magtheridon, moltencore, naxx, onyxia, ssc, tbc-ac, tempestkeep, ulduar, voa, wotlk-an, wotlk-cos, wotlk-dtk, wotlk-eoe, wotlk-fos, wotlk-gd, wotlk-hol, wotlk-hor, wotlk-hos, wotlk-nex, wotlk-occ, wotlk-ok, wotlk-os, wotlk-pos, wotlk-toc, wotlk-uk, wotlk-up, wotlk-vh, zulaman

So whenever these instance strategies are applied, if the bot is a resto druid, it will automatically apply blanketing. When they leave the instance/raid, it cleans up the strategy, removing it (similar to other raid/instance strategies).

2. **Prowl both existing in the combat strategy (catdruidstrategy) and noncombatstrategy has proven to be taxing/higher cpu usage than I would like** - @brighton-chi identified that prowl was high on the pmon output, and I confirmed locally that it was quite taxing. To fix this, I tested this by removing the prowl trigger/action from the non-combat strategy, and the numbers went down significantly. It's because it was constantly scanning for targets while not in combat, similar to how the pvp strategy does (not enabled by default, very taxing). Most rndbots aren't in combat, so this was scanning on all rndbots that were feral (cat) while out of combat. Now, it being only in the combat strategy is appropriate - the trigger returns early, and doesn't even scan if the bot is in combat. This is the prowl trigger:

Bot is not already prowling and not in combat
Bot has the Prowl spell and it's off cooldown
A valid target exists within engagement range (checks enemy player → grind target → dps target in priority order)
Distance check: 30y base, −10y if target has a victim, −10y more if target is also moving (floor 10y), +15y in BG or Arena

After I made this change locally, I tested it, and the only difference is that the cat will no longer go into prowl when outside of combat (walking around with master, unengaged to enemy target). It will still go into prowl when told to attack a target, or dps assist. We could potentially look at adding this back in, in a different way, when pvp becomes more of a focus. 

## Feature Evaluation

This reduces the processing cost of prowl and the blanketing strategy quite a bit, while still being functional in the situations that need those abilities/strategy.

## Impact Assessment

    - - [x] No, not at all

It reduces processing costs.

    - - [ ] No
    - - [x] Yes (**explain why**)

The resto druid will no longer blanket while in most 5-mans/solo/world/group play. This is fine, because the required heals for most of these instances is a lot lower than raids. 

The cat druid will no longer preemptively go into prowl before the enemy is set to be attacked. It was cool to see the cat go into prowl when getting close to a mob, but there was a lot of processing cost with this outside of combat, so it's being shelved.

    - - [x] No

## AI Assistance

- - [ ] No

I had it locate the applyinstancestrategy, and recommend the best way to inject the blanketing strategy specifically for resto druids. Fortunately the includes were already there for tab = resto druid, so it was just about adding a few lines of code.

## Final Checklist

### DIFF
--- a/src/Ai/Base/Actions/CheckMountStateAction.cpp
+++ b/src/Ai/Base/Actions/CheckMountStateAction.cpp
@@ -17,6 +17,7 @@
 #include "SpellAuraEffects.h"
 
 static constexpr uint32 SPELL_COLD_WEATHER_FLYING = 54197;
+static constexpr float PARACHUTE_LAND_THRESHOLD = 15.0f;
 
 // Define the static map / init bool for caching bot preferred mount data globally
 std::unordered_map<uint32, PreferredMountCache> CheckMountStateAction::mountCache;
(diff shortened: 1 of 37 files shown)


ThePenguinMan96:
I accidently merged into master instead of test staging. Reuploading. My bad!

PR #2454, opened 2026-06-08
Druid Update Hotfix

ThePenguinMan96:
## Pull Request Description

Hello playerbots community! Through testing and feedback, two issues have been brought to my attention with the druid update:

1. **The blanketing strategy is messing with the healer dps/healer should attack function**, as outlined in issue #2428 . It's also clunky in small groups/solo play. I have decided to remove the blanketing strategy from the list of default resto strategies. It is now an optional strategy only, enabled on resto druid by typing "co +blanketing" for those difficult healing fights.

2. **Prowl both existing in the combat strategy (catdruidstrategy) and noncombatstrategy has proven to be taxing/higher cpu usage than I would like** - @brighton-chi identified that prowl was high on the pmon output, and I confirmed locally that it was quite taxing. To fix this, I tested this by removing the prowl trigger/action from the non-combat strategy, and the numbers went down significantly. It's because it was constantly scanning for targets while not in combat, similar to how the pvp strategy does (not enabled by default, very taxing). Most rndbots aren't in combat, so this was scanning on all rndbots that were feral (cat) while out of combat. Now, it being only in the combat strategy is appropriate - the trigger returns early, and doesn't even scan if the bot is in combat. This is the prowl trigger:

Bot is not already prowling and not in combat
Bot has the Prowl spell and it's off cooldown
A valid target exists within engagement range (checks enemy player → grind target → dps target in priority order)
Distance check: 30y base, −10y if target has a victim, −10y more if target is also moving (floor 10y), +15y in BG or Arena

After I made this change locally, I tested it, and the only difference is that the cat will no longer go into prowl when outside of combat (walking around with master, unengaged to enemy target). It will still go into prowl when told to attack a target, or dps assist. We could potentially look at adding this back in, in a different way, when pvp becomes more of a focus. 

## Feature Evaluation

This reduces the processing cost of prowl and the blanketing strategy quite a bit, while still being functional in the situations that need those abilities/strategy.

## Impact Assessment

    - - [x] No, not at all

It reduces processing costs.

    - - [ ] No
    - - [x] Yes (**explain why**)

The cat druid will no longer preemptively go into prowl before the enemy is set to be attacked. It was cool to see the cat go into prowl when getting close to a mob, but there was a lot of processing cost with this outside of combat, so it's being shelved.

    - - [x] No

## AI Assistance

- - [ ] No

## Final Checklist

### DIFF
--- a/src/Ai/Class/Druid/Strategy/GenericDruidNonCombatStrategy.cpp
+++ b/src/Ai/Class/Druid/Strategy/GenericDruidNonCombatStrategy.cpp
@@ -22,7 +22,6 @@ public:
         creators["remove curse on party"] = &remove_curse_on_party;
         creators["abolish poison on party"] = &abolish_poison_on_party;
         creators["revive"] = &revive;
-        creators["prowl"] = &prowl;
         creators["aquatic form"] = &aquatic_form;
     }
 
@@ -90,14 +89,6 @@ private:
     {
         return new ActionNode("revive",
                               /*P*/ { NextAction("caster form") },
-                              /*A*/ {},
-                              /*C*/ {});
-    }
-
-    static ActionNode* prowl([[maybe_unused]] PlayerbotAI* botAI)
-    {
-        return new ActionNode("prowl",
-                              /*P*/ { NextAction("cat form") },
                               /*A*/ {},
                               /*C*/ {});
     }
@@ -189,10 +180,7 @@ void GenericDruidNonCombatStrategy::InitTriggers(std::vector<TriggerNode*>& trig
     if (specTab == DRUID_TAB_BALANCE || specTab == DRUID_TAB_RESTORATION)
         triggers.push_back(new TriggerNode("often", { NextAction("apply oil", 1.0f) }));
     if (specTab == DRUID_TAB_FERAL)
-    {
         triggers.push_back(new TriggerNode("often", { NextAction("apply stone", 1.0f) }));
-        triggers.push_back(new TriggerNode("prowl", { NextAction("prowl", ACTION_INTERRUPT) }));
-    }
 
 }
 
--- a/src/Bot/Factory/AiFactory.cpp
+++ b/src/Bot/Factory/AiFactory.cpp
@@ -351,7 +351,7 @@ void AiFactory::AddDefaultCombatStrategies(Player* player, PlayerbotAI* const fa
                 engine->addStrategiesNoInit("balance", "cure", "aoe", "cc", "dps assist", nullptr);
             }
             else if (tab == DRUID_TAB_RESTORATION)
-                engine->addStrategiesNoInit("resto", "cure", "dps assist", "blanketing", "tranquility", nullptr);
+                engine->addStrategiesNoInit("resto", "cure", "dps assist", "tranquility", nullptr);
             else
             {
                 if (player->HasSpell(SPELL_CAT_FORM) && !player->HasAura(SPELL_DRUID_THICK_HIDE))
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -1624,7 +1624,7 @@ void PlayerbotAI::ApplyInstanceStrategies(uint32 mapId, bool tellMaster)
 {
     static const std::vector<std::string> allInstanceStrategies =
     {
-        "aq20", "blacktemple", "bwl", "gruulslair", "hyjal", "icc", "karazhan",
+        "aq20", "blacktemple", "blanketing", "bwl", "gruulslair", "hyjal", "icc", "karazhan",
         "magtheridon", "moltencore", "naxx", "onyxia", "ssc", "tbc-ac", "tempestkeep",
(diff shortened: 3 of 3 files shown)


NoxMax:
Side note about branches because I saw you posted this to master first. If you click edit title, it will allow you to change the target branch, so you don't need to make a new PR.

ThePenguinMan96:
> Side note about branches because I saw you posted this to master first. If you click edit title, it will allow you to change the target branch, so you don't need to make a new PR.

Oof, this is amazing to know. Thanks for telling me. I looked around a fair bit too!

Celandriel:
Im not totally sure about adding the strategy to the instance strategies. Seems out of place.

ThePenguinMan96:
Totally understand. I was combing through to see how I could apply blanketing only to raids? It's not a very good strategy when by itself or with a few people, it's good in the group, but amazing in raids. 

It's also not addressing the root issue that blanketing conflicts with healer DPS. 

I'm happy to revisit the drawing board on this one. In this current PR, blanketing wouldn't be used in most five man content, and not at all otherwise. It does make the rdruid heal quite a bit more. 

That, and really only instance strategies should remain in that section of aifactory.

dillyns:
IMO blanketing should be removed from default druid strategies and just left as an option for the player to enable if they want to.

Dreathean:
> IMO blanketing should be removed from default druid strategies and just left as an option for the player to enable if they want to.

That's tough though because many players may never know about that strategy in that scenario and miss out on significantly improved resto druid performance in raids.

Would it not be possible to roll it into their overall resto strategy and have that portion only activate when in a raid? I guess that is an expensive check?

ThePenguinMan96:
Originally I tried this, but it was inconsistent with how strategies are run. If we have it say "use these two triggers when in a group of 5 or more", strategies are only initialized in certain scenarios -  "talents spec", "maintenance", reset botAI. Basically, it didn't effectively track if the druid was in a raid or not because strategies would initialize prior to changing group sizes.
I also tried adding checks literally in the triggers and actions themselves, but that was taxing on pmon - every time it went to blanket wild growth/rejuvenation, it would check group size. 

I think ultimately I am going to revisit this on my weekend, and make it play nice with healer DPS, and just have it be active all the time. Originally I made it a strategy because I was worried resto druids would go out of mana too often. But in testing, it really performed well. In low level 5 man groups, druid didn't even really have to drink often.

brighton-chi:
> > IMO blanketing should be removed from default druid strategies and just left as an option for the player to enable if they want to.
> 
> 
> 
> That's tough though because many players may never know about that strategy in that scenario and miss out on significantly improved resto druid performance in raids.
> 
> 
> 
> Would it not be possible to roll it into their overall resto strategy and have that portion only activate when in a raid? I guess that is an expensive check?

It wouldn't expect it to be an expensive check. But that type of framework doesn't currently exist.

brighton-chi:
> Originally I tried this, but it was inconsistent with how strategies are run. If we have it say "use these two triggers when in a group of 5 or more", strategies are only initialized in certain scenarios -  "talents spec", "maintenance", reset botAI. Basically, it didn't effectively track if the druid was in a raid or not because strategies would initialize prior to changing group sizes.
> I also tried adding checks literally in the triggers and actions themselves, but that was taxing on pmon - every time it went to blanket wild growth/rejuvenation, it would check group size. 
> 
> I think ultimately I am going to revisit this on my weekend, and make it play nice with healer DPS, and just have it be active all the time. Originally I made it a strategy because I was worried resto druids would go out of mana too often. But in testing, it really performed well. In low level 5 man groups, druid didn't even really have to drink often. 

Probably Druids should never use tree form when solo anyway. Maybe one possibility is to have the healer dps strategy disable blanketing. Since healer dps is by default auto removed in instances, that could be a way to automatically trigger blanketing.

Or is it as simple as having a trigger for blanketing-related actions be subject to a group check?

dillyns:
Strongly disagree with the idea of rolling blanketing into the core resto spec without a way to disable it.
As someone who runs individual progression at level 60, I had to immediately remove blanketing when it was added because it OOMed my resto druid real quick.
This is also why I dont think it should be autoadded from instance triggers. If a player has already disabled it for whatever reason, it shouldn't keep popping up on its own and have to constantly be removed everytime they enter an instance.
Also what is the point of blanketing for low level druids who dont even have revitalize yet? Unless there is alot of aoe, which is rare in vanilla dungeons, is it not just wasting mana spreading rejuvs on the group?

brighton-chi:
Wait... The parent blanketing class already returns nullptr if there is no group. So why is blanketing causing issue 2428?

ThePenguinMan96:
Not sure. Based on feedback and testing, I'm just going to make it an optional strategy. Won't have access to computer until a few days from now

ThePenguinMan96:
All right, just made a new commit that removes blanketing from applyinstancestrategies. It is strictly an optional (non-default) strategy the player can try out on healer-intensive fights. This PR is ready to be reviewed.

Celandriel (approved):

PR #2456, opened 2026-06-09
Implement remaining racials* + minor modifications to racials strategy

brighton-chi:
## Pull Request Description

*The title is a lie because I did not implement Cannibalize. I don't think that one is ever going to be worth the processing cost to find nearby corpses.

Anyway:
- Implemented Stoneform and PoisonDiseaseBleedTrigger.
- Implemented Escape Artist and MovementImpairedTrigger. The trigger excludes Stealth and Prowl, like with Hand of Freedom. I realize Gnomes cannot normally be Druids, but I have left Prowl in, in case somebody uses a mod to make it happen. I could easily be persuaded to remove the Prowl exclusion, though.
- Implemented Rogue version of Arcane Torrent, used based on the low energy trigger.
- Implemented DK version of Arcane Torrent; I don't know shit about DKs, and it doesn't look like there is any runic-energy-related trigger, so I just made this a generic boost trigger. That means it's used on cooldown while the boost strategy is active and the encounter trips the balance that causes boost trigger to fire. It's not great, but it's better than not using the ability at all.
- All racials, plus Lifeblood (which is also under the racials strategy), are gated behind HasSpell() checks. This stops bots from evaluating racials they don't have and reduces log spam.
- Removed the ActionNodeFactory for racials, which was used only for Lifeblood with Gift of the Naaru alternative. I split those abilities into their own TriggerNodes, behind their own spell checks.
- Increased threshold for Lifeblood and Gift of the Naaru use to medium health instead of low health. These are not strong heals, plus they are HoTs, so bots may as well get more use out of them. Gift of the Naaru can be used on allies as well, but I did not implement it (because of the effort and because I don't think it's worth the processing cost anyway for such an insignificant heal).
- Removed spell checks from EMFH and WoTF IsPossible() since they are checked before the trigger is evaluated.
- Removed IsUseful() overrides from EMFH and WoTF since the checks are already in the triggers. If we should have the checks done twice anyway, LMK.

## Feature Evaluation

## How to Test the Changes

I tested:
- EMFH with Fear
- Escape Artist with Entangling Roots
- Stoneform with Rupture
- Arcane Torrent (Rogue version)

I did not test the DK version of Arcane Torrent. I did not check all possible conditions for the CC breaks, but it can be done most easily by self-botting and using the .aura GM command. The racials strategy is a default combat strategy only so for testing at least you may want to add nc +racials to make it easier.

## Impact Assessment

    - - [ ] No, not at all

New triggers always add more processing, but this stuff is insignificant.

    - - [ ] No
    - - [x] Yes (**explain why**)

Racials is a default combat strategy so new racials/changes to existing racials modifies default behavior.

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

### DIFF
--- a/src/Ai/Base/Strategy/RacialsStrategy.cpp
+++ b/src/Ai/Base/Strategy/RacialsStrategy.cpp
@@ -4,45 +4,104 @@
  */
 
 #include "RacialsStrategy.h"
+#include "Playerbots.h"
 
-class RacialsStrategyActionNodeFactory : public NamedObjectFactory<ActionNode>
+namespace
 {
-public:
-    RacialsStrategyActionNodeFactory() { creators["lifeblood"] = &lifeblood; }
-
-private:
-    static ActionNode* lifeblood(PlayerbotAI* /*botAI*/)
-    {
-        return new ActionNode("lifeblood",
-                              /*P*/ {},
-                              /*A*/ { NextAction("gift of the naaru") },
-                              /*C*/ {});
-    }
-};
-
-void RacialsStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
-{
-    triggers.push_back(
-        new TriggerNode("low health", { NextAction("lifeblood", ACTION_NORMAL + 5) }));
-    triggers.push_back(
-        new TriggerNode("medium aoe", { NextAction("war stomp", ACTION_NORMAL + 5) }));
-    triggers.push_back(new TriggerNode(
-        "low mana", { NextAction("arcane torrent", ACTION_NORMAL + 5) }));
-
-    triggers.push_back(new TriggerNode(
-        "generic boost", { NextAction("blood fury", ACTION_NORMAL + 5),
-        NextAction("berserking", ACTION_NORMAL + 5),
-        NextAction("use trinket", ACTION_NORMAL + 4) }));
-
-    triggers.push_back(new TriggerNode(
-        "loss of control", { NextAction("every man for himself", ACTION_EMERGENCY + 1) }));
-
-    triggers.push_back(new TriggerNode(
-        "fear charm sleep", { NextAction("will of the forsaken", ACTION_EMERGENCY + 1) }));
-
+    constexpr uint32 SPELL_ARCANE_TORRENT_ENERGY = 25046;
+    constexpr uint32 SPELL_ARCANE_TORRENT_MANA = 28730;
+    constexpr uint32 SPELL_ARCANE_TORRENT_RUNIC_POWER = 50613;
+    constexpr uint32 SPELL_WAR_STOMP = 20549;
+    constexpr uint32 SPELL_BERSERKING = 26297;
+    constexpr uint32 SPELL_EVERY_MAN_FOR_HIMSELF = 59752;
+    constexpr uint32 SPELL_WILL_OF_THE_FORSAKEN = 7744;
+    constexpr uint32 SPELL_STONEFORM = 20594;
+    constexpr uint32 SPELL_ESCAPE_ARTIST = 20589;
 }
 
 RacialsStrategy::RacialsStrategy(PlayerbotAI* botAI) : Strategy(botAI)
 {
-    actionNodeFactories.Add(new RacialsStrategyActionNodeFactory());
+    // No custom ActionNodeFactory needed
+}
+
+void RacialsStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
+{
+    Player* bot = botAI->GetBot();
+
+    if (bot->HasSpell(SPELL_ARCANE_TORRENT_MANA))
+    {
+        triggers.push_back(new TriggerNode(
+            "low mana", { NextAction("arcane torrent", ACTION_NORMAL + 5) }));
+    }
+
+    if (bot->HasSpell(SPELL_ARCANE_TORRENT_ENERGY))
+    {
+        triggers.push_back(new TriggerNode(
+            "low energy", { NextAction("arcane torrent", ACTION_NORMAL + 5) }));
+    }
+
+    if (bot->HasSpell(SPELL_ARCANE_TORRENT_RUNIC_POWER))
+    {
+        // No low runic power trigger exists; this trigger should be modified if one is added
+        triggers.push_back(new TriggerNode(
+            "generic boost", { NextAction("arcane torrent", ACTION_NORMAL + 5) }));
+    }
+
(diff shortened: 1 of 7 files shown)


src/Ai/Base/Strategy/RacialsStrategy.cpp
```diff
@@ -4,45 +4,104 @@
...
+        triggers.push_back(new TriggerNode(
+            "generic boost", { NextAction("arcane torrent", ACTION_NORMAL + 5) }));
+    }
 
-}
+    if (bot->HasSpell(SPELL_WAR_STOMP))
+    {
+        triggers.push_back(new TriggerNode(
+            "medium aoe", { NextAction("war stomp", ACTION_NORMAL + 5) }));
+    }
```
ThePenguinMan96, line 54: It might be worth having the bot check it's shape shifted form. Cats and bears can't cast war stomp without shifting out of form. Might warrant a test. 

I worry if a bear is tanking a boss or something and the adds trigger war stomp during the filler part of the rotation.

src/Ai/Base/Strategy/RacialsStrategy.cpp
```diff
@@ -4,45 +4,104 @@
...
+    {
+        // Currently targets self only
+        triggers.push_back(new TriggerNode(
+            "medium health", { NextAction("gift of the naaru", ACTION_LIGHT_HEAL + 5) }));
+    }
+
+    if (botAI->HasSpell("lifeblood"))
+    {
+        triggers.push_back(new TriggerNode(
+            "medium health", { NextAction("lifeblood", ACTION_LIGHT_HEAL + 5) }));
```
ThePenguinMan96, line 102: These are going to help pvp and pve content alike. Great work. The dwarf rogue in the mirror with another rogue is gonna be OP.

src/Ai/Base/Trigger/GenericTriggers.cpp
```diff
@@ -491,6 +491,19 @@ bool FearSleepSapTrigger::IsActive()
...
+bool PoisonDiseaseBleedTrigger::IsActive()
+{
+    return botAI->HasAuraToDispel(bot, DISPEL_POISON) ||
+           botAI->HasAuraToDispel(bot, DISPEL_DISEASE) ||
+           bot->HasAuraWithMechanic(1 << MECHANIC_BLEED);
+}
+
+bool MovementImpairedTrigger::IsActive()
+{
+    return botAI->IsMovementImpaired(bot) &&
```
ThePenguinMan96, line 503: Does this check swimming status?

ThePenguinMan96 (approved):

src/Ai/Base/Strategy/RacialsStrategy.cpp
brighton-chi: Seems ok. I haven't seen any shifting out to use War Stomp.

src/Ai/Base/Trigger/GenericTriggers.cpp
brighton-chi: Good thought. I checked to see if swimming would trigger the ability, but it does not.

Celandriel (commented):
## Review: Implement remaining racials + racials-strategy changes

**Verdict: Request changes** — solid architectural direction (HasSpell-gating at init, splitting Lifeblood/Gift of the Naaru into their own TriggerNodes), but two blockers below.

### Blocking

**1. `botAI->HasSpell("...")` does not compile.** In `RacialsStrategy.cpp` the new `if (botAI->HasSpell("blood fury"))` / `"gift of the naaru"` / `"lifeblood"` blocks call a method that doesn't exist — `HasSpell` lives on `Player`/`Pet` and takes a `uint32` spell id, not a name string, and there's no `PlayerbotAI::HasSpell(std::string)` overload anywhere in the module. This fails under `-Werror`. The other racials in this same function already gate on id constants (`bot->HasSpell(SPELL_BLOOD_FURY)`, etc.) — mirror that here. The mix of id-constant gating for 9 spells and string lookups for 3 is itself a smell.

**2. Rogue Arcane Torrent is gated on a `"low energy"` trigger that isn't registered.** The `SPELL_ARCANE_TORRENT_ENERGY` block pushes a `TriggerNode("low energy", ...)`, but there is no `"low energy"` entry in `TriggerContext` and no `LowEnergyTrigger` class in the module (zero grep hits). `GetTrigger("low energy")` returns null and the node is silently skipped — so Rogue Arcane Torrent never fires. This contradicts the PR description ("used based on the low energy trigger") and means that path couldn't have been exercised in testing. Add a `LowEnergyTrigger` (+ `creators["low energy"]`) or reuse an existing energy-pool trigger. Note `"low mana"` is registered (used by the mana version); the energy equivalent just doesn't exist yet.

### Non-blocking

- **DK Arcane Torrent on `generic boost`** fires it on cooldown during any boost window and also silences/interrupts nearby casters as a side effect. Acceptable stopgap as you note, but worth an inline comment since it's the least predictable behavior here.
- **`poison disease bleed` → Stoneform at `ACTION_DISPEL` (50.0f).** In-band, but Stoneform is a 1-min CD that also grants armor; firing it at dispel priority for any trivial bleed/poison can preempt more valuable dispel-band actions and burn the CD. Consider a lower band or a magnitude check.
- **`MovementImpairedTrigger`** also fires for un-dispellable slows (the `GetSpeedRate(MOVE_RUN) < 1.0f` path). Escape Artist will just fail to cast, but it adds queue churn. Low impact.

### Positives

- Moving the `HasSpell(uint32)` gating into `InitTriggers()` (one-time, not per-tick) is the right call — it genuinely removes per-tick trigger evaluation for bots lacking the racial. Real perf win, not just log-spam reduction.
- Removing the `isUseful()` overrides and `isPossible()` spell checks on EMFH/WoTF is safe: the aura-type checks are exactly replicated by `LossOfControlTrigger`/`FearCharmSleepTrigger`, those actions are referenced only from RacialsStrategy, and the triggers are now only added when the bot has the spell. The base `CastSpellAction::isUseful()` still runs.
- New actions/triggers (`stoneform`, `escape artist`, `poison disease bleed`, `movement impaired`) all have matching `creators[]` entries and live in the correct generic dirs. The Stealth/Prowl exclusion mirrors the existing Hand of Freedom pattern.

Fix the two blockers and this is good to go.

brighton-chi:
The AI is wrong on both. Maybe it's looking at an outdated version of the mod.

Celandriel (approved):
Stupid AI

PR #2457, opened 2026-06-09
Add Death Talon Wyrmguard Strat to BWL

dillyns:
## Pull Request Description

Adds strategy for ranged and tank bots when fighting Death Talon Wyrmguards.
Ranged will stay 16 yards away so as not to get hit by war stomp.
Tanks that are actively tanking at least one wyrmguard will move away from other wyrmguards not they are not tanking.

The main issue I'm still having is healer positioning, sometimes the tanks will outrange the healers in their attempt to stay away from eachother, but imo the fight is already alot smoother.

## Feature Evaluation

## How to Test the Changes

Go to BWL and pull the Wyrmguard 3 packs.
Ranged should stay away from all Wyrmguards, and tanks should attempt to spread out the wyrmguards.
I still split the pull between tanks using rti targets and pull back, since the tanks done actively pull mobs off of eachother.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Claude was used to generate the tank actions that checks for nearby untanked wyrmguards. I have reviewed and refined claude output afterwards.

## Final Checklist

### DIFF
--- a/src/Ai/Raid/BWL/BWLActions.cpp
+++ b/src/Ai/Raid/BWL/BWLActions.cpp
@@ -51,3 +51,85 @@ bool BwlNefarianFearWardAction::Execute(Event /*event*/)
 
     return botAI->CastSpell("fear ward", victim);
 }
+
+// Trash
+
+static constexpr float WYRMGUARD_SAFE_DISTANCE = 16.0f;
+
+Unit* BwlDeathTalonWyrmguardTankMoveAwayAction::GetTarget()
+{
+    // Find a nearby wyrmguard within 16 yards being tanked by someone else
+    for (auto const& [guid, ref] : bot->GetThreatMgr().GetThreatenedByMeList())
+    {
+        Unit* unit = ref->GetOwner();
+        if (!unit || !unit->IsAlive() || unit->GetEntry() != NPC_DEATH_TALON_WYRMGUARD)
+            continue;
+        if (bot->GetDistance2d(unit) > WYRMGUARD_SAFE_DISTANCE)
+            continue;
+        Unit* victim = unit->GetVictim();
+        if (victim && victim != bot && victim->IsPlayer() && PlayerbotAI::IsTank(victim->ToPlayer()))
+            return unit;
+    }
+    return nullptr;
+}
+
+bool BwlDeathTalonWyrmguardTankMoveAwayAction::isUseful()
+{
+    if (!GetTarget())
+        return false;
+
+    // Must be actively tanking a wyrmguard before moving away from another tank's
+    for (auto const& [guid, ref] : bot->GetThreatMgr().GetThreatenedByMeList())
+    {
+        Unit* unit = ref->GetOwner();
+        if (unit && unit->IsAlive() && unit->GetEntry() == NPC_DEATH_TALON_WYRMGUARD && unit->GetVictim() == bot)
+            return true;
+    }
+    return false;
+}
+
+bool BwlDeathTalonWyrmguardTankMoveAwayAction::Execute(Event /*event*/)
+{
+    Unit* target = GetTarget();
+    if (!target)
+        return false;
+
+    float distToTravel = WYRMGUARD_SAFE_DISTANCE - bot->GetDistance2d(target);
+    if (distToTravel <= 0)
+        return false;
+
+    return MoveAway(target, distToTravel);
+}
+
+Unit* BwlDeathTalonWyrmguardRangedMoveAwayAction::GetTarget()
+{
+    // Find the closest wyrmguard within 16 yards
+    Unit* closest = nullptr;
+    float closestDist = WYRMGUARD_SAFE_DISTANCE;
+    for (auto const& [guid, ref] : bot->GetThreatMgr().GetThreatenedByMeList())
+    {
+        Unit* unit = ref->GetOwner();
+        if (!unit || !unit->IsAlive() || unit->GetEntry() != NPC_DEATH_TALON_WYRMGUARD)
+            continue;
+        float dist = bot->GetDistance2d(unit);
+        if (dist < closestDist)
+        {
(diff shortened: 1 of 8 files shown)


dillyns:
I tested this successfully with 20 yards safe distance, but am planning to test with 16 yards (the war stomp is 15) to see if they will still be safe and maybe minimize issues that come from the tanks being spread too far apart for healers.

Kyroth88:
Thank you so much for this work. I hate these fights!!

NoxMax:
16 yards may be too close if some rounding imprecision or whatever else causes some issue. 18 is bit more reasonable without a meaningful increase on spread distance.

brighton-chi:
It's unlikely to be an issue. In my experience, even no buffer works fine (i.e., 15 yards here). I don't remember the exact details, but playerbots has code that causes the distance to round up if it's just barely short.

Since there are two packs, I think it would be possible to hardcode 6 spots, designate 2 per tank, and select the nearest designated spot, which would probably create the most consistent positioning. It's probably not worth it for this situation, though.

Issues probably are just running up against inherent limitations of MoveAway(), which calls a direct move away from the target (though it will check for other angles if the move fails) and thus can be messy in a scenario with multiple bots trying to get away from each other. FleePosition() is a more dynamic approach and may be worth trying; it's a poor candidate for deadly AoEs but can result in better positioning overall.

brighton-chi:
Also, you of course don't have to do this, but you can automate the strategy more by having the tanks mark targets themselves. You could probably keep the threat list approach that you have but just build a vector and use that to assign Wyrmguards to marks and tanks.

Lurker's add phase has something like this where it builds a vector of the Coilfang Guardians with a "possible targets no los" grid search. In retrospect, that one should probably use deterministic sorting, though in practice maybe it's unlikely to make a difference.

dillyns:
> Also, you of course don't have to do this, but you can automate the strategy more by having the tanks mark targets themselves. You could probably keep the threat list approach that you have but just build a vector and use that to assign Wyrmguards to marks and tanks.
> 
> Lurker's add phase has something like this where it builds a vector of the Coilfang Guardians with a "possible targets no los" grid search. In retrospect, that one should probably use deterministic sorting, though in practice maybe it's unlikely to make a difference.

I didn't use marks because I actually have no idea how to get a tank to stick to a specific target beyond telling them to taunt and hoping it sticks.
It is my experience that tanks don't really care about rti target beyond the pull, assuming they are set to tank assist, but if that's wrong please let me know. It would make the fight smoother for sure.

brighton-chi:
> > Also, you of course don't have to do this, but you can automate the strategy more by having the tanks mark targets themselves. You could probably keep the threat list approach that you have but just build a vector and use that to assign Wyrmguards to marks and tanks.
> > Lurker's add phase has something like this where it builds a vector of the Coilfang Guardians with a "possible targets no los" grid search. In retrospect, that one should probably use deterministic sorting, though in practice maybe it's unlikely to make a difference.
> 
> I didn't use marks because I actually have no idea how to get a tank to stick to a specific target beyond telling them to taunt and hoping it sticks. It is my experience that tanks don't really care about rti target beyond the pull, assuming they are set to tank assist, but if that's wrong please let me know. It would make the fight smoother for sure.

You would need to use a multiplier to disable TankAssistAction for the encounter, and then there would need to be an action class that inherits from AttackAction to direct the tanks to attack their assigned targets. They should then stop trying to pick up the other Wyrmguards. I don't think you have to use marks if you're not going to direct DPS; I find it makes it easier for me to keep track of things, but I think you're right that it won't ultimately affect the tank behavior in and of itself.

I assume BwL doesn't have the scaffolding for multipliers right now, but that's really easy to copy from another instance strat. Then you could just do something like...

```
float BwLDeathTalonWyrmguardDisableTankAssistMultiplier::GetValue(Action* action)
{
    if (!botAI->IsTank(bot))
        return 1.0f;

    if (!AI_VALUE2(Unit*, "find target", "death talon wyrmguard"))
        return 1.0f;

    if (bot->GetVictim() != nullptr && dynamic_cast<TankAssistAction*>(action))
        return 0.0f;

    return 1.0f;
}
```
The GetVictim check isn't strictly necessary, but that would allow a tank to go attack another Wyrmguard after its Wyrmguard dies.

Celandriel (approved):

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -51,3 +51,85 @@ bool BwlNefarianFearWardAction::Execute(Event /*event*/)
...
+}
+
+bool BwlDeathTalonWyrmguardTankMoveAwayAction::Execute(Event /*event*/)
+{
+    Unit* target = GetTarget();
+    if (!target)
+        return false;
+
+    float distToTravel = WYRMGUARD_SAFE_DISTANCE - bot->GetDistance2d(target);
+    if (distToTravel <= 0)
```
Celandriel: Minor change. 
if (distToTravel <= 0.0f)

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -51,3 +51,85 @@ bool BwlNefarianFearWardAction::Execute(Event /*event*/)
...
+}
+
+bool BwlDeathTalonWyrmguardRangedMoveAwayAction::Execute(Event /*event*/)
+{
+    Unit* target = GetTarget();
+    if (!target)
+        return false;
+
+    float distToTravel = WYRMGUARD_SAFE_DISTANCE - bot->GetDistance2d(target);
+    if (distToTravel <= 0)
```
Celandriel: if (distToTravel <= 0.0f)

Celandriel (changes requested):

Celandriel (approved):

PR #2466, opened 2026-06-12
Tanks use RTI targets

dillyns:
## Pull Request Description

There are two changes in this PR.
Tank RTI Targetting:
Right now tanks don't really care about their RTI target other than when using "pull rti target"
With this change a tank who is given an RTI target will make keeping the rti target aggrod to them priority 1, even from other tanks. Regular tank assist targetting follows after.

"Main tank" ambiguous behaviour:
When running with multiple tank bots where no tank has main tank assignment, the "default" main tank (whichever tank comes first in the lookup) will taunt off of other tanks.
Imo this is ambiguous behaviour, since the player never actually set any main tank they would expect the tanks to never be taunting off of eachother.

## Feature Evaluation

## How to Test the Changes

RTI targets:
Get two tank bots and assign one bot an rti mark like "rti cross".
Go to a pack of at least two enemies and mark one with the rti mark. (I used the first pull in MC)
Mark skull on a different enemy. Both tanks will go and attack the skull.

After the change the rti mark tank will go and attack the marked target instead of skull.

Main tank behaviour:
This one is difficult to test because its hard to tell which tank is considered the default main tank.
Get two tank bots and do not mark any as main tank. 
Tell one tank to attack something, keeping the other tank on passive or flee until the first tank gets aggro. 
Send the second tank to attack.
If the second tank was the default main tank it will taunt the mob off the first tank. Otherwise reset and reverse the tank attack order to see the behavior.

After the change neither tank should tank off eachother.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Claude generated IsExplicitMainTank function. It was thoroughly reviewed and tested.

## Final Checklist

### DIFF
--- a/src/Ai/Base/Value/AttackerCountValues.cpp
+++ b/src/Ai/Base/Value/AttackerCountValues.cpp
@@ -22,7 +22,7 @@ bool HasAggroValue::Calculate()
     {
         return true;
     }
-    bool isMT = botAI->IsMainTank(bot);
+    bool isMT = botAI->IsExplicitMainTank(bot);
     if (victim &&
         (victim->GetGUID() == bot->GetGUID() || (!isMT && victim->ToPlayer() && botAI->IsTank(victim->ToPlayer()))))
     {
--- a/src/Ai/Base/Value/TankTargetValue.cpp
+++ b/src/Ai/Base/Value/TankTargetValue.cpp
@@ -104,6 +104,10 @@ public:
 
 Unit* TankTargetValue::Calculate()
 {
+    Unit* rtiTarget = RtiTargetValue::Calculate();
+    if (rtiTarget)
+        return rtiTarget;
+
     // FindTargetForTankStrategy strategy(botAI);
     FindTankTargetSmartStrategy strategy(botAI);
     return FindTarget(&strategy);
--- a/src/Ai/Base/Value/TankTargetValue.cpp
+++ b/src/Ai/Base/Value/TankTargetValue.cpp
@@ -9,6 +9,7 @@
 #include "AttackersValue.h"
 #include "Group.h"
 #include "PlayerbotAI.h"
+#include "Playerbots.h"
 
 class FindTargetForTankStrategy : public FindNonCcTargetStrategy
 {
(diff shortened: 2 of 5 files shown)


brighton-chi:
The default main tank is the last tank to join the raid. That's how it works with all the assistant roles too (last player/bot to join the raid that fits the criteria). That's a result of how AC's GetFirstMember() works. The iterative list works by inserting group members at the front as they join (so the function gets the first member on the list, but the list is populated in reverse order).

Re: RTI, is it right that tanks will still use default targeting if they have their RTI target aggroed? Returning the RTI target unconditionally in TankTargetValue::Calculate() doesn't make them stick to their RTI target no matter what? And I don't see any condition in RTITargetValue::Calculate() that would fit the bill.

Also, iirc skull is hardcoded to be an RTI target, though lower priority if you set a different RTI target. If you don't have any other RTI target set for bots but mark with skull, will the tanks also prioritize the skull? Because that would not be desirable imo--I use skull generically to direct DPS but would want tanks to continue doing their regular thing.

dillyns:
Added a fix to reach the fallback behavior when rti target is already aggroed, and added a guard when two tanks have the same rti set so they dont keep taunting off eachother (like when they are all set to skull by default for example)

Re: the default main tank. I understand how it is chosen and where it is useful, but the problem I have is the unexpected behavior that a player will experience because of it regarding tank taunting.
If I'm a player who knows nothing of the codebase and I have multiple tanks with none of them set as main tank explicity in the game raid menu, I would expect that all the tanks are considered equal. Essentially a tank free for all, whoever generates the most threat is the tank and tanks don't taunt off eachother. 
However because of the default main tank, which a player has no awareness of without reading the codebase, they will see that one of their tanks, the default main tank, keeps taunting off the other tanks.

The fix included in this PR just checks for an explicit main tank that the player set in the raid menu before it enables the main tank taunting behavior, so a player should never be confused as to why they are seeing tanks taunt off other tanks.

brighton-chi:
I agree with you. I was just explaining how the default group iteration works. I personally wouldn't mind going even further and having tanks never try to take aggro from other tanks, even the explicit main tank, but I digress.

Is your intent only to stop the aggro grabbing for non-explicit main tanks, or to more broadly treat them the same as other tanks? The latter makes more sense to me if the premise is that a player has no idea that there is an actual code-designated main tank. But right now, I think the result of the PR is a sort of hybrid approach for a non-explicit main tank. 

While the non-explicit main tank won't try to take aggro from other tanks after these changes, I think it will still try to stick to its own target instead of picking up adds (just like an explicit main tank) since no changes were made in TankTargetValue. I always found that behavior irritating, particularly in 10-player raids where you have just 2 tanks. In that case, you end up with a tank that picks up one mob and attacks it while your other tank runs around trying to pick up everything else. It's stupid behavior, and I think if this PR is overhauling the invisible MT concept (which I think is a good thing to do), we should just go all the way.

dillyns:
I only really thought of the aggro-grabbing issue  when making the PR, but I definately agree with you about also changing the tank targeting behavior to also check for explicit main tank being set. I'll look into adding that as well to these changes.

Celandriel (dismissed):

Celandriel:
> I only really thought of the aggro-grabbing issue when making the PR, but I definately agree with you about also changing the tank targeting behavior to also check for explicit main tank being set. I'll look into adding that as well to these changes.

I saw this after. are you still working on it?

dillyns:
Yes I do want to make the change that I said here before calling this complete

> also changing the tank targeting behavior to also check for explicit main tank being set. I'll look into adding that as well to these changes.

Celandriel (approved):

PR #2469, opened 2026-06-12
fix: Strip any [BRACKET] prefix from chat commands

scarecr0w12:
## Pull Request Description
Fixes an issue where chat commands prefixed with bracket-enclosed text (e.g., `[VIP]`, `[MOD]`, `[ADMIN]`) fail to parse and execute. Commands like `lfg 5`, `lfg 10`, `lfg 25`, and `lfg 40` now work correctly even when prefixed.

This is a generic solution that handles any future bracket-enclosed prefix without requiring code changes.

## Feature Evaluation

- **Minimum logic required:** Check if command text starts with `[`, find the matching `]`, and strip everything up to and including it. Then trim whitespace.
- **Processing cost:** Negligible. This logic executes once per chat command in the `HandleCommand()` method, using simple string operations (`find()` and `substr()`). No loops or complex processing. Cost is constant time O(n) where n is the length of the prefix (typically 3-6 characters).

## How to Test the Changes

1. Ensure a bot is configured to receive commands via VIP chat (or any bracket-prefixed channel).
2. Send commands like:
   - `[VIP] lfg 5`
   - `[VIP] lfg 10`
   - `[VIP] lfg 25`
   - `[VIP] lfg 40`
3. Expected behavior: Commands execute correctly and the bot joins the appropriate LFG queue.
4. Test with other bracket prefixes (e.g., `[MOD]`, `[ADMIN]`) to verify generic handling.

## Impact Assessment

  - ✅ **No, not at all** — The change adds only a single conditional check and string operations that execute once per chat command, not per tick.

  - ✅ **No** — This only fixes command parsing; no behavioral changes.

  - ✅ **No** — The fix is simple and handles all bracket prefixes generically.

## AI Assistance
- ✅ **No** — Manual implementation and testing.

## Final Checklist

- ✅ Stability is not compromised.
- ✅ Performance impact is understood, tested, and acceptable.
- ✅ Added logic complexity is justified and explained.
- ✅ No new bot dialogue lines added.
- ✅ No documentation changes needed.

## Notes for Reviewers

This fix addresses a real-world issue where VIP/MOD/ADMIN chat prefixes prevent command parsing. The solution is intentionally generic to future-proof against additional prefix types. The code follows AzerothCore style guidelines and passes all codestyle checks.


Celandriel:
My preference for this is to be a configurable with specific targeted strings.

scarecr0w12:
> My preference for this is to be a configurable with specific targeted strings.

Could set up a configurable regex? i just figured that would be too complex and most would use the bracketing setup.

Celandriel:
I think what I was trying to balance was not adding a search to every chat command, while also keeping it flexible. Similar to 
```
# Configure message prefixes which will be excluded in analysis in trade action to open trade window
AiPlayerbot.TradeActionExcludedPrefixes = "RPLL_H_,DBMv4,{звезда} Questie,{rt1} Questie"
```
So in the case where you dont actually use those prefixes it doesnt impact you. 
Im not totally sure what the best design.

kadeshar:
At the moment, we don’t plan to move forward with this change or merge it into the project, so I’m going to close the PR.

PR #2472, opened 2026-06-14
Add config to announce no reagents to buff

brighton-chi:
## Pull Request Description

Related to what was discussed with @kadeshar during the group buff/greater blessing rewrite PR. This adds a general announcement that bots will make when they try to cast a group buff or greater blessing but don't have the requisite ingredients and instead cast the single version (the announcement is not made unless the single version is successfully cast). The announcement is enabled by default, but I'm fine with disabling by default too. I have no real opinion there. The frequency of the announcement can be configured also and is 5 minutes by default.

I also set up the framework for the message to be translatable. I deleted an old sql that was related solely to the prior announcement system. There are other sqls that add the lines for the prior announcement system, but I didn't want to partially edit old sqls since that seems like a bad path. Also, I did not renumber entries so there are now some unused entries in the tables.

## Feature Evaluation

The announcement is made through GroupBuffSpellAction::Execute() for group buffs and CastGreaterBlessingAssignmentAction::Execute() for greater blessings. Helpers were added to get the announcements and handle the delay and pruning of the associated map.

Announcements are gated behind the bot actually attempting the group/greater variant so are controlled by the respective configs (disabled, raid only, group only).

## How to Test the Changes

Enable AutoGreaterBlessings and/or AutoPartyBuffs. Make bots destroy reagents if they have them. Check for announcements when they cast single variants. Check to see if gap between announcements matches MissingBuffReagentMessageCooldown.

## Impact Assessment

    - - [ ] No, not at all

There is a map used to throttle the announcements, but it's small and gets pruned whenever the next announcement is made or the bot can cast the greater buff again.

    - - [ ] No
    - - [x] Yes (**explain why**)

Announcements are made by default.

    - - [ ] No

Translations will be needed.

## AI Assistance

- - [ ] No

I'll be honest--this one was very much a "needed to be done but also super uninteresting" PR. So I totally took a backseat to AI here.

## Final Checklist


data/sql/playerbots/updates/2025_09_17_00_paladin_buff_reagent_texts.sql
NoxMax, line 1: Don't delete old SQL files that have been merged. Ever.
The worldserver will look at the updates table in the DB and see that there's a missing file, and it will cry about it.
Always create new files that amend the old ones. Never delete or edit old ones.

NoxMax (changes requested):

data/sql/playerbots/base/ai_playerbot_texts.sql
```diff
@@ -1725,10 +1725,7 @@ INSERT INTO `ai_playerbot_texts` (`id`, `name`, `text`, `say_type`, `reply_type`
 	(1705, 'cast_spell_command_error_failed', 'Failed to cast %spell', 0, 0, '', 'J’ai échoué à lancer %spell.', 'Konnte %spell nicht wirken.', '施放 %spell 失败了！', '', 'Fallo al lanzar %spell', '', 'Не удалось кастовать %spell.'),
 	(1706, 'cast_spell_command_amount', ' |cffffff00(x%amount left)|r', 0, 0, '', ' |cffffff00(x%amount restant)|r.', ' |cffffff00(x%amount übrig)|r', ' |cffffff00(剩余 x%amount)|r', '', ' |cffffff00(x%amount restante)|r', '', ' |cffffff00(x%amount осталось)|r'),
 	(1707, 'dummy_end', 'dummy', 0, 0, '', 'dummy', 'Dummy', '', '', '', '', 'dummy'),
-	(1708, 'rp_missing_reagent_greater_blessing', 'By the Light... I forgot my Symbols of Kings. We’ll make do with %base_spell!', 0, 0, '', 'Par la Lumière... J\'ai oublié mes Symboles du roi. On se contentera de %base_spell !', 'Beim Licht… Ich hab meine Königssymbole vergessen. Na gut, dann nehmen wir eben %base_spell!', '', '', '', '', ''),
```
NoxMax, line 1728: Same here. An edit changes the SQL hash, and now there's a mismatch with the hash of the old file. It doesn't break things, but it's bad practice.

data/sql/playerbots/updates/2025_09_17_00_paladin_buff_reagent_texts.sql
brighton-chi: That's not difficult at all to resolve. Is that really what AC does, keep old sqls around that are completely dead code?

Celandriel: I think they consolidate them after a while, which I think means that they must also remove the entries from the update table.

NoxMax: The files do get consolidated, and the individual ones are moved to the archives https://github.com/azerothcore/azerothcore-wotlk/blob/80fd8c057a5e175eb622d2cf53aba105b69792d8/src/server/database/Updater/UpdateFetcher.cpp#L288
I think the do it once a year, for files that are from the year before last. So earlier this year they archived 2024. Presumably next January they archive 2025.

brighton-chi: Ah, makes sense.

data/sql/playerbots/base/ai_playerbot_texts.sql
brighton-chi: Query: To bring this back, should I amend the initial commit instead of doing a new one? Will that change the scenario?

NoxMax: On your branch group-buff-texts, copy back the original `ai_playerbot_texts.sql` and `2025_09_17_00_paladin_buff_reagent_texts.sql` files as they were, commit that, and then squash that commit with your first commit. The diff would then be as if you have not touched those files at all.
Then modify the new `2026_06_13_00_ai_playerbot_missing_group_buff_reagent.sql` to have all the needed SQL changes for your PR, and commit the modification.
Finally force push to github.

NoxMax: PS: I use GitHub desktop to do all my branch operations. Cherry-pick, reorder, squash, amend/undo last commit. It makes things easy and clear.

brighton-chi: Yeah, that's what I use too. I operate in the world of GUIs only. Too amateur to go any deeper.

Hopefully this worked.

NoxMax: I don't think it's about being an amateur, despite what some would say. There are people who would kill me for saying this, but GUI > CLI, and I would die on that hill. If I have to do it with CLI, I will, but if I can do it with a GUI, I would do that every time.

data/sql/playerbots/updates/2026_06_13_00_ai_playerbot_missing_group_buff_reagent.sql
```diff
@@ -0,0 +1,29 @@
...
+  'rp_missing_reagent_generic',
+  'missing_group_buff_reagent'
+);
+
+INSERT INTO ai_playerbot_texts
+  (name, text, say_type, reply_type, text_loc1, text_loc2, text_loc3, text_loc4, text_loc5, text_loc6, text_loc7, text_loc8)
+VALUES
+  ('missing_group_buff_reagent',
+   'I am out of reagents for %group_spell and am casting %base_spell instead.',
+   0, 0, '', '', '', '', '', '', '', '');
```
kadeshar, line 24: Texts should be INSERTed with specific id.

src/Ai/Base/Util/GenericBuffUtils.cpp
```diff
@@ -205,9 +205,67 @@ namespace ai::buff
...
+        };
+
+        std::string const message = PlayerbotTextMgr::instance().GetBotTextOrDefault(
+            "missing_group_buff_reagent",
+            "I am out of reagents for %group_spell and am casting %base_spell instead.",
+            placeholders);
+
+        Group* group = bot->GetGroup();
+        const bool announced =
+            group->isRaidGroup() ? botAI->SayToRaid(message) : botAI->SayToParty(message);
```
kadeshar, line 255: group should have null check

kadeshar (changes requested):

data/sql/playerbots/updates/2026_06_13_00_ai_playerbot_missing_group_buff_reagent.sql
brighton-chi: Do you want id specified for ai_playerbot_texts_chance also? It looks like prior sqls did not do that.

kadeshar: All update scripts after 2025_10_09_00_ai_playerbot_texts_fix.sql should have id in INSERT

brighton-chi: It doesn't look like that has been done for ai_playerbot_texts_chance. It's not in 5/22/26, 5/19/26, 4/28/26, etc.

Is it going to be problematic if we start hardcoding the ids now when the prior ones have not been?

kadeshar: Aaaaa im blind. Only ai_playerbot_texts need ids, ai_playerbot_texts_chance not.

kadeshar (approved):

PR #2481, opened 2026-06-15
Fix re-entrant Engine::Init/Reset use-after-free on the bot AI tick

### LINKED ISSUE #2474: UAF / heap-corruption on map-update threads under high bot count + real-player combat (analysis + proposed fixes for two)
## Environment
- mod-playerbots @ `1041b0b0` (master), AzerothCore WotLK 3.3.5a, Win64 / MSVC / RelWithDebInfo
- ~3000 random bots, `MapUpdate.Threads = 6` (parallel map updates)
- Symbolicated minidumps + matching PDB; WER LocalDumps armed

## Summary
Under sustained high bot count — and especially when a **real player is in active combat among bots** — we hit a family of use-after-free / heap-corruption crashes on the MapUpdate worker threads. They surface in different subsystems (engine value lookup, spell-effect iteration, collision, unit searches) but share one root: a bot-related object is **accessed after free, or mutated from a foreign thread, during the AI tick.** Frequency: ~1 crash / 7h bots-only; **minutes** under active player combat. Three representative cases below — two we root-caused and fixed (fixes described, PRs available), one still open with a clean stack.

---

### 1) `AiObjectContext` value-context UAF in `Engine::ProcessTriggers` — STILL OPEN
Symbolicated stack:
```
std::_Hash<...>::_Find_last            ; AV, RAX=0
AiObjectContext::GetUntypedValue       ; AiObjectContext.cpp:121  (valueContexts.GetContextObject(name, botAI))
Trigger::GetTargetValue
RakeTrigger::IsActive
Engine::ProcessTriggers                ; Engine.cpp:467
PlayerbotAI::UpdateAI
Player::Update                         ; MapUpdate worker thread
```
A bot's `AiObjectContext` value container is freed/corrupted while `ProcessTriggers` walks it mid-tick (the hash lookup dereferences a null/garbage bucket). Not guardable at the call site (corrupt container). Occurs with **no other players online**. Sometimes it manifests instead as a stack-cookie fail-fast (`0xC0000409`) when the corrupted walk jumps through a bad pointer. **We can provide a full-memory dump.**

---

### 2) `HandleBotOutgoingPacket` runs on the *sending* thread (data race) — FIXED here
`WorldSession::SendPacket` → `OnPlayerbotPacketSent` → `PlayerbotAI::HandleBotOutgoingPacket` executes **synchronously on whatever thread sends the packet.** With parallel map updates, cross-map group/guild/chat broadcasts and bot→bot whispers therefore mutate a bot's AI state from a *foreign* thread:
- `PacketHandlingHelper`'s bare `std::stack` queue (no lock), and
- inline opcode handling (`SMSG_MESSAGECHAT` etc.) that touches `AiObjectContext` values + chat queues,
…racing the bot's own map-thread `UpdateAI`. Result: corrupted `unordered_map` (AV) or stack-cookie smash (fail-fast, no dump).

**Fix we applied:** (a) mutex `PacketHandlingHelper`'s queue — push under lock; `Handle()` swaps the queue out under lock then processes (handlers may re-enter `AddPacket` on the same thread); (b) defer the inline opcode handling into a per-bot list drained at the start of `UpdateAI` on the bot's own map thread. Stability improved markedly. **PR available.**

---

### 3) `LogoutPlayerBot` called from a map worker thread — FIXED here
`PlayerbotAI::UpdateAIInternal` (and other AI paths) call `PlayerbotHolder::LogoutPlayerBot` directly → `WorldSession::LogoutPlayer(true)` + `delete session`, freeing the bot's `Player`/`PlayerbotAI`/`WorldSession` **from a map worker thread** while other map threads may still be ticking that player, its auras or pets (the core mandates logout only in `World::UpdateSessions`).

**Fix we applied:** defer `LogoutPlayerBot` to the world thread via the existing `PlayerbotWorldThreadProcessor` (a new `BotDeferredLogoutOperation`) when called off the world thread; run inline (unchanged) on the world thread. **PR available.**

---

### Already on your side — thanks
We backported `571735cd` (null `spellInfo` in `TogglePetSpellAutoCastAction`) and `af0c5e7c` (`TellMaster` null-master) — both were **live** crashes here (`RandomBotSayWithoutMaster = 0` makes all random bots masterless). Appreciated.

### Repro
High bot count + parallel map updates + (ideally) a real player fighting among bots.

### Offer
Symbolicated stacks, full-memory dumps, and PRs for #2 and #3 on request.

timoinglin:
## Pull Request Description

Fixes the recurring playerbot use-after-free that takes the worldserver down on the map update threads. Root cause is a re-entrancy bug in the AI engine: `Engine::Reset()` deletes the `triggers` / `multipliers` / `queue` containers, and it is reached through `Engine::Init()` (addStrategy / removeStrategy / removeAllStrategies / ChangeStrategy / ResetStrategies). Some actions call `ChangeStrategy` on their own engine from inside `Execute()`, which runs while `DoNextAction` / `ProcessTriggers` is still iterating those same containers. When the changed strategy is the one currently ticking, `Reset()` frees the containers mid-iteration -> dangling `TriggerNode*` / `Multiplier*` -> heap corruption. It reproduces with `MapUpdate.Threads = 1`, so it is single-thread re-entrancy, not a thread race. Full write-up + a symbolicated crash dump are on #2474.

## Feature Evaluation

- Minimum logic: a depth counter. `DoNextAction` and `ExecuteAction` push a small RAII `TickScope` that increments `tickDepth`. While `tickDepth > 0`, `Init()` sets `initPending = true` and returns instead of running `Reset()`; the rebuild runs once when the outermost tick unwinds.
- Processing cost: negligible. One integer increment/decrement per engine tick, no allocation, no per-bot state. In the re-entrant case it does slightly less work, because a mid-tick rebuild becomes a single deferred one.

## How to Test the Changes

- High random-bot count with combat specs active (rogue and feral druid are the most reliable; idle fishing bots and arena exits also hit it).
- Optionally set `MapUpdate.Threads = 1` - it still crashes, which is what proves it is re-entrancy and not a race.
- Before: ACCESS_VIOLATION on a map worker thread in the AI engine (e.g. `Engine::ProcessTriggers`, `AiObjectContext::GetUntypedValue`, `Engine::MultiplyAndPush`); ~1 crash per several hours bots-only, minutes with a real player fighting among bots.
- After: no crash on those paths. Quick confirm without waiting for a crash: temporarily `LOG_ERROR` in `Engine::Init()` when `tickDepth > 0` - it lights up on any rogue/feral/fishing session, showing the re-entrant `Reset()` was firing.

## Impact Assessment

    - - [x] No, not at all

One int inc/dec per tick via RAII, no allocation, no per-bot data; if anything it reduces work in the re-entrant case.

    - - [x] No
    - - [ ] Yes (**explain why**)

Behavior is identical; a strategy change requested from inside a tick rebuilds one tick later instead of immediately, which is not observable in bot behavior.

    - - [ ] No

Minimal: one `if (tickDepth > 0)` guard in `Init()` plus a small RAII `TickScope`. Two files (`Engine.cpp` / `Engine.h`), commented.

## AI Assistance

- - [ ] No

Yes. I run a private WotLK server and used an AI assistant for the crash-log analysis, root-causing this re-entrancy bug, and drafting the patch. I reproduced the crash myself (single-threaded, `MapUpdate.Threads = 1`, which is what proves it is re-entrancy and not a thread race), reviewed the change, and it is running live on my server now. It is small and self-contained - a depth counter that defers `Init()`'s rebuild while a tick is on the stack.

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (N/A - no new dialogue)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (N/A - no conf/command changes)

## Notes for Reviewers

The full root-cause analysis, the in-tick offenders found (`CheckStealthAction`, the `FishingAction` family, `ArenaTactics`), and a symbolicated crash dump are on #2474. The fix is running on my live server now.

### DIFF
--- a/src/Bot/Engine/Engine.cpp
+++ b/src/Bot/Engine/Engine.cpp
@@ -114,8 +114,29 @@ void Engine::Reset()
     actionNodeFactories.creators.clear();
 }
 
+Engine::TickScope::TickScope(Engine* engine) : engine(engine) { ++engine->tickDepth; }
+
+Engine::TickScope::~TickScope()
+{
+    // When the outermost tick unwinds, apply any Init() that was deferred during it.
+    if (--engine->tickDepth == 0 && engine->initPending)
+    {
+        engine->initPending = false;
+        engine->Init();
+    }
+}
+
 void Engine::Init()
 {
+    // WL: never delete/rebuild the triggers/multipliers/queue while a tick is iterating them
+    // (a strategy change requested from inside an Action::Execute / Trigger::Check). Defer the
+    // rebuild to the TickScope at the bottom of the tick instead. See Engine.h for the rationale.
+    if (tickDepth > 0)
+    {
+        initPending = true;
+        return;
+    }
+
     Reset();
 
     for (std::map<std::string, Strategy*>::iterator i = strategies.begin(); i != strategies.end(); i++)
@@ -140,6 +161,8 @@ void Engine::Init()
 
 bool Engine::DoNextAction(Unit* /*unit*/, uint32 /*depth*/, bool minimal)
 {
+    TickScope tickScope(this);  // WL: defer any mid-tick Init() until this tick unwinds (re-entrancy UAF fix)
+
     LogAction("--- AI Tick ---");
 
     if (sPlayerbotAIConfig.logValuesPerTick)
@@ -304,6 +327,8 @@ bool Engine::MultiplyAndPush(
 
 ActionResult Engine::ExecuteAction(std::string const name, Event event, std::string const qualifier)
 {
+    TickScope tickScope(this);  // WL: defer any mid-tick Init() until this unwinds (re-entrancy UAF fix)
+
     bool result = false;
 
     ActionNode* actionNode = CreateActionNode(name);
--- a/src/Bot/Engine/Engine.h
+++ b/src/Bot/Engine/Engine.h
@@ -104,6 +104,24 @@ private:
     void LogAction(char const* format, ...);
     void LogValues();
 
+    // WL: re-entrancy guard (single-thread use-after-free fix).
+    // An Action::Execute() or Trigger::Check() can call ChangeStrategy / addStrategy /
+    // removeStrategy / removeAllStrategies / Init on THIS engine while DoNextAction /
+    // ProcessTriggers is still iterating triggers/multipliers/queue. Init() -> Reset()
+    // then deletes the very containers being iterated -> dangling pointers -> heap
+    // corruption (the recurring bot-AI crash; reproduced even single-threaded, proving
+    // it is re-entrancy, not a thread race). Fix: while a tick is on the stack
+    // (tickDepth > 0) Init() only flags initPending and returns; the rebuild runs once,
+    // after the outermost tick unwinds.
+    int tickDepth = 0;
+    bool initPending = false;
+    struct TickScope
+    {
+        explicit TickScope(Engine* engine);
+        ~TickScope();
+        Engine* engine;
+    };
+
     ActionExecutionListeners actionExecutionListeners;
 
 protected:


Celandriel:
You must use the PR template.

timoinglin:
Done, filled in the template and rebased onto test-staging. (The windows-build red is the runner failing to install the VS toolchain in Configure OS — looks like a CI hiccup, not the change; linux/macos build fine.)

NoxMax:
A couple of us have been trying to crash our servers both on Linux and Windows based on your initial parameters, without success. Which is a good thing, but also a bad thing if there is genuinely a unique situation here that can cause a crash.
What are any other AC mods you are running, and could you please share both of your worldserver.conf and playerbots.conf so we can see your full settings.

noisiver:
I'd also like to know what you are doing when it crashes. Are you playing? Is that a requirement? I've had a Windows system running now for 3 hours, 40 minutes with 2000 random bots and map update threads set to 4 and no crashes. I have them automatically join battlegrounds and arenas and I can see that they are doing that just fine.

timoinglin:
Mods: mod-playerbots, mod-ah-bot, mod-individual-xp, and our own in-house module (mod-wowlegends). mod-wowlegends is just LLM bot-chat + a hardcore mode + a couple of . chat commands — no strategy/engine code, no playerbots headers, and it never runs on the bot AI tick, so it isn't involved here. The other two are stock.

Configs attached (worldserver.conf + playerbots.conf), DB passwords stripped.

@noisiver — a real player fighting among the bots is by far the fastest trigger for us (minutes). Bots-only it's much rarer, order of hours, so a 3h40m bots-only run staying up matches what we see — so no, playing isn't a hard requirement, just a big accelerant. My read: a live player forces nearby bots into the combat/stealth/transition states that fire the offender actions — CheckStealthAction (rogue stealth toggles in combat), the FishingAction family, ArenaTactics on arena exit — and each calls ChangeStrategy on its own ticking engine, which is the re-entrancy. Wandering and auto-BG bots hit those paths a lot less. It also still reproduces at MapUpdate.Threads=1, so it isn't a thread race.

Been running the fix live since yesterday (500 bots), no sign of the crash since.


[playerbots_sanitized.conf.txt](https://github.com/user-attachments/files/29015256/playerbots_sanitized.conf.txt)
[worldserver_sanitized.conf.txt](https://github.com/user-attachments/files/29015257/worldserver_sanitized.conf.txt)

noisiver:
We need to reproduce this before we proceed with this pull request. The bots running arena on their own should still trigger whatever happens when they leave the arena. They have been running arenas for 4 hours now and nothing. If something happened after the arena match ends, you'd think it would've happened by now, wouldn't you?

noisiver:
I grabbed 4 altbots, formed a 5v5 team and will use selfbot to run some arena matches. If it doesn't crash after like 10 matches I will say I can't reproduce it.

timoinglin:
@noisiver here's exactly how I trigger it: go to the human starting zone where there's a big cluster of low level bots, and just kill mobs right in the middle of them. The more bots packed in there, the faster it crashes. A real player fighting inside a dense bot pack is what does it for me - minutes, not hours. Wandering/arena bots barely hit it, which is probably why your runs stay up.

Also no crash doesn't mean it's not there. It's a use-after-free, not a guaranteed crash. `ResetStrategies` frees the `triggers`/`multipliers`/`queue` while `DoNextAction` is still looping over them, so you get a dangling pointer every time, but it only actually crashes when that freed memory gets reused before the next read. So it can run for hours looking fine, then crash later.

If you want to see it without waiting for a crash: add a `tickDepth` counter (`++` at the top of `DoNextAction` and `ExecuteAction`, `--` at the end) and `LOG_ERROR` in `Engine::Init()` when `tickDepth > 0`. Run a busy combat session and it fires = the re-entrant `Reset()` freeing the containers mid-loop. That's the bug. That counter is basically my fix, just logging instead of deferring.

noisiver:
We need to be able to reproduce the issue if we are to consider merging this pull request. If we have no evidence of the crash we don't want to add code that from our view does nothing. That's the issue.

noisiver:
What you should do is to run only mod-playerbots and no other modules to make sure it will actually crash with that.

NoxMax:
> A real player fighting inside a dense bot pack is what does it for me

It might be because the real player is overriding BotActiveAlone with their presence, rather than them actually doing something there. For what it's worth, I copy/pasted your entire conf and only edited the mysql/network configs. Still no crash.

noisiver:
I reset all the bots to make them all level 1, created a human and started killing stuff in the starting zone with 190ish bots around me. I had to fight for spawns of course but I got many kills and all kinds of classes were around me - including stealthed rogues - and no crash so far. I am using 100% bot activity with smart scale disabled so they are all always active.

This is what the area looks like while I'm doing this and you can clearly see they are all fairly close to me.

image

This is the amount of time I've **so far** spent doing this.

{A6ABE08E-98B8-47A0-9C42-2F23A62AA02B}

I will continue for 10 more minutes or so but if it still hasn't crashed I'm sorry to say I can't reproduce this.

noisiver:
Yeah I can't reproduce it.

What I tried is I added this debuff to myself

{AD4DAA3C-0714-4CF6-A3C1-EC5D98415BD6}

to make sure I didn't die, went into echo ridge mine and attacked 3 creatures, added the debuff to them and walked backwards slowly while tabbing to make sure none of them reset or died while I moved them into the large pack of bots fighting for spawns.

When I was in the middle of the pack I removed the debuff from the 3 creatures and killed them. I did this 8-9 times (or possibly more) and there was no crash.

I then tried getting killed by a creature while in the pack of bots and ran back and revived also in the pack (without GM commands) several times and still no crash.

This is on top of the normal situation where you fight for spawns that I also tried for a while.

timoinglin:
Thanks for putting real time into this, especially @noisiver - you tried harder to reproduce it than I expected.

Since you can't trigger it on the same code, it's most likely something specific to my build (my core fork + my own module), not stock mod-playerbots - so the problem may well be on my end. I don't want you adding code for a crash you can't see, so let's drop this one. The guard's running fine on my own server, so I'm covered. Thanks again, and sorry for the back and forth.

NoxMax:
Hey we're glad you're looking out. And if you have questions or ideas, we're on Discord.

PR #2486, opened 2026-06-18
Immune targets =/= possible targets

brighton-chi:
## Pull Request Description

This is mainly to address the issue of bots continuing to attack Mages with Ice Block and Paladins with Divine Shield. I made the actual check a little broader to capture all cases of when a target is both physical and magical immune.

## Feature Evaluation

Just added immunity checks to AttackersValue::IsPossibleTarget()

## How to Test the Changes

The easiest way is to play a Mage or a Paladin and go zerg into the opposing team, then put up Ice Block/Divine Shield.

## Impact Assessment

    - - [ ] No, not at all

This is a new check when determining possible targets, but it should be very cheap.

    - - [ ] No
    - - [x] Yes (**explain why**)

Default bot behavior is idiotic as they stick to immune targets like glue.

    - - [x] No

## AI Assistance

- - [ ] No

I wanted to capture more than just Ice Block and Divine Shield (which would have been simple aura checks) so I had DeepSeek *shudder* look at the spell flags.

## Final Checklist

### DIFF
--- a/src/Ai/Base/Value/AttackersValue.cpp
+++ b/src/Ai/Base/Value/AttackersValue.cpp
@@ -154,6 +154,11 @@ bool AttackersValue::IsPossibleTarget(Unit* attacker, Player* bot, float /*range
     if (attacker->HasUnitFlag(UNIT_FLAG_IMMUNE_TO_PC) || attacker->HasUnitFlag(UNIT_FLAG_NOT_SELECTABLE))
         return false;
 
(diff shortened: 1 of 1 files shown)


Vortikai:
I was just thinking about how this was happening, only it was from an altbot pet, not a bot :D

brighton-chi:
This won't stop pets, unfortunately. Bots don't really control their pets. They are mysterious creatures.

Celandriel (approved):

PR #2494, opened 2026-06-20
Create framework for excluding targets (SWP staging)

brighton-chi:
## Pull Request Description

**The tl;dr version**: This PR lets you make dungeon/raid strategies that allow tanks or dps to do what they normally do, except completely ignore [insert target] under [insert conditions].

**The longer version:**
This PR implements the ability for scripts to modify values used for bot target selection in order to exclude specified NPCs. Currently, doing this is possible only through assigning the moon icon, and even that doesn’t block actions that target through the “on attackers” path (i.e., actions relating to multi-dotting). Under this PR's approach, a bot evaluates its target, exclusions are gathered from all active combat strategies, and then they are applied before the strategy's own selection logic runs. This means excluded NPCs never enter the candidate pool in the first place. The bot simply acts as if they aren't there, rather than the strategy reacting after a bad target has already been chosen.

This capability is useful because it allows dungeon/raid scripts to command bots (or certain bots or types of bots) to ignore particular targets without completely suppressing their standard target selection behavior. Previously this could be done only with blanket overrides of TankAssistAction or DpsAssistAction and then creating an entirely new custom targeting scheme, which practically would have to be much simpler than the default targeting schemes, especially for tanks. 

I use this for SWP strategies, and therefore they cannot be implemented until this system (or a similar one) is in place. The system is very easy to use and can be done entirely in the strategy.cpp and strategy.h files for the dungeon/raid strategy you are working on. 

1. Declare a HasTargetExclusions member of the dungeon/raid strategy class (bool HasTargetExclusions() const override { return true; }).
2. Declare an AppendTargetExclusions member (void AppendTargetExclusions(GuidSet& exclusions, TargetValueExclusionType type) override)
3. Make a void function for each exclusion that contains the conditions for your exclusion and the target(s) being excluded.
4. In the definition of AppendTargetExclusions for the dungeon/raid strategy class, you can then select between any or all of tank, dps, and attackers value exclusions with a switch statement.

For example, the following sets up an exclusion that prevents tanks and melee dps from considering Shield Orbs as valid targets during Kil’jaeden:

```
void AppendKiljaedenShieldOrbExclusions(PlayerbotAI* botAI, GuidSet& exclusions)
{
    if (!botAI->IsMelee(botAI->GetBot()))
        return;

    if (!botAI->GetAiObjectContext()->GetValue<Unit*>("find target", "kil'jaeden")->Get())
        return;

    for (ObjectGuid const guid : botAI->GetAiObjectContext()->GetValue<GuidVector>("attackers")->Get())
    {
        Unit* attacker = botAI->GetUnit(guid);
        if (!attacker)
            continue;

        if (attacker->GetEntry() == static_cast<uint32>(SunwellNpcs::NPC_SHIELD_ORB))
            exclusions.insert(guid);
    }
}
```

The above is needed because I need tanks and (depending on positioning) melee dps to switch to Sinister Reflections when they spawn and back to KJ, but they should not go after the Orbs. There are four Sinister Reflections that spawn at a time so manually setting up a system for tanks and melee dps to decide between which to attack on every tick is far less effective and efficient than just telling them to ignore Shield Orbs and letting them pick between KJ and each Sinister Reflection.

Another example: I have another exclusion on KJ so that dps doesn’t attack Sinister Reflections unless a tank has aggro. If I used just a multiplier or action to implement such a block, it would result in dps rapidly swapping targets if their standard logic would want them to attack Sinister Reflections (they would attempt to attack the Sinister Reflection, then either be blocked (multiplier) or forced back to another target (action), and there would just be an ongoing fight between the raid implementation and standard strategy). But this way, dps simply does their own thing until a tank gets aggro, which is crucial because the Sinister Reflection mechanic is that they do not aggro immediately unless attacked first, and they will kill non-tanks almost immediately if aggroed onto them.

Conditions can basically be whatever you’d use for strategies anyway—presence of particular mob, role of bot, position of bot or mob, aura present on bot or mob, casting status, casting target, victim identity, distance comparisons, blah blah blah.

## Feature Evaluation

I think this is already pretty much covered by the description above and "Impact Assessment" below, but LMK if anything else is needed.

## How to Test the Changes

Right now, just play with this merged and make sure there isn't some terrible performance impact.

## Impact Assessment

    - - [ ] No, not at all

When no strategy with target exclusions is active (i.e., right now), there is no performance impact. During Engine::Init(), the presence of a boolean flag is checked by scanning loaded strategies to see if any of them use target exclusions. Any strategy that doesn't call HasTargetExclusions gets ignored. This doesn't happen per-tick.

When a strategy with exclusions is active, there is a performance impact, but I think it is minimal. GatherStrategyTargetExclusions iterates each bot's combat strategies and calls AppendTargetExclusions on each. If there is no exclusion, the cost is just iterating the strategies. The real cost comes with implementing exclusions and the conditions on which the exclusion kicks in (but that's basically the same consideration as everything else with respect to dungeon/raid strategies).

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

I wouldn’t say so. There’s not much code added here, and it has no impact on anybody who isn’t seeking to utilize it for a specific purpose. 

## AI Assistance

- - [ ] No

I had GPT-5.4 assist with analyzing the best sites in the target valuation paths to hook into and help with the overall implementation.

## Final Checklist

### DIFF
--- a/src/Ai/Base/Value/AttackerWithoutAuraTargetValue.cpp
+++ b/src/Ai/Base/Value/AttackerWithoutAuraTargetValue.cpp
@@ -6,26 +6,27 @@
 #include "AttackerWithoutAuraTargetValue.h"
 
 #include "Playerbots.h"
+#include "Strategy.h"
+#include "TargetValue.h"
 
 Unit* AttackerWithoutAuraTargetValue::Calculate()
 {
     GuidVector attackers = botAI->GetAiObjectContext()->GetValue<GuidVector>("attackers")->Get();
+    GuidSet const dynamicExclusions = GatherStrategyTargetExclusions(botAI, TargetValueExclusionType::Attacker);
     // Unit* target = botAI->GetAiObjectContext()->GetValue<Unit*>("current target")->Get();
     uint32 max_health = 0;
     Unit* result = nullptr;
     for (ObjectGuid const guid : attackers)
     {
         Unit* unit = botAI->GetUnit(guid);
-        if (!unit || !unit->IsAlive())
+        if (!unit || !unit->IsAlive() || dynamicExclusions.find(guid) != dynamicExclusions.end())
             continue;
 
         if (!bot->IsWithinCombatRange(unit, botAI->GetRange(range)))
             continue;
 
         if (unit->GetHealth() < max_health)
-        {
             continue;
-        }
 
         if (!botAI->HasAura(qualifier, unit, false, true))
         {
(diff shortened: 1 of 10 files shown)


kadeshar:
Conflict to resolve

kadeshar (approved):

Issue #1954, opened 2025-12-27
Since update(pull) 1928, loot does not work as expected in RDF


Ben19882:
**Bug Description**
When doing (Classic) RDF with bots, gear that you or your bots need is bugged.
Bots do not roll on loot, and sometimes the loot disappears.
This seems to happen around 60% of the time.
It was working correctly before pull request #1928.

**How To Reproduce**

Invite bots to your party.

Queue for (Classic) RDF.

Kill bosses and observe loot behavior.

**Expected Behavior**
Bots should correctly roll Need, Greed,(depends on your config) or Pass, and loot should remain available until rolls are completed.

**Config:**
Bots' loot roll level (0 = pass, 1 = greed, 2 = need)
Default: 1 (greed)
AiPlayerbot.LootRollLevel = 1

**I also see this:**
https://github.com/mod-playerbots/mod-playerbots/pull/1911
Will that fix it?

PS:
I am now updating to the latest build.

Ben19882:
Okay, I just tested the latest build of AzerothCore with PlayerBots, and it seems the issue is gone.
I will ask some real players to test it again. Thank you!

Ben19882:
The problem is still there.. i got some feedback from some players

Wishmaster117:
If ou merge this pr: https://github.com/mod-playerbots/mod-playerbots/pull/1911 the issue persists?

Ben19882:
@Wishmaster117 Is it production ready?

Wishmaster117:
Yes I've finished it, it not still merged because we are working one module core stability before, but give it a try and report if it solve your problem,  I'm using it and my bots rool

Ben19882:
@Wishmaster117 It is running in production now.
I will let you know whether it solved the issue or not. Thank you for the help so far!

Ben19882:
Good morning @Wishmaster117 

I have run some dungeons myself, and the players who reported the issues also ran several dungeons. After merging the PR that was mentioned, the problem seems to be resolved.

I also really like the new style! It’s fast and feels even more like a real player. Nice work ,it must have taken a lot of effort. I really like this project and truly appreciate the support you all provide.

Have a great day! I hope this new loot system will fully getting merged soon so more people can try it out, because it’s a huge improvement. We’re not even a full day further yet, so it might be a bit early to say everything is completely fixed, but so far no one has experienced the issue again after the compile.

hermensbas:
[...]

PR #2496, opened 2026-06-20
Improve performance of LootRollAction::Execute

ArkosLUL:
## Pull Request Description

Vote for all items in one pass instead of recreating the rolls vector each time and only calculating the roll for the first item. This significantly improves the roll times especially in raids where bots could even hit the roll timeout.

## Feature Evaluation

Now bots will roll on all the items that are in group->GetRolls() vector in one pass instead of voting only for a single item. This reduces the overhead of recreating the vector each time the method is called.

## How to Test the Changes

1) Enter a raid or do some quests in the world with bots. 
2) Check how fast the bots roll on dropped items

## Impact Assessment

    - - [ ] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

AI was used to identify the root cause of the slowness.

## Final Checklist

### DIFF
--- a/src/Ai/Base/Actions/LootRollAction.cpp
+++ b/src/Ai/Base/Actions/LootRollAction.cpp
@@ -20,6 +20,7 @@ bool LootRollAction::Execute(Event /*event*/)
         return false;
 
     std::vector<Roll*> rolls = group->GetRolls();
+    bool voted = false;
     for (Roll*& roll : rolls)
     {
         auto voteItr = roll->playerVote.find(bot->GetGUID());
@@ -102,11 +103,10 @@ bool LootRollAction::Execute(Event /*event*/)
                 group->CountRollVote(bot->GetGUID(), guid, vote);
                 break;
         }
-        // One item at a time
-        return true;
+        voted = true;
     }
 
-    return false;
+    return voted;
 }
 
 RollVote LootRollAction::CalculateRollVote(ItemTemplate const* proto, ItemUsage usage)


Jellypowered:
FWIW I have tested this on my personal fork and found it to cause no issues, speedy looting now, no more timeouts and loot missing because of timeouts. It has not broken anything, modules included.

noisiver:
I read something on Discord saying it degraded performance for several seconds while the rolls are handled. Is that no longer a thing? If so, can that be proven? The last thing we need are people complaining about the server hanging (not literally) when loot is being handled.

noisiver:
I managed to find the post and ironically it's you who said it @Jellypowered 
{DCCBF255-1D8A-440C-BBDB-63F66DA67FE7}

Or are you saying without this change it hangs, but with it it doesn't?

Celandriel:
It must be causing some lag. We know that was the case when we were dealing with all the skills being added and we had to rewrite some of it. I would like to see before and after pmon.

Jellypowered:
I feel some more context is needed before you use past comments of mine in the discord 🤣 

My first test was to simply remove the following all together. 
```
// One item at a time
        return true;
```
I misunderstood the scope of the change and it wasn't until Arkos posted his [Gist](https://gist.github.com/ArkosLUL/1eaf6b83d80b06058fd0245f14bccb07) link showing the whole function that I figured it out. 

I also had some custom changes that introduced my own set of lag/issues and have since resolved those as well. 

So there's the whole picture and why I had said that at the time 👍

noisiver:
> It must be causing some lag. We know that was the case when we were dealing with all the skills being added and we had to rewrite some of it. I would like to see before and after pmon.

I think this makes sense, a comparison of pmon before and after these changes.

Jellypowered:
```
Stock: (1b5c178)
-------------------------------------[TOTAL BOT]-------------------------------------------------------
percentage     time  |     min ..     max (      avg  of      count) - type      : name
-------------------------------------------------------------------------------------------------------
 0.694%      0.525s |     0.0 ..    55.2 (     0.195 of       2696) - Action    : loot roll

With PR 2496: 
-------------------------------------[TOTAL BOT]-------------------------------------------------------
percentage     time  |     min ..     max (      avg  of      count) - type      : name
-------------------------------------------------------------------------------------------------------
 0.347%      0.109s |     0.0 ..    54.1 (     0.111 of        987) - Action    : loot roll
``` 
Both tests was performed running around doing quests in the Eastvale Logging Camp in Elwynn Forest with a 40 man raid group. 
 
Observed in stock: Loot rolls took considerably longer without this PR, while there have been improvements and it does not delay looting as long as I remembered, there's still perceivable delay.

Observed with this PR: the loot rolls were done before I moved to the next group of murlocs.

If anyone wants a longer test, All my team is level 8, I can run around for 30 mins on each and then do the stack.

Celandriel:
#1954

Celandriel:
Also, I suspect that the difference will come up more in a raid with a boss where you have 39 bots rolling on 4-5 pieces each

Jellypowered:
> Also, I suspect that the difference will come up more in a raid with a boss where you have 39 bots rolling on 4-5 pieces each
Yeah I don't have any alts at dungeon/raid level yet so I can't test that part of it properly 😆

ArkosLUL:
@noisiver @Celandriel Here is pmon results with the following setup:
1) PMon was enabled in config and not during the runtime
2) I have Loot Rates x3 which means even trash mobs drop lots of items in terms of quantity (even better for this test case)
3) Both times I've run Molten Core from start to Gehenass including the boss kill itself
4) 40 man raid with 23 being my alt bots, the rest are addclass bots
[StockLootRoll.txt](https://github.com/user-attachments/files/29174870/StockLootRoll.txt)
[ChangedLootRoll.txt](https://github.com/user-attachments/files/29174871/ChangedLootRoll.txt)

noisiver:
So stock is:
`1.647%      3.212s |     0.0 ..    37.0 (     0.460 of       6985) - Action    : loot roll`

With this change it is:
`0.980%      2.246s |     0.0 ..    37.4 (     0.221 of      10186) - Action    : loot roll`

Am I reading it right when it says takes 0.980% of processing time compared to 1.647%? But what is the 37.4? And what does 0.221 of 10186 mean? I'm asking because I don't know it, but there are others here who do so I'm asking them.

Celandriel:
> So stock is: `1.647% 3.212s | 0.0 .. 37.0 ( 0.460 of 6985) - Action : loot roll`

% is the percentage of the total action time across all the sum action times. Its hard to use this number to compare anything especially with the second case having 50% less calls. 
Time is the sum time across all ticks for that action. (So 6985 loot roll actions took 3.212s to run)
0.0 .. 37.4 is the min/max range in ms for the action to execute. So max execution for this action is  37.4.
0.221 is the average execution time for the action. 

The thing to compare are the max execution time and the average when comparing this sort of change.

ArkosLUL:
So uh, is there something else you would like me to do? From the data provided above the changed action average execution time is lower than the stock. I can record longer PMon sessions in TBC raids if needed.

Celandriel:
Sorry, I want to test myself.

kadeshar:
For me is ok. @Celandriel how tests?

Celandriel (approved):

PR #2501, opened 2026-06-22
Remove nested loops from group tank functions

brighton-chi:
## Pull Request Description

Three functions, IsAssistTankOfIndex, GetGroupMainTank, and GetGroupAssistTank, check IsMainTank for every group member when iterating, thus creating a second layer of looping. This is bad for performance and is easily fixed by a per-call cache to get the main tank before iterating. I'm sure there are better ways of doing it, but this one is simple for a plebe like me to do and should be enough to improve the situation.

## Feature Evaluation

I added a new helper, GetMainTankGuid, which is now used in the boolean tank role checks in PlayerbotAI.cpp and tank role getters in RaidBossHelpers.cpp.

The ignoreMemberFlag parameter was removed from IsMainTank. This is because I couldn't think of a simple way to keep it in a fashion that makes sense. Previously, IsMainTank would influence determination of assist tanks too, but now it doesn't, and so keeping the parameter there potentially creates an inconsistency between who the main tank is considered to be for purposes of being the main tank vs. purposes of being an assist tank. Moving the parameter to the new helper preserves the old pattern, but since nobody is actually going to directly call that helper, it's useless to exist there. Ultimately, the parameter has not been used a single time, and it's hard for me to imagine why anybody would want to use it, so it's safe to remove and it doesn't seem like anybody would be bothered by it.

I tested ingame by temporarily adding an IsAssistTankOfIndex call at the beginning of all 61 Black Temple triggers and sitting in the instance with 24 bots in a raid, though there were only 3 tanks so practically there were only 3 callers of relevance. Over 60s, pmon showed a reduction of 0.125s on triggers after the change (1.868s reduced to 1.743s). 

## How to Test the Changes

1. Run instances with strategies that rely on designated main tanks and assist tanks and confirm the strategies still work.
2. Check pmon before/after changes in the PR to evaluate performance differences.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

I used DeepSeek v4 Pro, but it's terrible and might actually have caused me to spend more time on this than I otherwise would have. I ended up practically doing this all myself so please check carefully.

## Final Checklist

### DIFF
--- a/src/Ai/Raid/RaidBossHelpers.cpp
+++ b/src/Ai/Raid/RaidBossHelpers.cpp
@@ -108,13 +108,14 @@ Player* GetGroupMainTank(PlayerbotAI* botAI, Player* bot)
     if (!group)
         return nullptr;
 
+    ObjectGuid const mainTankGuid = botAI->GetMainTankGuid(group);
+    if (mainTankGuid.IsEmpty())
+        return nullptr;
+
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-        if (!member || !member->IsAlive())
-            continue;
-
-        if (botAI->IsMainTank(member))
+        if (member && member->IsAlive() && member->GetGUID() == mainTankGuid)
             return member;
     }
 
@@ -129,28 +130,32 @@ Player* GetGroupAssistTank(PlayerbotAI* botAI, Player* bot, uint8 index)
     if (!group)
         return nullptr;
 
+    ObjectGuid const mainTankGuid = botAI->GetMainTankGuid(group);
+    if (mainTankGuid.IsEmpty())
+        return nullptr;
+
     uint8 assistantCount = 0;
     std::vector<Player*> nonAssistantTanks;
 
     for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
     {
         Player* member = ref->GetSource();
-        if (!member || !member->IsAlive())
-            continue;
-
-        if (botAI->IsAssistTank(member))
+        if (!member || !member->IsAlive() || !botAI->IsTank(member) ||
+            member->GetGUID() == mainTankGuid)
         {
-            if (group->IsAssistant(member->GetGUID()))
-            {
-                if (assistantCount == index)
-                    return member;
+            continue;
+        }
 
-                assistantCount++;
-            }
-            else
-            {
-                nonAssistantTanks.push_back(member);
-            }
+        if (group->IsAssistant(member->GetGUID()))
+        {
+            if (assistantCount == index)
+                return member;
+
+            assistantCount++;
+        }
+        else
+        {
+            nonAssistantTanks.push_back(member);
         }
     }
 
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -2369,47 +2369,38 @@ bool PlayerbotAI::IsDps(Player* player, bool bySpec)
     return false;
 }
 
-bool PlayerbotAI::IsMainTank(Player* player, bool ignoreMemberFlag)
+ObjectGuid PlayerbotAI::GetMainTankGuid(Group* group)
+{
+    if (!group)
+        return ObjectGuid::Empty;
+
+    // Check for main tank flag
+    Group::MemberSlotList const& slots = group->GetMemberSlots();
+    for (Group::member_citerator itr = slots.begin(); itr != slots.end(); ++itr)
+    {
+        if (itr->flags & MEMBER_FLAG_MAINTANK)
+            return itr->guid;
+    }
+
+    // Fallback if no flag is set: return the first alive tank
+    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+    {
+        Player* member = ref->GetSource();
+        if (member && IsTank(member) && member->IsAlive())
+            return member->GetGUID();
+    }
+
+    return ObjectGuid::Empty;
+}
+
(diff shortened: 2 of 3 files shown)


nl-saw:
The optimization itself looks sound, but this is not a pure refactor. It changes role-detection logic in a few places, and those are exactly the kinds of changes that can produce subtle AI behavior regressions even when the code appears cleaner. My suggestion is to split this up in multiple PR's, or just leave it at removing nested loops, as the title suggests. The gained performance would probably be negligible as there are just not so many groups.

brighton-chi:
Are you referring to the removal of the ignoreMemberFlag parameter? I suppose it could be kept by passing it to the new helper through the tank functions, though that would change the signature of them all, other than IsMainTank, and it's all tied together so I'm not sure how the PR would be split.

The value of that seems questionable, given that the parameter is currently unused, and I cannot think of a reason that somebody would want to use the main tank flag and then not have the flagged character be the main tank (which instead means the last tank to join the group becomes the main tank).

Especially given recent concerns raised on Discord about performance issues in raids, it seems to me to be worth it to implement optimizations. But of course it's not my call.

nl-saw:
That flag is a more minor concern, I am more concerned about the changes that have little to do with removing nested loops. About splitting; i.e. 1 pr to eliminate the nested loops by refactor (but do not change code logic) and 2nd for the rest. I'd then also love to see some ppl confirming the change would resolve the alleged performance issues in raids.

brighton-chi:
I was not intending to making any other functional changes. If that's the result then it should be fixed. Can you elaborate on what else is changing here?

I did run a raid test noted in the description. The difference is not enormous, but I wouldn't say it's irrelevant either, especially since it's in the context of a single raid group.

nl-saw:
I just want to make sure there aren't any subtle behavioral changes. I think the performance gains may be outweighed by potential regressions.

Can you verify that the following scenarios remain behaviorally equivalent to the previous implementation?

- `IsAssistTank()` behavior
  - Grouped vs ungrouped tanks
  - Cases where no main tank is explicitly assigned

- `GetGroupAssistTank()` selection/ordering semantics
  - Assistant tanks are still preferred as before
  - Selection order is unchanged when multiple eligible tanks exist

- Dead-player edge cases
  - Dead flagged main tank
  - Dead fallback-selected main tank
  - Alive secondary tank becoming the effective tank
  - Assistant flags on dead players

- Regression coverage / validation
  - Explicitly flagged main tank
  - No flagged main tank (fallback selection)
  - Dead main tank
  - Multiple tanks in group
  - Assistant tank ordering and selection

As this looks like more than a pure performance refactor, so I'd appreciate confirmation that these cases were validated and that behavior remains unchanged.

brighton-chi:
**`IsAssistTank()` behavior**

- Grouped vs ungrouped tanks
  - Doesn't apply in groups, same as before. The prior IsMainTank() started with a group check and returned IsTank() if there was no group. So if there was no group before, there was no assist tank (IsAssistTank() resolved to IsTank() && !IsTank()). In-group behavior is same also. Both old and new start with an IsTank() check, and the old function directly excluded a bot that met IsMainTank() and the new only excludes a bot with a guid matching GetMainTankGuid(), which now contains the looping logic that was previously in IsMainTank().

- Cases where no main tank is explicitly assigned
  - Same as before. Before, `IsMainTank()` would fall back to getting the first player that satisfied `IsTank()` and `IsAlive()` and had the same guid as the caller. Now, `GetMainTankGuid()` gets the guid of the first player that satisfies `IsTank()` and `IsAlive()`, and then `IsMainTank()` just returns that guid.

**`GetGroupAssistTank()` selection/ordering semantics**

- Assistant tanks are still preferred as before
  - Yes. Like before, the loop excludes members that are dead or are not tanks, and the maintank. Before this was done through an `IsAssistTank` check, now to avoid the nested looping, it's done by skipping members if they fail `IsTank()` or match the main tank's guid. The remaining code block to check assistants first is the exact same as before, just the indenting changed due to the prior changes.

- Selection order is unchanged when multiple eligible tanks exist
  - Yes, the iteration code block is unchanged.

**Dead-player edge cases**

- Dead flagged main tank
  - Same as before, still considered main tank. The first loop for the flagged member doesn't check `IsAlive()`, the looping is just in `GetMainTankGuid()` now instead of directly in `IsMainTank()`. The code is the same for the loop, except that the new function returns the flagged member's guid if one is found and passes it to  IsMainTank()`, whereas before, `IsMainTank()` ran the loop for the flagged member's guid directly with a `break` if one was found.

- Dead fallback-selected main tank
  - Same as before, need to be alive to count as fallback. Before, `IsMainTank()` directly returned `false` if the caller
failed `IsAlive()`. Now, `GetMainTankGuid()` checks `IsAlive()` in the fallback loop and returns `ObjectGuid::Empty` if no member satisfies `IsAlive()` (and the empty guid is passed to `IsMainTank()`, which returns `false`).

- Alive secondary tank becoming the effective tank
  - Same as before. The prior fallback loop and current fallback loop contain the same checks for `IsTank()` and `IsAlive()`.

- Assistant flags on dead players
  - Unchanged. Whether this matters depends on the third parameter of `IsAssistTankOfIndex`, which remains the same. Both the old and new `IsAssistTankOfIndex()` return `false` before looping if `ignoreDeadPlayers` is `true` and the player fails `IsAlive()`.

**Regression coverage / validation**

Are you asking about in-game testing? I have been in raids where the 1st, 2nd, 4th, and 5th scenarios came into play, and everything behaved as expected. I haven't tested the 3rd scenario, but I can do that.

kadeshar:
Conflicts to resolve

kadeshar (approved):

PR #2504, opened 2026-06-22
Change license and PR template

Celandriel:
**Pull Request Description**
The admin team have voted to relicensing from GNU AGPL v3.0 to GNU GPL v2.0, aligning mod-playerbots with the licenses of both its upstreams — AzerothCore (GPLv2) and the CMaNGOS playerbots Bot AI core it is derived from (GPLv2). This PR begins that process by defining all future code at GPLv2 while any previous code undergoes approval by the contributors, or has its licenses fixed to the correct attribution. 

**What this PR does**

- Replaces the root LICENSE with GNU GPL v2.0 text.
- Updates PULL_REQUEST_TEMPLATE.md with a Code Provenance / Attribution question (was any code copied/adapted from a sister project — if so, name the project and original authors) plus checklist items for attribution and the GPLv2 header on new files.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -114,7 +114,7 @@ update, whose content are called within the codebase at src/strategy/actions/Fis
 - - [ ] Added logic complexity is justified and explained.
 - - [ ] Any new bot dialogue lines are translated.
 - - [ ] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).
-- - [ ] New source files use the GPLv2 header (see NOTICE.md).
+- - [ ] New source files use the GPLv2 header.
 - - [ ] Documentation updated if needed (Conf comments, WiKi commands).
 
 ## Notes for Reviewers
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -77,6 +77,26 @@ If yes, please specify:
 
 
 
+## Code Provenance / Attribution
+<!--
+mod-playerbots is GPLv2 and is derived from the CMaNGOS playerbots (ike3). Code may be ported or
+adapted from other GPLv2 projects, but upstream attribution MUST be preserved (GPLv2 §1 and §2(a)).
+Never replace an upstream copyright notice with the project header.
+-->
+Was any code in this PR copied or adapted from a sister / upstream project (e.g. CMaNGOS playerbots,
+ MaNGOS, another module)?
+- - [ ] No, all code in this PR is original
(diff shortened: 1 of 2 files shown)


NoxMax:
This remains one of my facepalm moments (#1573 and subsequently #1673) "hey everyone! we should match our license to AC. I am sure they will stay on v3 from now on."
AC two months later: "actually, GPLv2 https://github.com/azerothcore/azerothcore-wotlk/pull/23655"
🤦

PULL_REQUEST_TEMPLATE.md
```diff
@@ -93,6 +113,8 @@ update, whose content are called within the codebase at src/strategy/actions/Fis
 - - [ ] Performance impact is understood, tested, and acceptable.
 - - [ ] Added logic complexity is justified and explained.
 - - [ ] Any new bot dialogue lines are translated.
+- - [ ] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).
+- - [ ] New source files use the GPLv2 header (see NOTICE.md).
```
NoxMax: I think these can be combined into one, as it touches the same licensing matter. But on "in-file note", does this mean actually putting attributions within the file itself, not just the commit? Is that needed? Did AC do that with the RBAC port?

Celandriel: See this for what I'm thinking.
 https://github.com/azerothcore/azerothcore-wotlk/blob/master/src%2Fserver%2Fgame%2FCombat%2FCombatManager.cpp

NoxMax: I see. I think it's overkill to add it into the file given that highlighting any line would show information about the commit that modified it, but if that's how it's done so be it. On making it a single point, what do you think of

`New source files use GPLv2 (see NOTICE.md) and ported code is attributed (project + author(s) in commit trailer + in-file note).`

Celandriel: Im not sure how to handle the case where just a small section of the file is incorporated. But for large file ports I think it should follow the  same approach.

NoxMax: What about large ports in-file, and minor things in commit message only? The detailed instructions would then be in NOTICE.md.

`New source files use GPLv2 and ported code is correctly attributed (see NOTICE.md for full details).`

Celandriel: Small section as well as coauthored trailers in the commit history.

NoxMax: for the large ports? yeah of course. Co-authored for all, in-file too but for large ports only.

kadeshar (approved):

PR #2506, opened 2026-06-22
Fix: prot paladins get no blessing without Sanctuary; unify party-blessing path

Yuof:
## Pull Request Description

Fixes Protection paladins maintaining **no blessing at all** when the Blessing of Sanctuary talent isn't trained (guaranteed at level 20-39, and for any prot build that skips the talent), and reworks the paladin party-blessing path into a single, consistent dispatcher.

Previously the four blessing-on-party strategies (`bsanc`, `bmight`, `bwisdom`, `bkings`) each had their own bespoke action with divergent `hasBwisdom`/`hasBkings`/group logic, including a Kings action that bailed when the bot had no group (so solo bots got nothing). They're now one shared `CastBlessingOnPartyAction` that picks the best blessing per target from the role-priority table, leading with the strategy's blessing where it fits and complementing other paladins instead of duplicating.

### Base logic

Each blessing strategy keeps a single "lead" trigger; the shared action routes through `GetDesiredBlessingCategory(target, preferred)`:

1. **Lead**: if the strategy's preferred blessing suits the target's role, the bot can cast it, and no other paladin already provides it, use it.
2. **Renewed Hope rule**: with no Discipline priest carrying Renewed Hope in the raid, tanks promote Sanctuary to #1 (its 3% reduction doesn't stack with Renewed Hope).
3. **Fallback**: walk `BASE_BLESSING_PRIORITIES[role]` for the highest blessing the bot can cast that isn't already covered.

Target selection picks a party member (or self, when solo) that lacks that blessing **from us** (caster = bot), so we maintain our own blessing, can upgrade our own lower-priority blessing, and complement rather than duplicate. The "already covered" check reuses the greater-blessing system's strength-aware `HasEquivalentOrStrongerSameFamilyBlessing`, so we won't displace an equal-or-stronger blessing but will upgrade a weaker Might/Wisdom (Improved Might/Wisdom aware).

Party members' pets are included as targets too (second commit), scanned after all players so players are blessed first, matching how generic party buffs (Power Word: Fortitude, Mark of the Wild, ...) already walk pets. A pet isn't role-classifiable, so it's treated as physical DPS (Might → Kings).

**Role priority table** (`BASE_BLESSING_PRIORITIES`, reused from the greater-blessing system):

| Role | Priority order |
| --- | --- |
| Casters | Kings → Wisdom → Sanctuary → Might |
| Physical DPS (rogue/warrior/DK) | Might → Kings → Sanctuary |
| Hybrid DPS (enh/ret/hunter/cat) | Might → Kings → Wisdom → Sanctuary |
| Druid tank | Kings → Might → Sanctuary → Wisdom |
| Warrior/DK tank | Kings → Might → Sanctuary |
| Paladin tank | Sanctuary → Might → Wisdom → Kings |

**Strategy lead applicability** (`PreferredLeadAppliesTo`):

| Strategy (lead) | Leads on |
| --- | --- |
| `bkings` (Kings) | all roles |
| `bmight` (Might) | physical & hybrid DPS |
| `bwisdom` (Wisdom) | casters & hybrid DPS |
| `bsanc` (Sanctuary) | tanks |

### What changed vs. the old behavior

| Aspect | Old | New |
| --- | --- | --- |
| Strategy actions | 4 separate actions, bespoke `hasBwisdom`/`hasBkings`/group logic | 1 shared dispatcher keyed off the lead blessing |
| Prot paladin w/o Sanctuary talent | no blessing (esp. solo) | falls through to Might/Kings; self-buffs |
| Solo bot | Kings action bailed on "no group", no buff | self-targets and buffs |
| Multiple paladins | could duplicate / fight over blessings | complement via a strength-aware "another paladin has equal-or-stronger" check (won't duplicate; upgrades a weaker Might/Wisdom) |
| Re-cast / overwrite | could re-cast or be blocked by others' blessings | by-us check: never overwrites our own, upgrades our own lower-priority blessing |
| Sanctuary recipients | all tank specs equally | role-priority (paladin tanks lead; other tanks get it as fallback / when no Renewed Hope priest) |
| Disc-priest Renewed Hope | not considered | tanks promote Sanctuary when no Renewed Hope priest is present |
| Greater-blessing mode | per-action checks | single `IsGreaterBlessingMode` gate on trigger + target |
| Pets | never blessed (only players scanned) | party pets blessed too, after players, as physical DPS (Might → Kings) |

## Feature Evaluation

  - A per-target priority walk reusing the existing `ai::gbless` role table and `ResolveRoleProfile`, plus name-based `HasSpell` ("can I cast it") and the greater-blessing system's strength-aware `HasEquivalentOrStrongerSameFamilyBlessing` / by-us aura checks ("is it already covered"). One group scan for a Renewed Hope priest. Shared blessing utilities live in a new header-only `ai::blessing` namespace (`PaladinBlessingHelper.h`), consumed by both the dispatcher and the greater-blessing system. Strategy selection is unchanged (prot uses `bsanc`, holy uses `bwisdom`, ret uses `bmight`).
  - The walk is a handful of `HasSpell`/`HasAura` calls over at most 4 categories, run during the blessing trigger's existing **4s** check interval, and only for the single-blessing fallback path; raids run the cached greater-blessing system instead, where these triggers are inert. Net code is a reduction (one path replaces four). No new DB queries or allocations.

## How to Test the Changes

1. **Spell not available**: Prot paladin bot level 20-39 (no Sanctuary talent) receives Kings/Might (not nothing). Below 20: receives Might.
2. **Solo**: A lone paladin self-buffs its role blessing (prot gets Sanctuary if trained else Might; ret gets Might; holy gets Wisdom).
3. **Greater blessings active**: In a group/raid eligible for auto greater blessings, the single-blessing dispatcher does nothing (greater system owns blessings). Verify no double-casting.
4. **Multiple paladins**: Two+ paladins on the same group complement each other (e.g. casters end up with Kings from one and Wisdom from another) rather than both casting the same blessing.
5. **Multiple paladins, same strategy**: Two `bkings` paladins still diverge (one leads Kings, the other falls to the next role priority), no permanent fighting.
6. **Renewed Hope**: With a Disc priest carrying Renewed Hope in the raid, non-paladin tanks get Kings/Might; remove that priest and tanks shift to Sanctuary.
7. **Pets**: A hunter/warlock/DK in the group has their pet blessed (Might, or Kings under `bkings`) after the players are blessed; the pet's blessing isn't re-cast once present.

## Impact Assessment

    - - [ ] No, not at all

  The per-target walk and the Renewed Hope group scan run only on the blessing trigger's existing 4s interval, only in the non-greater (small-group/solo) path; raids use the cached greater-blessing system where these triggers are inactive. Overall this removes code paths rather than adding them.

    - - [ ] No
    - - [x] Yes (**explain why**)

  Paladin party-blessing selection is unified and corrected: prot paladins always get a blessing (the core bug), blessings are chosen by role priority, multiple paladins complement instead of duplicate, and Sanctuary is promoted for tanks when no Renewed Hope priest is present.

    - - [ ] No

  It adds the role-priority walk and a Renewed Hope check, but **net** complexity drops: four bespoke blessing actions collapse into one shared dispatcher and several now-dead helpers are removed.

## AI Assistance

- - [ ] No

AI (Claude Code) was used for investigation, design discussion, and code generation across the paladin blessing files. Every change was reviewed and is understood and owned by the contributor.

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (N/A, no new dialogue.)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (N/A, no conf/wiki changes.)

## Notes for Reviewers

- **Edge cases covered:** spell-not-known (falls through the priority walk), solo (self-target via `FindBlessingTarget`; never "greater mode" since `IsEligibleGroupForAutoBlessings(nullptr)` is false), greater-blessing mode (dispatcher gated off on both trigger and target), multiple paladins / same strategy (complement via the strategy-agnostic "from another paladin" aura check).
- **Known caveat (transient multi-paladin race):** in the uncoordinated single-blessing path, two paladins evaluating the same 4s tick before either casts can both pick a target's top blessing and briefly double-cast; it self-corrects within a cycle as the "from another" check kicks in. The greater-blessing system is what pre-assigns to avoid this.
- The `Renewed Hope` talent ids (`57470`/`57472`) are hardcoded (verified WotLK 3.3.5a talent ranks).
- The greater-blessing assignment logic and role table are unchanged; shared blessing utilities (aura/strength comparison, role/category mapping, Renewed Hope detection) were extracted into a new header-only `ai::blessing` namespace (`PaladinBlessingHelper.h`) consumed by both the greater path and the new dispatcher.
- Self-cast / combat blessing actions (`blessing of might/wisdom/kings` and their remaps) are intentionally left unchanged; only the on-party single-blessing path was reworked.

### DIFF
--- a/src/Ai/Class/Paladin/Actions/PaladinActions.cpp
+++ b/src/Ai/Class/Paladin/Actions/PaladinActions.cpp
@@ -8,6 +8,7 @@
 #include "AiFactory.h"
 #include "Event.h"
 #include "GenericBuffUtils.h"
+#include "PaladinBlessingHelper.h"
 #include "PaladinGreaterBlessingAction.h"
 #include "PaladinHelper.h"
 #include "Playerbots.h"
@@ -24,18 +25,6 @@ static bool IsBlessingTargetCandidate(Player* bot, Player* player)
     return bot->GetDistance(player) < sPlayerbotAIConfig.spellDistance * 2 &&
            bot->IsWithinLOS(player->GetPositionX(), player->GetPositionY(),
                             player->GetPositionZ());
-}
-
-static bool HasBlessingAura(
-    PlayerbotAI* botAI, Unit* target, std::initializer_list<char const*> auraNames)
-{
-    for (char const* auraName : auraNames)
-    {
-        if (botAI->HasAura(auraName, target))
-            return true;
-    }
-
-    return false;
 }
 
 static bool IsGreaterBlessingMode(Player* bot)
@@ -90,48 +79,6 @@ static Unit* FindBlessingTarget(
     }
 
     return nullptr;
-}
-
-static inline bool IsTankRole(Player* player)
-{
-    if (!player)
-        return false;
-
-    if (player->HasTankSpec())
-        return true;
-
-    if (PlayerbotAI* otherAI = GET_PLAYERBOT_AI(player))
-    {
-        if (otherAI->HasStrategy("tank", BOT_STATE_NON_COMBAT) ||
-            otherAI->HasStrategy("tank", BOT_STATE_COMBAT)     ||
-            otherAI->HasStrategy("tank face", BOT_STATE_NON_COMBAT) ||
-            otherAI->HasStrategy("tank face", BOT_STATE_COMBAT)     ||
-            otherAI->HasStrategy("bear", BOT_STATE_NON_COMBAT) ||
-            otherAI->HasStrategy("bear", BOT_STATE_COMBAT))
-            return true;
-    }
-
-    return false;
-}
-
-static inline bool IsOnlyPaladinInGroup(Player* bot)
-{
-    if (!bot)
-        return false;
-
-    Group* group = bot->GetGroup();
-    if (!group)
-        return true;
-
-    uint32 paladins = 0u;
-    for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
-    {
-        Player* player = ref->GetSource();
-        if (!player || !player->IsInWorld()) continue;
-        if (player->getClass() == CLASS_PALADIN) ++paladins;
-    }
-
-    return paladins == 1u;
 }
 
 inline std::string const GetActualBlessingOfMight(Unit* target)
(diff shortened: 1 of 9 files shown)


kadeshar:
Conflict to resolve

Yuof:
This should be mergeable and is working, but just adding a note that #2530 completely replaces this system. 
So we can skip this if the other PR get merged.

Yuof:
I kept this open as an earlier alternative to #2530, but after both being open for a longer time: I can't see any value in somebody reviewing this instead of the bigger PR.

PR #2513, opened 2026-06-26
Fix helms and cloaks slowly disappearing

wetbrownsauce:
## Pull Request Description

Currently, the code for showing helm/cloak on bots has a 20% chance of hiding their helm/cloak. This happens in Prepare so it gets called on every randomization that happens, resulting in every bot eventually losing their helm and cloak visually. Nothing ever toggles it back on if that is desired.

This change makes it so if the conf option for show helm/cloak are off, then the bots always randomize with no helm or cloak. If the option is on, I've switched it from set flag to toggle flag, so there will be a 20% chance on each randomization to toggle their helm/cloak.

The aim is so that bots will be closer to how real players might act, where some would have helmets shown, some wouldn't. But of course, if the player wants them to never show helm/cloak, that's an option too.

## Feature Evaluation

## How to Test the Changes

1. Toggle the option off, run server with freshly made bots. None should have helm or cloak.
2. Randomize bots a few times, helm and cloak should never show.
3. Toggle the option on, randomize bots a few times. Some should have helms and cloaks shown now.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

## Final Checklist

- - [ ] Stability is not compromised.

## Notes for Reviewers

If different behavior is desired, let me know. I haven't had the chance to test this locally, but if the concept is liked, then I can when I have more time.


wetbrownsauce:
closing. wrote it on the wrong branch and I've since improved. Will post new PR

PR #2517, opened 2026-06-27
BWL Bot Strategies: Razorgore & Vaelastrasz

HennyWilly:
## Pull Request Description

Implements bot behavior for _Razorgore the Untamed_ and _Vaelastrasz the Corrupt_ in Blackwing Lair.

**Razorgore**:
- General:
  - Fire resistance aura: Reduces damage from [Fireball Volley](https://www.wowhead.com/classic/spell=22425/fireball-volley) and [Conflagration](https://www.wowhead.com/classic/spell=16805/conflagration).
- Phase 1: Eggs
  - One tank bot marks Razorgore with the moon rti and tanks him whenever he is not mind-controlled.
    => All other bots ignore the boss (He wipes the raid if he dies before all eggs are destroyed!)
  - Mind-controlling the boss and destroying the eggs must be done by the real player.
- Phase 2: Razorgore
  - The moon rti is removed => Bots engage the boss
  - All bots except the current tank dodge the frontal cone AoE ([Cleave](https://www.wowhead.com/classic/spell=19632/cleave)) => move behind the boss
  - Additionally, ranged bots avoid [War Stomp](https://www.wowhead.com/classic/spell=24375/war-stomp) by moving away
  - Tanks not currently tanking the boss suppress TankFaceAction to avoid cleave damage

**Vaelastrasz**:
- General:
  - Fire resistance aura: Reduces damage from [Fire Nova](https://www.wowhead.com/classic/spell=23462/fire-nova)  and [Flame Breath](https://www.wowhead.com/classic/spell=23461/flame-breath).
  - All bots except the current tank position themselves on the rear flank to avoid damage from _Flame Breath_, [Cleave](https://www.wowhead.com/classic/spell=19983/cleave), and [Tail Sweep](https://www.wowhead.com/classic/spell=15847/tail-sweep)
  - Tanks not currently tanking the boss suppress TankFaceAction to avoid AoE damage
  - ~~Because the boss starts at 30% hp, warlocks suppress Drain Soul until 3% HP (negligible damage before that)~~
- Handling of [Burning Adrenaline](https://www.wowhead.com/classic/spell=18173/burning-adrenaline):
  - The current tank stays in place. The rear flank positioning ensures that other bots won't get hit by the explosion.
  - Other bots will move away from raid members.
    - Bots will prefer a position within LOS.
    - Ranged bots will try to find a spot where they can still reach the boss with their attacks.
    - If a bot gets pushed back by Tail Sweep, they will try to reach the boss without coming near other raid members.
    - If the boss is dead, bots with Burning Adrenaline will still try to avoid other raid members.
    - Movement direction is calculated using weighted repulsion.

**Miscellaneous**:
- _BwlTurnOffSuppressionDeviceAction_ restricted to Rogues (or raid cheat)
- New helpers _IsNonBABotNearPosition_, _ClearTargetIcon_
- Multiplier infrastructure enabled (was commented out)

## Feature Evaluation

  - Razorgore: check if eggs are alive (scans nearest game objects), boss position + frontal cone check for AoE avoidance, moon marker assignment by the off-tank
  - Vaelastrasz: group position comparison for weighted repulsion, aura check for Burning Adrenaline, tank comparison for flank positioning
  - _AreRazorgoreEggsAlive()_: Scans all nearest game objects. Expensive, but only called by the Razorgore tank action and multiplier
  - Frontal cone check (_HasInArc_): inexpensive geometric calculations
  - Group iteration in _BwlVaelastraszMoveAwayAction_: O(n) over raid members, only active during Vaelastrasz encounter or while _Burning Adrenaline_ aura is active
  - All triggers/multipliers are encounter-bound, so they should never fire outside the instance

## How to Test the Changes

1. Enter BWL with enough bots with the appropriate level.
2. Engage Razorgore in the first room: The player must mind control the boss and destroy the eggs
    Expected behavior:
    - A paladin bot should enable their fire resistance aura.
    - The off-tank should immediately mark and tank Razorgore until he gets mind controlled
    - Bots should handle all enemy NPCs
    - When all eggs are destroyed, the rti mark should be removed and all bots should engage the boss. 
    - Bots should position themselves to avoid AOE damage.
3. Engage Vaelastrasz in the next room: Talk to him and wait for the RP to finish.
    Expected behavior:
    - A paladin bot should enable their fire resistance aura.
    - A tank should gain threat. Every other bot will position itself at the rear flank.
    - If the current tank gets Burning Adrenaline, the bot will not run away.
    - Every other bot with Burning Adrenaline will try to find a safe spot to explode.
    - ~~Warlocks should not Drain Soul before 3%~~
    - If the boss is dead and a bot still has Burning Adrenaline, it will still try to avoid other raid members.
4. Reach the suppression room. If cheats are enabled, all bots should be able to deactivate the suppression devices. Otherwise only rogues will do that.

## Impact Assessment

    - - [ ] No, not at all

Encounter-bound triggers/actions, only active in BWL instance

    - - [ ] No
    - - [x] Yes (**explain why**)
              
New boss behavior for Razorgore & Vaelastrasz (BWL-specific)

    - - [ ] No
              
Self-contained within BWL modules (triggers/actions/multipliers), no changes to global systems

## AI Assistance

- - [ ] No

**brainstorming and code generation**
The code in this PR is mostly written by me.
Originally, when bots got Burning Adrenaline from Vaelastrasz, they went to one of 12 hard-coded positions inside Vaelastrasz' room. But I didn't like that approach, so I asked Big Pickle from OpenCode for some advice and it generated the initial version of the current weighted repulsion approach. The following finalization of this code was done by me again.

**review** 
I used Big Pickle during development for code reviews. Reasonable suggestions were reviewed and applied by me.

Also the first iteration of the text of this PR was generated by AI.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

I know that this PR only contains the first two bosses but I wanted to take a break from playing AzerothCore, so I decided to submit and merge my current progress.

### DIFF
--- a/src/Ai/Raid/BWL/BWLActions.cpp
+++ b/src/Ai/Raid/BWL/BWLActions.cpp
@@ -19,19 +19,19 @@ static constexpr float VAELASTRASZ_REPULSION_RANGE = 20.0f;
 static constexpr float VAELASTRASZ_TAIL_SWEEP_RECOVERY_RANGE = 30.0f;
 static constexpr float VAELASTRASZ_RANGED_BIAS_THRESHOLD = 25.0f;
 static constexpr float VAELASTRASZ_RANGED_BIAS_FLEE_WEIGHT = 0.7f;
-static constexpr float VAELASTRASZ_RANGED_BIAS_BOSS_WEIGHT = 1 - VAELASTRASZ_RANGED_BIAS_FLEE_WEIGHT;
+static constexpr float VAELASTRASZ_RANGED_BIAS_BOSS_WEIGHT = 1.0f - VAELASTRASZ_RANGED_BIAS_FLEE_WEIGHT;
 
 // General
 
 bool BwlOnyxiaScaleCloakAuraCheckAction::Execute(Event /*event*/)
 {
-    bot->AddAura(SPELL_ONYXIA_SCALE_CLOAK, bot);
+    bot->AddAura(static_cast<uint32>(BlackwingLairSpells::SPELL_ONYXIA_SCALE_CLOAK), bot);
     return true;
 }
 
 bool BwlOnyxiaScaleCloakAuraCheckAction::isUseful()
 {
-    return !bot->HasAura(SPELL_ONYXIA_SCALE_CLOAK);
+    return !bot->HasAura(static_cast<uint32>(BlackwingLairSpells::SPELL_ONYXIA_SCALE_CLOAK));
 }
 
 bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
@@ -48,23 +48,17 @@ bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
     return true;
 }
 
-bool BwlTurnOffSuppressionDeviceAction::isUseful()
-{
-    // Originally, only rogues could disarm suppression devices.
-    // If raid cheats are enabled, any bot can disarm the devices.
-    return bot->IsClass(CLASS_ROGUE) || botAI->HasCheat(BotCheatMask::raid);
-}
-
 // Razorgore the Untamed
 
-bool BwlRazorgoreAvoidAoe::Execute(Event /*event*/)
+bool BwlRazorgoreAvoidAoeAction::Execute(Event /*event*/)
 {
     Unit* boss = AI_VALUE2(Unit*, "find target", "razorgore the untamed");
     if (!boss)
         return false;
 
-    // The current tank should not move
-    if (PlayerbotAI::IsTank(bot) && boss->GetVictim() == bot)
+    // The current target should not move.
+    // Also prevents non-tanks from rotating the boss if they have aggro.
+    if (boss->GetVictim() == bot)
         return false;
 
     float distance = bot->GetDistance2d(boss);
@@ -91,7 +85,7 @@ bool BwlRazorgoreAvoidAoe::Execute(Event /*event*/)
         float moveY = bot->GetPositionY() + (dY / dist) * INCREMENTAL_MOVE_STEP_DISTANCE;
 
         return MoveTo(boss->GetMapId(), moveX, moveY, bot->GetPositionZ(), false, false,
-                      false, false, MovementPriority::MOVEMENT_COMBAT);
+                      false, false, MovementPriority::MOVEMENT_COMBAT, true);
     }
 
     // Ranged outside the cone but too close => back off to avoid War Stomp
@@ -167,7 +161,7 @@ bool BwlVaelastraszMoveAwayAction::CalculateFleeDirection(const Unit* boss, floa
                 continue;
 
             // Boss alive: ignore other Burning Adrenaline bots so they can group together
-            if (!bossDead && p->HasAura(SPELL_BURNING_ADRENALINE))
+            if (!bossDead && p->HasAura(static_cast<uint32>(BlackwingLairSpells::SPELL_BURNING_ADRENALINE)))
                 continue;
 
             // Compute a flee direction via weighted repulsion from other bots.
@@ -254,7 +248,7 @@ bool BwlVaelastraszMoveAwayAction::MoveAlongFleeDirection(const Unit* boss, floa
                 continue;
 
             if (MoveTo(bot->GetMapId(), targetX, targetY, bot->GetPositionZ(), false, false,
-                       false, false, MovementPriority::MOVEMENT_FORCED))
+                       false, false, MovementPriority::MOVEMENT_FORCED, true))
                 return true;
         }
     }
@@ -266,7 +260,7 @@ bool BwlVaelastraszMoveAwayAction::MoveAlongFleeDirection(const Unit* boss, floa
 
 bool BwlUseHourglassSandAction::Execute(Event /*event*/)
 {
-    return botAI->CastSpell(SPELL_HOURGLASS_SAND, bot);
+    return botAI->CastSpell(static_cast<uint32>(BlackwingLairSpells::SPELL_HOURGLASS_SAND), bot);
 }
 
 bool BwlNefarianFearWardAction::Execute(Event /*event*/)
@@ -292,7 +286,7 @@ Unit* BwlDeathTalonWyrmguardTankMoveAwayAction::GetTarget()
     for (auto const& [guid, ref] : bot->GetThreatMgr().GetThreatenedByMeList())
     {
         Unit* unit = ref->GetOwner();
-        if (!unit || !unit->IsAlive() || unit->GetEntry() != NPC_DEATH_TALON_WYRMGUARD)
+        if (!unit || !unit->IsAlive() || unit->GetEntry() != static_cast<uint32>(BlackwingLairNPCs::NPC_DEATH_TALON_WYRMGUARD))
             continue;
         if (bot->GetDistance2d(unit) > WYRMGUARD_SAFE_DISTANCE)
             continue;
@@ -312,7 +306,8 @@ bool BwlDeathTalonWyrmguardTankMoveAwayAction::isUseful()
     for (auto const& [guid, ref] : bot->GetThreatMgr().GetThreatenedByMeList())
     {
         Unit* unit = ref->GetOwner();
-        if (unit && unit->IsAlive() && unit->GetEntry() == NPC_DEATH_TALON_WYRMGUARD && unit->GetVictim() == bot)
+        if (unit && unit->IsAlive() && unit->GetEntry() == static_cast<uint32>(BlackwingLairNPCs::NPC_DEATH_TALON_WYRMGUARD) &&
+            unit->GetVictim() == bot)
             return true;
     }
     return false;
@@ -339,7 +334,7 @@ Unit* BwlDeathTalonWyrmguardRangedMoveAwayAction::GetTarget()
     for (auto const& [guid, ref] : bot->GetThreatMgr().GetThreatenedByMeList())
     {
         Unit* unit = ref->GetOwner();
-        if (!unit || !unit->IsAlive() || unit->GetEntry() != NPC_DEATH_TALON_WYRMGUARD)
+        if (!unit || !unit->IsAlive() || unit->GetEntry() != static_cast<uint32>(BlackwingLairNPCs::NPC_DEATH_TALON_WYRMGUARD))
             continue;
         float dist = bot->GetDistance2d(unit);
         if (dist < closestDist)
--- a/src/Ai/Raid/BWL/BWLActions.cpp
+++ b/src/Ai/Raid/BWL/BWLActions.cpp
@@ -1,9 +1,25 @@
 #include "BWLActions.h"
 
+#include "RtiTargetValue.h"
 #include "Playerbots.h"
 #include "BWLHelpers.h"
+#include "RaidBossHelpers.h"
 
 using namespace BlackwingLairHelpers;
+
+static constexpr float INCREMENTAL_MOVE_STEP_DISTANCE = 3.0f;
+
+static constexpr float RAZORGORE_CONE_RADIUS = 15.0f;
+static constexpr float RAZORGORE_CONE_ARC = M_PI;
+static constexpr float RAZORGORE_BACK_FUZZ_ARC = M_PI_4;
+static constexpr float RAZORGORE_RANGE_DISTANCE = 15.0f;
+static constexpr float RAZORGORE_MELEE_DISTANCE = 3.0f;
+
+static constexpr float VAELASTRASZ_REPULSION_RANGE = 20.0f;
+static constexpr float VAELASTRASZ_TAIL_SWEEP_RECOVERY_RANGE = 30.0f;
+static constexpr float VAELASTRASZ_RANGED_BIAS_THRESHOLD = 25.0f;
+static constexpr float VAELASTRASZ_RANGED_BIAS_FLEE_WEIGHT = 0.7f;
+static constexpr float VAELASTRASZ_RANGED_BIAS_BOSS_WEIGHT = 1 - VAELASTRASZ_RANGED_BIAS_FLEE_WEIGHT;
 
 // General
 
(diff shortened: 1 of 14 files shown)


src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -32,6 +48,220 @@ bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
...
+        float dist = sqrt(dX * dX + dY * dY);
+        if (dist == 0.0f)
+            dist = 0.1f;
+
+        float moveX = bot->GetPositionX() + (dX / dist) * INCREMENTAL_MOVE_STEP_DISTANCE;
+        float moveY = bot->GetPositionY() + (dY / dist) * INCREMENTAL_MOVE_STEP_DISTANCE;
+
+        return MoveTo(boss->GetMapId(), moveX, moveY, bot->GetPositionZ(), false, false,
+                      false, false, MovementPriority::MOVEMENT_COMBAT);
+    }
```
brighton-chi, line 89: Curious why you're going with the default false for the lessDelay parameter. I haven't done any comparisons in a long time, but back when I was playing with the parameter, I felt passing true made movement noticeably smoother and couldn't really think of any disadvantage to just passing true all the time.

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -32,6 +48,220 @@ bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
...
+        ClearTargetIcon(bot, RtiTargetValue::moonIndex);
+    }
+
+    return false;
+}
+
+bool BwlRazorgoreMarkBossAction::isUseful()
+{
+    return PlayerbotAI::IsTank(bot);
+}
```
brighton-chi, line 130: Is there a benefit to doing the IsTank check here instead of at the top of the trigger?

src/Ai/Raid/BWL/BWLActions.h
```diff
@@ -2,7 +2,7 @@
 #define PLAYERBOTS_BWLACTIONS_H
 
 #include "Action.h"
-#include "MovementActions.h"
+#include "AttackAction.h"
```
brighton-chi, line 6: Since MovementAction is used too, should it still be included? I know it's included transitively through AttackAction.h, but on that basis, Action.h is transitively included too.

src/Ai/Raid/BWL/BWLHelpers.h
```diff
@@ -2,6 +2,7 @@
 #define PLAYERBOTS_BWLHELPERS_H
 
 #include "Player.h"
+#include "PlayerbotAI.h"
 
 namespace BlackwingLairHelpers
 {
```
brighton-chi, line 8: The people in charge (or maybe just Keleborn...?) have decreed that we should start using enum classes instead of enums for boss strategies

src/Ai/Raid/BWL/BWLHelpers.h
```diff
@@ -31,6 +41,9 @@ namespace BlackwingLairHelpers
     };
 
     bool IsActiveSuppressionDeviceInRange(const GameObject* go, const Player* bot);
+    bool AreRazorgoreEggsAlive(PlayerbotAI* botAI);
+    bool IsRazorgoreOffTank(Player* bot);
+    bool IsNonBABotNearPosition(const Player* bot, Position const& position, float distance);
 }
 
 #endif //_PLAYERBOT_RAIDBWLHELPERS_H
```
brighton-chi: Comment needs update (or just removal)

src/Ai/Raid/BWL/BWLMultipliers.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (!bot->IsClass(CLASS_WARLOCK))
+        return 1.0f;
+
+    Unit* boss = AI_VALUE2(Unit*, "find target", "vaelastrasz the corrupt");
+    if (!boss)
+        return 1.0f;
+
+    // Vaelastrasz starts at 30% hp. Warlocks will try to use Drain Soul at 20% hp.
+    // Delay usage of Drain Soul until 3% hp because of low damage.
+    if (dynamic_cast<CastDrainSoulAction*>(action) && action->GetTarget() == boss &&
```
brighton-chi: Is usage/effectiveness of execute-type abilities against Vael calculated based on the percentage of his current HP vs. his starting HP, not vs. his max HP?

src/Ai/Raid/BWL/BWLTriggers.cpp
```diff
@@ -14,13 +14,36 @@ bool BwlSuppressionDeviceTrigger::IsActive()
...
+
+bool BwlVaelastraszPositioningTrigger::IsActive()
+{
+    // Prevent non-tanks from rotating the boss while the tanks gain thread.
+    if (Unit* boss = AI_VALUE2(Unit*, "find target", "vaelastrasz the corrupt"))
+        return boss->GetVictim() != bot;
+    return false;
+}
+
+bool BwlVaelastraszBurningAdrenaline::IsActive()
```
brighton-chi: BwlVaelastraszBurningAdrenalineTrigger?

src/Ai/Raid/BWL/BWLMultipliers.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "HunterActions.h"
+#include "Playerbots.h"
+#include "ReachTargetActions.h"
+#include "WarlockActions.h"
+
+using namespace BlackwingLairHelpers;
+
+static constexpr float VAELASTRASZ_BA_SAFE_DISTANCE = 20.0f;
+static constexpr float VAELASTRASZ_BA_BOSS_DISTANCE = 30.0f;
+static constexpr float VAELASTRASZ_WARLOCK_DRAIN_SOUL_THRESHOLD = 3;
```
brighton-chi: I know it doesn't matter functionally, but just for consistency I think this should be 3.0f

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -32,6 +48,220 @@ bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
...
+// Razorgore the Untamed
+
+bool BwlRazorgoreAvoidAoe::Execute(Event /*event*/)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "razorgore the untamed");
+    if (!boss)
+        return false;
+
+    // The current tank should not move
+    if (PlayerbotAI::IsTank(bot) && boss->GetVictim() == bot)
```
brighton-chi: Maybe this should just be anybody who is the current target?

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -32,6 +48,220 @@ bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
...
+        return false;
+
+    float distance = bot->GetDistance2d(boss);
```
brighton-chi, line 64: Since this local is used only once, I'm not sure it's useful

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -32,6 +48,220 @@ bool BwlTurnOffSuppressionDeviceAction::Execute(Event /*event*/)
     return true;
 }
 
+bool BwlTurnOffSuppressionDeviceAction::isUseful()
```
brighton-chi: I feel like these checks would benefit from moving to the top of the trigger to exclude unnecessary gameobject searches

brighton-chi (commented):
Awesome, I feel like I learned some stuff here. I did not know about the HasInArc function, and the custom flee action you've put in is interesting.

src/Ai/Raid/BWL/BWLHelpers.h
Celandriel: We follow ac style 😉

brighton-chi: Despite the fact that AC does not even begin to follow AC style in this regard. I think it's better this way but find it funny.

isitsimon:
👏🏻👏🏻👏🏻

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: Most of the calls in raid strategies use the default value for _lessDelay_, except some actions in ICC. So I might have just copied the call.
After some tests, I didn't notice much of an improvement but it can't hurt, so I set the _lessDelay_ parameter to _true_ in both _MoveTo_ calls.

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: I didn't put any thought into it. But after checking [_DoNextAction_ in _Engine.cpp_](https://github.com/mod-playerbots/mod-playerbots/blob/049f35906ed0b66ea5fcdd7fdc9f12ca2ab480ca/src/Bot/Engine/Engine.cpp#L179), it looks like we can skip some code by returning false inside _isUseful_ instead of _Execute_.

src/Ai/Raid/BWL/BWLActions.h
HennyWilly: Included _MovementActions.h_

src/Ai/Raid/BWL/BWLHelpers.h
HennyWilly: Changed enums to enum classes.
Don't get me wrong, I like enum classes and strong typing but theses static casts feel a bit redundant when there is an underlying type on the enum class. Not complaining about the comment but about how it is handled in C++.
There is also `std::to_underlying` in C++ 23, but AC uses C++ 20...

src/Ai/Raid/BWL/BWLHelpers.h
HennyWilly: Removed. This was the only header in BWL that had this kind of comment anyway.

src/Ai/Raid/BWL/BWLMultipliers.cpp
HennyWilly: TL;DR: The percentage is based on his max HP.

As far as I can tell, `CastDrainSoulAction` is triggered by `TargetCriticalHealthTrigger`. One of the base classes (`ValueInRangeTrigger`) uses the "health" value. And the `HealthValue` class calculates the targets health in percent relative to the max health:
`return (static_cast<float>(target->GetHealth()) / target->GetMaxHealth()) * 100`

src/Ai/Raid/BWL/BWLTriggers.cpp
HennyWilly: Whoops, I missed that one.
I also missed "Action" on `BwlRazorgoreAvoidAoeAction`

src/Ai/Raid/BWL/BWLMultipliers.cpp
HennyWilly: Fixed

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: Good point. I did the same for Vaelastrasz. Works as intended.

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: It's also used inside `if (PlayerbotAI::IsRanged(bot))`:
`float distToTravel = RAZORGORE_RANGE_DISTANCE - distance;`

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: Good catch. Moved the check to the top of the trigger.
But since raid cheats are [enabled by default](https://github.com/mod-playerbots/mod-playerbots/blob/049f35906ed0b66ea5fcdd7fdc9f12ca2ab480ca/conf/playerbots.conf.dist#L590), there would only be benefits if raid cheats were manually disabled.

HennyWilly:
@brighton-chi Thanks for the review. Much appreciated.

src/Ai/Raid/BWL/BWLActions.cpp
brighton-chi: I was wondering if it would be better in the trigger and not in the action at all, but honestly I don't know if it makes a difference in performance to have it in the trigger instead of isUseful or isPossible

src/Ai/Raid/BWL/BWLActions.cpp
brighton-chi: True. I never even considered the suppression room being difficult because I main a Rogue, and I just assumed the Playerbots version was always disabling based on proximity, regardless of class.

src/Ai/Raid/BWL/BWLActions.cpp
brighton-chi: I think it's some pretty specific scenarios where you really notice it and it actually makes a difference. I first did for BBW in Karazhan, when movement was through a path of fixed waypoints. There was a noticeably delay upon reaching one before going to the next. I've just not been able to think of any disadvantage to always using true there.

src/Ai/Raid/BWL/BWLMultipliers.cpp
brighton-chi: Sorry, I failed to express what I actually was thinking about. What I mean is that if execute abilities that are gated on health checks inherently would come into effect at Vael's percentage health compared to his max, then don't we want drain soul to kick in when it normally does? Only Affliction Warlocks are coded to use it, and it is part of their standard rotation in execute range (it gets 4x damage when the target is below 25% health).

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: Ah, sorry. I need to start reading more carefully...

There might be some overhead when doing the check inside of _isUseful_ instead of the trigger. I also tend to do those checks as early as possible.
In this case I did not because `BwlRazorgoreNotMindControlledTrigger` is used for two actions: `BwlRazorgoreAvoidAoeAction` and `BwlRazorgoreMarkBossAction`. In theory, I could parameterize the trigger or implement a second trigger that does the same but only for tanks. But that felt redundant.

src/Ai/Raid/BWL/BWLMultipliers.cpp
HennyWilly: Wow, I wasn't aware that Drain Soul was that good of an execute ability during Wrath. I distinctly remember that Drain Soul was only useful for Soul Shard generation during TBC, so I thought bots were wasting DPS during this fight.

Quotes from WotLK affliction warlock guides:
- [Drain Soul is the hardest hitting execute ability in the game](https://www.icy-veins.com/wotlk-classic/affliction-warlock-dps-pve-guide)
- [Drain Soul only saw use for grinding out Soul Shard in the past. Those days are over, Drain Soul now is major high damage dealing execute ability.](https://www.wowhead.com/wotlk/guide/classes/warlock/affliction/dps-overview-pve)

In my tests without the multiplier, the Drain Soul dot accounted for about 40% of the warlock's total damage while fighting Vaelastrasz.
So long story short, I removed the multiplier, so warlocks will Drain Soul at 20% hp again.

kadeshar (approved):

PR #2518, opened 2026-06-28
Rewrite Magtheridon Strategy

brighton-chi:
## Pull Request Description

I hated this strategy, and when I tried Magtheridon the other day, it didn't even work, so I decided it was time to just rewrite it. Notable changes with respect to functionality include:

- The main tank's waiting position and actual tanking position have been changed so now the tank should wait between the boss and the group when the Channeler phase ends, and then the move to the final tank position should leave the tank in range of healers easily.
- Ranged positioning logic was significantly simplified, with a priority of staying away from the boss and then a minor spread, instead of the overly complicated ranged/healer locations that the prior strategy attempted.
- Various silly behaviors have been solved through multipliers (e.g., main tank charging and taunting on the way to the waiting position, tank running back and forth to avoid Conflagration while tanking).
- Debris avoidance actually works now.
- Blast Nova timer no longer becomes desync'd by a failed or late interrupt, and Quake and ceiling collapse delays are accounted for.

## Feature Evaluation

The main changes with respect to the code are:
- Using spell hooks for Debris and Quake. Debris is cast by a trigger NPC, but the triggers pat and what remains is a dynamic object. When the ceiling actually falls 5s later, it is a one-shot. So this is crucial to avoid and cannot be detected without a spell hook. Quake is an instant cast by Magtheridon with no associated NPC or object and thus cannot be detected without a spell hook. Whenever it is cast, it pushes the next Blast Nova out by 7 seconds, so bots need to read it to determine timing for cube clicking.
- Calling RequestSpellInterrupt() to move out of Debris.
- Having dedicated, standalone functions for assigning and replacing cube clicker bots instead of the prior version of those processes being scattered haphazardly through actions and triggers.

A couple of maps were added, and a couple were removed. I tried to scope locally where possible, but the remaining maps need to be shared, either across actions and/or bots (for example, the MechanicTracker manages the Blast Nova timer, and that needs to be shared in case the MechanicTracker dies and is reassigned).

## How to Test the Changes

Fight Magtheridon. Try killing off cube clickers to make sure replacement assignments are made and timers are not desynced. 

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [ ] Yes (**explain why**)
Only with the “magtheridon” strategy active.

    - - [x] No

## AI Assistance

- - [ ] No

Used for Debris spell hook and cube clicker assignment logic

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

There is a super weird bug with Magtheridon that seems inexplicable based on his boss script. After about 20% HP, instead of casting Blast Nova every ~54s or so, he starts casting it with a repeating ~42s/~21s pattern, and without delays for Quake. This screws up the ability to stop Blast Nova, and actually makes it impossible with a single cube-clicker group for the 21s cast because the Mind Exhaustion debuff has not expired yet.

There are ways I could try to work around this, but this is a clear bug and I don't have the patience to try to adapt something for this purpose. The fight is beatable anyway.


src/Ai/Raid/Mag/Util/MagScripts.cpp
```diff
@@ -0,0 +1,76 @@
...
+using namespace MagtheridonHelpers;
+
+class MagtheridonBotSpellScript : public AllSpellScript
+{
+public:
+    MagtheridonBotSpellScript() : AllSpellScript("MagtheridonBotSpellScript") {}
+
+    void OnSpellCast(
+        Spell* /*spell*/, Unit* caster, SpellInfo const* spellInfo, bool /*skipCheck*/) override
+    {
```
kadeshar, line 25: AC implementation have null checks 

Image

src/Ai/Raid/Mag/Util/MagHelpers.cpp
```diff
@@ -0,0 +1,139 @@
...
+const Position RANGED_SPREAD_POSITION =           { -14.890f,   1.995f, -0.406f,   0.0f };
+const Position HEALER_SPREAD_POSITION =           {  -2.265f,   1.874f, -0.404f,   0.0f };
+
+std::unordered_map<uint32, time_t> blastNovaTimer;
+std::unordered_map<uint32, time_t> dpsWaitTimer;
+std::unordered_map<uint32, bool> ceilingCollapseApplied;
+std::unordered_map<uint32, bool> lastBlastNovaState;
+std::unordered_map<uint32, std::unordered_map<ObjectGuid, CubeInfo>>
+    botToCubeAssignments;
+std::unordered_map<uint32, std::vector<DebrisData>> activeDebrisPositions;
```
kadeshar, line 30: That is cleared somewhere?

src/Ai/Raid/Mag/MagActionContext.h
```diff
@@ -9,29 +15,74 @@ class RaidMagtheridonActionContext : public NamedObjectContext<Action>
...
+        creators["magtheridon main tank attack first three channelers"] =
+            &RaidMagtheridonActionContext::magtheridon_main_tank_attack_first_three_channelers;
+
+        creators["magtheridon first assist tank attack nw channeler"] =
+            &RaidMagtheridonActionContext::magtheridon_first_assist_tank_attack_nw_channeler;
+
+        creators["magtheridon second assist tank attack ne channeler"] =
+            &RaidMagtheridonActionContext::magtheridon_second_assist_tank_attack_ne_channeler;
+
+        creators["magtheridon misdirect hellfire channelers"] =
```
kadeshar: Should be "magtheridon misdirect hellfire channelers to main tank"?

src/Ai/Raid/Mag/Util/MagScripts.cpp
```diff
@@ -0,0 +1,76 @@
...
+        const uint32 instanceId = caster->GetMap()->GetInstanceId();
```
kadeshar: Maybe early check on mapId is good idea?

src/Ai/Raid/Mag/Util/MagHelpers.h
```diff
@@ -0,0 +1,91 @@
...
+extern const Position RANGED_SPREAD_POSITION;
+extern const Position HEALER_SPREAD_POSITION;
+
+extern std::unordered_map<uint32, time_t> dpsWaitTimer;
+extern std::unordered_map<uint32, time_t> blastNovaTimer;
+extern std::unordered_map<uint32, bool> ceilingCollapseApplied;
+extern std::unordered_map<uint32, bool> lastBlastNovaState;
+extern std::unordered_map<uint32, std::unordered_map<ObjectGuid, CubeInfo>>
+    botToCubeAssignments;
+extern std::unordered_map<uint32, std::vector<DebrisData>> activeDebrisPositions;
```
kadeshar, line 77: All this objects have proper cleaning while bots leave instance, logout etc?

kadeshar (changes requested):

src/Ai/Raid/Mag/Util/MagHelpers.cpp
brighton-chi: Good call, I should probably make a separate out-of-combat cleanup action like I've done for other strategies.

src/Ai/Raid/Mag/MagActionContext.h
brighton-chi: Yep, thanks.

src/Ai/Raid/Mag/Util/MagScripts.cpp
brighton-chi: It's totally impossible for any of those to be null for OnSpellCast(). OnSpellCast(this, m_caster, m_spellInfo, skipCheck) is called in Spell::cast(). "this" cannot be null, and m_caster and m_spellInfo are set in the constructors and immediately dereference in the constructors if null. So the server crashes long before getting here if any are null.

src/Ai/Raid/Mag/Util/MagScripts.cpp
brighton-chi: What do you mean? Are you concerned about performance? I don't think GetInstanceId() has any real cost, it's just one integer check. I could move the instanceId inside each spell check I guess, it would be doubled but then it wouldn't be checked until the spellInfo is checked (and those checks should be free since we already have the spellInfo by the time OnSpellCast is called).

kadeshar: Yeap, stricte performance to early exit method which is often called

brighton-chi: Ok well I think the way I changed it to should be the best for performance then? Since we already have spell info so it's just a direct integer comparison and there's nothing to retrieve.

kadeshar (approved):

PR #2526, opened 2026-06-30
Rename IsAlt to IsAltBot, following other bot check conventions; drop some confusing local variables

NoxMax:
## Pull Request Description

1. IsAlt is standardized into the convention of having "Bot" added to it like IsRandomBot and IsAddClassBot.
2. One local definition of IsRandomBot was dropped. That local variable should techically be named isRandomBot, but the whole usage here is confusing. Used inline directly. It's better and cleaner given the usage here.
3. One local definition of isSelfBot was dropped. Globally IsRealPlayer is used to refer to self-bots, so for now lets follow that convention. In the future I would like to globally define IsSelfBot for self-bots, and IsRealPlayer would then be defined as a real player under the players control (no botAI).

## Feature Evaluation

## How to Test the Changes

Check that everything compiles correctly. If it fails, it must be that you have another module that calls IsAlt and requires the term over there to be updated.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

If you know of a module that calls onto Playerbots' `IsAlt`, please share it.

### DIFF
--- a/src/Ai/Base/Actions/LeaveGroupAction.cpp
+++ b/src/Ai/Base/Actions/LeaveGroupAction.cpp
@@ -126,7 +126,7 @@ bool LeaveFarAwayAction::isUseful()
     if (trueMaster && !GET_PLAYERBOT_AI(trueMaster))
         return false;
 
-    if (botAI->IsAlt() &&
+    if (botAI->IsAltBot() &&
         (!groupLeaderBotAI || groupLeaderBotAI->IsRealPlayer()))  // Don't leave group when alt grouped with player groupLeader.
         return false;
 
--- a/src/Ai/Base/Actions/PassLeadershipToMasterAction.cpp
+++ b/src/Ai/Base/Actions/PassLeadershipToMasterAction.cpp
@@ -34,7 +34,7 @@ bool PassLeadershipToMasterAction::Execute(Event /*event*/)
 
 bool PassLeadershipToMasterAction::isUseful()
 {
-    return botAI->IsAlt() && bot->GetGroup() && bot->GetGroup()->IsLeader(bot->GetGUID());
+    return botAI->IsAltBot() && bot->GetGroup() && bot->GetGroup()->IsLeader(bot->GetGUID());
 }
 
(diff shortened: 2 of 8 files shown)


brighton-chi:
I must be missing something really obvious here, but it looks to me like public members are supposed to be PascalCase in AC.

https://www.azerothcore.org/wiki/cpp-code-standards

NoxMax:
NOPE, you are right. I seems to have read "Methods are always UpperCamelCase", somehow did not finish the sentence and deleted the "Upper" out of my head. Ending up with "Methods are always ~Upper~CamelCase" 🤦

So this is mostly moot. The only one that is relevant is `IsAlt` should follow the other ones and be `IsAltBot`. Will update shortly.

brighton-chi:
This isn't directly about your PR, but AC code style seems very loosely adhered to in AC itself and PB, even with ongoing PRs. It's not at a stage where things are currently consistent, much less retroactively.

There are things that are allegedly AC style but are almost never followed, like using enum classes, putting const after the type name, or not having blank spaces before/after braces in arrays.

On top of that, it doesn't help that AC code style has some what I understand are very questionable choices, such as leading underscores for private members (even functions), which I understand is pretty much universally considered bad practice.

I'm normally kind of a stickler for technicalities, but the current situation is disorganized enough that it is difficult to take seriously. This isn't a knock on anybody, and I'm not proposing any useful ideas here. It may just be the reality of this type of project (a lot of code, many contributors over time who didn't follow unified approaches, a long history that far predates anybody who is still involved, and a small active community).

NoxMax:
There. It's like nothing weird happened at all 😂

But yeah as you said, there are endless style inconsistencies in AC itself, let alone for us over here. Obviously there can't be a "standardize the style everywhere" project. It's futile. But more of fix things here and there on your way while you are doing something else.

kadeshar (approved):

PR #2529, opened 2026-07-01
fix(CORE/Druid): Fall back to Bear Form for pull below level 40

ppaslan:
## Pull Request Description

The druid pull pre-action was hardcoded to "dire bear form", which is impossible for druids under ~level 40. Because the pull invokes it via DoSpecificAction() (which does not honor an action's getAlternatives() fallback), the shift failed and PullStartAction aborted, so lower-level feral bots could not pull at all.

Fall back to Bear Form when Dire Bear Form is untrained; behavior at level 40+ is unchanged (still pulls in Dire Bear Form).

## Feature Evaluation

Minimum logic: one extra check in DruidPullStrategy::GetPreActionName(). If the bot doesn't know Dire Bear Form, return "bear form" instead of "dire bear form". Nothing else changes.

Processing cost: one HasSpell() call, and only when a pull actually starts, not on the per-tick loop. It sits next to HasSpell()/CanCastSpell() calls the same function already makes, so there's no meaningful cost even with a lot of bots.

## How to Test the Changes

**Setup:**

- 1 player + 1 Playerbot druid (alt bot or invited to group).
- The druid must be below level 40 (i.e., has not yet trained Dire Bear Form) and specced/set as feral tank.
- Give it the tank kit - whisper talents spec bear pve, then co +bear,+tank assist,+pull,+pull back.

**Steps:**

1. Put the bot in caster form.
2. Target a nearby enemy.
3. Whisper the bot - pull my target.

**Expected behavior:**

- Before this change: the bot fails to pull — it does nothing (the dire bear form pre-action returns "impossible", aborting the pull). If out of Bear Form, whispering do dire bear form also returns dire bear form: impossible.
- After this change: the bot shifts into Bear Form as the pull pre-action, casts its pull ability (Faerie Fire/Growl), and returns to its pull position — pulling works normally.

**Regression check (level 40+):**

Repeat with a druid that has trained Dire Bear Form. It should still pull using Dire Bear Form exactly as before confirming max-level behavior is unchanged.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Below level 40 the druid now shifts into Bear Form and pulls, instead of failing to pull. Level 40+ is unchanged.

    - - [ ] No

One if-branch, the same shape PaladinPullStrategy::GetPreActionName() already uses.

## AI Assistance

- - [ ] No

Used it to help trace the pull code path and to tidy up this write-up. The actual fix is one line, and I understand it and can explain it.

 MaNGOS, another module)?

## Final Checklist

-  [x] Stability is not compromised.
-  [x] Performance impact is understood, tested, and acceptable.
-  [x] Added logic complexity is justified and explained.
-  [x] Any new bot dialogue lines are translated. (N/A - no new dialogue.)
-  [x] Any code ported/adapted from another project is attributed. (N/A - original.)
-  [x] New source files use the GPLv2 header. (N/A - no new files.)
-  [x] Documentation updated if needed (Conf comments, WiKi commands). (N/A.)

## Notes for Reviewers

CastDireBearFormAction already lists "bear form" in getAlternatives(), but that fallback is only applied by the engine's normal action-selection loop (DoNextAction), not by the DoSpecificAction() path the pull pre-action uses. That's why the fix goes in GetPreActionName() rather than relying on the alternative.

### DIFF
--- a/src/Ai/Class/Druid/Strategy/DruidPullStrategy.cpp
+++ b/src/Ai/Class/Druid/Strategy/DruidPullStrategy.cpp
@@ -8,6 +8,8 @@
 #include "PlayerbotAI.h"
 #include "Playerbots.h"
 
+constexpr uint32 SPELL_DIRE_BEAR_FORM = 9634;
+
 std::string DruidPullStrategy::GetPullActionName() const
 {
     std::string const pullActionName = PullStrategy::GetPullActionName();
@@ -32,7 +34,7 @@ std::string DruidPullStrategy::GetPreActionName() const
         return "";
 
     // Fall back to Bear Form when Dire Bear Form isn't trained yet (< ~level 40).
-    if (!botAI->HasSpell("dire bear form"))
+    if (!botAI->GetBot()->HasSpell(SPELL_DIRE_BEAR_FORM))
         return "bear form";
 
     return PullStrategy::GetPreActionName();
--- a/src/Ai/Class/Druid/Strategy/DruidPullStrategy.cpp
+++ b/src/Ai/Class/Druid/Strategy/DruidPullStrategy.cpp
@@ -31,5 +31,9 @@ std::string DruidPullStrategy::GetPreActionName() const
     if (GetPullActionName() == "faerie fire")
         return "";
 
+    // Fall back to Bear Form when Dire Bear Form isn't trained yet (< ~level 40).
+    if (!botAI->HasSpell("dire bear form"))
+        return "bear form";
+
     return PullStrategy::GetPreActionName();
 }


src/Ai/Class/Druid/Strategy/DruidPullStrategy.cpp
```diff
@@ -31,5 +31,9 @@ std::string DruidPullStrategy::GetPreActionName() const
     if (GetPullActionName() == "faerie fire")
         return "";
 
+    // Fall back to Bear Form when Dire Bear Form isn't trained yet (< ~level 40).
+    if (!botAI->HasSpell("dire bear form"))
```
brighton-chi: I can't really speak to the functionality of this PR, but assuming it's good, then for this HasSpell check, if Dire Bear Form is only one rank, we should use a compile-time constant for the spell id and use the core's Unit:: HasSpell since PlayerbotAI's string-based version is more expensive.

Celandriel: Isnt it better to use an action node instead of a specific line here? Make it a more global change?

ppaslan: > I can't really speak to the functionality of this PR, but assuming it's good, then for this HasSpell check, if Dire Bear Form is only one rank, we should use a compile-time constant for the spell id and use the core's Unit:: HasSpell since PlayerbotAI's string-based version is more expensive.

Agreed, will make the necessary change and yes the change is functional and confirmed.

ppaslan: > Isnt it better to use an action node instead of a specific line here? Make it a more global change?

Good point that the alternative already exists the reason it isn't used is that ExecuteAction (the DoSpecificAction path) doesn't consult getAlternatives(), only DoNextAction does. 

I can do the global version, but I want to flag, the only other class with a pull pre-action is Paladin, and it solves it the same scoped GetPreActionName way this PR does. dire bear form is actually the only level-gated pre-action in the system, so a global ExecuteAction change would touch a hot shared path for a problem that only exists in this one spot (as far as i can tell). 

Happy to do the global refactor if you want to establish that pattern deliberately, but the scoped fix matches the existing Paladin precedent and keeps the blast radius tiny which seemed more in line with the predictability-first philosophy. Your call on scope.

src/Ai/Class/Druid/Strategy/DruidPullStrategy.cpp
```diff
@@ -31,5 +33,9 @@ std::string DruidPullStrategy::GetPreActionName() const
     if (GetPullActionName() == "faerie fire")
         return "";
 
+    // Fall back to Bear Form when Dire Bear Form isn't trained yet (< ~level 40).
+    if (!botAI->GetBot()->HasSpell(SPELL_DIRE_BEAR_FORM))
+        return "bear form";
```
kadeshar, line 38: That also not require check HasSpell?

PR #2531, opened 2026-07-01
feat: concentrate random bots in players' current zone (opt-in)

NSO73:
## Pull Request Description

Adds an opt-in config `AiPlayerbot.RandomBotConcentrateInPlayerZone` (default `0`). When enabled, a random
bot that is about to random-teleport to a level-appropriate location prefers a location whose zone currently
holds a real (non-GM) player. It falls back to the normal world-wide selection whenever no real player is in
a matching zone (or nobody is online).

Why: on modest hardware, an admin often wants the "living world" illusion focused *around* the player rather
than spread across zones nobody visits. Concentrating teleport targets on occupied zones keeps the visible
population where the player actually is, instead of scattering bots across the map. It pairs naturally with
`AiPlayerbot.SyncLevelWithPlayers` so the concentrated bots also match the players' level range.

## Feature Evaluation

  Inside the existing `RandomTeleportForLevel(Player*)`, after the level-appropriate teleport locations are
  gathered, and *only when the opt-in flag is set*: build the set of `(mapId, zoneId)` pairs currently
  occupied by online non-GM players, then keep only the candidate locations that fall in one of those zones.
  If the resulting subset is empty, keep the original list unchanged (normal behavior). That is the whole
  feature: one config-gated filter plus one helper method (`GetPlayerZoneTeleportLocations`).

  It does **not** run per tick. It only runs inside `RandomTeleportForLevel`, which is already throttled per
  bot by `MinRandomBotTeleportInterval`/`Max` (seconds to thousands of seconds apart). When it does run, the
  added work is: one pass over online real players to build two small sets (`O(P)`), then one pass over the
  bounded candidate-location list (`O(L)`) doing cheap set lookups. The only non-trivial call is
  `sMapMgr->GetZoneId(...)`, and it is guarded by a prior `mapId` membership check so it is invoked only for
  locations that are on a map where a player currently is (usually very few). When the flag is `0`, the added
  cost is a single boolean check.

## How to Test the Changes

1. In `playerbots.conf`, set:
   - `AiPlayerbot.RandomBotConcentrateInPlayerZone = 1`
   - `AiPlayerbot.SyncLevelWithPlayers = 1`
   - a random-bot pool large enough to be visible (e.g. `MinRandomBots`/`MaxRandomBots = 150`).
   - (optional, to observe faster) temporarily lower `MinRandomBotTeleportInterval`/`Max`.
2. Log in a single real character in an open-world zone (e.g. a level-30 character in Stranglethorn Vale).
3. Wait for the random-teleport cycle (or `.playerbots rndbot init`) and observe where bots end up.

Expected: the large majority of eligible random bots (roughly those above the early racial-start band, see
Notes) converge into the player's current zone instead of being scattered world-wide. With the flag off, the
distribution is the stock world-wide spread. Turning `SyncLevelWithPlayers` on as well makes those bots match
the player's level.

Observed on AzerothCore 3.3.5a: with the flag on, eligible random bots visibly converge into the player's
current zone instead of the stock world-wide spread. The number that concentrates is bounded by how many
bots in the pool are level-appropriate for that zone: with `SyncLevelWithPlayers` on, a small pool yields a
noticeably more populated zone around the player; with it off, only the fraction already near the player's
level can converge.

## Impact Assessment

    - - [ ] No, not at all

  Not on any per-tick path. The extra work happens only inside `RandomTeleportForLevel`, which is already an
  infrequent, interval-throttled event, and only when the opt-in flag is enabled. The added cost is a bounded
  `O(P + L)` filter with an early map-membership guard around the only non-trivial call (`GetZoneId`). With
  the flag off it is a single boolean check, so default behavior is unchanged in cost.

    - - [x] No
    - - [ ] Yes (**explain why**)

  The config defaults to `0`; with it disabled the teleport selection is byte-for-byte the previous behavior.

    - - [ ] No

  Minimal and isolated: one new private helper method and one config-gated `if` inside
  `RandomTeleportForLevel`. It introduces no new persistent state, touches no other code paths, and is fully
  short-circuited when the flag is off.

## AI Assistance

- - [ ] No

AI assistance was used; the change was fully reviewed, tested, and is understood and owned by the contributor.

 MaNGOS, another module)?

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (N/A — this change adds no bot dialogue.)
- - [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note). (N/A — all code is original.)
- - [x] New source files use the GPLv2 header. (N/A — no new source files; only edits to existing ones.)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (Config option documented in `playerbots.conf.dist`.)

## Notes for Reviewers

- Opt-in and default-off, so there is zero behavioral or cost change unless an admin explicitly enables it.
- Graceful fallback: if nobody is online or no candidate location matches a player's zone, the original
  world-wide list is used unchanged.
- GM players are intentionally excluded, so a roaming GM does not attract the random-bot population.
- Known engine-level limitation (not introduced by this PR): bots at/below the early racial-start band tend
  to stay pinned near their starting area, so concentration is most visible for bots above that band. Nothing
  in this change alters that behavior; it only filters teleport targets when a teleport already occurs.
- By design, this shifts bot density toward the player's current zone (that is the intended effect). It adds
  no per-tick cost to the teleport logic itself; it only changes *which* location is chosen when a random
  teleport already happens.


Celandriel:
You need to fill out the full PR template.

NSO73:
Done, the full PR template is now filled in. Thanks for the review!

NoxMax:
To be honest, I don't see the justification for it as valid. Even on sub-optimal hardware, you can increase the number of bots in world, but keep a low `BotActiveAlone` value, set it to zero for what its worth and have 10 thousand bots in world, and just the default
```
AiPlayerbot.BotActiveAloneForceWhenInRadius = 150
AiPlayerbot.BotActiveAloneForceWhenInZone = 1
```
It wouldn't be demanding on the server with such settings.

NSO73:
Thanks, that's a fair point and you're right that on a large server it makes this redundant.

The thing is that `BotActiveAlone` + `ForceWhenInZone/InRadius` and this flag solve two different problems:

- `BotActiveAlone` (and the force rules) control **activity/CPU**: which bots run their full AI. `ForceWhenInZone`/`Radius` activate the bots that are **already** in the player's zone; they don't attract or move any bot there.
- This flag controls **distribution**: *which* teleport target a bot picks, i.e. how many bots physically end up in the player's zone in the first place.

So your recipe keeps CPU low, but it doesn't change the spatial spread. With the stock ZoneBracket distribution, bots are scattered across every zone of the bracket, so to get a *visibly* populated zone around the player you need a pool large enough that a big share lands in that one specific zone statistically. Each logged-in bot is a full `Player` in memory even while inactive, so that density costs **RAM**, not just CPU. On a 16 GB box, the "just run thousands of bots" route isn't really available.

What this flag does is decouple density from pool size: a small pool (e.g. 150) + concentrate → a crowded zone around the player on modest hardware, without loading thousands of bots.

It's opt-in and default-off, so it's a no-op for setups like yours where a big pool already gives the density for free. It's really aimed at the low-RAM / small-pool case. Happy to drop it if that niche is out of scope for the project, though.

NoxMax:
I can see the issue of the RAM. But even then, are you aware of [mod-player-bot-level-brackets](https://github.com/DustinHendrickson/mod-player-bot-level-brackets)? I don't use it but a lot of other people do. Take a look at it if you haven't, particularly its config `UseDynamicDistribution`.

NSO73:
Thanks a lot for pointing me to level-brackets, that's a genuinely useful lead. For my actual use case it does cover the need: with `UseDynamicDistribution` concentrating the pool into the player's level bracket, a small pool on modest hardware ends up with enough level-matched bots around me to get the "living world" feel I was after. So from a purely practical standpoint, I'm covered, and I appreciate it.

For the record on how it relates to this PR, in case it's useful to whoever triages it: the two work on different axes and don't overlap.

- level-brackets / `UseDynamicDistribution` acts on the **level** axis: how many bots sit in each level bracket. It raises the count of level-matched bots, but they stay spread across every zone valid for that level, weighted by the static ZoneBracket distribution.
- This PR acts on the **spatial** axis: it filters the random-teleport target to the zone a real player is actually in.

The reason I think this specific behavior belongs in mod-playerbots itself rather than a satellite module: level distribution *can* live in an external module because the level APIs (`Randomize`, `IncreaseLevel`, ...) are public and drivable from outside. The teleport-*destination* selection isn't. The curated candidate-location list and the location-taking teleport are internal to `RandomPlayerbotMgr`, and there's no hook to override the chosen destination. So an external module can only approximate it by re-teleporting bots itself (fighting the built-in teleporter and rebuilding the valid-location data), which is exactly the kind of fragile workaround a small in-core, opt-in filter avoids. If deterministic zone concentration is ever wanted, core is really the only clean home for it.

I'll leave the PR open in case that low-RAM / small-pool niche is worth covering, but I completely understand if you'd rather not add the surface given level-brackets already handles most setups. Either way, thanks for the thoughtful review.

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -1779,13 +1781,55 @@ void RandomPlayerbotMgr::RandomTeleportForLevel(Player* bot)
...
+    std::set<uint32> playerMaps;
+    std::set<std::pair<uint32, uint32>> playerMapZones;
+    for (Player* player : GetPlayers())
+    {
+        if (!player || !player->IsInWorld() || player->IsGameMaster())
+            continue;
+
+        playerMaps.insert(player->GetMapId());
+        playerMapZones.insert(std::make_pair(player->GetMapId(), player->GetZoneId()));
+    }
```
kadeshar: On big servers that will be problematic. Should be somehow cached per tick or something.
You measured performance impact?

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -1779,13 +1781,55 @@ void RandomPlayerbotMgr::RandomTeleportForLevel(Player* bot)
...
+// vector when no player is online or none of the locations match, letting the caller fall back
+// to the default world-wide behaviour.
+std::vector<WorldLocation> RandomPlayerbotMgr::GetPlayerZoneTeleportLocations(std::vector<WorldLocation> const& locs)
+{
+    std::set<uint32> playerMaps;
+    std::set<std::pair<uint32, uint32>> playerMapZones;
+    for (Player* player : GetPlayers())
+    {
+        if (!player || !player->IsInWorld() || player->IsGameMaster())
+            continue;
```
kadeshar, line 1812: What about instances? Bots will teleport there?

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -1779,13 +1781,55 @@ void RandomPlayerbotMgr::RandomTeleportForLevel(Player* bot)
...
+    std::vector<WorldLocation> filtered;
+    if (playerMapZones.empty())
+        return filtered;
+
+    for (WorldLocation const& loc : locs)
+    {
+        if (playerMaps.find(loc.GetMapId()) == playerMaps.end())
+            continue;
+
+        uint32 zoneId = sMapMgr->GetZoneId(PHASEMASK_NORMAL, loc);
```
kadeshar, line 1839: What about zones with phases?

src/Bot/RandomPlayerbotMgr.cpp
NSO73: No, I haven't measured it, and I'd rather say so than hand you a number from my single-player box that wouldn't answer your question. What I can do is be precise about what this code actually touches.

`GetPlayers()` is not a world-wide scan: it returns `RandomPlayerbotMgr::players`, the module's own list of **non random bot** players, pushed in `OnPlayerLogin` and erased in `OnPlayerLogout`. So the loop is `O(P)` over the real players online, not over all sessions or all bots.

It is also not on a per-tick path: the only caller is `RandomTeleportForLevel`, which is throttled per bot by `MinRandomBotTeleportInterval`/`Max` (seconds to thousands of seconds apart). So the frequency is "once per bot teleport", not "once per bot per tick".

That said, your comment did point at a real defect: `GetPlayers()` returns the vector **by value**, so every call was copying the whole player list. Fixed, it now iterates `players` directly.

On the per-tick cache: I deliberately didn't add one. It would introduce persistent state needing invalidation on zone change, logout and GM toggle, to memoize a set that is rebuilt at most once per bot teleport interval, which felt like a worse trade than the scan it saves. But that's a judgement call on my side, not a strong position: if you still want it cached, say so and I'll add it.

src/Bot/RandomPlayerbotMgr.cpp
NSO73: Good catch, that was a real bug. A player sitting in a dungeon, raid, battleground or arena put that instance's `mapId` into the candidate set, and since `WorldLocation` carries no instance id, any bot filtered onto that map would have been sent to a *different* instance of it, not to the player.

Fixed: players on instanceable maps are now skipped entirely, so only open-world zones can ever be concentration targets.

```cpp
Map* map = player->GetMap();

// Instanceable maps (dungeons, raids, battlegrounds, arenas) are never valid targets: a
// WorldLocation carries no instance id, so a bot would be sent to another instance of the
// same map rather than to the player.
if (map->Instanceable())
    continue;
```

Documented in the conf comment as well.

src/Bot/RandomPlayerbotMgr.cpp
NSO73: You're right, the two sides of the comparison weren't using the same reference. The candidate locations were resolved with `sMapMgr->GetZoneId(PHASEMASK_NORMAL, loc)` (unphased terrain) while the player side used `Player::GetZoneId()`, which reflects the player's actual phase. A player standing in a phased area therefore never matched anything and the feature silently fell back to the world-wide behaviour.

Not dangerous, but silently degrading, which is worse than either working or not. Fixed by resolving the player's zone the same way as the locations, so both sides compare unphased terrain and a phased player still matches its underlying zone:

```cpp
uint32 zoneId = map->GetZoneId(PHASEMASK_NORMAL, player->GetPositionX(), player->GetPositionY(),
                               player->GetPositionZ());
```

src/Bot/RandomPlayerbotMgr.cpp
NoxMax: > I'd rather say so than hand you a number from my single-player box that wouldn't answer your question.

And we would rather that instead of just copy/pasting what the LLM says, that you acquaint yourself with the mechanics of the performance monitor and make some before/after test plan (making sure smart scaling is disabled in the test). My friend a main objective of the entire project is for people to actually learn the ins and outs of coding with all of this, not just to have LLMs unleashed on the codebase.

NSO73: Fair enough, so let me be clear about it.

Yes, I use an LLM, English isn't my first language. I also used one to get here in the first place: I asked an AI agent how to run a private server on my own machine and it pointed me to AzerothCore and to this module. My hardware was struggling with the default settings, which is where the idea came from. I hadn't studied the module's code before that, I own that.

The practical driver: I re-apply this patch by hand on every module update. Upstreaming it removes that, and it may help others running small setups. That's why the option is opt-in and default-off, and why the PR is written around the small-setup case, it was stated deliberately, not added to make the change look harmless.

I know projects get flooded with thoughtless LLM-generated submissions. I didn't want to be one of them, and I never hid the AI assistance, it's ticked in the template. But your point stands: a feature isn't three lines from an LLM, it's something you maintain long after I'm gone. Apologies if I came across as careless.

## Numbers, measured the way you asked

Perf monitor, before/after, `botActiveAloneSmartScale = 0`.

One limit I'd rather state up front than have you find it: I'm on a single-player box (i5-8350U, 16 GB), so this is the per-call cost with one real player online. The loop is `O(players online)`, so these numbers do **not** validate the 500-player case, and I won't pretend otherwise. What they bound is the unit cost of the added work on modest hardware, which is a ceiling: on real server hardware it will be lower.

`RandomTeleportForLevel` had no `PerfMonitorOperation`, and the filter runs before the call to `RandomTeleport`, so nothing on that path was being measured at all. I added two counters locally to get these numbers (not part of this PR).

Test setup: ~49 random bots, level-22 character parked in Ashenvale (zone 331), `SyncLevelWithPlayers = 1`, teleport interval temporarily lowered to `5..15s` to accumulate samples, two 10-minute windows, `.playerbots pmon reset` / `pmon tick`.

**Enclosing function, before/after** (`RandomTeleportForLevel`, which now wraps the filter):

| | avg / call | max / call | % of bot tick |
|---|---|---|---|
| flag OFF (baseline) | 4.737 ms | 9.0 ms | 0.060 % |
| flag ON (feature)   | 4.203 ms | 8.4 ms | 0.054 % |

Flag-on is actually *lower*: the added filter is buried under the run-to-run variance of the teleport itself. No measurable overhead at this level.

**The added helper itself** (`GetPlayerZoneTeleportLocations`): it never appears in the pmon output, in either pass. Not because it doesn't run (the concentration is visibly working in-game), but because it falls below pmon's own display threshold:

```cpp
// PerfMonitor::PrintStats
if (perc >= 0.1f || avg >= 0.25f || pd->maxTime > 1000) // µs
```

So, straight from the monitor's own visibility line, the helper stays **under 0.1 % of bot tick time, under 0.25 ms average per call, and under 1 ms worst-case per call** on this hardware, while actively concentrating bots. It also runs at most once per bot per `MinRandomBotTeleportInterval` (300..1800 s by default), never on a per-tick path.

That's the measurement. It confirms what the PR claimed (negligible, opt-in, zero cost when off), but now it's measured on modest hardware rather than asserted, and default-off means the helper is never even reached unless an admin opts in.

To be clear about where I stand: I proposed this only to help the community and to scratch my own small-hardware itch, nothing more. If you decide it doesn't belong in the module, I'll close it, no hard feelings, and thanks for the review either way.

NoxMax: Thank you for understanding. Though I should give more details about the usual parameters we use to test. Typically we do a minimum of 500 bots if not few thousand to test things. You're not really going to get useful data below a 500 sample size. To standardize the measurement, set the configs `BotActiveAlone = 100` & `botActiveAloneSmartScale = 0`. After server boot, toggle the monitor right after the bots finished logging-in, and run the stack command `playerbot pmon stack` 5 minutes after that.
The before and after of "PlayerbotAI::UpdateAIInternal 571" is the main measurement to look for.

> I re-apply this patch by hand on every module update.

Elaborate on your workflow. Does your server use the master branch of mod-playerbots directly, and you edit the files one by one after each update to reapply your concentrate edit?

NSO73: Measured, following your protocol: 500 bots, `BotActiveAlone = 100`, `botActiveAloneSmartScale = 0`, pmon toggled once all bots had logged in, `pmon stack` 5 minutes later. i5-8350U (4c/8t), 16 GB.

Two things I had to figure out first, worth stating since they shape the numbers:

* `UpdateAIInternal <mapId>` is per-map, so with the default level range most bots are outside Northrend and `571` is a marginal counter. I set `RandomBotMinLevel = 75` and `SyncLevelWithPlayers = 0` so that map 571 carries 80% of bot time and the measurement is actually about the bots being concentrated.
* My first flag-off run was on a server in a different state than the flag-on one, and that alone accounted for a third of the difference. I discarded that comparison and ran a second flag-off window under identical conditions.

`PlayerbotAI::UpdateAIInternal 571`, the two runs made under identical conditions:

| | flag off | flag on |
|---|---|---|
| avg per call | 1.999 ms | 2.488 ms |
| total | 124.901 s | 137.604 s |
| calls | 62469 | 55306 |

**+24.5% per call.** The spread between my two flag-off runs was 8%, so that is the noise floor and the effect is real.

The filter's own cost is not where it happens: `GetPlayerZoneTeleportLocations` is **125 calls, 0.022 s, 0.176 ms per call, 0.013% of bot time**. It never appears in `pmon stack` because `PrintStats` only prints counters above `perc >= 0.1f || avg >= 0.25f || maxTime > 1000`; I recovered it from the residue of the `RndBot : Total` line, which is exactly 0 calls with the flag off and 125 with it on.

@kadeshar this also answers your caching question: 125 calls over 5 minutes with 500 bots. A per-tick cache would add invalidation state (zone change, logout, GM toggle) to save 0.022 s. I'd rather leave it uncached.

So the +24.5% is not the filter, it's the density it creates: bots packed into one zone see each other, and that shows up in every tick. It's inherent to the feature and it scales with how many bots you concentrate. At 500 bots forced fully active it's +24.5%; the flag is opt-in and defaults to off.

One more measurement, since the point of the feature is what the player sees. Same 500 bots, same measured server latency (mean diff 344 ms flag off, 307 ms flag on), destination zones of random teleports over the same windows:

| destination zone | flag off | flag on |
|---|---|---|
| **Icecrown, where I stand** | **14** (8%) | **89** (62%) |
| The Storm Peaks | 37 | 0 |
| Zul'Drak | 20 | 0 |
| Wintergrasp | 12 | 0 |
| Sholazar Basin | 11 | 1 |
| Crystalsong Forest | 4 | 0 |
| Dalaran (`ProbTeleToBankers` path) | 28 | 24 |
| other zones (low-level areas, capitals) | 43 | 29 |
| **total teleports in the window** | **169** | **143** |

14 bots reaching the player's zone versus 89, at the same latency. The interesting part is where they come from: the other Northrend zones drop to zero, so the filter is not adding bots, it is redirecting the ones that would otherwise have scattered across the continent. Dalaran is untouched (28 vs 24) because the banker path returns before the filter is reached, and the remaining zones are teleports that do not go through this code path at all.

What I have not measured: whether a smaller concentrated pool beats a larger scattered one on total CPU. Happy to run it if it would help the decision.

On your question about my workflow: my earlier sentence was sloppy, sorry. My server does not track master directly and I do not edit files one by one. The module is a normal clone, my changes are commits on a dedicated branch, and updating is a `git rebase origin/master`, so the patch replays itself. Upstreaming would remove a step from my routine, but it is a cheap one, so that is not an argument for the PR and I should not have presented it as one.

To close: I may well be wrong here. As I said earlier, I'm not a developer and I don't have the full picture of the project's constraints and direction. Your first reaction was that the justification wasn't convincing, and I've tried to answer each point on its merits rather than argue around them, but it is entirely possible that I'm pushing something that isn't worth the surface it adds. If you conclude it's out of scope, close it without hesitation. Either way the numbers above are yours to keep, and thank you both for the time you've put into this.

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -1780,13 +1782,70 @@ void RandomPlayerbotMgr::RandomTeleportForLevel(Player* bot)
...
+        if (playerMapZones.find(std::make_pair(loc.GetMapId(), zoneId)) != playerMapZones.end())
+            filtered.push_back(loc);
+    }
+
+    return filtered;
+}
```
NoxMax, line 1859: I would also add here a check for hostile territories, similar to https://github.com/mod-playerbots/mod-playerbots/blob/3fa1c1e49f8f1324b72461e576bce7c89b0a6521/src/Bot/RandomPlayerbotMgr.cpp#L1662
Shouldn't teleport Horde bots to Elwynn for example.

conf/playerbots.conf.dist
```diff
@@ -708,6 +708,16 @@ AiPlayerbot.RandomBotMaxLevel = 80
...
 AiPlayerbot.SyncLevelWithPlayers = 0
 
+# Concentrate random bots in the zone(s) where real players currently are, instead of spreading
+# them across the whole world. When enabled, a random bot about to random-teleport to a
+# level-appropriate location will prefer a location whose zone currently holds a real (non-GM)
+# player. Players inside instanceable maps (dungeons, raids, battlegrounds, arenas) are ignored:
+# bots are only ever concentrated in open-world zones. It falls back to the normal world-wide
+# behaviour when no real player is in a matching zone. Best combined with
+# AiPlayerbot.SyncLevelWithPlayers so bots share the players' zones.
+# Default: 0 (disabled)
```
NoxMax: Comment is a bit long and verbose. I suggest simply this

```
# Concentrate randombots teleporting into level-appropriate open-world zones that hold a real
# (non-GM) player. Falls back to normal teleporting behaviour when no player's zone matches.
# Best paired with AiPlayerbot.SyncLevelWithPlayers.
# Default: 0 (disabled)
```

NoxMax (changes requested):
Thank you for your extensive testing. Numbers look good. I just have a couple more notes.

src/Bot/RandomPlayerbotMgr.cpp
NSO73: Done in 740f0050. I put the check inside `GetPlayerZoneTeleportLocations` rather than relying only on the one in `RandomTeleport`: the caller falls back to normal teleporting only when the filtered set is empty, so a hostile-only match would otherwise let the downstream team check drain the set and skip the teleport entirely.

conf/playerbots.conf.dist
NSO73: Done, used your wording verbatim.

kadeshar (dismissed):

kadeshar:
Resolve conflict

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -1777,13 +1779,85 @@ void RandomPlayerbotMgr::RandomTeleportForLevel(Player* bot)
...
+    {
+        if (!player || !player->IsInWorld() || player->IsGameMaster())
+            continue;
+
+        Map* map = player->GetMap();
+
+        // Instanceable maps (dungeons, raids, battlegrounds, arenas) are never valid targets: a
+        // WorldLocation carries no instance id, so a bot would be sent to another instance of the
+        // same map rather than to the player.
+        if (map->Instanceable())
```
kadeshar, line 1821: Best add condition checking that map is not nullptr

Celandriel (approved):

PR #2544, opened 2026-07-05
fix(Group): make RandomBotGroupNearby actually work

suikoden1434-sys:
## What

`AiPlayerbot.RandomBotGroupNearby` has been marked *"currently not functioning properly"* in the config for a while. This PR fixes the four issues that broke it, so random bots can once again form persistent open-world groups with each other.

## The bugs

**1. Grouper/guilder identity re-rolled every 30 seconds**
`GetFixedBotNumber()` mixes a time slot into its hash (added for the BotActiveAlone rotation in #2250 / #2288 — correct for that use). But `GetGrouperType()` / `GetGuilderType()` use the same function for what should be a bot's *stable* social personality. Result: every `BotActiveAloneDurationSeconds` (default 30s) every bot re-rolled its grouper type; grouped bots re-rolling into `SOLO` triggered leave-group logic and every group dissolved within seconds.

Fix: `GetFixedBotNumber(maxNum, BotTypeNumber)` is now stable per bot and salted per trait (grouper and guilder are no longer the same number, restoring the intent of the `BotTypeNumber` enum which was left orphaned in the header). New `GetRotatingBotNumber()` keeps the time-slot reshuffle and is used only by `AllowActive()`.

**2. Null-pointer crash on real players**
Both invite loops called `GET_PLAYERBOT_AI(player)->IsRealPlayer()` unguarded. `GET_PLAYERBOT_AI` returns `nullptr` for real players, so with `RandomBotInvitePlayer = 0` (default) a bot evaluating a nearby ungrouped human crashed the server. Now null-safe in both `InviteNearbyToGroupAction` and `InviteGuildToGroupAction`.

**3. Wrong player checked in the nearby-invite filter**
`InviteNearbyToGroupAction::Execute` fetched `GET_PLAYERBOT_AI(bot)` (the inviter, already validated in `isUseful()`) where the sibling guild-invite code checks the *candidate*. SOLO-type bots were therefore invited, accepted, and immediately left. Now checks the candidate's AI.

**4. The strategy never ran**
The `group` strategy was commented out in `AiFactory`, so the invite trigger never fired at all. It's now added for random bots when `RandomBotGroupNearby` is enabled — same gating pattern as the `lfg` strategy — so default behavior is unchanged for everyone who hasn't opted in.

**Bonus:** `BroadcastGuildGroupOrRaidInvite` passed literal English sentences as bot-text keys (the `//TODO move texts to sql!` block), spamming `Can't get bot text ...` in the log and silencing the invite chatter. Switched to `GetBotTextOrDefault` with proper keys.

## Testing

WotLK 3.3.5a, AzerothCore Playerbot branch (c62dc737), 1000 random bots, `RandomBotGroupNearby = 1`, MSVC 2022 / Win64 build:

- Baseline before fix: **0 rows ever** in `acore_characters.groups` (with the strategy force-enabled, groups formed but dissolved within one rotation window; plus reproducible crash from bug 2 when a real player stood near an inviting bot with default `RandomBotInvitePlayer`).
- After fix: **24 groups formed within ~10 minutes; 24/24 still intact 8+ minutes later** (16 rotation windows); population grew to 33. No crashes with a real player online. Guild invite broadcasts now appear in guild chat instead of log errors.

Default-config users see no behavior change: the strategy is only added when `RandomBotGroupNearby` is on, and the hash split preserves the existing activity-rotation behavior exactly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -970,7 +970,8 @@ AiPlayerbot.botActiveAloneSmartScaleWhenMaxLevel = 80
 AiPlayerbot.RandomBotQuestIds = "3802,5505,6502,7761,7848,10277,10285,11492,13188,13189,24499,24511,24710,24712"
 
 # Randombots will group with nearby randombots to do shared quests
-# Note: Currently not functioning properly
+# Group composition follows each bot's grouper type:
+# ~20% of bots stay solo, ~60% only join, ~20% lead groups of 2-5.
 AiPlayerbot.RandomBotGroupNearby = 0
 
 # Randombots will pick quests on their own and try to complete them
--- a/data/sql/playerbots/updates/2026_07_05_00_ai_playerbot_broadcast_invite_texts.sql
+++ b/data/sql/playerbots/updates/2026_07_05_00_ai_playerbot_broadcast_invite_texts.sql
@@ -1,0 +1,17 @@
+-- Texts for group/raid invite broadcasts (BroadcastGuildGroupOrRaidInvite)
+-- These keys replace the literal sentences previously passed to GetBotText().
+DELETE FROM ai_playerbot_texts WHERE name IN ('broadcast_group_invite_any', 'broadcast_group_invite_name', 'broadcast_raid_invite_any', 'broadcast_raid_invite_name');
+
+INSERT INTO `ai_playerbot_texts`
+    (`name`, `text`, `say_type`, `reply_type`,
+     `text_loc1`, `text_loc2`, `text_loc3`, `text_loc4`,
+     `text_loc5`, `text_loc6`, `text_loc7`, `text_loc8`)
+VALUES
+    ('broadcast_group_invite_any', 'Hey anyone wanna group up in %zone_name?', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_group_invite_any', 'LFM around %zone_name, whisper me', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_group_invite_any', 'Anyone questing in %zone_name? Looking for a group', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_group_invite_name', 'Hey %name, do you want to join my group? I''m heading for %zone_name!', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_group_invite_name', '%name, wanna group up? Heading to %zone_name', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_raid_invite_any', 'Hey anyone want to raid in %zone_name?', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_raid_invite_any', 'Putting a raid together for %zone_name, who''s in?', 0, 0, '', '', '', '', '', '', '', ''),
+    ('broadcast_raid_invite_name', 'Hey %name, I''m raiding in %zone_name, want to join?', 0, 0, '', '', '', '', '', '', '', '');
--- a/src/Ai/Base/Actions/InviteToGroupAction.cpp
+++ b/src/Ai/Base/Actions/InviteToGroupAction.cpp
@@ -61,7 +61,10 @@ bool InviteNearbyToGroupAction::Execute(Event /*event*/)
         if (player->GetGroup())
             continue;
 
-        if (!PlayerbotAIConfig::instance().randomBotInvitePlayer && GET_PLAYERBOT_AI(player)->IsRealPlayer())
+        PlayerbotAI* playerAI = GET_PLAYERBOT_AI(player);
+
+        // Real players have no playerbot AI - never dereference it blindly.
+        if (!PlayerbotAIConfig::instance().randomBotInvitePlayer && (!playerAI || playerAI->IsRealPlayer()))
             continue;
 
         Group* group = bot->GetGroup();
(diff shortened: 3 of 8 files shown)


NoxMax:
Unfortunately No. Group strategy didn't just stop working on its own one day. It has been intentionally disabled as it was having issues with other systems, and last I recall there are now sections of the code that depend on there not being such groups. Implementing grouping back is not easily done without checking its effect on many other things.

suikoden1434-sys:
Thanks for the context — that's really helpful, and it matches what I found when I dug deeper. You're right that it was deliberately disabled: the lifecycle manager still carries an explicit cleanup (`ProcessBot` in `RandomPlayerbotMgr.cpp`, the *"leave group if leader is rndbot"* block) that disbands any non-LFG group led by a random bot.

For what it's worth, here's what I mapped of the "systems that depend on no such groups," in case it's useful to anyone revisiting this later:

1. **Lifecycle rotation** — grouped bots early-out of `ProcessBot` (`GetGroup()` check) before the update/randomize/teleport/logout events, so bots in persistent groups stop being re-randomized, teleported, or rotated out. That early-out is also why the explicit disband cleanup never actually fires for them — the two checks currently shield each other.
2. **BG/LFG strategies** — invite acceptance swaps members to `+follow,-lfg,-bg`, but the *leader* keeps its `bg` strategy, so a leader can queue into a battleground and leave members idle until it returns.
3. The disband cleanup itself, which this PR effectively bypasses (unintentionally — happy to make that explicit instead, e.g. gating the cleanup on `!randomBotGroupNearby`).

Observational data so far (single server, 1000 random bots, `RandomBotGroupNearby = 1`, ~24h uptime): 103 groups / 269 members formed organically, size distribution 60x2 / 28x3 / 10x4 / 5x5 matching the grouper-type weights, zero groups with offline leaders, zero orphaned members, no related errors in the log. The stability fix (stable grouper identity) is what changed the picture — previously groups dissolved within one `BotActiveAloneDurationSeconds` window, which I suspect is part of why the feature was more trouble than it was worth when it was last evaluated.

I understand if re-enabling isn't wanted — in that case, would you be open to splitting this PR? Items 1-3 in the description (stable grouper/guilder identity, the `GET_PLAYERBOT_AI` null-deref crash, and the wrong-player check) are standalone correctness fixes that apply even with the strategy disabled — the grouper types feed `LeaveGroupAction`, guild creation, and the invite paths regardless. I can drop the `AiFactory`/conf changes from this PR and keep the feature opt-in-at-your-own-risk, or file the bugfixes separately — whichever you'd prefer.

Either way, thanks for the quick review — and if there are specific systems you recall breaking (arena teams? guild tasks? the RPG travel state?), I'm happy to test against them on my server and report back.

NoxMax:
Well my first question is this: Do you understand the system you are modifying? Not Claude, you. You can copy my comment to Claude and ask it to explain how grouping works and then copy back its response, yeah I guess I can chat with Claude through you 😂 However, if you have read the PR template (notwithstanding the other issues in this PR) we require contributors to understand the thing they are submitting, to know how it works.

As would the template tell you, you can use an LLM, but you need to understand what the LLM is doing. My recommendation is that you close this PR for now, learn how the things you want to modify work, then come with whatever you want to change. You can use Claude as a learning tool, that's perfectly valid, but you do need to actually learn about how your changes work and why they are needed or not.

suikoden1434-sys:
I am doing both, in a way.  My knowledge of programming is....dated, let's say.  I am using Claude, and alot of book reading, to learn new coding languages and techniques.  As for this particular mod and PR, I noticed the original bot group feature off and its note, then decided to look at it independently.  I have things set up where Claude and I bounce thoughts back and forth with regards to methodology in code changes, and Claude also explains any coding that I have trouble following it's suggestions.

Im getting old and I hate seeing great games from before go silent or get changed into oblivion, so I saw this mod and it's features as a great potential tool to safeguard those games as well as a way to keep learning.  I apologize if I am bothering you.

suikoden1434-sys:
I will try again once I learn some more and how better to express in my own words.  Thank you for your consideration.

NoxMax:
> I apologize if I am bothering you.

No no, not at all. It's just because in the age of LLMs, we have to check that contributors are aware of their changes. Last year we accepted quite a few PRs that were LLM written and our lack of scrutiny made us miss a whole bunch of issues in them and forced us to revert many of the changes, and reverts can be a mess to untangle. So you can understand why these days we are asking contributors to know about the thing they are modifying, and it's also why we implemented test-staging to test out PRs before pushing them to master. Plus the whole ethos of AzerothCore is that this project should be a learning experience for people to understand how these gaming systems work.

So we look forward to your contributions in the future, and if you have questions about the mechanics of anything, [AC has their own discord](https://discord.gg/4TNdBqmcy), and we have [our own discrod](https://discord.gg/PqBswcJeN) for all things Playerbots related.

Celandriel:
General thoughts on this. 
Its possible that it was disabled because of  issues related to crashing when using multiple threads that has since been resolved. This is the biggest risk I see.

I have a branch where I re-enabled it,but honestly I havent played enough on that branch to know whether there are issues or not. I know others have just added group strat back in and it works. 

I'm supportive of this effort, but like Nox says PR authors need to be able to understand what they are submitting.

suikoden1434-sys:
Understood, and thank you all for your insights.  I will keep learning!  Always keep the mind sharp, folks.

seedpodsw:
good start to making RandomBotGroupNearby actually work, just needs more @suikoden1434-sys 

try this 

Add central IsBotLedNearbyGroup / IsInBotLedNearbyGroup helper used by all gates
Split GetFixedBotNumber (stable, salted by BotTypeNumber) and add GetRotatingBotNumber for AllowActive
Fix null-deref, wrong-player check, and pending-invite races in InviteToGroupAction.cpp
Queue group invite/accept opcodes through PlayerbotWorldThreadProcessor instead of calling them on the AI thread
Make AcceptInvitationAction decline invites when bot is already grouped
Add 'group' strategy in AiFactory gated on randomBotGroupNearby; keep -lfg,-bg persistent for grouped members
Lifecycle integration in RandomPlayerbotMgr/PlayerbotMgr: disband gates, update-bypass fix, logout rotation, Refresh/Randomize/Revive/teleport guards, OnBotLogin validation
Mirror leader activity state onto nearby-group members so followers don't freeze mid-travel
Let FindNewMaster return a bot group leader so follow/master chains survive leader changes and BG exits
Priority model: BG/LFG/arena events make bots leave the nearby group first (dump), then queue solo; groups re-form after (retry)
Fix BroadcastGuildGroupOrRaidInvite text keys and update conf.dist note

seedpodsw:
i have continued over here with great progress on many fronts for grouping https://github.com/seedpodsw/mod-playerbots/tree/feature/bot-grouping

krashtestrich:
@seedpodsw this is a feature that I'd love to see working - I've checked out your fork and it looks pretty good - are you going to continue working on it ?

It's fallen behind the master player-bots branch quite a bit now, I might play around with merging those commits into a local branch of your fork

Issue #2523, opened 2026-06-29
Add locale tables for Playerbots DB data such as bot texts, names, guild names, arena team names, dungeon suggestions, and speech


hilch1981-prog:
## Feature Request

Please add official locale/localization table support for Playerbots database data.

Currently, several Playerbots DB tables contain text data that server administrators need to translate manually, especially for non-English servers.

Examples of localized/customized data:

```text
ai_playerbot_texts.sql
playerbots_names.sql
playerbots_guild_names.sql
playerbots_arena_team_names.sql
playerbots_dungeon_suggestion_definition.sql
playerbots_speech.sql
```

For Korean servers, these tables often need to be fully translated or replaced with Korean data.

## Problem

At the moment, localization is difficult to maintain because many Playerbots text/name tables either store English text directly or require administrators to overwrite the original data.

This causes several problems:

1. Localized data can be overwritten by future Playerbots DB updates.
2. Server administrators must reapply custom translation SQL files after updates.
3. It is difficult to maintain multiple languages cleanly.
4. Some text systems already support locale-style fields, but other Playerbots tables do not have a consistent locale structure.
5. Custom Korean, French, German, Chinese, Russian, Spanish, and other localization packs cannot be managed cleanly without modifying base data.

## Requested Solution

Please add official locale tables for Playerbots DB text/name data.

Suggested examples:

```sql
ai_playerbot_texts_locale
playerbots_names_locale
playerbots_guild_names_locale
playerbots_arena_team_names_locale
playerbots_dungeon_suggestion_definition_locale
playerbots_speech_locale
```

Example structure:

```sql
CREATE TABLE playerbots_names_locale (
    name_id INT NOT NULL,
    locale VARCHAR(4) NOT NULL,
    name VARCHAR(255) NOT NULL,
    PRIMARY KEY (name_id, locale)
);
```

Example:

```sql
CREATE TABLE playerbots_guild_names_locale (
    id INT NOT NULL,
    locale VARCHAR(4) NOT NULL,
    name VARCHAR(255) NOT NULL,
    PRIMARY KEY (id, locale)
);
```

For text-based tables:

```sql
CREATE TABLE playerbots_speech_locale (
    id INT NOT NULL,
    locale VARCHAR(4) NOT NULL,
    text VARCHAR(1024) NOT NULL,
    PRIMARY KEY (id, locale)
);
```

The module could then load localized values based on the server locale setting, with English/default data as fallback.

## Expected Behavior

[...]

PR #2545, opened 2026-07-05
refactor(chat): unify chat noun locale with GetBotText via GetLocalePriority()

chwt9299:
## 🍰 Pullrequest
Currently, ChatHelper formats quest/item/spell/gameobject names using
`sWorld->GetDefaultDbcLocale()`, while bot dialogue text (GetBotText)
uses `PlayerbotTextMgr::GetLocalePriority()` which detects the majority
locale among online real players. This causes a mismatch on mixed-locale
servers — for example, bot dialogue outputs Chinese but item/spell names
remain English.

This PR consolidates all 10 locale call sites in ChatHelper and
BroadcastHelper to use a single `PlayerbotAIConfig::GetLocalePriority()`
method. Internally it delegates to `PlayerbotTextMgr::GetLocalePriority()`,
making chat nouns and bot dialogue always speak the same language.

### Proof
- Compiled with RelWithDebInfo on Windows 11, no errors.

### Issues
- None

### How2Test
1. Build and start worldserver on a zhCN server with enUS DBC.
2. Have some zhCN clients connected.
3. Summon bots and observe chat output — quest/item/spell/gameobject
   names should now match the bot dialogue language instead of falling
   back to English.

### Todo / Checklist
- [X] None


hilch1981-prog:
Hello,

I am also working on Korean localization for Playerbots and I think this PR is related to the same problem I am currently facing.

I recently opened a Korean localization PR here:

https://github.com/mod-playerbots/mod-playerbots/pull/2547


And the related issue is here:

https://github.com/mod-playerbots/mod-playerbots/issues/2523

My first PR only adds Korean localization for `ai_playerbot_texts`, because that table already has locale columns such as `text_loc1`.

However, the remaining Playerbots tables do not currently seem to have proper locale columns or locale tables:
- `playerbots_names`
- `playerbots_guild_names`
- `playerbots_arena_team_names`
- `playerbots_dungeon_suggestion_definition`
- `playerbots_speech`

So I was not sure whether Korean data for those tables should be added as normal SQL updates, optional SQL files, or handled through a new locale/fallback system.

This PR seems to address a similar issue by adding a configurable text locale layer through `AiPlayerbot.TextLocale`.

If this approach is accepted, I would like to follow the same direction for Korean localization instead of replacing the default English data.

For Korean, the expected locale would be `koKR`, or the AzerothCore Korean locale index if the project prefers numeric locale values.

Please let me know if this PR is the correct direction for future localization work. If so, I can adjust the remaining Korean localization data to follow this structure.

conf/playerbots.conf.dist
```diff
@@ -2221,23 +2258,31 @@ AIPlayerbot.GuildFeedback = 1
 AiPlayerbot.GuildRepliesRate = 100
 # Bots without a master will say their lines
 AiPlayerbot.RandomBotSayWithoutMaster = 0
+#
+# Locale used for bot chat text (quest names, item names, spell names, etc.)
+# Uses LocaleConstant values: 0=enUS, 4=zhCN, 5=zhTW, etc.
+# Default: 0 (enUS) - set to your DB locale (e.g. 4 for zhCN) for localized chat nouns
+AiPlayerbot.TextLocale = 0
```
NoxMax, line 2261: Please use the PR template and target test-staging, not master. That being said, I am not the most versed with the localization system, but I don't think decoupling from DBC.Locale with another global language config on Playerbots is the way to go. The better approach is for the bots to know what is the player's client language (which AC knows and it should already be publicly accessible) and have the bot reply to the player accordingly. We technically already do that at login to figure out what most real players have set their client's language set to, but there is no 1-on-1 check, which is a valid thing to have.

As to the missing localized spell names, that is better solved by loading the translations into the normal place the game looks for them, rather than a lookup in the DB, which is not efficient.

chwt9299: Thanks for the feedback, NoxMax. I've retargeted to test-staging and updated the PR description accordingly. I also rethought the approach based on your suggestions. Pushed the changes, would appreciate another look when you get a chance.

src/PlayerbotAIConfig.cpp
```diff
@@ -1046,3 +994,10 @@ std::vector<std::vector<uint32>> PlayerbotAIConfig::ParseTempPetTalentsOrder(uin
 
     return orders;
 }
+
+#include "PlayerbotTextMgr.h"
+
```
NoxMax, line 999: If we include something, it's always at the top of the file.

src/Bot/Cmd/ChatHelper.cpp
```diff
@@ -420,8 +419,25 @@ std::string const ChatHelper::FormatWorldEntry(int32 entry)
...
+    uint8 dbLocale = sPlayerbotAIConfig.GetLocalePriority();
+    std::string spellName = spellInfo->SpellName[dbLocale] ?
+        spellInfo->SpellName[dbLocale] : spellInfo->SpellName[LOCALE_enUS];
+
+    // If DB locale is not enUS and DBC has no localized data, try spell_locale table
+    if (dbLocale != LOCALE_enUS && !spellInfo->SpellName[dbLocale])
+    {
+        QueryResult result = WorldDatabase.Query(
+            "SELECT NameLoc1 FROM spell_locale WHERE ID = %u AND Locale = '%s'",
+            spellInfo->Id, localeNames[dbLocale]);
```
NoxMax, line 431: Going with GetLocalePriority is a good direction, however, the main issue is the runtime targeting of DB. Typically if we call onto the DB during runtime, its for something that happens infrequently.For something that frequently happens, the data from the DB is loaded onto some set or whatever else at server start.

And the more glaring point, we don't have a `spell_locale` table. If you want to add code that tries to utilize it in another fork, then your PR needs to be at that fork. If you think that it's best to have it upstream here, then that table needs to be added here too. I'm not the one most versed about modifying the DB though, so hopefully another contributor can take a look and weigh-in.

PR #2550, opened 2026-07-07
feat: world pvp 

Soulnvkz:


seedpodsw:
see https://github.com/seedpodsw/mod-playerbots/tree/feature/bot-grouping for world pvp additions including world bot groups levels, questing, etc -- good start here, need to compile what you did here @Soulnvkz for some adds to the bot-grouping system (pvp in mind ofc)

Soulnvkz:
@seedpodsw Thanks, I'll take a look at your work. 
Actually, I accidentally created a PR here and closed it right away. 

In my fork, there's a cleaner and simpler working version that I was satisfied with for playing several days. Of course, it's nowhere near perfect.
The main issue was that even with a large number of bots on the server, they're still quite rare near actual players. So I added a recruiter system for that (excluding capitals, low-level zones, and safe zones). But the behavior could be much better. Your PvP groups seems like a pretty good idea.

UPD: plus, yes, I'm using a hack with caching PvP bots ( to be sure they are always active and helping each other), but implementing a PvP strategy with the existing RPG system should be a much cleaner and better solution to play with.

PR #2565, opened 2026-07-10
Cross class fixes

shenkai4257:
Pull Request Description
Three independent cross-class fixes that address buff conflicts and missing combat behaviors:

Warrior — Battle Shout vs Blessing of Might conflict Battle Shout and Blessing of Might both provide attack power and do not stack. When a warrior and a paladin are in the same party, they endlessly overwrite each other (ping-pong). This adds an early-out in BattleShoutTrigger::IsActive() — if the bot already has Blessing of Might (or Greater), skip Battle Shout entirely.

Priest (Holy) — Shadow Word: Pain in default rotation Holy Priests with high spell power benefit significantly from SW:P as filler DPS, but the default action list only had Smite/Mana Burn/Starshards. SW:P is added with highest default priority. Also lowers AlmostFullHealth from 85→70 so the "healer should attack" trigger fires more often in dungeons.

Shaman — Totemic Recall with distance-based trigger The existing trigger had restrictive conditions (no boss fight, all group members out of combat) that prevented totem recall even when the bot had moved far away — leaving totems behind to pull patrols. Replaced with a simple distance check: any totem >40 yards away triggers recall. Also fixed CastTotemicRecallAction::GetTarget() to return self (the spell is self-only; the base class returned nullptr outside combat, causing PREREQ failures).

Feature Evaluation
Minimum logic: Three single-condition guards added to existing IsActive() / isUseful() / GetTarget() overrides. No new loops or complex state machines.
Processing cost: Negligible — one or two HasAura() calls or GetDistance() per trigger evaluation. Existing behavior unchanged when conditions are not met.
How to Test the Changes
Warrior:

Add a warrior bot and a paladin bot to party.
Let the paladin cast Blessing of Might on the warrior.
Verify the warrior does not cast Battle Shout (check Playerbots.log — no battle shout action pushed).
Priest (Holy):

Add a Holy Priest bot. Engage in combat.
Verify SW:P is cast on the target when party HP is ≥70% (was 85%).
Shaman:

Add a shaman bot. Drop any totem manually.
Run away >40 yards.
Verify the shaman casts Totemic Recall (check Playerbots.log for totemic recall without USELESS or PREREQ).
Impact Assessment
Does this change increase per-bot/per-tick processing or risk scaling poorly with thousands of bots?
 No, not at all
 Minimal impact (explain below)
 Moderate impact (explain below)
 No
 Yes (explain why)
Warrior: skips Battle Shout when Might is present (correct stacking behavior).
Priest: casts SW:P more often; healer DPS threshold lowered.
Shaman: recalls totems based on distance instead of combat state.
Does this change add new decision branches or increase maintenance complexity?
 No
 Yes (explain below)
AI Assistance

 No
 Yes (explain below)
AI was used for brainstorming the approach, debugging via log analysis (identifying USELESS / PREREQ status), and code generation. All generated code was reviewed, tested, and verified against in-game behavior.

Code Provenance / Attribution

 No, all code in this PR is original
 Yes (name the project and the original author(s) below)
Final Checklist
 Stability is not compromised.
 Performance impact is understood, tested, and acceptable.
 Added logic complexity is justified and explained.
 Any new bot dialogue lines are translated. (N/A — no new dialogue)
 Any code ported/adapted from another project is attributed. (N/A)
 New source files use the GPLv2 header. (N/A — no new files)
 Documentation updated if needed. (N/A)
 New and modified files do not introduce new compiler warnings.
Notes for Reviewers
All three fixes are intentionally small and isolated — each touches a single trigger or action in a single class file. No shared infrastructure is modified. The AlmostFullHealth config default change from 85→70 is the only behavioral change that affects non-Shaman/Priest/Warrior bots.


noisiver:
Why are there so many different changes to files not affected by the main reason for this PR? Like the license changes, shouldn't they be in a separate PR just like the changes to the translation workflow? The same for files moved to other folders, like MagHelpers. This PR covers so many different things and it might end up being a lot to review that could be split into different PRs to make it a lot simpler.

I'm also curious if battle shout shouldn't take priority and allow the paladin(s) to cast different blessings instead of might? Since warrior only has 2 buffs, battle and commanding shout, while paladins have several different ones I mean. At level 80 battle shout does give the same amount of attack power as blessing of might so it's not really a downside to allowing it to take priority, is there?

brighton-chi:
Number one is simply incorrect if you look at the existing code.

I also disagree with lowering almostfullhealth threshold. That hugely impacts gameplay beyond healer DPS. It's the initial healing threshold so you're telling healers not to start healing until 70% instead of 85%. It's configurable so I'm not sure why the default threshold should be changed in such an extreme way.

PR #2403, opened 2026-05-19
Expand PWS Usage by Disc Priests

brighton-chi:
## Pull Request Description

Disc Priests currently do not use Power Word: Shield until a party member is at "medium health" (configurable, default 65%). That is the second-level healing threshold; at the highest level, "almost full health" (configurable, default 85%), Disc Priests will use only Prayer of Mending and Renew. This PR adds PWS as the highest priority healing action starting at "almost full health."

## Feature Evaluation

- Describe the **minimum logic** required to achieve the intended behavior.
- Describe the **processing cost** when this logic executes across many bots.
Added new action "power word: shield on party" to "party member almost full health" trigger in HealPriestStrategy.cpp (the "heal" strategy, which is the default strategy for Disc Priests, as Holy Priests use the "holy heal" strategy).

## How to Test the Changes

Group up with a Disc Priest and take damage; confirm that PWS is cast when you drop below your configured almost full health threshold.

## Impact Assessment

- Does this change increase per-bot/per-tick processing or risk scaling poorly with thousands of bots?
    - - [x] No, not at all
    - - [ ] Minimal impact (**explain below**)
    - - [ ] Moderate impact (**explain below**)

- Does this change modify default bot behavior?
    - - [ ] No
    - - [x] Yes (**explain why**)

PWS is the signature ability of Disc. Priests, and delaying it to 65% is significantly limiting their effectiveness in group content, particularly with raid bosses that deal raidwide damage and in heroic dungeons, where waiting until the tank hits 65% is a death wish. This single change makes an enormous difference in their effectiveness.

See below for before/after change for Eredar Twins (focus on the activity).

Before:

Screenshot 2026-05-18 201949

After:

image

- Does this change add new decision branches or increase maintenance complexity?
    - - [x] No
    - - [ ] Yes (**explain below**)

## AI Assistance

Was AI assistance used while working on this change?
- - [x] No
- - [ ] Yes (**explain below**)

## Final Checklist

- - [x] Stability is not compromised.
- - [x] Performance impact is understood, tested, and acceptable.
- - [x] Added logic complexity is justified and explained.
- - [x] Any new bot dialogue lines are translated.
- - [x] Documentation updated if needed (Conf comments, WiKi commands).

PR #2566, opened 2026-07-10
fix: Warrior Battle Shout vs Blessing of Might; Priest SW:P; Shaman T…

shenkai4257:
…otemicRecall

Warrior: skip BattleShout if target has Blessing of Might (unstoppable overwrite cycle)
Priest (Holy): add SW:P default, lower AlmostFullHealth 85->70
Shaman: TotemicRecall distance trigger + self-target fix

Pull Request Description
Three independent cross-class fixes that address buff conflicts and missing combat behaviors:

Warrior — Battle Shout vs Blessing of Might conflict Battle Shout and Blessing of Might both provide attack power and do not stack. When a warrior and a paladin are in the same party, they endlessly overwrite each other (ping-pong). This adds an early-out in BattleShoutTrigger::IsActive() — if the bot already has Blessing of Might (or Greater), skip Battle Shout entirely.

Priest (Holy) — Shadow Word: Pain in default rotation Holy Priests with high spell power benefit significantly from SW:P as filler DPS, but the default action list only had Smite/Mana Burn/Starshards. SW:P is added with highest default priority. Also lowers AlmostFullHealth from 85→70 so the "healer should attack" trigger fires more often in dungeons.

Shaman — Totemic Recall with distance-based trigger The existing trigger had restrictive conditions (no boss fight, all group members out of combat) that prevented totem recall even when the bot had moved far away — leaving totems behind to pull patrols. Replaced with a simple distance check: any totem >40 yards away triggers recall. Also fixed CastTotemicRecallAction::GetTarget() to return self (the spell is self-only; the base class returned nullptr outside combat, causing PREREQ failures).

Feature Evaluation
Minimum logic: Three single-condition guards added to existing IsActive() / isUseful() / GetTarget() overrides. No new loops or complex state machines.
Processing cost: Negligible — one or two HasAura() calls or GetDistance() per trigger evaluation. Existing behavior unchanged when conditions are not met.
How to Test the Changes
Warrior:

Add a warrior bot and a paladin bot to party.
Let the paladin cast Blessing of Might on the warrior.
Verify the warrior does not cast Battle Shout (check Playerbots.log — no battle shout action pushed).
Priest (Holy):

Add a Holy Priest bot. Engage in combat.
Verify SW:P is cast on the target when party HP is ≥70% (was 85%).
Shaman:

Add a shaman bot. Drop any totem manually.
Run away >40 yards.
Verify the shaman casts Totemic Recall (check Playerbots.log for totemic recall without USELESS or PREREQ).
Impact Assessment
Does this change increase per-bot/per-tick processing or risk scaling poorly with thousands of bots?
No, not at all
Minimal impact (explain below)
Moderate impact (explain below)
No
Yes (explain why)
Warrior: skips Battle Shout when Might is present (correct stacking behavior).
Priest: casts SW:P more often; healer DPS threshold lowered.
Shaman: recalls totems based on distance instead of combat state.
Does this change add new decision branches or increase maintenance complexity?
No
Yes (explain below)
AI Assistance

No
Yes (explain below)
AI was used for brainstorming the approach, debugging via log analysis (identifying USELESS / PREREQ status), and code generation. All generated code was reviewed, tested, and verified against in-game behavior.

Code Provenance / Attribution

No, all code in this PR is original
Yes (name the project and the original author(s) below)
Final Checklist
Stability is not compromised.
Performance impact is understood, tested, and acceptable.
Added logic complexity is justified and explained.
Any new bot dialogue lines are translated. (N/A — no new dialogue)
Any code ported/adapted from another project is attributed. (N/A)
New source files use the GPLv2 header. (N/A — no new files)
Documentation updated if needed. (N/A)
New and modified files do not introduce new compiler warnings.
Notes for Reviewers
All three fixes are intentionally small and isolated — each touches a single trigger or action in a single class file. No shared infrastructure is modified. The AlmostFullHealth config default change from 85→70 is the only behavioral change that affects non-Shaman/Priest/Warrior bots.

### DIFF
--- a/src/Ai/Class/Priest/Strategy/HolyPriestStrategy.cpp
+++ b/src/Ai/Class/Priest/Strategy/HolyPriestStrategy.cpp
@@ -33,6 +33,7 @@ HolyPriestStrategy::HolyPriestStrategy(PlayerbotAI* botAI) : HealPriestStrategy(
 std::vector<NextAction> HolyPriestStrategy::getDefaultActions()
 {
     return {
+        NextAction("shadow word: pain", ACTION_DEFAULT + 0.5f),
         NextAction("smite", ACTION_DEFAULT + 0.2f),
         NextAction("mana burn", ACTION_DEFAULT + 0.1f),
         NextAction("starshards", ACTION_DEFAULT)
--- a/src/Ai/Class/Shaman/ShamanActions.h
+++ b/src/Ai/Class/Shaman/ShamanActions.h
@@ -413,6 +413,9 @@ class CastTotemicRecallAction : public CastSpellAction
 public:
     CastTotemicRecallAction(PlayerbotAI* botAI) :
         CastSpellAction(botAI, "totemic recall") {}
+
+    // Self-only spell — no combat target required.
+    Unit* GetTarget() override { return bot; }
 };
 
 class CastStrengthOfEarthTotemAction : public CastTotemAction
--- a/src/Ai/Class/Shaman/ShamanTriggers.cpp
+++ b/src/Ai/Class/Shaman/ShamanTriggers.cpp
@@ -189,76 +189,44 @@ bool TotemicRecallTrigger::IsActive()
     if (!bot->HasSpell(SPELL_TOTEMIC_RECALL))
         return false;
 
-    Map* map = bot->GetMap();
-    if (map && map->IsDungeon())
-    {
-        InstanceScript* instance = ((InstanceMap*)map)->GetInstanceScript();
-        if (instance)
-        {
-            for (uint32 i = 0; i < instance->GetEncounterCount(); ++i)
-            {
-                if (instance->GetBossState(i) == IN_PROGRESS)
-                    return false;
-            }
-        }
-    }
-
-    Group* group = bot->GetGroup();
-    if (group)
-    {
-        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
-        {
-            Player* member = ref->GetSource();
-            if (!member)
-                continue;
-
-            if (member->IsInCombat())
-                return false;
-
-            Pet* pet = member->GetPet();
-            if (pet && pet->IsInCombat())
-                return false;
-        }
-    }
-
+    // Mana Tide Totem protection: if active and within 30 yards, don't recall.
     ObjectGuid guid = bot->m_SummonSlot[SUMMON_SLOT_TOTEM_WATER];
     if (!guid.IsEmpty())
     {
         Creature* totem = bot->GetMap()->GetCreature(guid);
-        uint32 currentSpell = 0;
-        if (totem)
-        {
-            currentSpell = totem->GetUInt32Value(UNIT_CREATED_BY_SPELL);
-        }
-
+        uint32 currentSpell = totem ? totem->GetUInt32Value(UNIT_CREATED_BY_SPELL) : 0;
         for (size_t i = 0; i < MANA_TIDE_TOTEM_COUNT; ++i)
-        {
             if (currentSpell == MANA_TIDE_TOTEM[i] && totem && totem->GetDistance(bot) <= 30.0f)
                 return false;
-        }
     }
 
+    // Fire Elemental Totem protection: if active and within 30 yards, don't recall.
     guid = bot->m_SummonSlot[SUMMON_SLOT_TOTEM_FIRE];
     if (!guid.IsEmpty())
     {
         Creature* totem = bot->GetMap()->GetCreature(guid);
-        uint32 currentSpell = 0;
-        if (totem)
-        {
-            currentSpell = totem->GetUInt32Value(UNIT_CREATED_BY_SPELL);
-        }
-
+        uint32 currentSpell = totem ? totem->GetUInt32Value(UNIT_CREATED_BY_SPELL) : 0;
         for (size_t i = 0; i < FIRE_ELEMENTAL_TOTEM_COUNT; ++i)
-        {
             if (currentSpell == FIRE_ELEMENTAL_TOTEM[i] && totem && totem->GetDistance(bot) <= 30.0f)
                 return false;
-        }
     }
 
-    return !bot->m_SummonSlot[SUMMON_SLOT_TOTEM_EARTH].IsEmpty() ||
-           !bot->m_SummonSlot[SUMMON_SLOT_TOTEM_FIRE].IsEmpty() ||
-           !bot->m_SummonSlot[SUMMON_SLOT_TOTEM_WATER].IsEmpty() ||
-           !bot->m_SummonSlot[SUMMON_SLOT_TOTEM_AIR].IsEmpty();
+    // Any totem more than 40 yards away → recall (lost buffs, might pull patrols).
+    uint8 const slots[] = { SUMMON_SLOT_TOTEM_EARTH, SUMMON_SLOT_TOTEM_FIRE,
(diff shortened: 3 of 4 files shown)


brighton-chi:
> Warrior — Battle Shout vs Blessing of Might conflict Battle Shout and Blessing of Might both provide attack power and do not stack. When a warrior and a paladin are in the same party, they endlessly overwrite each other (ping-pong). This adds an early-out in BattleShoutTrigger::IsActive() — if the bot already has Blessing of Might (or Greater), skip Battle Shout entirely.

Again, this is not accurate to the trigger. Battle Shout casts only if there is not a stronger Blessing of Might applied. I checked under various scenarios and never got any overwriting. What situation is producing this for you?

This simplified approach would make it so that a much stronger Battle Shout would not be cast if a weaker BoM was on the bot. I don't think that's the right approach, and even if we wanted to do that, you added an early return that makes the remainder of the trigger dead code. The remainder would need to be removed if we were to make this change.

> Priest (Holy) — Shadow Word: Pain in default rotation Holy Priests with high spell power benefit significantly from SW:P as filler DPS, but the default action list only had Smite/Mana Burn/Starshards. SW:P is added with highest default priority. Also lowers AlmostFullHealth from 85→70 so the "healer should attack" trigger fires more often in dungeons.

Again, I don't understand why we would make such an aggressive change to AlmostFullHealth when it is configurable. I don't know why we would change it at all, given the impact. The gate on healer dps is only one that is contingent on this setting. Far more impactful is that this is the top-level healing threshold. If you reduce this to 70%, bots would not heal until 70% (assuming you leave it at the default) since PartyMemberAlmostFullHealthTrigger is based on the setting for AlmostFullHealth. That is way too late for any type of competitive content.

I don't really know enough to comment on whether added SWP to default actions is a good idea, though it looks to me right now that only Smite would ever cast as a default action (the Mana Burn and Starshards casts are dead code since Smite has no cooldown and is always known by a Priest). FWIW, if we do make any changes to SWP, we should probably rename the associated actions/triggers since amusingly, it is named PowerWordPain for them.

> Shaman — Totemic Recall with distance-based trigger The existing trigger had restrictive conditions (no boss fight, all group members out of combat) that prevented totem recall even when the bot had moved far away — leaving totems behind to pull patrols. Replaced with a simple distance check: any totem >40 yards away triggers recall. Also fixed CastTotemicRecallAction::GetTarget() to return self (the spell is self-only; the base class returned nullptr outside combat, causing PREREQ failures).
> 

Note that Shaman already will drop a new totem in combat if the old one is at least 30 yards away and they don't have the aura from the totem. I suppose this change would have functional impact on behavior if a Shaman moved more than 40 yards away and hadn't dropped a totem due to the totem being on cooldown, but I'm not sure why the Shaman should then pick up all of its totems instead of just the one that is out of range.

I believe the assertion about Totemic Recall returning nullptr is accurate so that fix makes sense. However, I am not familiar with the details, but I recall this ability having problems in the past when it was "working" because it was causing Shaman to lock up. If we are to make changes to it, it would require careful review and testing in various scenarios.

shenkai4257:
Warrior: Oh, Battle Shout finally got AP comparison logic? That's great to hear. Even easier then — just make the paladin's Blessing of Might trigger skip when Battle Shout is already up. Problem solved.

Priest SW:P: Regarding SW:P being "dead code": your analysis assumes getDefaultActions() executes actions sequentially without checking isUseful(). That's not how this system works. Each action's isUseful() is evaluated before execution. Once SW:P is applied to the target (21s duration), CastPowerWordPainAction::isUseful() returns false, and the action engine falls through to Smite. SW:P gets refreshed every ~18-20s. This is exactly the intended filler behavior — the DoT uptime is free damage during GCDs that would otherwise be spent on wanding.
On AlmostFullHealth=70: I've tested it across several dungeons and it works fine. If you're concerned about the impact, feel free to run a few bosses with the change and check the healing behavior. It's configurable anyway — users can set it to whatever they want.
And about the naming — PowerWordPainTrigger is a C++ class name, not the spell identifier. The actual spell name passed to the game system is "shadow word: pain" (see PriestTriggers.h lines 17-18), which is exactly how it's named in the DBC for WotLK. Funny thing is — even if I wrote it as the spell name in the DBC, it would still work in-game. Which it does.

Shaman Totemic Recall: Shaman totems are not placed one by one, nor can they be recalled one by one... right? As for why 40 yards? Because that's the totem's effective range. If you're out of range, you're not getting the buff anyway — so you have to re-drop. Might as well recall first for the mana refund and then re-drop. Isn't that just better?

brighton-chi:
I'm rather confused as to why you seem to be taking this so personally. In any case, to respond to your points:

> Warrior: Oh, Battle Shout finally got AP comparison logic? That's great to hear. Even easier then — just make the paladin's Blessing of Might trigger skip when Battle Shout is already up. Problem solved.
> 

What do you mean finally? It's already in the trigger you edited. Why would we always skip BoM when BS is up? If a Paladin's BoM is stronger due to level difference or talents, why shouldn't the Paladin cast BoM? 

Again, what is the situation in which you are observing the buffs overwriting each other? I haven't observed this problem but would be happy to try to reproduce.

> Priest SW:P: Regarding SW:P being "dead code": your analysis assumes getDefaultActions() executes actions sequentially without checking isUseful(). That's not how this system works. Each action's isUseful() is evaluated before execution. Once SW:P is applied to the target (21s duration), CastPowerWordPainAction::isUseful() returns false, and the action engine falls through to Smite. SW:P gets refreshed every ~18-20s. This is exactly the intended filler behavior — the DoT uptime is free damage during GCDs that would otherwise be spent on wanding. On AlmostFullHealth=70: I've tested it across several dungeons and it works fine. If you're concerned about the impact, feel free to run a few bosses with the change and check the healing behavior. It's configurable anyway — users can set it to whatever they want. 

I didn't say SWP was dead code. I said Mana Burn and Starshards are. I don't have an opinion one way or another whether to add SWP to the default actions.

Re: AlmostFullHealth, I'm not sure why the fact that it is configurable means we should change it. That argument cuts both ways.

I'm quite aware that the change would be very meaningful in practice. As an example, Disc Priests used to not use PWS at almost full; they would not until the high threshold (65%). I did a PR to change that, and their effectiveness increased drastically. I have an image in that PR from Eredar Twins where my Disc Priest increased Absorb + Healing by 50%, and PWS went from 36K to 215K. And this is in a scenario where they would heal at almost full; they just didn't use PWS to heal.

https://github.com/mod-playerbots/mod-playerbots/pull/2403

> And about the naming — PowerWordPainTrigger is a C++ class name, not the spell identifier. The actual spell name passed to the game system is "shadow word: pain" (see PriestTriggers.h lines 17-18), which is exactly how it's named in the DBC for WotLK. Funny thing is — even if I wrote it as the spell name in the DBC, it would still work in-game. Which it does.
> 
I am well aware the error is non-functional. That doesn't mean we shouldn't correct it.

> Shaman Totemic Recall: Shaman totems are not placed one by one, nor can they be recalled one by one... right? As for why 40 yards? Because that's the totem's effective range. If you're out of range, you're not getting the buff anyway — so you have to re-drop. Might as well recall first for the mana refund and then re-drop. Isn't that just better?

They can be place one-by-one. Placing them all together is by casting Call of the Elements. The trigger for that spell doesn't fire unless there are at least 2 totems missing.

Totems have a 30-yard range, except Totem of Wrath has a 40-yard range IIRC. The range isn't really the point though. The point is that there are encounters with very large areas, and even if the Shaman is not themselves getting the buff, that doesn't mean nobody is benefitting. I'm not totally opposed to the concept you're implementing; I'm just noting that it's not cut and dry since it could result in picking up a totem that is still useful but can't be recast due to cooldown (the big one is Fire Elemental Totem).

PR #2568, opened 2026-07-11
Revise general boss helpers

brighton-chi:
## Pull Request Description

I've become aware of various issues that need to be addressed in raid strategies so I'm hoping to revisit older ones like Karazhan and TK to make some changes and general code updates (like moving to enum class).

Before I do so, I'm going to correct a few things I should have done in the first place, which affect many raid strategies.
1. Change the generic boss marking helper from a void to a bool so methods that do nothing but mark can actually return true. Note I'm actually moving away from using RTI in strategies as I think it's problematic and largely unnecessary in light of workarounds. But until/unless I get to addressing them all, it's better for this to be a bool.
2. Remove exceptions from IsMechanicTrackerBot for healers, tanks, and specific bots. I limited this helper a lot at first because I was worried about making more important roles have to run separate actions just to manage containers. But now that I can consolidate erasing into a single out-of-combat action that can erase everything in the instance in one tick, it's not really a problem. Plus, I didn't appreciate that AI ticks were just 100ms apart so losing a tick here or there even in combat is hardly relevant. Better to make sure there is always a bot available to perform these tasks.
3. For out-of-combat container erasure methods, use a 3-layer guard of (1) bot is in the instance, (2) bot is not in combat and neither are nearby raid members in combat, and (3) bot doesn't have the applicable boss on its threat list. Bots being in/out of combat is very different from players since it's driven by their own code, and bots often end up in their non-combat engines in certain phases of fights. What exists is mostly fine, but having all of these checks helps prevent timers and such being reset or erased at the wrong time in the middle of a fight, so this is the unified approach I'm going with now.

## Feature Evaluation

Update RaidBossHelpers to change void functions to bool functions, remove role checks from IsMechanicTrackerBot and update signatures and callsites, update triggers for standalone container management functions.

## How to Test the Changes

Run TBC instances with raid strategies and check for bots locking up and not responding to any input, or not properly responding to timed mechanics in encounters.

## Impact Assessment

    - - [ ] No, not at all

Tiny, tiny bit from IsMechanicTracker not being limited to dps bots only.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Raid/Kara/KaraTriggers.cpp
+++ b/src/Ai/Raid/Kara/KaraTriggers.cpp
@@ -110,8 +110,25 @@ bool RomuloAndJulianneBothBossesRevivedTrigger::IsActive()
 
 bool WizardOfOzNeedTargetPriorityTrigger::IsActive()
 {
-    return IsMechanicTrackerBot(bot, KARAZHAN_MAP_ID) &&
-           AI_VALUE2(Unit*, "find target", "moroes");
+    if (!IsMechanicTrackerBot(bot, KARAZHAN_MAP_ID))
+        return false;
+
+    static const std::array<const char*, 5> ozTargets =
+    {
+        "dorothee",
+        "tito",
+        "roar",
+        "strawman",
+        "tinhead",
+    };
+
+    for (const char* name : ozTargets)
+    {
+        if (AI_VALUE2(Unit*, "find target", name))
+            return true;
+    }
+
+    return false;
 }
 
 bool WizardOfOzStrawmanIsVulnerableToFireTrigger::IsActive()
--- a/src/Ai/Raid/Kara/KaraTriggers.cpp
+++ b/src/Ai/Raid/Kara/KaraTriggers.cpp
@@ -14,11 +14,7 @@ bool ManaWarpIsAboutToExplodeTrigger::IsActive()
 
 bool AttumenTheHuntsmanNeedTargetPriorityTrigger::IsActive()
 {
-    if (botAI->IsHeal(bot))
-        return false;
-
-    Unit* midnight = AI_VALUE2(Unit*, "find target", "midnight");
-    return midnight != nullptr;
+    return AI_VALUE2(Unit*, "find target", "midnight");
 }
 
 bool AttumenTheHuntsmanAttumenSpawnedTrigger::IsActive()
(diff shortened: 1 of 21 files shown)


src/Ai/Raid/RaidBossHelpers.cpp
```diff
@@ -217,3 +229,29 @@ Unit* GetNearestPlayerInRadius(Player* bot, float radius)
 
     return nearestPlayer;
 }
+
+// Grid search for dynamic objects for methods to avoid dynobj-based AoE hazards
+std::vector<Position> GetDynamicObjectPositions(Player* bot, float searchRadius, uint32 spellId)
```
kadeshar, line 233: That is declared for future strategies usage?

src/Ai/Raid/Kara/KaraTriggers.cpp
```diff
@@ -127,18 +110,8 @@ bool RomuloAndJulianneBothBossesRevivedTrigger::IsActive()
...
-    Unit* tito = AI_VALUE2(Unit*, "find target", "tito");
-    Unit* roar = AI_VALUE2(Unit*, "find target", "roar");
-    Unit* strawman = AI_VALUE2(Unit*, "find target", "strawman");
-    Unit* tinhead = AI_VALUE2(Unit*, "find target", "tinhead");
-    Unit* crone = AI_VALUE2(Unit*, "find target", "the crone");
-
-    Unit* target = GetFirstAliveUnit({ dorothee, tito, roar, strawman, tinhead, crone });
-    return target != nullptr;
+    return IsMechanicTrackerBot(bot, KARAZHAN_MAP_ID) &&
+           AI_VALUE2(Unit*, "find target", "moroes");
```
kadeshar: moroes in Wizard of Oz trigger?

brighton-chi: lmao

src/Ai/Raid/RaidBossHelpers.cpp
brighton-chi: Yep. I'm actually using it for a new Nightbane strategy in a Karazhan rewrite that I can PR once this one is merged.

src/Ai/Raid/Kara/KaraTriggers.cpp
```diff
@@ -110,8 +110,25 @@ bool RomuloAndJulianneBothBossesRevivedTrigger::IsActive()
...
+        return false;
+
+    static const std::array<const char*, 5> ozTargets =
+    {
+        "dorothee",
+        "tito",
+        "roar",
+        "strawman",
+        "tinhead",
+    };
```
kadeshar: Duplicate list. Not better keep in helper?

brighton-chi: Sure, simple enough

kadeshar (approved):

PR #2557, opened 2026-07-09
[CI/CD] promote warnings to errors.

Celandriel:
## Pull Request Description

To promote a warning free environment this will force all prs to be error free. 

## Feature Evaluation

- Describe the **minimum logic** required to achieve the intended behavior.
- Describe the **processing cost** when this logic executes across many bots.

## Impact Assessment

- Does this change increase per-bot/per-tick processing or risk scaling poorly with thousands of bots?
    - - [ ] No, not at all
    - - [ ] Minimal impact (**explain below**)
    - - [ ] Moderate impact (**explain below**)

- Does this change modify default bot behavior?
    - - [ ] No
    - - [ ] Yes (**explain why**)

- Does this change add new decision branches or increase maintenance complexity?
    - - [ ] No
    - - [ ] Yes (**explain below**)

## AI Assistance

Was AI assistance used while working on this change?
- - [ ] No
- - [ ] Yes (**explain below**)

## Code Provenance / Attribution

Was any code in this PR copied or adapted from a sister / upstream project (e.g. CMaNGOS playerbots,
 MaNGOS, another module)?
- - [ ] No, all code in this PR is original
- - [ ] Yes (**name the project and the original author(s) below**)

## Final Checklist

- - [ ] Stability is not compromised.
- - [ ] Performance impact is understood, tested, and acceptable.
- - [ ] Added logic complexity is justified and explained.
- - [ ] Any new bot dialogue lines are translated.
- - [ ] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).
- - [ ] New source files use the GPLv2 header.
- - [ ] Documentation updated if needed (Conf comments, WiKi commands).
- - [ ] New and modified files do not introduce new compiler warnings.

PR #2569, opened 2026-07-12
[CI/CD] Clear remaining warnings for the upcoming promotion of warnings to errors in #2557

NoxMax:
## Pull Request Description

A couple warnings about casting and unused `bot` from #2518 is addressed. And with the upcoming change of CI in #2557, may as well address the points it brings up in its own compile.
1. Position in MovementActions.h, should be a struct as the core defines it.
2. All the action weights in Strategy.h should be constexpr.

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Base/Actions/MovementActions.h
+++ b/src/Ai/Base/Actions/MovementActions.h
@@ -17,7 +17,7 @@ class Player;
 class PlayerbotAI;
 class Unit;
 class WorldObject;
-class Position;
+struct Position;
 
 #define ANGLE_45_DEG (static_cast<float>(M_PI) / 4.f)
 #define ANGLE_90_DEG M_PI_2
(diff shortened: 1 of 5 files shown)


brighton-chi:
Should we just change the signature then and delete bot from the call sites (probably not many).

NoxMax:
Even more so I think `Player*` itself can be dropped. Take a second look at `IsPositionInActiveConflagration` just to double check that this wouldn't break things.

brighton-chi:
I'll take a look.

Overall, I should probably just start using local variables more instead of parameters for standalone helpers. Avoid the potential for this problem altogether.

Celandriel (approved):

PR #2570, opened 2026-07-12
Feat/mechanar sepethrea kite

jrad7:
## Pull Request Description

Adds a combat strategy for the Nethermancer Sepethrea fight in The Mechanar.

Sepethrea summons Raging Flames elementals that fixate on a random party member with an unbreakable taunt and drop a trail of ground fire as they move. By default, bots treat the flame as a normal add: they attack it, stand in its fire, and the tank chases the fixate off the boss, which wipes the group.

The strategy:
- Kites the fixated bot's flame down the length of the boss room, clear of the party, and casts or heals between hops.
- Keeps the kite inside the room so it never drags the boss up the corridor and starts the Pathaleon gauntlet.
- Leashes a fixated healer to the tank so the heal range holds.
- Steps bystanders out of both Infernos and off the fire trail.
- Holds all DPS on the boss and keeps the tank off the untankable flame, with a path to reclaim a loose boss.

It follows the existing per-dungeon pattern, so it is inert everywhere except inside The Mechanar.

## Feature Evaluation

  Four triggers with their actions and two multipliers, installed only inside The Mechanar. Each trigger exits on the cheapest check first (in combat, then a bounded grid search for the boss or a flame). Hazard sets are gathered once per action and reused across scoring instead of being re-queried per candidate.

  Nothing runs outside The Mechanar, so the open-world population is unaffected. Inside, the cost per bot per tick is a few range-bounded grid searches behind early-outs; the scoring loops run only while a bot is fixated or standing in fire, for the few bots actually in the dungeon.

## How to Test the Changes

With a group covering tank, healer, and DPS, clear to Nethermancer Sepethrea and pull.

Expected:
- The fixated bot walks the flame up and down the room, out of the party, and keeps casting or healing.
- No bot attacks a flame; all DPS stay on the boss.
- Unfixated bots step out of the Inferno and off the fire trail.
- A fixated healer stays in the heal range of the tank.
- The fight never moves toward the elevator corridor.
- If the boss comes loose, the tank breaks off to regain aggro, then resumes.

## Impact Assessment

    - - [ ] No, not at all

  The added work exists only for bots inside The Mechanar, so it does not scale with global bot count. Within the fight, it is a few bounded grid searches behind early-outs, with heavier scoring gated on being fixated or in fire.

    - - [ ] No
    - - [x] Yes (**explain why**)

  It overrides targeting and movement during the Sepethrea fight (do not attack the flame, kite the fixate, dodge the trail, tank stays on the boss), because the default behavior wipes the group. The override is confined to that fight and mirrors the existing per-dungeon strategies.

    - - [ ] No

  It adds a self-contained `Ai/Dungeon/Mechanar/` unit plus one registration line each in the existing dungeon strategy/context files. The complexity is isolated to the new directory and follows the established layout.

## AI Assistance

- - [ ] No

Used for code generation, iteration on the kite and dodge geometry, and documentation. The behavior was refined against the live fight, and all logic has been reviewed and is owned by the contributor.

 MaNGOS, another module)?

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (N/A: no dialogue added.)
- - [x] Any code ported/adapted from another project is attributed. (N/A: all original.)
- - [ ] New source files use the GPLv2 header.
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (N/A: no config or commands added.)

## Notes for Reviewers

- New files under `src/Ai/Dungeon/Mechanar/` follow the same header convention as the existing dungeon strategy files (header guard, no per-file GPLv2 block). I can add GPLv2 headers if preferred.
- The room boundary is a fixed rectangle with every interior cell verified against the live navmesh.
- `SegmentInRoom` also samples the hop midpoint; this is redundant while the boundary stays convex and is kept as a guard against future non-convex shapes.

### DIFF
--- a/src/Ai/Dungeon/Mech/MechActions.cpp
+++ b/src/Ai/Dungeon/Mech/MechActions.cpp
@@ -2,6 +2,7 @@
 #include "AiFactory.h"
 #include "MechActions.h"
 #include "MechShared.h"
+#include "Creature.h"
 #include "Group.h"
 #include "Timer.h"
 
@@ -97,12 +98,13 @@ namespace
         1.75f, -1.75f, 2.10f, -2.10f, 2.45f, -2.45f, 2.80f, -2.80f, 3.14f };
 
     bool SpotSafe(std::vector<std::pair<float, float>> const& flames,
-                  std::vector<MechanarFlames::TrailPatch> const& patches, float x, float y)
+                  std::vector<MechanarFlames::TrailPatch> const& patches, float x, float y,
+                  float flameClear = MechanarFlames::INFERNO_SAFE_DIST)
     {
         for (auto const& f : flames)
         {
             float const dx = x - f.first, dy = y - f.second;
-            if (dx * dx + dy * dy < MechanarFlames::INFERNO_SAFE_DIST * MechanarFlames::INFERNO_SAFE_DIST)
+            if (dx * dx + dy * dy < flameClear * flameClear)
                 return false;
         }
         for (MechanarFlames::TrailPatch const& p : patches)
@@ -115,14 +117,19 @@ namespace
         return true;
     }
 
+    int SegSamples(float ax, float ay, float bx, float by)
+    {
+        return std::max(5, static_cast<int>(Dist2d(ax, ay, bx, by) / 3.0f) + 1);
+    }
+
     bool SegSafe(std::vector<std::pair<float, float>> const& flames,
                  std::vector<MechanarFlames::TrailPatch> const& patches,
                  float ax, float ay, float bx, float by)
     {
-        constexpr int N = 5;
-        for (int k = 1; k <= N; ++k)
+        int const n = SegSamples(ax, ay, bx, by);
+        for (int k = 1; k <= n; ++k)
         {
-            float const t = static_cast<float>(k) / N;
+            float const t = static_cast<float>(k) / n;
             if (!SpotSafe(flames, patches, ax + (bx - ax) * t, ay + (by - ay) * t))
                 return false;
         }
@@ -130,26 +137,58 @@ namespace
     }
 
     float HazardClearance(std::vector<std::pair<float, float>> const& flames,
-                          std::vector<MechanarFlames::TrailPatch> const& patches, float x, float y)
+                          std::vector<MechanarFlames::TrailPatch> const& patches, float x, float y,
+                          float flameClear = MechanarFlames::INFERNO_SAFE_DIST)
     {
         float clear = 1e9f;
         for (auto const& f : flames)
-            clear = std::min(clear, Dist2d(x, y, f.first, f.second) - MechanarFlames::INFERNO_SAFE_DIST);
+            clear = std::min(clear, Dist2d(x, y, f.first, f.second) - flameClear);
         for (MechanarFlames::TrailPatch const& p : patches)
             clear = std::min(clear, Dist2d(x, y, p.x, p.y) - (p.radius + MechanarFlames::TRAIL_DANGER_MARGIN));
+        return clear;
+    }
+
+    float SegClearance(std::vector<std::pair<float, float>> const& flames,
+                       std::vector<MechanarFlames::TrailPatch> const& patches,
+                       float ax, float ay, float bx, float by)
+    {
+        int const n = SegSamples(ax, ay, bx, by);
+        float clear = 1e9f;
+        for (int k = 0; k <= n; ++k)
+        {
+            float const t = static_cast<float>(k) / n;
+            clear = std::min(clear, HazardClearance(flames, patches, ax + (bx - ax) * t, ay + (by - ay) * t));
+        }
         return clear;
     }
 }
 
 bool SepethreaKiteFlameAction::Execute(Event)
 {
-    Unit* flame = MechanarFlames::GetFixatingFlame(bot);
-    if (!flame)
+    std::vector<Creature*> chasers;
+    MechanarFlames::CollectChasingFlames(bot, chasers);
+    if (chasers.empty())
         return false;
 
-    float const dist = bot->GetDistance2d(flame);
     float const bx = bot->GetPositionX();
     float const by = bot->GetPositionY();
+
+    Creature* flame = chasers.front();
+    float dist = bot->GetDistance2d(flame);
+    for (size_t i = 1; i < chasers.size(); ++i)
+    {
+        float const d = bot->GetDistance2d(chasers[i]);
+        if (d < dist)
+        {
+            dist = d;
+            flame = chasers[i];
+        }
+    }
+
+    std::vector<std::pair<float, float>> chaserPos;
+    chaserPos.reserve(chasers.size());
+    for (Creature* chaser : chasers)
+        chaserPos.emplace_back(chaser->GetPositionX(), chaser->GetPositionY());
 
     float fx = bx - flame->GetPositionX();
     float fy = by - flame->GetPositionY();
@@ -164,13 +203,19 @@ bool SepethreaKiteFlameAction::Execute(Event)
         }
     }
 
-    float px, py;
-    bool const haveParty = PartyCentroid(bot, px, py);
-    float const groupDist = haveParty ? bot->GetExactDist2d(px, py) : 1000.0f;
+    std::vector<std::pair<float, float>> avoidFlames;
+    MechanarFlames::CollectAvoidFlames(bot, avoidFlames);
+    std::vector<MechanarFlames::TrailPatch> patches;
+    MechanarFlames::CollectTrailPatches(bot, 60.0f, patches);
 
     float tankX = 0.0f, tankY = 0.0f;
-    bool const leashed = botAI->IsHeal(bot) && TankPosition(bot, tankX, tankY);
+    bool const haveTank = TankPosition(bot, tankX, tankY);
+    bool const leashed = botAI->IsHeal(bot) && haveTank;
     float const tankDist = leashed ? bot->GetExactDist2d(tankX, tankY) : 0.0f;
+
+    float px = tankX, py = tankY;
+    bool const haveParty = haveTank || PartyCentroid(bot, px, py);
+    float const groupDist = haveParty ? bot->GetExactDist2d(px, py) : 1000.0f;
 
     uint32 const nowMs = getMSTime();
     bool turning = _turnUntilMs != 0 && nowMs < _turnUntilMs;
@@ -192,12 +237,7 @@ bool SepethreaKiteFlameAction::Execute(Event)
                                         : by - MechanarFlames::ROOM_Y_MIN;
         if (endDist < MechanarFlames::KITE_TURN_RUNWAY)
         {
-            std::vector<std::pair<float, float>> avoidFlames;
-            MechanarFlames::CollectAvoidFlames(bot, flame, avoidFlames);
-            std::vector<MechanarFlames::TrailPatch> patches;
-            MechanarFlames::CollectTrailPatches(bot, 60.0f, patches);
-
-            float bestKey = -1e18f;
+            float bestKey = -1e18f, bestLaneClear = -1e18f;
             for (float side : { 1.0f, -1.0f })
             {
                 float const sideX = side > 0.0f
@@ -206,21 +246,26 @@ bool SepethreaKiteFlameAction::Execute(Event)
                 float laneEndY = flame->GetPositionY() - endSign * MechanarFlames::KITE_TURN_PASS_BEHIND;
                 laneEndY = std::min(std::max(laneEndY, MechanarFlames::ROOM_Y_MIN + 3.0f),
                                     MechanarFlames::ROOM_Y_MAX - 3.0f);
-                bool const laneSafe = SegSafe(avoidFlames, patches, bx, by, sideX, by) &&
-                                      SegSafe(avoidFlames, patches, sideX, by, sideX, laneEndY);
-                float const key = (laneSafe ? 1000.0f : 0.0f) +
+                float const laneClear =
+                    std::min(SegClearance(avoidFlames, patches, bx, by, sideX, by),
+                             SegClearance(avoidFlames, patches, sideX, by, sideX, laneEndY));
+                float const key = laneClear * MechanarFlames::KITE_TURN_CLEAR_WEIGHT +
                                   std::fabs(sideX - flame->GetPositionX());
                 if (key > bestKey)
                 {
                     bestKey = key;
+                    bestLaneClear = laneClear;
                     _turnSide = side;
                 }
             }
-            _turnEnd = endSign;
-            _turnFlame = flame->GetGUID();
-            _turnLateral = true;
-            _turnUntilMs = nowMs + MechanarFlames::KITE_TURN_FAILSAFE_MS;
-            turning = true;
+            if (bestLaneClear > 0.0f)
+            {
+                _turnEnd = endSign;
+                _turnFlame = flame->GetGUID();
+                _turnLateral = true;
+                _turnUntilMs = nowMs + MechanarFlames::KITE_TURN_FAILSAFE_MS;
+                turning = true;
+            }
         }
     }
     if (turning)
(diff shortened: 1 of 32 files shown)


src/Ai/Dungeon/Mechanar/MechanarActions.cpp
```diff
@@ -0,0 +1,631 @@
...
+        return true;
+    }
+
+    float Dist2d(float ax, float ay, float bx, float by)
+    {
+        float const dx = ax - bx, dy = ay - by;
+        return std::sqrt(dx * dx + dy * dy);
+    }
+
+    // Closest approach of segment a->b to point (px,py): how near a hop's path passes a
```
brighton-chi: Could you please remove the comments for the PR?

src/Ai/Dungeon/Mechanar/MechanarActions.cpp
```diff
@@ -0,0 +1,631 @@
...
+            fx = 1.0f;
+            fy = 0.0f;
+        }
+    }
+
+    // Distance to the huddle of the rest of the party (the boss-DPS stack).
+    float px, py;
+    bool const haveParty = PartyCentroid(bot, px, py);
+    float const groupDist =
+        haveParty ? Dist2d(bot->GetPositionX(), bot->GetPositionY(), px, py) : 1000.0f;
```
brighton-chi: When calculating distance from a WorldObject to coordinates, or between two WorldObjects, you can use Azerothcore functions. For this, you could replace with GetExactDist2d().

src/Ai/Dungeon/Mech/MechActions.cpp
```diff
@@ -0,0 +1,631 @@
...
+
+    if (bestX == bx && bestY == by)
+        return false;  // nowhere better reachable; hold rather than oscillate
+
+    // Heal-capable bots use a gentler priority so a needed heal can still fire once out;
+    // everyone else moves FORCED so combat formation can't immediately drag them back in.
+    MovementPriority const priority =
+        botAI->IsHeal(bot) ? MovementPriority::MOVEMENT_COMBAT : MovementPriority::MOVEMENT_FORCED;
+    return MoveTo(bot->GetMapId(), bestX, bestY, bot->GetPositionZ(), false, false, false, false,
+                  priority, true, false);
```
brighton-chi, line 459: Is this to allow healers to enter fire in order to get to healing range? No issues with healers dying due to this?

src/Ai/Dungeon/Mechanar/MechanarActions.cpp
```diff
@@ -0,0 +1,631 @@
...
+    // everyone else moves FORCED so combat formation can't immediately drag them back in.
+    MovementPriority const priority =
+        botAI->IsHeal(bot) ? MovementPriority::MOVEMENT_COMBAT : MovementPriority::MOVEMENT_FORCED;
+
+    return MoveTo(bot->GetMapId(), destX, destY, bot->GetPositionZ(), false, false, false, false,
+                  priority, true, false);
+}
+
+// Force every bot's DPS onto Sepethrea (never the flame).
+bool SepethreaFocusBossAction::Execute(Event /*event*/)
```
brighton-chi: I think the entire mechanism regarding keeping targets on the boss is much more complicated than it needs to be. All you need to do is disable DpsAssistAction, TankAssistAction, and CastDebuffSpellOnAttackerAction in multipliers, and then this action can just run for all bots and issue Attack() if Sepethrea is not the AI's current target. That will just start combat for them and get them to target Sepethrea.

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
```diff
@@ -0,0 +1,81 @@
...
+        dynamic_cast<SepethreaAvoidTrailAction*>(action) ||
+        dynamic_cast<SepethreaAvoidFlameAction*>(action))
+        return 1.0f;
+
+    // Only a movement action (or the gap-closer casts Charge/Intercept, which are
+    // CastSpellActions, not MovementActions) can fight the kite or dodge. Every other action
+    // (the whole spell rotation and heals) is never touched, so skip the grid searches
+    // entirely for them. This is the common case (most of the queue).
+    bool const isMover =
+        dynamic_cast<MovementAction*>(action) || dynamic_cast<CastReachTargetSpellAction*>(action);
```
brighton-chi: AttackAction inherits from MovementAction so you're going to be blocking attacking for the scenarios below. If that's intended, that's fine, but I just wanted to point that out.

src/Ai/Dungeon/Mech/MechShared.cpp
```diff
@@ -0,0 +1,261 @@
...
+        {
+            float const dx = bot->GetPositionX() - p.x;
+            float const dy = bot->GetPositionY() - p.y;
+            if (std::sqrt(dx * dx + dy * dy) < p.radius + TRAIL_DANGER_MARGIN)
+                return true;
+        }
+        return false;
+    }
+
+    Unit* GetSepethrea(Player* bot)
```
brighton-chi, line 256: You don't need a grid search for the boss. You can just use FindTargetValue, which scans the bot's threat list and is much cheaper. Plus with find target, you also have no need to do IsAlive() or IsInCombat() checks.

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // search for Sepethrea must only happen for the handful of target actions.
+    if (!PlayerbotAI::IsTank(bot))
+        return 1.0f;
+
+    // The stock target-acquisition actions that would (re-)grab the flame. The tank's
+    // boss target is supplied instead by SepethreaFocusBossAction; its spell rotation
+    // (CastSpellAction and similar, not AttackActions) is left fully intact.
+    if (!dynamic_cast<TankAssistAction*>(action) &&
+        !dynamic_cast<AggressiveTargetAction*>(action) &&
+        !dynamic_cast<AttackAnythingAction*>(action))
```
brighton-chi: AggressiveTargetAction and AttackAnythingAction are noncombat methods. I don't think they're relevant here.

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    return 1.0f;
+}
+
+// Keep the tank off the un-tankable flame for the duration of the fight.
+float SepethreaTankFocusMultiplier::GetValue(Action* action)
+{
+    // Cheap gates first: GetValue runs for every queued action every tick, so the grid
+    // search for Sepethrea must only happen for the handful of target actions.
+    if (!PlayerbotAI::IsTank(bot))
```
brighton-chi: botAI is already available for multipliers

src/Ai/Dungeon/Mechanar/MechanarShared.cpp
```diff
@@ -0,0 +1,261 @@
...
+            resX += tx / len * 0.5f;
+            resY += ty / len * 0.5f;
+        }
+        x = resX;
+        y = resY;
+    }
+
+    bool HealerHoldsFire(Player* bot)
+    {
+        if (!bot)
```
brighton-chi: bot and botAI null checks don't have any value for dungeon strategies; you can get rid of all of them.

src/Ai/Dungeon/Mech/MechShared.cpp
```diff
@@ -0,0 +1,261 @@
...
+        float t = (len2 < 1e-6f) ? 0.0f : ((px - ax) * dx + (py - ay) * dy) / len2;
+        t = std::max(0.0f, std::min(1.0f, t));
+        cx = ax + t * dx;
+        cy = ay + t * dy;
+    }
+}
+
+namespace MechanarFlames
+{
+    bool InRoom(float x, float y)
```
brighton-chi, line 45: I think for limiting scope it's best to just inline single-line helpers.

src/Ai/Dungeon/Mechanar/MechanarTriggers.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "MechanarTriggers.h"
+#include "MechanarShared.h"
+#include "AiObject.h"
+#include "AiObjectContext.h"
+
+// Nethermancer Sepethrea: Raging Flames kite.
+
+bool SepethreaKiteFlameTrigger::IsActive()
+{
+    if (!MechanarFlames::GetFixatingFlame(bot))
```
brighton-chi: Triggers are the first line of checks, and they should be ordered by checks based on expense and probability of them occuring. Usually, you'd want to do any role or class checks first, and then most of the time a FindTargetValue check for the boss right after is good to exclude the triggers from firing for the rest of the instance. Here, 3/4 of the triggers are starting with grid searches, and those ideally shouldn't be contained in triggers at all because you're going to then double run them because you need them for the actions anyway.

src/Ai/Dungeon/Mechanar/MechanarTriggers.cpp
```diff
@@ -0,0 +1,65 @@
...
+    if (!MechanarFlames::InTrailDanger(bot))
+        return false;
+
+    // The healer holds its ground to keep healing (same rule as the elemental dodge).
+    return !MechanarFlames::HealerHoldsFire(bot);
+}
+
+bool SepethreaFocusBossTrigger::IsActive()
+{
+    if (!bot->IsInCombat())
```
brighton-chi: This function can just be 
```
return AI_VALUE2(Unit*, "find target", "nethermancer sepethrea");

**Review by brighton-chi — COMMENTED** [review]
Awesome, thank you for doing this. I don't know why I thought you couldn't get dynamic objects from the grid. That's huge. I'm not sure if the performance hit in raids will be too much, but I definitely want to try it out.

File: src/Ai/Dungeon/Mechanar/MechanarActionContext.h [hunk 13]
```diff
@@ -0,0 +1,30 @@
+#ifndef PLAYERBOTS_MECHANARACTIONCONTEXT_H
```
brighton-chi: Forgot to mention, to follow naming convention and because of stupid Windows and file length limitations, can you rename the folder to "Mech" and the Mechanar in Files to Mech? And the strategy string name would be "tbc-mech" instead of "mechanar."

src/Ai/Dungeon/Mechanar/MechanarActionContext.h
```diff
@@ -0,0 +1,30 @@
...
+#define PLAYERBOTS_MECHANARACTIONCONTEXT_H
+
+#include "AiObjectContext.h"
+#include "Action.h"
+#include "MechanarActions.h"
+
+class TbcDungeonMechanarActionContext : public NamedObjectContext<Action>
+{
+public:
+    TbcDungeonMechanarActionContext() : NamedObjectContext<Action>(false, true)
```
brighton-chi: The arguments here are unneeded and can be deleted

jrad7:
Pushed three commits for review convenience, can squash if that's preferred.

First is pure comment stripping, all git diff noise.
Second are the smaller review changes.
Third is the change I was most unsure about, getting rid of the multiple grid searches across both triggers and actions.

src/Ai/Dungeon/Mechanar/MechanarActionContext.h
jrad7: fixed, switch from Mechanar to Mech just about everywhere.

switched the strat name to tbc-mech

src/Ai/Dungeon/Mechanar/MechanarActionContext.h
jrad7: fixed

src/Ai/Dungeon/Mechanar/MechanarActions.cpp
jrad7: all comments are stripped out

src/Ai/Dungeon/Mechanar/MechanarActions.cpp
jrad7: distances now use `GetExactDist2d()`

src/Ai/Dungeon/Mechanar/MechanarActions.cpp
jrad7: ooo thank you, this works better, i was struggling here.

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
jrad7: this stops fixated bots from attacking

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
jrad7: fixed

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
jrad7: these are removed

src/Ai/Dungeon/Mechanar/MechanarShared.cpp
jrad7: removed them all

src/Ai/Dungeon/Mechanar/MechanarTriggers.cpp
jrad7: This is the fix I was most unsure of. Every trigger leads with a "find target" boss check so the search doesn't run outside the fight.

and the flame search is cached per bot and the trigger and action share that value.

src/Ai/Dungeon/Mechanar/MechanarTriggers.cpp
jrad7: switched and removed the checks

src/Ai/Dungeon/Mech/MechActions.cpp
jrad7: The healer will tolerate standing in fire until it is under 50% hp if there is someone who needs heals.

It felt a bit better for the healer to take on more risk, but I don't know... This strat is not nearly 100% successful so I'm sure it could be improved more.

brighton-chi: I'll give this a spin hopefully sometime this week. In practice, does the group tend to get kinda split up and out of healing range?

I understand movement priority is independent from action priority and involves whether a movement will interrupt and override a prior movement. So I figured how this would work would be the healer would run through fire to go heal if needed but if below 50% the avoidance would override the attempt to reach the target. It's a sensible idea, iirc I did pretty much this thing for Archimonde. I'm not sure how intolerable the flames are to stand in at all; I seem to recall them killing you very quickly, but probably it's just best to observe via testing.

src/Ai/Dungeon/Mechanar/MechanarMultipliers.cpp
brighton-chi: Sounds good, I wasn't sure if the flames were fast enough that they should just not attack at all.

src/Ai/Dungeon/Mechanar/MechanarTriggers.cpp
brighton-chi: What I meant was you don't need the GetSepethrea helper at all, and when you need to check for her or need the pointer in an action, multiplier, or trigger you can just inline:

```AI_VALUE2(Unit*, "find target", "nethermancer sepethrea")```

That being said, I think what you have with keeping but changing the helper works fine. With a standalone helper, you have to go through AiObjectContext so how you did it is right. Ultimately, I don't think using the helper provides value in this case but if you like it there's no problem, totally up to you.

brighton-chi:
I'll let Keleborn comment on the value caching if he wants =P

It's definitely a good idea, but I'm just not proficient or knowledgeable at all with the technical aspects of the value system, which is a major gap for sure.

src/Ai/Dungeon/Mech/MechActions.cpp
jrad7: Yes, Claude found the Archimonde fight and took some inspiration from it. That's how we ended up with this solution. The healer would kite away, get trapped in a corner due to the flame trails, and the tank/party would die. The flames do a lot of damage quickly, but the healer has to at least try to take a little risk to save the party.

src/Ai/Dungeon/Mechanar/MechanarTriggers.cpp
jrad7: got it, I'll take another look here.

jrad7: Looked at the git diff with and without the helper, and it just looks cleaner overall with the literal strings contained in the helper. I'm going to keep the helper for now.

src/Bot/Engine/BuildSharedValueContexts.cpp
jrad7, line 1: This is the bit about the value registration I was most unsure about. This looks totally out of place to drop a Mech specific value into. But maybe it's just the first time this has been needed?

I think the other place to register these values is over in ValueContext.h, it contains a long list of values and didn't feel much better thowring it there, mixed in with everything, but if that's where it belongs, I can make that change.

brighton-chi: Yeah I'm not sure the approach has ever been taken with strategies before. Doesn't mean it shouldn't be done, but I don't know enough to have any opinion as to where this should go.

brighton-chi:
I manage to test this on both Normal and Heroic. On Normal, the avoidance works very well. The fight is just super safe overall. Here is a video with IP nerfs of 20% damage/30% healing (due to SWP gear): https://youtu.be/sj1by8IGvNk

However, I noticed two issues that you can see in the video. First is that AoE needs to be suppressed--you can see my selfbot attempting to AoE when the adds are near the boss. Second is that the Rogue never entered combat. The log shows that every time he attempted to engage with SepethreaFocusBossAction, it was blocked by SepethreaKiteFlameMultiplier. It's not immediately obvious to me why.

Heroic is, as expected, much messier since there are 3 adds and they move faster. I tried 5 times with 40% damage/50% healing before finally downing her, but I need to analyze more closely to be able to offer some good feedback there.

jrad7:
Typical rogue, always sandbagging 😄 

I thought I had suppressed aoe, but maybe I did it in dungeonclear and not the stock strategy. I can get that fixed up for sure.

I did zero testing on heroic, so that's certainly a gap I need to start paying more attention to. Thanks for the test runs and feedback, it is very much appreciated.

I'll see what I can clean up over the next couple of days.

jrad7:
Pushed an update:

- Fixed the rogue not entering combat issue. It was due to the bot not entering combat on the pull, then getting stuck out of combat.
- Also made melee more aggressive in general, as they were being super passive.
- Made the entire thing multi-flame aware to help with heroic.
- Some general improvements on the kiting, how the bot picks a kite lane, and how it turns at the end of a lane.
- AOE should be suppressed now.
- On normal, I get over 95% clear rates, for heroic it's still pretty bad, but it should be better.

kadeshar (approved):

PR #2571, opened 2026-07-13
Force rebuff on ready check

Yuof:
## Pull Request Description

Allows refreshing buffs semi-manually: bots don't refresh buffs in combat, so buffs can drop mid-combat on boss fights.
When a ready check arrives out of combat, bots whisper their readiness to the master and start topping off missing or short buffs. The confirm goes out only once nothing is left to do. A bot that still has buff work it can perform stays not-ready until the check ends; that includes a bot out of reagents, whose casts are attempted and fail. Whoever shows not-ready has a reason. The `rebuff` chat command starts the same buff pass without a ready check.

The automatic ready-check behavior is gated behind `AiPlayerbot.ForceRebuffOnReadyCheck` (**off by default**); the manual `rebuff` command works regardless of the config.

How it works:
- `ForceRebuffState` (new `src/Bot/ForceRebuff.h/.cpp`) tracks one rebuff window per bot. While a window is open, buff triggers re-evaluate every tick and the ready reply waits on three gates: a buff action the engine executed this cycle (stamped at the top of the buff `Execute`s; a failed cast is retried, which is how missing reagents keep a bot not-ready), the global cooldown of the last cast (the GCD hides the next round of work from the triggers), and any cast still in progress. Work the bot can never perform does not block, because a wrong-spec talent buff or another caster's stronger aura would otherwise hold the reply forever.
- Buffs are topped off toward full duration by a shared refresh predicate with an adaptive margin, so a raid-sized pass converges instead of re-staling its own first targets. The same predicate now serves every aura-staleness check in the generic buff code.
- An always-on `ForceRebuffStrategy` hosts the reply trigger and a multiplier that suppresses heals while buff work is proposed or the last cast's GCD runs. Bots buff first, heal second, report ready last.
- A two-minute ceiling closes a window that can never finish (a manual rebuff has no ready-check expiry to end it).

**Known limitation:** paladin blessings (including greater blessings) do not participate in the pass. The blessing triggers and actions are not rebuff-aware: blessings are not topped off and do not defer the ready reply, so a paladin can report ready with blessings unrefreshed. Blessing support is planned as a follow-up PR on top of #2530.

## Feature Evaluation

- **Minimum logic:** one small state object per bot (5 fields plus the bot pointer). When no window is open, which is the normal case, the added per-tick cost is a handful of boolean short-circuits: in `Trigger::needCheck` for buff triggers, in the heal multiplier, inside the shared refresh predicate, and at the top of each buff `Execute`, plus one flag reset per engine cycle and one pointer store per successful cast. All constant-time. Opening a window requires an explicit ready check (config-gated) or an explicit chat command.
- **Processing cost across many bots:** idle cost is effectively zero for all bots at all times. During a window, buff triggers evaluate per tick instead of on their throttle interval. That is bounded by the window's lifetime (a ready check's ~30 s, hard-capped at 2 min) and applies only to bots in the group being ready-checked. Bots outside a ready-checked group never open windows.

## How to Test the Changes

Setup: `AiPlayerbot.ForceRebuffOnReadyCheck = 1`, a party or raid with bots (priest/mage/druid buffers make it easy to observe), out of combat.

1. Let buffs age (or dispel a few), then issue a ready check. Expected: each bot whispers its readiness stats immediately, re-casts its missing/short buffs, and its ready confirm arrives only after its casts stop.
2. Repeat with a 25-player raid: the pass completes and every bot reports ready, with no re-cast loop on the first targets.
3. Take a priest bot's candles and ready check: it falls back to single-target Fortitude where possible; if buff work remains that it cannot finish, it stays not-ready until the check expires.
4. Set `ForceRebuffOnReadyCheck = 0` and say `rebuff`: same buff pass, no ready-check reply. During an open ready check, `rebuff` must not eat the pending ready confirm.
5. Ready check while in combat: unchanged legacy behavior (immediate reply).
6. Bots restored from strategy sets saved before this feature answer instantly (legacy behavior); opt one in with a strategy reset or `nc +force rebuff` and verify it then runs the pass.
7. Paladin blessings are expected NOT to refresh during the pass (see known limitation above).

## Impact Assessment

    Idle: one boolean short-circuit per buff-trigger `needCheck` and per multiplier evaluation. Full evaluation happens only inside an explicitly opened, time-bounded window, for the checked group only.

    - - [x] Yes (**explain why**)

    With the config off, two things change: (1) the `rebuff` chat command exists (explicit master command, same trust gate as other commands); (2) the consolidated refresh predicate makes `GroupBuffSpellAction` re-cast a group buff whose group-wide aura is about to expire instead of treating it as covered, a small correctness improvement to existing behavior.

    One new state class + strategy + trigger + two actions, deliberately kept in one file pair (`src/Bot/ForceRebuff.h/.cpp`) plus standard context registrations. The ready-reply decision is four conditions in one place (`ReadyReplyAction::isUseful`).

## AI Assistance

Used for design exploration and code generation across the change (Claude Code). All code was reviewed line by line and tested in-game by the contributor. The design decisions are the contributor's, from what counts as blocking work to the heal suppression and the timeout ceiling.

## Final Checklist

- - [x] Any new bot dialogue lines are translated. *(no new dialogue lines; the readiness report reuses existing output)*
- - [x] Any code ported/adapted from another project is attributed. *(nothing ported)*
- - [x] Documentation updated if needed (Conf comments, WiKi commands). *(conf.dist documents both settings and the `rebuff` command)*

## Notes for Reviewers

- The not-ready signal counts only executed buff work plus the last cast's GCD and in-flight casts. Reagent starvation still blocks, since those casts are attempted and fail. The 2-minute ceiling covers the manual command's window.
- Bots keep saved strategy sets (the playerbots DB store restores them on login), so existing bots do not pick up the new always-on "force rebuff" strategy automatically. It felt wrong to migrate those, so AltBots will need a reset.
- I am not sure how to edit the wiki to add the new command — pointers welcome.

### DIFF
--- a/src/Bot/Engine/Trigger/Trigger.cpp
+++ b/src/Bot/Engine/Trigger/Trigger.cpp
@@ -7,6 +7,7 @@
 #include "Trigger.h"
 #include "AiObjectContext.h"
 #include "Event.h"
+#include "PlayerbotAI.h"
 
 Trigger::Trigger(PlayerbotAI* botAI, std::string const name, int32 checkInterval)
     : AiNamedObject(botAI, name),
@@ -33,6 +34,10 @@ Unit* Trigger::GetTarget() { return GetTargetValue()->Get(); }
 
 bool Trigger::needCheck(uint32 now)
 {
+    // During an out-of-combat force-rebuff, evaluate every buff trigger each tick
+    if (IsBuffTrigger() && !IsDebuffTrigger() && botAI->forceRebuff.IsPending() && !bot->IsInCombat())
+        return true;
+
     if (checkInterval < 2)
         return true;
 
--- a/src/Bot/ForceRebuff.h
+++ b/src/Bot/ForceRebuff.h
@@ -1,0 +1,62 @@
+/*
+ * This file is part of the mod-playerbots module for AzerothCore. See AUTHORS file for Copyright
+ * information; released under GNU GPL v2 license, redistribute/modify under version 2 of the License,
+ * or (at your option) any later version.
+ */
+#ifndef PLAYERBOTS_FORCEREBUFF_H
+#define PLAYERBOTS_FORCEREBUFF_H
+
+#include "Define.h"
+#include "Strategy.h"
+#include <string>
+#include <vector>
+
+class Player;
+class PlayerbotAI;
+class SpellInfo;
+
+class ForceRebuffState
+{
+public:
+    explicit ForceRebuffState(Player* owner = nullptr) : bot(owner) {}
+
+    ForceRebuffState(ForceRebuffState const&) = delete;
+    ForceRebuffState& operator=(ForceRebuffState const&) = delete;
+
+    void Begin(bool reply);
+    void End() { pending = false; }
+    bool IsPending() const;
+    bool ShouldReplyToReadyCheck() const { return replyToReadyCheck; }
+
+    void NoteBuffWork();
+    void NoteBuffProposed();
+    void RollBuffPendingCycle() { buffPendingThisCycle = false; buffProposedThisCycle = false; }
+    bool IsBuffPendingThisCycle() const { return buffPendingThisCycle; }
+    bool IsBuffProposedThisCycle() const { return buffProposedThisCycle; }
+
+    void NoteCast(SpellInfo const* spellInfo) { lastGcdSpell = spellInfo; }
+    bool IsOnGlobalCooldown() const;
+
+    bool BuffBelowRefreshTarget(int32 remaining, int32 maxDuration, uint32 baseBeforeDuration) const;
+
+private:
+    Player* bot;
+    bool pending = false;
+    bool replyToReadyCheck = false;
+    bool buffPendingThisCycle = false;
+    bool buffProposedThisCycle = false;
+    SpellInfo const* lastGcdSpell = nullptr;
+    uint32 beginMs = 0;
+};
+
+class ForceRebuffStrategy : public Strategy
+{
+public:
+    ForceRebuffStrategy(PlayerbotAI* botAI) : Strategy(botAI) {}
+
+    std::string const getName() override { return "force rebuff"; }
+    void InitTriggers(std::vector<TriggerNode*>& triggers) override;
+    void InitMultipliers(std::vector<Multiplier*>& multipliers) override;
+};
+
+#endif
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -509,6 +509,22 @@ AiPlayerbot.TellWhenMissingBuffReagents = 1
 # Default: 300 (5 minutes)
 AiPlayerbot.MissingBuffReagentMessageCooldown = 300
 
+# When a ready check is issued out of combat, bots refresh missing/short buffs before reporting ready.
+# A bot that still has buff work it can perform stays not-ready until the ready check ends; this
+# includes missing reagents (the cast is attempted and fails). Uses the always-on "force rebuff"
+# strategy: bots with strategy sets saved before this feature keep their saved set - reset their
+# strategies or use "nc +force rebuff" to opt them in. The "rebuff" chat command starts the same
+# buff pass without a ready check and works regardless of this setting.
+# Default: 0 (disabled)
+AiPlayerbot.ForceRebuffOnReadyCheck = 0
+
+# During a force-rebuff pass (ready check or the "rebuff" command), a buff is topped off (recast)
+# once it drops more than this many seconds below its full duration, so buffs are refreshed toward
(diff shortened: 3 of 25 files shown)


brighton-chi:
Based on a very quick peek at the code, this doesn't include weapon enhancements, right? So it doesn't include poisons, Shaman weapon buffs, using spell stones or oils, etc.?

I don't think this needs to be both a config and a strategy. I don't know if it really needs to be either of them--you have a chat command so I'm not sure the ready check element adds much. If this was only a chat command, you wouldn't need the config or the strategy since there's no reason somebody would need to disable its usage (or disable it serverwide since this isn't a cheat of any sort). But if you do keep the ready check component, I'd lean to it just being a config or strategy, and for that component only. Between the two, config is nice to limit strategy list noise, or alternatively it could be a strategy but disabled by default.

Yuof:
> I don't think this needs to be both a config and a strategy. I don't know if it really needs to be either of them--you have a chat command so I'm not sure the ready check element adds much. If this was only a chat command, you wouldn't need the config or the strategy since there's no reason somebody would need to disable its usage (or disable it serverwide since this isn't a cheat of any sort). But if you do keep the ready check component, I'd lean to it just being a config or strategy, and for that component only. Between the two, config is nice to limit strategy list noise, or alternatively it could be a strategy but disabled by default.

It started as a convenience feature as a ready-check extension, as that felt like a natural in-game flow.
It also provides some nice visual feedback.
I additionally added the (default off) config option: a ready check before every boss uses up spell reagents very quickly and that might not be desired by some users.
I later added the command as a second convenient way to use this if the config is false.

It being a strategy is necessary as I need a place to add ForceRebuffBuffFirstMultiplier  (delay heals after buffs). It didn't feel correct to bolt that to any other strategy. Without the multiplier my test setup with a 25man lvl 60 raid (with individual progression, as that disables group buffs till 61) went from ~90s down to ~25s for full buffs.

anyways I feel like a "default off" strategy will not be discovered by most users so strategy + config option + chat command felt like a nice balance.

brighton-chi:
What is the issue with healing? Priests stopping to heal after single Fortitude?

Group buffs exist at 60 even with IP. They were around in Vanilla, and the first ranks are pre-60. I think they are just not available from trainers (need books).

Yuof:
> What is the issue with healing? Priests stopping to heal after single Fortitude?
yes, they will oscillate between buffing an healing which makes them take ages to finish the whole raid.


> Group buffs exist at 60 even with IP. They were around in Vanilla, and the first ranks are pre-60. I think they are just not available from trainers (need books).

that was just an example (the story how I found this). This can happen for other reasons.

src/Bot/Engine/Trigger/Trigger.cpp
```diff
@@ -34,6 +35,10 @@ Unit* Trigger::GetTarget() { return GetTargetValue()->Get(); }
 
 bool Trigger::needCheck(uint32 now)
 {
+    // During an out-of-combat force-rebuff, evaluate every buff trigger each tick
+    if (IsBuffTrigger() && botAI->forceRebuff.IsPending() && !bot->IsInCombat())
```
kadeshar: DebuffTrigger is also BuffTrigger. Should be excluded here i think

src/Bot/ForceRebuff.h
```diff
@@ -0,0 +1,59 @@
...
+    bool BuffBelowRefreshTarget(int32 remaining, int32 maxDuration, uint32 baseBeforeDuration) const;
+
+private:
+    Player* _bot;
+    bool _pending = false;
+    bool _replyToReadyCheck = false;
+    bool _buffPendingThisCycle = false;
+    bool _buffProposedThisCycle = false;
+    SpellInfo const* _lastGcdSpell = nullptr;
+    uint32 _beginMs = 0;
```
kadeshar: Mostly in module private variables starting from small letter not _

kadeshar (changes requested):

src/Bot/Engine/Trigger/Trigger.cpp
```diff
@@ -34,6 +36,10 @@ Unit* Trigger::GetTarget() { return GetTargetValue()->Get(); }
 
 bool Trigger::needCheck(uint32 now)
 {
+    // During an out-of-combat force-rebuff, evaluate every buff trigger each tick
+    if (IsBuffTrigger() && !dynamic_cast<DebuffTrigger*>(this) && botAI->forceRebuff.IsPending() && !bot->IsInCombat())
```
kadeshar: Add IsDebuffTrigger to follow pattern (like IsBuffTrigger)

kadeshar (changes requested):

kadeshar (approved):

PR #2573, opened 2026-07-14
MC Improvements

Yuof:
## Pull Request Description

Molten Core strategy improvements, measured with a bot-only raid (27 bots, one per talent tree, no real player) at individual-progression vanilla tuning (0.5 power / 0.5 healing). Every change was benchmarked with 10 runs per configuration.

**Golemagg**
- Core Ragers are excluded from dps target selection while Golemagg lives (target-exclusion framework from #2494): they auto-heal to full at 50% and cannot die before he does, so any dps sent there is wasted.
- Healers hold the midpoint between the boss camp and the Core Rager tank camp. Heal target selection drops anyone beyond 40y or out of LOS; with healers parked at one camp, the other camp's tank died unhealed.
- The default "melee when out of mana / in dead zone" fallback is blocked for non-melee on this fight (Magma Splash makes walking into melee stupid for casters), and melee now back off once their Magma Splash stack reaches 20, returning after it expires (threshold swept below).

**Majordomo Executus**
- Excluded from dps target selection entirely: he alternates Magic/Damage Reflection and cannot die; the encounter ends when his eight adds are dead, so attacks on him are reflected damage for nothing.

**Lava escape (MC-wide, mainly Ragnaros)**
- A bot standing or swimming in magma (knocked into a lava pool by Wrath of Ragnaros or an Eruption) runs to the nearest raid member on dry ground; without this, bots swim in place and burn to death.

**Shazzrah**
- The ranged-spread trigger gates on Arcane Explosion range instead of firing unconditionally (every already-safe ranged bot attempted a failing move each tick).

### Back-Off Threshold Sweep

Without a back-off (current behavior) melee die to Golemagg's stacking dot. The sweep looked for the highest threshold that still prevents the deaths; each row is 10 runs.

| Back-off | Kill avg (blues) | Deaths avg (blues) | Kill avg (BiS 76) | Deaths avg (BiS 76) |
|---|---|---|---|---|
| none (current) | 108s | 3.4 | 112s | 5.2 |
| at 5 stacks | 139s | 0.3 | 146s | 0.7 |
| at 10 stacks | 126s | 0.2 | 133s | 0.5 |
| **at 20 stacks (chosen)** | **118s** | **0.3** | **121s** | **0.7** |

### Full Clear Results

Pre-raid blues, 10 runs per boss.

| Encounter | Pass | Kill avg | Deaths avg |
|---|---|---|---|
| Lucifron | 10/10 | 53s | 1.0 |
| Magmadar | 10/10 | 94s | 0.0 |
| Gehennas | 10/10 | 59s | 0.1 |
| Garr | 10/10 | 83s | 1.2 |
| Baron Geddon | 10/10 | 72s | 0.1 |
| Shazzrah | 10/10 | 49s | 0.2 |
| Sulfuron Harbinger | 10/10 | 64s | 0.0 |
| Golemagg | 10/10 | 121s | 0.3 |
| Majordomo Executus | 10/10 | 75s | 4.5 |
| Ragnaros | 10/10 | 106s | 0.2 |

Before the Golemagg changes the same comp went 0/10 on him at this tuning; with them he is a stable 10/10. The Majordomo deaths are the add burst at pre-raid gear.

## Feature Evaluation

- **Minimum logic:** two entries in the MC strategy's target-exclusion set (a guid lookup during dps target selection, only while the MC raid strategy is active); one healer position action gated by a distance check; one aura-stack read in the Golemagg multiplier; a liquid-flag check for the lava trigger; a range gate on the Shazzrah trigger. No new polling loops, no per-tick scans.
- **Processing cost:** all logic lives behind the MC raid strategy, so bots outside Molten Core pay nothing. Inside MC the additions are constant-time checks (aura stack amount, 2D distance, liquid flags) evaluated on the normal trigger cadence. The Shazzrah change is a net reduction: it removes thousands of failing move attempts per fight.

## How to Test the Changes

1. Enable instance strategies (`AiPlayerbot.ApplyInstanceStrategies = 1`) and raid Molten Core with a bot raid (tested with 27 bots at individual-progression 0.5 tuning, pre-raid gear).
2. Golemagg: melee leave the boss around 20 Magma Splash stacks and return after the stack expires; healers stand between the two tank camps; no bot attacks a Core Rager while Golemagg lives; below 10% everyone burns the boss.
3. Majordomo: bots kill the eight adds and never target Majordomo himself.
4. Ragnaros: a bot knocked into a lava pool runs to the nearest dry raid member instead of swimming in place until it dies.
5. Shazzrah: ranged bots reposition only when actually inside Arcane Explosion range.

The tables above come from an automated raid harness (10 runs per configuration on a dedicated test server); the harness itself is a separate branch/PR.

## Impact Assessment

Constant-time checks on the MC strategy's trigger cadence; bots outside Molten Core are untouched.

    - - [x] Yes (**explain why**)

Only inside Molten Core with the MC raid strategy active (the changes described above); behavior elsewhere is unchanged.

Encounter-specific triggers/actions following the established MC strategy pattern, each gated to its boss.

## AI Assistance

Used for the benchmark automation and for drafting. I reviewed and tested all strategy code myself; the tables above are those test runs.

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (no new dialogue)
- - [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note). (nothing ported)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (no config changes)

## Notes for Reviewers

- All numbers were produced at individual-progression 0.5 vanilla tuning, which is harsher than stock; stock-tuning results should only be better.

### DIFF
--- a/src/Ai/Raid/MC/MCActions.cpp
+++ b/src/Ai/Raid/MC/MCActions.cpp
@@ -9,15 +9,18 @@
 #include "Playerbots.h"
 #include "RtiTargetValue.h"
 
+#include <algorithm>
+
 static constexpr float LIVING_BOMB_DISTANCE = 20.0f;
 static constexpr float INFERNO_DISTANCE = 20.0f;
-
-// don't get hit by Arcane Explosion but still be in casting range
-static constexpr float ARCANE_EXPLOSION_DISTANCE = 26.0f;
 
 // dedicated tank positions; prevents assist tanks from positioning Core Ragers on steep walls on pull
 static const Position GOLEMAGG_TANK_POSITION{795.7308, -994.8848, -207.18661};
 static const Position CORE_RAGER_TANK_POSITION{846.6453, -1019.0639, -198.9819};
+
+// Midpoint of the two tank camps (56y apart): healers standing here reach both
+static const Position GOLEMAGG_HEALER_POSITION{821.2f, -1007.0f, -203.0f};
+static constexpr float HEALER_POSITION_TOLERANCE = 8.0f;
 
 static constexpr float GOLEMAGGS_TRUST_DISTANCE = 30.0f;
 static constexpr float CORE_RAGER_STEP_DISTANCE = 5.0f;
@@ -55,6 +58,75 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
             return MoveAway(boss, distToTravel);
     }
     return false;
+}
+
+bool McMoveFromLavaAction::Execute(Event /*event*/)
+{
+    // Escape to the nearest raid member on dry ground: raids fight on the
+    // rock shelves, so a dry teammate marks the closest safe floor.
+    Unit* dryTarget = nullptr;
+    float bestDist = 100.0f;
+
+    if (Group* group = bot->GetGroup())
+    {
+        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+        {
+            Player* member = ref->GetSource();
+            if (!member || member == bot || !member->IsAlive() || member->GetMapId() != bot->GetMapId())
+                continue;
+
+            if (member->GetLiquidData().Flags & MAP_LIQUID_TYPE_MAGMA)
+                continue;
+
+            float dist = bot->GetDistance(member);
+            if (dist < bestDist)
+            {
+                bestDist = dist;
+                dryTarget = member;
+            }
+        }
+    }
+
+    // No dry raid member in range: stay put and let the in-lava trigger fire
+    // again — any other destination (the current target included) can path
+    // straight back into the pool.
+    if (!dryTarget)
+        return false;
+
+    botAI->InterruptSpell();
+
+    if (!MoveTo(dryTarget->GetMapId(), dryTarget->GetPositionX(), dryTarget->GetPositionY(),
+                dryTarget->GetPositionZ(), false, false, false, true, MovementPriority::MOVEMENT_FORCED))
+        return false;
+
+    // Hold the AI so rotation/facing actions can't cancel the escape spline
+    // mid-lava. Capped at 2s on purpose: the in-lava trigger refires and
+    // re-issues the escape while the bot is still swimming.
+    constexpr uint32 MAX_ESCAPE_HOLD_MS = 2000;
+    constexpr float MIN_ESCAPE_SPEED = 0.1f;
+
+    float const speed = bot->GetSpeed(MOVE_RUN);
+    if (speed > MIN_ESCAPE_SPEED)
+        botAI->SetNextCheckDelay(std::min(MAX_ESCAPE_HOLD_MS, uint32(1000.0f * bot->GetDistance(dryTarget) / speed)));
+
+    return true;
+}
+
+bool McGolemaggBackOffAction::Execute(Event /*event*/)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "golemagg the incinerator");
+    if (!boss)
+        return false;
+
+    // Out of swing range the stack stops growing and expires 30s after
+    // the last application; the multiplier blocks re-engaging until then.
+    botAI->InterruptSpell();
+
+    float distToTravel = MAGMA_SPLASH_BACK_OFF_DISTANCE - bot->GetDistance2d(boss);
+    if (distToTravel <= 0.0f)
+        return true;
+
+    return MoveAway(boss, distToTravel);
 }
 
 bool McGolemaggMarkBossAction::Execute(Event /*event*/)
@@ -134,6 +206,16 @@ bool McGolemaggMainTankAttackGolemaggAction::Execute(Event /*event*/)
             return MoveUnitToPosition(boss, GOLEMAGG_TANK_POSITION, boss->GetCombatReach());
     }
     return false;
+}
+
+bool McGolemaggHealerPositionAction::Execute(Event /*event*/)
+{
+    if (bot->GetExactDist2d(GOLEMAGG_HEALER_POSITION) <= HEALER_POSITION_TOLERANCE)
+        return false;
+
+    return MoveTo(bot->GetMapId(), GOLEMAGG_HEALER_POSITION.GetPositionX(), GOLEMAGG_HEALER_POSITION.GetPositionY(),
+                  GOLEMAGG_HEALER_POSITION.GetPositionZ(), false, false, false, false,
+                  MovementPriority::MOVEMENT_COMBAT, true, false);
 }
 
 bool McGolemaggAssistTankAttackCoreRagerAction::Execute(Event event)
--- a/src/Ai/Raid/MC/MCMultipliers.cpp
+++ b/src/Ai/Raid/MC/MCMultipliers.cpp
@@ -5,16 +5,21 @@
  */
 
 #include "MCMultipliers.h"
+#include "AttackAction.h"
 #include "ChooseTargetActions.h"
 #include "DKActions.h"
 #include "DruidActions.h"
+#include "GenericActions.h"
 #include "GenericSpellActions.h"
 #include "HunterActions.h"
 #include "MCActions.h"
 #include "MCHelpers.h"
+#include "MovementActions.h"
 #include "PaladinActions.h"
 #include "Playerbots.h"
+#include "ReachTargetActions.h"
 #include "ShamanActions.h"
+#include "SpellAuras.h"
 #include "WarriorActions.h"
 
 using namespace MoltenCoreHelpers;
@@ -100,8 +105,11 @@ static bool IsSingleLivingTankInGroup(Player* bot)
 
 float GolemaggMultiplier::GetValue(Action* action)
 {
-    if (AI_VALUE2(Unit*, "find target", "golemagg the incinerator"))
+    if (Unit* golemagg = AI_VALUE2(Unit*, "find target", "golemagg the incinerator"))
     {
+        // Below 10% (soft enrage: Earthquake + converging Core Ragers) the
+        // fight is a burn race, everyone dps, no splash management.
+        bool burnPhase = golemagg->GetHealthPct() <= 10.0f;
         if (PlayerbotAI::IsTank(bot) && IsSingleLivingTankInGroup(bot))
         {
             // Only one tank => Pick up Golemagg and the two Core Ragers
@@ -116,6 +124,21 @@ float GolemaggMultiplier::GetValue(Action* action)
                 return 0.0f;
         }
         if (IsDpsBotWithAoeAction(bot, action))
+            return 0.0f;
+
+        // Golemagg's Magma Splash stacks an uncapped 30s dot on anyone striking
+        // him in melee. This blocks the default "melee when out of mana / in dead zone" fallback
+        // that walks casters and hunters onto the boss.
+        if (!burnPhase && PlayerbotAI::IsRanged(bot) && dynamic_cast<MeleeAction*>(action))
+            return 0.0f;
+
+        // Backed-off melee stay out until their whole Magma Splash stack expires (30s after the last application).
+        Aura* splash = bot->GetAura(SPELL_MAGMA_SPLASH);
+        bool backedOff = splash && splash->GetStackAmount() >= MAGMA_SPLASH_BACK_OFF_STACKS;
+        bool engagesBoss = !dynamic_cast<McGolemaggBackOffAction*>(action) &&
+                           (dynamic_cast<AttackAction*>(action) || dynamic_cast<MeleeAction*>(action) ||
+                            dynamic_cast<CastReachTargetSpellAction*>(action));
+        if (!burnPhase && !PlayerbotAI::IsTank(bot) && backedOff && engagesBoss)
             return 0.0f;
     }
     return 1.0f;
--- a/src/Ai/Raid/MC/MCStrategy.cpp
+++ b/src/Ai/Raid/MC/MCStrategy.cpp
@@ -5,8 +5,12 @@
  */
 
 #include "MCStrategy.h"
+#include "MCHelpers.h"
 #include "MCMultipliers.h"
+#include "Playerbots.h"
 #include "Strategy.h"
+
+using namespace MoltenCoreHelpers;
 
 void RaidMcStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
 {
@@ -66,6 +70,12 @@ void RaidMcStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
     triggers.push_back(
         new TriggerNode("mc golemagg is assist tank",
                         { NextAction("mc golemagg assist tank attack core rager", ACTION_RAID) }));
+    triggers.push_back(
+        new TriggerNode("mc golemagg magma splash",
+                        { NextAction("mc golemagg back off", ACTION_RAID + 1) }));
+    triggers.push_back(
+        new TriggerNode("mc golemagg is healer",
+                        { NextAction("mc golemagg healer position", ACTION_RAID) }));
 
     // Majordomo Executus
     triggers.push_back(
@@ -81,6 +91,11 @@ void RaidMcStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
     triggers.push_back(
         new TriggerNode("mc core hound mark",
                         { NextAction("mc core hound mark", ACTION_RAID) }));
+
+    // Anywhere in MC: escaping a lava pool outranks everything else.
+    triggers.push_back(
+        new TriggerNode("mc in lava",
+                        { NextAction("mc move from lava", ACTION_RAID + 1) }));
 }
 
 void RaidMcStrategy::InitMultipliers(std::vector<Multiplier*>& multipliers)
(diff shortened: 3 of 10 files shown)


src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +58,71 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
     return false;
 }
 
+bool McMoveFromLavaAction::Execute(Event /*event*/)
+{
+    // Escape to the nearest raid member on dry ground: raids fight on the
+    // rock shelves, so a dry teammate marks the closest safe floor.
+    Unit* dryTarget = nullptr;
+    float bestDist = 100.0f;
+
```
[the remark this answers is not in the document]
Yuof: A bot more than 100y from every raid member while swimming in magma is a wipe, not a pathing problem.

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +58,71 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+    if (!MoveTo(dryTarget->GetMapId(), dryTarget->GetPositionX(), dryTarget->GetPositionY(),
+                dryTarget->GetPositionZ(), false, false, false, true, MovementPriority::MOVEMENT_FORCED))
+        return false;
+
+    // Hold the AI for the travel time so rotation/facing actions can't cancel
+    // the escape spline mid-lava
+    float const speed = bot->GetSpeed(MOVE_RUN);
+    if (speed > 0.1f)
+        botAI->SetNextCheckDelay(std::min(2000u, uint32(1000.0f * bot->GetDistance(dryTarget) / speed)));
```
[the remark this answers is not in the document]
Yuof: The in-lava trigger refires and re-issues the escape if we are still in lava after 2s.

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +58,71 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+                bestDist = dist;
+                dryTarget = member;
+            }
+        }
+
+    if (!dryTarget)
+        dryTarget = AI_VALUE(Unit*, "current target");
+
+    if (!dryTarget)
+        return false;
```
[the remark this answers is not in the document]
Yuof: fixed

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -4,16 +4,21 @@
 #include "RtiTargetValue.h"
 #include "MCHelpers.h"
 
+#include <algorithm>
+
 static constexpr float LIVING_BOMB_DISTANCE = 20.0f;
 static constexpr float INFERNO_DISTANCE = 20.0f;
 
 // don't get hit by Arcane Explosion but still be in casting range
```
kadeshar: Comment to delete

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +57,72 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+            if (member->GetLiquidData().Flags & MAP_LIQUID_TYPE_MAGMA)
+                continue;
+
+            float dist = bot->GetDistance(member);
+            if (dist < bestDist)
+            {
+                bestDist = dist;
+                dryTarget = member;
+            }
+        }
```
kadeshar, line 87: Multiline if should have { }

src/Ai/Raid/MC/MCMultipliers.cpp
```diff
@@ -112,6 +118,21 @@ float GolemaggMultiplier::GetValue(Action* action)
...
+            return 0.0f;
+
+        // Backed-off melee stay out until their whole Magma Splash stack expires (30s after the last application).
+        if (!burnPhase && !PlayerbotAI::IsTank(bot))
+            if (Aura* splash = bot->GetAura(SPELL_MAGMA_SPLASH))
+                if (splash->GetStackAmount() >= MAGMA_SPLASH_BACK_OFF_STACKS &&
+                    !dynamic_cast<McGolemaggBackOffAction*>(action) &&
+                    (dynamic_cast<AttackAction*>(action) || dynamic_cast<MeleeAction*>(action) ||
+                     dynamic_cast<CastReachTargetSpellAction*>(action)))
+                    return 0.0f;
```
kadeshar: Merge if's

kadeshar (changes requested):

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +55,74 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+    }
+
+    // No dry raid member in range: stay put and let the in-lava trigger fire
+    // again — any other destination (the current target included) can path
+    // straight back into the pool.
+    if (!dryTarget)
+        return false;
+
+    bot->AttackStop();
```
brighton-chi: AttackStop() doesn't have any impact on avoidance. You only need it if you have a situation where you need to stop autoattacking but are still in range. It's just cancelling the autoattack.

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +55,74 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+    bot->InterruptNonMeleeSpells(false);
```
brighton-chi: It's pretty much always better to use botAI->InterruptSpell(), which covers channeling melee spells also like Whirlwind. In either case, pending casts won't be interrupted. Not sure if that really matters here.

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -131,6 +202,17 @@ bool McGolemaggMainTankAttackGolemaggAction::Execute(Event /*event*/)
     return false;
 }
 
+bool McGolemaggHealerPositionAction::Execute(Event /*event*/)
+{
+    if (bot->GetExactDist2d(GOLEMAGG_HEALER_POSITION.GetPositionX(), GOLEMAGG_HEALER_POSITION.GetPositionY()) <=
```
brighton-chi: You don't need to pass the X and Y positions as arguments for GetExactDist2d(). This can be just bot->GetExactDist2d(GOLEMAGG_HEALER_POSITION).

src/Ai/Raid/MC/MCMultipliers.cpp
```diff
@@ -112,6 +118,21 @@ float GolemaggMultiplier::GetValue(Action* action)
         }
         if (IsDpsBotWithAoeAction(bot, action))
             return 0.0f;
+
+        // Golemagg's Magma Splash stacks an uncapped 30s dot on anyone striking
+        // him in melee. This blocks the default "melee when out of mana / in dead zone" fallback
+        // that walks casters and hunters onto the boss.
+        if (!burnPhase && !PlayerbotAI::IsTank(bot) && !PlayerbotAI::IsMelee(bot) && dynamic_cast<MeleeAction*>(action))
```
brighton-chi: Could consolidate by checking PlayerbotAI::IsRanged(bot) here

src/Ai/Raid/MC/MCMultipliers.cpp
```diff
@@ -112,6 +118,21 @@ float GolemaggMultiplier::GetValue(Action* action)
...
+            return 0.0f;
+
+        // Backed-off melee stay out until their whole Magma Splash stack expires (30s after the last application).
+        Aura* splash = bot->GetAura(SPELL_MAGMA_SPLASH);
+        bool backedOff = splash && splash->GetStackAmount() >= MAGMA_SPLASH_BACK_OFF_STACKS;
+        bool engagesBoss = !dynamic_cast<McGolemaggBackOffAction*>(action) &&
+                           (dynamic_cast<AttackAction*>(action) || dynamic_cast<MeleeAction*>(action) ||
```
brighton-chi, line 139: You return true when the bot is out of range so does blocking AttackAction and MeleeAction do anything here?

src/Ai/Raid/MC/MCStrategy.cpp
```diff
@@ -84,3 +99,26 @@ void RaidMcStrategy::InitMultipliers(std::vector<Multiplier*>& multipliers)
     multipliers.push_back(new BaronGeddonAbilityMultiplier(botAI));
     multipliers.push_back(new GolemaggMultiplier(botAI));
 }
+
+void RaidMcStrategy::AppendTargetExclusions(GuidSet& exclusions, TargetValueExclusionType type)
+{
+    if (type != TargetValueExclusionType::Dps && type != TargetValueExclusionType::Attacker)
```
brighton-chi, line 110: I need to fix this, but AppendTargetExclusions doesn't actually block bots from considering the excluded targets as valid for determining whether to AoE. So there would need to be a separate multiplier to block AoE actions against Golemagg I think.

src/Ai/Raid/MC/MCStrategy.cpp
```diff
@@ -84,3 +99,26 @@ void RaidMcStrategy::InitMultipliers(std::vector<Multiplier*>& multipliers)
...
+        return;
+
+    // Damage into these is wasted: Core Ragers are unkillable while Golemagg
+    // lives (full heal at 50%), and Majordomo reflects and cannot die; his
+    // encounter ends when the eight adds are dead. Tanks still pick them up
+    // through their own target selection.
+    AiObjectContext* context = botAI->GetAiObjectContext();
+    bool const golemaggAlive =
```
brighton-chi: You already got context so this can just use the AI_VALUE2 macro to find target

src/Ai/Raid/MC/MCTriggers.cpp
```diff
@@ -20,7 +21,41 @@ bool McBaronGeddonInfernoTrigger::IsActive()
 
 bool McShazzrahRangedTrigger::IsActive()
 {
-    return AI_VALUE2(Unit*, "find target", "shazzrah") && PlayerbotAI::IsRanged(bot);
+    // Only fire inside Arcane Explosion range. The move-away action no-ops
+    // beyond it, so an unconditional trigger makes every already-safe ranged
+    // bot attempt a failing move each tick.
+    Unit* boss = AI_VALUE2(Unit*, "find target", "shazzrah");
+    return boss && PlayerbotAI::IsRanged(bot) && bot->GetDistance2d(boss) < ARCANE_EXPLOSION_DISTANCE;
```
brighton-chi: Role checks are cheaper than FindTargetValue so it's better for them to come first.

src/Ai/Raid/MC/MCTriggers.cpp
```diff
@@ -39,6 +74,11 @@ bool McGolemaggIsAssistTankTrigger::IsActive()
     return AI_VALUE2(Unit*, "find target", "golemagg the incinerator") && PlayerbotAI::IsAssistTank(bot);
 }
 
+bool McGolemaggIsHealerTrigger::IsActive()
+{
+    return AI_VALUE2(Unit*, "find target", "golemagg the incinerator") && PlayerbotAI::IsHeal(bot);
```
brighton-chi: Same here re: role check

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +55,74 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+    // again — any other destination (the current target included) can path
+    // straight back into the pool.
+    if (!dryTarget)
+        return false;
+
+    bot->AttackStop();
+    bot->InterruptNonMeleeSpells(false);
+
+    if (!MoveTo(dryTarget->GetMapId(), dryTarget->GetPositionX(), dryTarget->GetPositionY(),
+                dryTarget->GetPositionZ(), false, false, false, true, MovementPriority::MOVEMENT_FORCED))
```
brighton-chi, line 99: Should add argument for 10th parameter passing true

Edit: Actually, I realized you're not doing the movement incrementally here, so actually it doesn't matter one way or another.

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -52,6 +55,74 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+        return false;
+
+    // Hold the AI so rotation/facing actions can't cancel the escape spline
+    // mid-lava. Capped at 2s on purpose: the in-lava trigger refires and
+    // re-issues the escape while the bot is still swimming.
+    float const speed = bot->GetSpeed(MOVE_RUN);
+    if (speed > 0.1f)
+        botAI->SetNextCheckDelay(std::min(2000u, uint32(1000.0f * bot->GetDistance(dryTarget) / speed)));
```
brighton-chi: Best to use constexpr locals for these numbers

brighton-chi (commented):
A few comments as I'm helping out the maintainers with reviewing strats

src/Ai/Raid/MC/MCMultipliers.cpp
Yuof: The action returning true just stops it re-issuing a move, it doesn't hold the bot out. Without blocking `AttackAction`/`MeleeAction` the rotation re-targets Golemagg next tick and chases back into swing range, so the stack never gets to expire. Deliberately distance-independent for the full 30s.

src/Ai/Raid/MC/MCStrategy.cpp
Yuof: Golemagg is already covered from the other side: `GolemaggMultiplier` (`MCMultipliers.cpp:119`) zeroes all dps AoE for the whole encounter while he's alive.

Skipping an equivalent for Majordomo on purpose, his adds are the kill condition and AoEing them is the intended play. The exclusion is only there to stop bots single-targeting Majordomo himself.

src/Ai/Raid/MC/MCMultipliers.cpp
brighton-chi: Once an action returns true, it ends the AI tick. In practice, very few actions are above ACTION_RAID (60) in priority. Basically it's just emergency stuff like Lay on Hands and some random things that don't need to be that high priority but are (like Life Tap). So whether or not an action that returns true stops the bot from acting depends on whether the action is conditioned (in the trigger or action itself) to no longer return true once the objective is reached.

Usually you would return false in such a situation and then block just the undesirable actions with a multiplier. If you legit don't want the bot to do anything (which maybe is the case here; I'm not familiar with the mechanics of the fight), then it's fine to unconditionally return true, but then you don't need a multiplier.

Yuof: I went back and tested both versions: 
Current version: 
- Trigger when close to boss and 20+ stacks -> queues movement
- Trigger every tick until we are far enough( No trigger while at range)
- Bot still can do non melee combat related actions: heals (pots), hybrid dps (shaman shock, paladin judgment, etc.)
- Action does not trigger during this just the multiplier suppresses any action that would result in refreshing the stacks

Action always triggers and returns true:
- Trigger when close to boss and 20+ stacks -> queues movement
- Trigger every tick until we lose stacks
- Bot can't do any actions that are lower in prio than the mechanic

src/Ai/Raid/MC/MCStrategy.cpp
```diff
@@ -90,3 +105,25 @@ void RaidMcStrategy::InitMultipliers(std::vector<Multiplier*>& multipliers)
...
+void RaidMcStrategy::AppendTargetExclusions(GuidSet& exclusions, TargetValueExclusionType type)
+{
+    if (type != TargetValueExclusionType::Dps && type != TargetValueExclusionType::Attacker)
+        return;
+
+    // Damage into these is wasted: Core Ragers are unkillable while Golemagg
+    // lives (full heal at 50%), and Majordomo reflects and cannot die; his
+    // encounter ends when the eight adds are dead. Tanks still pick them up
+    // through their own target selection.
+    AiObjectContext* context = botAI->GetAiObjectContext();
```
kadeshar, line 117: Unused

src/Ai/Raid/MC/MCActions.cpp
```diff
@@ -58,6 +61,75 @@ bool McShazzrahMoveAwayAction::Execute(Event /*event*/)
...
+    // rock shelves, so a dry teammate marks the closest safe floor.
+    Unit* dryTarget = nullptr;
+    float bestDist = 100.0f;
+
+    if (Group* group = bot->GetGroup())
+    {
+        for (GroupReference* ref = group->GetFirstMember(); ref; ref = ref->next())
+        {
+            Player* member = ref->GetSource();
+            if (!member || member == bot || !member->IsAlive() || member->GetMapId() != bot->GetMapId())
```
kadeshar, line 75: You checking mapId but not instanceId.

brighton-chi: We don't have that convention in any boss strategies, though. It would be super weird to have a group member that is in a different version of the same raid. They'd have to be locked to another version, and you would have had to have summoned them in. And it would be clear immediately that you screwed something up in that case since they wouldn't actually show up.

kadeshar: Fair. Ok, ignore that.

src/Ai/Raid/MC/MCStrategy.cpp
Yuof: It looks unused, but it isn't — it's what the `AI_VALUE` macros expand against:

```cpp
#define AI_VALUE(type, name)         context->GetValue<type>(name)->Get()
#define AI_VALUE2(type, name, param) context->GetValue<type>(name, param)->Get()
```

They name `context` unqualified, so it has to be a name in scope at the call site. In most files that's the `AiObjectContext* context` member `AiObject` provides, but `Strategy` derives from `PlayerbotAIAware`, not `AiObject`, so it has no such member — the local is the only thing the macro can bind to.

I did remove it once, on 2026-08-05, exactly on this reasoning. It failed CI on all five build jobs, and the local went back in.

kadeshar: Ok, my bad. Closing :)

kadeshar (approved):

PR #2578, opened 2026-07-17
fix: read COUNT(*) as uint64 to silence Field type warning

biship:
## Pull Request Description

`COUNT(*)` returns a LONGLONG (64-bit) column, but the account type check in `RandomPlayerbotFactory::CalculateTotalAccountCount()` read it with `Get<uint32>()`, causing the core to log a warning on every server startup:

```
Warning: Field::GetData<unsigned int> on LONGLONG field .COUNT(*) (.) at index 1.
> Please use GetData<uint64>()
```

This reads the field as `uint64` as the core asks (matching the identical query pattern already used two lines above, at line 337, in the same file), then narrows to `uint32` for storage in the existing `uint32` counters (`existingUnassignedAccounts`, `existingRndBotAccounts`, `existingAddClassAccounts`) that the rest of the function already uses. No behavioral change — the count of bot accounts of a given type always fits comfortably in 32 bits.

## Feature Evaluation

This is a one-line type-correctness fix, not new logic. It changes how a single already-executing query result is read (`Get<uint64>()` + `static_cast<uint32>` instead of `Get<uint32>()`); the query itself, its frequency, and the surrounding control flow are all unchanged. It runs once at server startup during bot account provisioning (`CalculateTotalAccountCount()`), not per-bot or per-tick, so there is no per-bot/per-tick cost at all.

## How to Test the Changes

1. Start a worldserver with mod-playerbots enabled and `AiPlayerbot.RandomBotAccounts` (or the relevant random-bot-account-creation config) turned on, on a `playerbots_account_type` table that already has rows (i.e. not a fresh empty install).
2. Observe the startup log during "Creating random bot accounts...".
3. **Before the fix:** the log shows `Warning: Field::GetData<unsigned int> on LONGLONG field .COUNT(*) (.) at index 1.` / `> Please use GetData<uint64>()`, once per distinct `account_type` value present in the table.
4. **After the fix:** the warning no longer appears, and account counts (`existingUnassignedAccounts`, `existingRndBotAccounts`, `existingAddClassAccounts`) are unchanged — bot account creation proceeds identically.

## Impact Assessment

    - - [X] No, not at all

    - - [X] No
    - - [ ] Yes (**explain why**)

    - - [X] No

## AI Assistance

- - [ ] No

Claude (Anthropic) was used to diagnose the root cause of a startup log warning observed while running a local worldserver, trace it to this specific `Get<uint32>()` call, and propose/apply the one-line fix. The change was reviewed and is fully understood: it is a narrow type-correctness fix consistent with the existing `Get<uint64>()` usage on the neighboring query in the same file.

 MaNGOS, another module)?
- - [X] No, all code in this PR is original

## Final Checklist

- - [X] Any new bot dialogue lines are translated. (N/A — no new dialogue)
- - [X] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note). (N/A — no ported code)
- - [X] New source files use the GPLv2 header. (N/A — no new files)
- - [X] New and modified files do not introduce new compiler warnings.

## Notes for Reviewers

Branched from and targeting `test-staging` per the contribution guide.

### DIFF
--- a/src/Bot/Factory/RandomPlayerbotFactory.cpp
+++ b/src/Bot/Factory/RandomPlayerbotFactory.cpp
@@ -353,7 +353,7 @@ uint32 RandomPlayerbotFactory::CalculateTotalAccountCount()
         {
             Field* fields = typeCheck->Fetch();
             uint8 accountType = fields[0].Get<uint8>();
-            uint32 count = fields[1].Get<uint32>();
+            uint32 count = static_cast<uint32>(fields[1].Get<uint64>());
 
             if (accountType == 0) existingUnassignedAccounts = count;
             else if (accountType == 1) existingRndBotAccounts = count;


NoxMax:
What is your OS and what is your config in world server conf for `Logger.sql.sql`?

biship:
Windows 11 x64
Logger.sql.sql=3,Console Errors

NoxMax:
> Logger.sql.sql=3,Console Errors

That explains things. The default value is 2, errors and crashes only, which is why most of us don't see this. You set it to 3, which is warning level. Regardless, your edit is valid. Thanks.

kadeshar (approved):

Issue #1200, opened 2025-04-13
Critical loot bug (cheat) in Halls of Stone


Regrad:
**Bug Description**
In the Halls of Stone dungeon, there is a chest from which bots can collect loot indefinitely.

**Commit Hash**
6ae96f19b96d36b59e9a8a9d3372bf9694085520

**How To Reproduce Bug**
Подвести ботов к сундуку, чтобы они могли открыть его.

**Expected Behavior**
Loot must be obtained at most once. Bots must not open the chest before completing the boss.

**Screenshots**

![Image](https://github.com/user-attachments/assets/94d23b38-1188-4af1-a85b-013eb18ad29c)

**Modules**
Transmog

**Playerbot Settings**
100+ players, 1000 bots

**System**
OS: Windows
Hardware: Xeon 22 cores

**Additional Context**
This way you can get an infinite amount of gold and emblems.

kadeshar:
You as player also can loot this chest indefinitely?

Regrad:
> You as player also can loot this chest indefinitely?

Yup

kadeshar:
Could you change this line and check that fix issue? 

![Image](https://github.com/user-attachments/assets/a4a1b60b-dbfa-4a64-a16d-6d4063f7d2be)

Regrad:
I haven't noticed this bug recently. It may have been fixed.

Regrad:
The bug in Halls of Stone can still be reproduced.

Regrad:
> Could you change this line and check that fix issue?
> 
> ![Image](https://github.com/user-attachments/assets/a4a1b60b-dbfa-4a64-a16d-6d4063f7d2be) and this
> 
> ![Image](https://github.com/user-attachments/assets/8477b497-1670-48ed-ac85-772dfb9c68d6)

Can you add new pull request, please? Or at least drop the difference file. Add changes from screenshots is not very convenient.

kadeshar:
> > Could you change this line and check that fix issue?
> > ![Image](https://github.com/user-attachments/assets/a4a1b60b-dbfa-4a64-a16d-6d4063f7d2be) and this
> > ![Image](https://github.com/user-attachments/assets/8477b497-1670-48ed-ac85-772dfb9c68d6)
> 
> Can you add new pull request, please? Or at least drop the difference file. Add changes from screenshots is not very convenient.

You tested that changes? They work?

Regrad:
[...]

Issue #1162, opened 2025-04-01
Bot's can loot chest before gunship event ends.


Jaccob7:
**Bug Description**
Bot's can open the chest in ICC before they gunship event is completed

**Commit Hash**
AzerothCore rev. 95fc3f0932c5+ 2025-03-29 17:25:52 +0800 (Playerbot branch) (Unix, RelWithDebInfo, Static)

**How To Reproduce Bug**
Go into the ICC raid and start the gunship event

**Expected Behavior**
They're not to supposed to be able to loot the chest until the event is over.

**Screenshots**
![Image](https://github.com/user-attachments/assets/affc7759-07f9-4d48-8b87-1ebb8212e7c0)

**Modules**
![Image](https://github.com/user-attachments/assets/5e0261ff-a7ac-41b9-814f-927a80961494)

**Playerbot Settings**

**System**
OS: Linux

kadeshar:
Its not playerbots issue. This issue is related with azerothcore.

Dreathean:
If you mention bots at all in their issues, they (honestly understandably) won't look at it and will say to post here. This issue would need to be replicated on a clean, non-playerbot server to confirm the issue is on the main core.

So the main question is, are you able to open the chest early or only the bots? If only the bots then that does make it clear it's a playerbot issue.

Jaccob7:
I'll test and report back

noisiver:
I'm trying to understand the issue more specifically. You say they can open the chest. You can't? The bots can and will interact with objects that the player can't, including stuff like doors, chests. If you can't open the chest then it's a playerbots issue because the bots shouldn't be able to interact with it to begin with.

Jaccob7:
From player

I'm unable to until the fight is over but they can get into it early I stopped them from doing it by making them stay but still attack and that stopped them last night

kadeshar:
I had similiar situation with hodir chest loot. But there i can loot chest as player.

parazitenew:
This is an old issue, it always existed and related to playerbots, any chest can be looted by bots before any event starts, not only in ICC.

PR #2579, opened 2026-07-17
fix(loot): let bots loot quest objects again

Yuof:
## Pull Request Description

Bots cannot loot **any** quest chest, and haven't been able to since #1280.

Example: a bot on *Filthy Paws* (quest 307) walks to a Miners' League Crate and stands on it:

```
chest 271: added false canLoot false empty false possible false
           skill 0 reqItem 0 dz 0.0 dist 0.0 goflags 4 heldFor 44757ms
→ wrote off quest target ... Entry: 271
```

Spawned, `GO_STATE_READY`, no skill and no key required, distance 0. The loot pipeline
refuses it every tick.

**Cause.** `GO_FLAG_INTERACT_COND` (goflags 4 above) means *interaction is conditional*. The
server never checks the flag because the **client** evaluates the condition, and for these objects the
condition is quest state. That is why the flag sits on quest objects generally:

| type | objects carrying the flag | condition |
|------|--------------------------|-----------|
| chest | 788 (764 with quest items) | quest loot / `chest.questId` |
| questgiver | 429 | quest dialog status |
| goober | 257 | `goober.questId` |
| spell focus / generic / … | ~50 | `questID` |

Two guards read the flag as "never touch": `LootObject::IsLootPossible()`, so `"can loot"`
never goes true, and `OpenLootAction::DoLoot()`, so the object is never opened. A bot has no
client and therefore never gets the exemption a player gets.

**Where the guards came from.** 97f582b9 (#1280) added both for excluding three chests, but only one of
them carries the flag:

| chest | addon flags | actually guarded by |
|-------|-------------|--------------------|
| Gunship Armory (×8) | 16 `NOT_SELECTABLE` | the next guard, already present |
| Deathbringer's Cache (×4) | 16 / 50 | the next guard, already present |
| Tribunal Chest (×2) | **4 `INTERACT_COND`** | this guard |

So the check was doing exactly one job (keeping bots out of the Tribunal Chest), and every
quest object in the game was collateral.

**Fix.** A bot has to make the call its client would, so ask the function the server uses to
make it: `GameObject::ActivateToQuest()` is what `BuildValuesUpdate` calls to set
`GO_DYNFLAG_LO_ACTIVATE` / `GO_DYNFLAG_LO_SPARKLE`, i.e. whether a player's client lights the
object up. It covers every type in the table above and goes false again once the objective is
satisfied.

`GO_FLAG_NOT_SELECTABLE` and the `isSpawned()` checks are untouched, so #1162 (Gunship Armory)
is unaffected: that chest is gated by `NOT_SELECTABLE`, which the ICC script removes when the
gunship event ends.

## Feature Evaluation

**Minimum logic:** one condition on each of the two guards that already test the flag:
`&& !go->ActivateToQuest(bot)`. No new helper, no new state, no new lookup path.

**Processing cost:** only objects carrying `GO_FLAG_INTERACT_COND` reach the call; everything
else short-circuits on the flag test exactly as before. For those, `ActivateToQuest()` is the
quest-log scan the server already performs per player per gameobject update in
`BuildValuesUpdate`. It adds one such call per loot evaluation of a conditional object in
loot range, which for a typical bot is none, occasionally one.

## How to Test the Changes

**A quest chest can be looted again.** Quest 307 has no kill objective, so it isolates the
chest path:

1. Put a level 15+ bot in Loch Modan at ~(-4936, -2967, 318), with the `loot` strategy.
2. Give it quest 307 *Filthy Paws*: 4x Chewed Bone from Miners' League Crates (GO 271; 10
   crates, unpooled, in a small cluster).
3. Before: the bot walks onto a crate and never opens it, 0/4 indefinitely. After: it opens
   them and completes the objective in a few minutes.

**The Tribunal Chest is still refused** (#1200 regression check):

1. Put a bot in Halls of Stone at the chest, ~(880, 345, 204).
2. Test twice: with no quest, and while holding *Ancient Relics* (12870, repeatable, wants the
   Relic of Ulduar that is in that chest's loot table).
3. Expected both times: `ActivateToQuest` false, `can loot` false, the chest never opens, no
   matter how long the bot stands there, and regardless of the Tribunal of Ages event state.

Measured with a bot parked at the object for 10 minutes, sampled every 30s:

| build | quest | activate | can loot | times looted | relics |
|-------|-------|----------|----------|--------------|--------|
| this PR | none | false | false | **0** | 0 |
| this PR | Ancient Relics | false | false | **0** | 0 |
| **guard removed entirely** (control) | none | n/a | n/a | **4** | **3** |

The control row is the point: remove the `INTERACT_COND` check and #1200 is back within 30
seconds. The chest is consumable with a 180s respawn, so a bot farms it roughly every 150s
with no quest at all. The guard is load-bearing; this PR keeps it and fixes only its condition.

The Tribunal Chest never activates because its Relic of Ulduar is **ordinary** loot
(`QuestRequired=0`), so `HaveQuestLootForPlayer()` is false even for a bot on the relic quest,
the same reason a player's client never lights that chest up. Quest chest items are
`QuestRequired=1`, so those activate for a bot on the quest, and stop activating once it has
enough.

## Impact Assessment

    - - [ ] No, not at all

Gated behind the existing `GO_FLAG_INTERACT_COND` test, so it only runs for conditional objects
in loot range, so for the vast majority of loot evaluations nothing changes at all. When it does
run, it is a single `ActivateToQuest()` call, the same one the core already makes per player
per gameobject update.

    - - [ ] No
    - - [x] Yes (**explain why**)

Bots can loot quest chests again, which is the point of the PR; they could before #1280. They
now touch exactly the conditional objects a player's client would offer them, and stop once the
objective is satisfied.

    - - [x] No

The same two guards, with one condition added to each. The combined
`INTERACT_COND | NOT_SELECTABLE` test in `IsLootPossible()` is split so the two flags can be
answered separately, which is why that hunk looks bigger than it is.

## AI Assistance

- - [ ] No

Used for the investigation and for drafting this description: reproducing the failure, tracing
it to the flag, checking how the core uses `GO_FLAG_INTERACT_COND` and `ActivateToQuest()`,
surveying which objects carry the flag, and running the before/after and control measurements
above. The change itself is two conditions and was reviewed and verified by me.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

- Not addressed here, but worth knowing while reviewing the control row: the Tribunal Chest is
  a plain DB spawn with a 180s respawn, no `chest.questId`, no `eventId`, no `conditions` rows
  and no instance-script handling, so nothing server-side gates it on the Tribunal of Ages
  event, and forcing the event to DONE changes nothing. Only the client-side condition keeps
  players out of it, which is why bots needed a guard in the first place. That looks like an
  AzerothCore data issue rather than a mod-playerbots one, and I have not touched it.

### DIFF
--- a/src/Ai/Base/Actions/LootAction.cpp
+++ b/src/Ai/Base/Actions/LootAction.cpp
@@ -140,8 +140,9 @@ bool OpenLootAction::DoLoot(LootObject& lootObject)
     if (go && (go->GetGoState() != GO_STATE_READY))
         return false;
 
-    // This prevents dungeon chests like Tribunal Chest (Halls of Stone) from being ninja'd by the bots
-    if (go && go->HasFlag(GAMEOBJECT_FLAGS, GO_FLAG_INTERACT_COND))
+    // This prevents dungeon chests like Tribunal Chest (Halls of Stone) from being ninja'd by the bots.
+    // Quest objects carry the same flag but are gated on quest state, which ActivateToQuest answers.
+    if (go && go->HasFlag(GAMEOBJECT_FLAGS, GO_FLAG_INTERACT_COND) && !go->ActivateToQuest(bot))
         return false;
 
     // This prevents raid chests like Gunship Armory (ICC) from being ninja'd by the bots
--- a/src/Mgr/Item/LootObjectStack.cpp
+++ b/src/Mgr/Item/LootObjectStack.cpp
@@ -308,7 +308,12 @@ bool LootObject::IsLootPossible(Player* bot)
     // Prevent bot from running to chests that are unlootable (e.g. Gunship Armory before completing the event) or on
     // respawn time
     GameObject* go = botAI->GetGameObject(guid);
-    if (go && (go->HasFlag(GAMEOBJECT_FLAGS, GO_FLAG_INTERACT_COND | GO_FLAG_NOT_SELECTABLE) || !go->isSpawned()))
+    if (go && (go->HasFlag(GAMEOBJECT_FLAGS, GO_FLAG_NOT_SELECTABLE) || !go->isSpawned()))
+        return false;
+
+    // Conditional objects (quest chests, goobers, ...) are gated client-side on quest state.
+    // A bot has no client, so make the same call the server makes for one.
+    if (go && go->HasFlag(GAMEOBJECT_FLAGS, GO_FLAG_INTERACT_COND) && !go->ActivateToQuest(bot))
         return false;
 
     if (skillId == SKILL_NONE)


Celandriel (approved):

Sakuragi85:
is there an option to disable this behavior? I dont want bots picking up my items.

Yuof:
if you don't want bots contesting your quests, IMHO the easiest way would be to disable bot questing (AiPlayerbot.RpgStatusProbWeight.DoQuest)

Celandriel:
> is there an option to disable this behavior? I dont want bots picking up my items.

While in a group with you, or overall? In a group with you setting loot to free for all disables their loot.

PR #2585, opened 2026-07-20
Honor EnableGreet at all native greeting call sites

DiNaSoR:
## What changed

Guard the four native Playerbots greeting call sites with the existing `AiPlayerbot.EnableGreet` configuration value.

## Why

The option is already exposed to administrators, but these login, invitation, and group-master paths emitted a greeting even when it was disabled. That creates duplicate/private greetings for modules that provide their own party greeting and makes the setting inconsistent.

## Impact

Default behavior is unchanged because `AiPlayerbot.EnableGreet` defaults to enabled. Servers that set it to `0` now suppress all four native greeting paths consistently.

## Validation

- Built AzerothCore `worldserver` successfully on Windows x64, RelWithDebInfo, static modules.
- Changes are limited to the four greeting call sites.


Celandriel:
Please Fill out the PR template

PR #2592, opened 2026-07-22
refactor: Amend IsRealPlayer and introduce IsSelfBot

NoxMax:
## Pull Request Description

1. `IsRealPlayer` used to be a check for selfbots. That check is now named `IsSelfBot`
2. `IsRealPlayer` is now a check for for a regular player without any bot AI. That was done with `HasActivePlayerMaster` which is now dropped.
3. `HasRealPlayerMaster` (which checked for `IsRealPlayer || IsSelfBot`) is now dropped and replaced with the more explicit `HasGameClientMaster`.
4. Some C++ bracket code standard changes were made.

## Feature Evaluation

Point # 1 is the main confusing thing that needed fixing, but otherwise the whole change doesn't affect performance.

## How to Test the Changes

Everything should compile. You should summon any form of bots without issue, and if you convert your character to a selfbot there should also be no issue.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Search for relevant places in the codebase regarding the status of selfbots.

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Base/Actions/AcceptDuelAction.cpp
+++ b/src/Ai/Base/Actions/AcceptDuelAction.cpp
@@ -19,7 +19,7 @@ bool AcceptDuelAction::Execute(Event event)
     p >> playerGuid;
 
     // Do not auto duel with low health
-    if ((!(IsRealPlayer(botAI->GetMaster()) || IsSelfBot(botAI->GetMaster())) ||
+    if ((!botAI->HasGameClientMaster() ||
         (botAI->GetMaster() && botAI->GetMaster()->GetGUID() != playerGuid)) &&
         AI_VALUE2(uint8, "health", "self target") < 90)
     {
--- a/src/Ai/Base/Actions/AcceptDuelAction.cpp
+++ b/src/Ai/Base/Actions/AcceptDuelAction.cpp
@@ -18,8 +18,9 @@ bool AcceptDuelAction::Execute(Event event)
     ObjectGuid playerGuid;
     p >> playerGuid;
 
-    // do not auto duel with low hp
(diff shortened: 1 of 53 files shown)


NoxMax:
Going into this, I really didn't care for `HasRealPlayerMaster` and wanted to inline it to clearly `IsRealPlayer || IsSelfBot`. On Discord that was the consensus as well. After applying it and looking at the bracketing situation, I don't feel as strongly that this makes things simpler.

I will keep the inline now, but if others think the bracketing is now clunky, then I though of the term `HasGameClientMaster`, which would unambiguously cover a master from a real human side, whether they are a regular player or a selfbot.

brighton-chi:
Maybe the IsAltBot() definition could use a comment since the inlining does make that one look confusing (bot is not a randombot or a selfbot, and bot has a master that is either a selfbot or a not selfbotting person).

Although is it even possible to have a master that is not one of those?

NoxMax:
Honestly the confusion might be speaking some merit for adding `HasGameClientMaster`, still it should be clear now. Also added the point that an addclassbot is also (by design) considered an altbot, but the inverse is not necessarily true.

NoxMax:
Resolved merged conflicts and added HasGameClientMaster for IsRealPlayer || IsSelfBot. It's honestly cleaner, and that's the whole point of the PR. And of course it much more obvious than the old `HasRealPlayerMaster`. But if people still think inline is best, I will revert it.

kadeshar (approved):

NoxMax:
Merge looks clean as far as the co-PRs merged along in test-staging. None of them use IsRealPlayer or HasRealPlayerMaster.

Issue #893, opened 2025-01-23
Bots keep spawning out of bounds in twisting nether and other places


Abnormalpingu:
**Bug Description**
I've been seeing this issue after I setup the online playermap (https://www.azerothcore.org/catalogue.html#/details/230741658), just to see where the bots are (yes it's not 100% accurate but it gives somewhat of an idea). The playermap is just a php site that queries the database for player locations, it doesn't seem to update over time though. 

Its a bit random how many bots ends up in these locations but anywhere from 5-10 after a server restart. There's also some bots stuck outside various other locations in kalimdor (desolace along the coast and directly south of un'goro). 
Eastern kingdom (silverpine forest - south east along the coast towards The Greymane Wall). 
Normally i just invite them to my group and summon them to move them. But it seems there's some out of bounds spawn rules that need to me made. 

**Commit Hash**
:~/azerothcore-wotlk# git show
commit 36987b8a964a84b4981f97198eab0c58c9fc1c5d (HEAD -> Playerbot, origin/Playerbot, origin/HEAD)
Merge: 8e5802da4 42e0ab5a5

**How To Reproduce Bug**
Restart server and verify location in playermap or /who ingame

**Expected Behavior**
Bots shouldn't spawn outside the map and also in inaccessible places. Maybe there should be some sort of check that if bot ends up within unexpected coordinates they get teleported to the nearest graveyard. 

**Screenshots**
https://imgur.com/a/GAWak8s

**Modules**
mod-ahbot
mod-playerbots
mod-solocraft

**Playerbot Settings**
500

**System**
OS: debian 12
Hardware: virtual machine, 10 core, 16gb ram

**Additional Context**
Thank you for the work you're doing on this mod, it's really really cool ! :D

privatecore:
Is anyone else experiencing this problem? You can turn on logging for the bot's location (bot_location.csv): https://github.com/liyunfan1223/mod-playerbots/blob/5a0c27637e1d2b450637df3909d84e9f6e3b9199/conf/playerbots.conf.dist#L2040 It'll be interesting to see the actual coordinates for those bots.

PR #2597, opened 2026-07-24
fix random bots getting stuck outside the map

darksoon:
## Pull Request Description

been dealing with this on my own server too (~1000 bots), same as reported in #893 - bots slowly pile up in spots like twisting nether or just off the coast in random zones and never move again since nothing ever picks them for anything.

couldn't figure out where exactly they get placed there in the first place, but instead of chasing that down I just added a check that runs on a handful of bots each tick (rotates through everyone so it doesn't add up even with a lot of bots online). it looks at whether there's solid ground under the bot and whether the zone/area actually resolves for their coords. if not, they get sent to the nearest graveyard for their team instead of sitting there forever.

instances are skipped on purpose - dungeon/raid geometry looks "wrong" to a flat ground-height check all the time and that's not what this is about, only open world bots.

been running it for a bit and haven't seen any of the previously stuck bots reappear. not a root cause fix, just stops the symptom, so feel free to close this out whenever the actual cause gets found instead.

## Feature Evaluation

**Minimum logic required:** per checked bot it's three lookups and a comparison - `GetZoneId`/`GetAreaId` (both resolve through the already-loaded AreaTable DBC) and one `Map::GetHeight`. Bots that are not in world, mid-teleport or in combat are skipped before any of that runs. Nothing is stored per bot, no new state, no timers.

**Processing cost across many bots:** the batch size is fixed (`maxChecksThisTick = 15`), and it rotates through the online bots with a cursor, refilling the list once it has wrapped around. So the per-tick cost is **constant whether 50 or 1000 bots are online** - more bots only means each individual bot is revisited less often, which is fine because this is a slow-drift problem, not something that needs to be caught within a second. This runs on the RandomPlayerbotMgr tick, not per bot AI tick.

## How to Test the Changes

1. Run a realm with random bots for a while (the longer the uptime, the more reliable - stuck bots accumulate slowly).
2. Before the change: `.gm on`, then look for bots in Twisting Nether or off the coast of random zones, e.g. via a query on `characters.position_x/y/z` for coordinates that resolve to no zone. They will be standing still indefinitely and never get picked for anything.
3. Apply the change and restart.
4. Watch the server log for lines mentioning the teleport to the nearest graveyard - on a realm that has been up for a while these appear during the first minutes as the rotation walks through everyone.
5. Verify those bots resume normal activity (they get picked up by the usual random bot logic again once they are on valid ground).

On my realm 14 bots were recovered within the first pass after startup, and none of the previously stuck ones came back over the following days.

## Impact Assessment

    - - [ ] No, not at all

Fixed cost of 15 checks per manager tick regardless of bot count, each check being two DBC lookups and one height query, with in-combat/in-transit bots skipped beforehand. It does not scale with the number of bots - only the revisit interval per bot does.

    - - [ ] No
    - - [x] Yes (**explain why**)

Only for bots that are already broken: a bot standing at coordinates with no resolvable zone/area or no ground beneath it gets teleported to its team's nearest graveyard. Bots at valid positions are never touched. Instances are excluded entirely, since dungeon and raid geometry regularly fails a flat ground-height check and that is not the case being targeted here.

    - - [x] No

It's one self-contained function called from the existing manager tick, no new strategies, triggers or values.

## AI Assistance

- - [ ] No

Yes - I used an AI assistant for narrowing down the problem, for writing the code, and for this description. I reviewed the change and I understand what it does and why; the behaviour was verified on my own realm (~1000 bots) rather than only reasoned about - the 14 recovered bots mentioned above are an actual observation, not an estimate.

What I explicitly do **not** claim to know is the root cause: I never found where the bots get placed at those coordinates in the first place. This is a symptom fix and I'd rather say that openly than present it as more than it is.

## Final Checklist

(no new bot dialogue and no new source files in this PR - the change lives in the existing RandomPlayerbotMgr, so those two points don't apply.)

## Notes for Reviewers

The batch size is a hardcoded default parameter (`ValidateBotPositions(uint32 maxChecksThisTick = 15)`) rather than a config option - happy to move it into the conf if you'd prefer it tunable, I just didn't want to add a config entry for something most realms shouldn't need to touch.

The "drifted from ground" part of the condition exists because a bot standing perfectly still far below or above the terrain is not caught by a plain "is there ground here" query - that was the case for the ones sitting under the world on my realm.


Celandriel:
Use the PR template

src/Bot/RandomPlayerbotMgr.h
```diff
@@ -99,6 +99,11 @@ class RandomPlayerbotMgr : public PlayerbotHolder
 
     void LogPlayerLocation();
     void UpdateAIInternal(uint32 elapsed, bool minimal = false) override;
+    // RyanWoW fix (GH#893 - "Out-of-Bounds Bot Spawning"): periodically rescue bots that
+    // ended up at invalid world coordinates (Twisting Nether, map edges, underground, ...)
+    // by teleporting them to the nearest graveyard. Runs a small batch each tick to keep
+    // the cost bounded regardless of total bot count.
+    void ValidateBotPositions(uint32 maxChecksThisTick = 15);
```
NoxMax, line 107: No need to add a comment in the header that just repeats what is said in the .cpp

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -173,6 +174,78 @@ botPIDImpl::~botPIDImpl() {}
 
 uint32 RandomPlayerbotMgr::GetMaxAllowedBotCount() { return GetEventValue(0, "bot_count"); }
 
+// RyanWoW fix for GH mod-playerbots#893 ("Out-of-Bounds Bot Spawning"): bots occasionally
+// end up at invalid coordinates after a restart (Twisting Nether, ocean edges, underground).
+// We validate a small rotating batch of online bots each tick and teleport any bot whose
+// current position has no valid ground beneath it (or sits far below/above the terrain,
+// which VMAP/MMAP height queries can't otherwise catch for a bot standing still) to the
+// nearest graveyard for their team - the same remedy the bug reporter suggested upstream.
+void RandomPlayerbotMgr::ValidateBotPositions(uint32 maxChecksThisTick)
```
NoxMax, line 183: We don't really write verbose comments explaining the issue on GH and who fixed it and the entire context. For any highlighted code in the editor, it will point to the originating PR anyway which will have the full details. For here, the comment can be condensed to one or two lines concisely explaining what this is for.

src/Bot/RandomPlayerbotMgr.cpp
```diff
@@ -173,6 +174,78 @@ botPIDImpl::~botPIDImpl() {}
...
+        float x = bot->GetPositionX();
+        float y = bot->GetPositionY();
+        float z = bot->GetPositionZ();
+
+        AreaTableEntry const* zone = sAreaTableStore.LookupEntry(map->GetZoneId(bot->GetPhaseMask(), x, y, z));
+        AreaTableEntry const* area = sAreaTableStore.LookupEntry(map->GetAreaId(bot->GetPhaseMask(), x, y, z));
+
+        float ground = map->GetHeight(bot->GetPhaseMask(), x, y, z + 2.0f, true, 50.0f);
+        bool noValidGround = (ground <= INVALID_HEIGHT);
+        bool driftedFromGround = !noValidGround && std::fabs(z - ground) > 40.0f && !bot->IsFlying() && !bot->GetTransport();
```
NoxMax, line 231: I feel like this is going to catch a lot of false positives. Water surfaces least of all where it could be valid that the bot is there, but because ground is calculated to the ocean floor, this kicks in an marks them in an invalid position. Bots relationship with mmaps is tenacious at best, and this asks a lot of that relationship.

Right now there are some existing guards that would teleport a bot out of a place if they stay idle in there for some time https://github.com/mod-playerbots/mod-playerbots/blob/3fa1c1e49f8f1324b72461e576bce7c89b0a6521/src/Bot/RandomPlayerbotMgr.cpp#L1446. They don't deal with these situations right as they happen, but they are a safe approach.

NoxMax:
And as Celandriel said, please use the PR template and target test-staging instead of master.

darksoon:
Sorry for the noise - template filled in and the base is now test-staging as asked. Should be good for another look.

kadeshar:
Conflict to resolve

PR #2600, opened 2026-07-25
Prevent silent random bot account creation failures

demansou:
## Pull Request Description

Random bot account creation could silently fail when the configured prefix plus its decimal suffix exceeded
[AzerothCore's account limits](https://github.com/mod-playerbots/azerothcore-wotlk/blob/test-staging/src/server/game/Accounts/AccountMgr.h#L23-L35).
Account names are limited to 17 characters, and passwords to 16. When
`AiPlayerbot.RandomBotRandomPassword = 0`, the generated account name is also used as the password, so the effective
generated-name limit is 16 characters.

The current loop ignores the result of `AccountMgr::CreateAccount`, increments the creation count before the call,
and logs success unconditionally. During local load-test setup, `codexloadbot20260723` produced no bot accounts
because even `codexloadbot202607230` is 21 characters. The shorter `codexlt23` prefix worked, but startup did not
identify the invalid generated name.

This change:

- collects the account names that are actually missing;
- validates the longest missing generated name before creating any accounts, including suffix digit boundaries;
- reports the applicable account-name or password limit with an actionable configuration error;
- checks every `CreateAccount` result and counts only successful calls;
- drains prior asynchronous inserts before aborting character creation after an unexpected failure; and
- documents the generated-name limits next to `AiPlayerbot.RandomBotAccountPrefix`.

Existing accounts are not rejected merely because the password mode changed; validation applies only to names that
would actually be created. Generated names are rejected rather than truncated to avoid collisions and unsafe prefix
ownership ambiguity.

## Feature Evaluation

  Reuse the existing startup account-existence scan to collect missing names, check the longest missing name against
  AzerothCore's account/password limits, and handle the existing `CreateAccount` return value.

  No per-bot or per-tick logic is added. This runs only during random-bot account initialization. It performs the same
  account-existence queries as before and temporarily stores one short string per missing account.

## How to Test the Changes

Use disposable databases or take a snapshot first because these tests create accounts. Start with no matching
random-bot accounts. Set `AiPlayerbot.MaxRandomBots = 0` and `AiPlayerbot.AddClassAccountPoolSize = 0` so the explicit
`AiPlayerbot.RandomBotAccountCount` drives these focused cases.

| Random password | Account count | Prefix | Longest generated name | Expected result |
|---|---:|---|---|---|
| `0` | `1` | `rndprefix123456` | `rndprefix1234560` (16) | Account creation succeeds |
| `0` | `1` | `rndprefix1234567` | `rndprefix12345670` (17) | Reports the 16-character password limit; creates no pending accounts |
| `1` | `1` | `rndprefix1234567` | `rndprefix12345670` (17) | Account creation succeeds |
| `1` | `1` | `rndprefix12345678` | `rndprefix123456780` (18) | Reports the 17-character account-name limit; creates no pending accounts |
| `0` | `10` | `rndprefix123456` | `rndprefix1234569` (16) | Account creation succeeds |
| `0` | `11` | `rndprefix123456` | `rndprefix12345610` (17) | Rejects before creation, covering the `9` → `10` suffix boundary |

For each case, start `worldserver` and verify valid cases create the expected accounts. For invalid cases, verify the
log includes the generated name, actual length, applicable limit, and corrective configuration guidance, and that
random-bot character creation does not begin.

Finally, create a valid 17-character account with random passwords enabled, then disable random passwords and restart
with no accounts missing. Existing account initialization should not be rejected.

Checks run locally:

- `python3 apps/codestyle/codestyle-cpp.py`
- AzerothCore parent-tree C++ codestyle check
- `git diff --check`

The module has no registered unit-test target, and its CI does not enable `BUILD_TESTING`. Adding useful automated
coverage would require new module CMake/CI plumbing and a new public test seam, so no unit test is included in this
startup-only fix. Cross-platform compilation is left to this draft PR's existing CI.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

  Valid default configurations behave as before. Invalid configurations now stop with an actionable error.

    - - [ ] No

  It adds startup-only validation and error-result branches. They do not execute in bot AI or update loops.

## AI Assistance

- - [ ] No

Codex was used to trace the silent failure through the module and AzerothCore account manager, inspect the relevant
Git history and open PRs, draft the implementation and documentation, and perform a second-pass review. The submitted
diff was reviewed against the current source, core limits, asynchronous database behavior, project style rules, and
the boundary cases listed above.

 MaNGOS, another module)?

## Final Checklist

No new source files, bot dialogue, translations, or ported code are included. The macOS, Windows, and Ubuntu
cross-platform CI builds passed.

## Notes for Reviewers

This is a latent error path rather than a regression from the recent account-count work:

- The unchecked account-creation loop dates to #971 (`624e4806`, February 2025).
- #2354 (`ccce1423`) only migrated the call to the `AccountMgr` singleton and preserved the ignored result.
- #2438 (`58b67038`) added user-error handling for insufficient `RandomBotAccountCount`, but does not handle account creation
  failures.

Open PR #2284 also changes this account-selection/creation area for realm-scoped ownership and currently retains the
unchecked `CreateAccount` call. This focused fix may need a straightforward rebase or equivalent handling folded into
that PR if #2284 lands first.

Malformed UTF-8 and empty/SQL-wildcard prefixes are intentionally outside this PR. Prefixes also participate in
`LIKE`-based ownership/deletion logic, so those cases need a broader safety design rather than being mixed into the
length/error-handling fix.

### DIFF
--- a/src/Bot/Factory/RandomPlayerbotFactory.cpp
+++ b/src/Bot/Factory/RandomPlayerbotFactory.cpp
@@ -21,6 +21,33 @@
 #include "SharedDefines.h"
 #include "SocialMgr.h"
 #include "Timer.h"
+#include "Util.h"
+
+namespace
+{
+char const* GetAccountCreationError(AccountOpResult result)
+{
+    switch (result)
+    {
+        case AOR_OK:
+            return "no error";
+        case AOR_NAME_TOO_LONG:
+            return "account name is too long";
+        case AOR_PASS_TOO_LONG:
+            return "password is too long";
+        case AOR_EMAIL_TOO_LONG:
+            return "email is too long";
+        case AOR_NAME_ALREADY_EXIST:
+            return "account name already exists";
+        case AOR_NAME_NOT_EXIST:
+            return "account name does not exist";
+        case AOR_DB_INTERNAL_ERROR:
+            return "database error";
+    }
+
+    return "unknown error";
+}
+}
 
 constexpr RandomPlayerbotFactory::NameRaceAndGender RandomPlayerbotFactory::CombineRaceAndGender(uint8 race,
                                                                                                 uint8 gender)
@@ -595,8 +622,9 @@ void RandomPlayerbotFactory::CreateRandomBots()
 
     LOG_INFO("playerbots", "Creating random bot accounts...");
     std::unordered_map<NameRaceAndGender, std::vector<std::string>> nameCache;
-    std::vector<std::future<void>> account_creations;
+    std::vector<std::string> accountNamesToCreate;
     int account_creation = 0;
+    bool accountCreationFailed = false;
 
     // Calculates the total number of required accounts.
     uint32 totalAccountCount = CalculateTotalAccountCount();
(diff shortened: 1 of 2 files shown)


kadeshar:
Conflict to resolve

src/Bot/Factory/RandomPlayerbotFactory.cpp
```diff
@@ -611,11 +639,38 @@ void RandomPlayerbotFactory::CreateRandomBots()
...
+        std::size_t const longestAccountNameLength = utf8length(accountNameForLength);
+
+        if (longestAccountNameLength > MAX_ACCOUNT_STR)
         {
-            continue;
+            LOG_ERROR("playerbots",
+                "Cannot create random bot accounts: generated account name '{}' is {} characters, exceeding the "
+                "{}-character account name limit. Shorten AiPlayerbot.RandomBotAccountPrefix.",
+                longestAccountName, longestAccountNameLength, MAX_ACCOUNT_STR);
+            return;
```
kadeshar, line 658: That leaves sPlayerbotAIConfig.randomBotAccounts empty even if there is one account to create and fail validation and 5000 other accounts is existing?

src/Bot/Factory/RandomPlayerbotFactory.cpp
```diff
@@ -21,6 +21,33 @@
...
+        case AOR_NAME_TOO_LONG:
+            return "account name is too long";
+        case AOR_PASS_TOO_LONG:
+            return "password is too long";
+        case AOR_EMAIL_TOO_LONG:
+            return "email is too long";
+        case AOR_NAME_ALREADY_EXIST:
+            return "account name already exists";
+        case AOR_NAME_NOT_EXIST:
+            return "account name does not exist";
```
kadeshar, line 43: Create return such code?

PR #2604, opened 2026-07-26
feat(ub): add Underbog (tbc-ub) dungeon strategy for Hungarfen

jrad7:
## Pull Request Description

Adds a combat strategy for the Hungarfen fight in The Underbog.

Hungarfen periodically spawns an Underbog Mushroom under a random party member.
The mushroom is harmless while it grows but aggroes whoever it spawned on, so
every stock target value treats it as a live add: bots peel onto it, the tank
taunts it, and dps flip into aoe rotations. Ten Grow stacks later it blooms into
a Spore Cloud and the group is standing in it. At 20% health the boss roots
itself and health-leeches everyone within 20yd for 11 seconds; by default the
whole group stands in that and dies.

The strategy:

- Ignores mushrooms entirely on both difficulties. Party dps is too low to make
  killing them worthwhile even on Normal, so the group burns the boss and steps
  out of blooms instead. This runs on two fronts: strategy target exclusions
  keep dps/tank/attacker selection off them, and a multiplier suppresses dps aoe
  rotations while the boss is up so the inflated attacker count cannot flip bots
  into splashing the adds.
- Vacates a Spore Cloud pre-emptively. Grow stacks are an exact countdown (one
  every 2s, bloom on the tenth), so at 8 stacks the bot already moves. A tank who
  is holding the boss walks him out of the cloud rather than fleeing, so the boss
  does not sit in it either.
- Retreats every bot, tank included, past the Foul Spores leech radius and holds
  there for the 11s. There is nothing to tank while he is rooted. The retreat
  fans across candidate lanes so the run out never crosses a spore cloud and
  never parks the bot next to a mushroom that is about to bloom.

It follows the existing per-dungeon pattern, so it is inert everywhere except
inside The Underbog.

## Feature Evaluation

  Two triggers with their actions, two multipliers, and one target-exclusion
  hook, installed only inside The Underbog. Every one of them leads with the
  threat-list `find target` check for the boss, so outside the Hungarfen fight
  they cost one cached lookup and exit. The single grid search (nearby
  mushrooms) lives behind a short-interval per-bot value, so the trigger, the
  action and the exclusion hook share one search instead of repeating it.

  Nothing runs outside The Underbog, so the open-world population is unaffected.
  Inside, everything before the boss is a cached `find target` miss. During the
  fight the per-bot cost is one 40yd creature search on a short interval plus
  short loops over that list; the Foul Spores retreat scoring runs only for the
  ~11s the aura is up.

## How to Test the Changes

With a group covering tank, healer and dps, clear to Hungarfen and pull.

Expected:

- No bot ever attacks or taunts an Underbog Mushroom, and dps do not fire aoe
  rotations at them.
- Bots step out of a mushroom before it blooms, not after it starts ticking.
- A tank holding the boss walks the boss out of a cloud instead of running away
  from him.
- At 20% the whole group, tank included, walks out past 20yd and holds until the
  aura drops, and the run out does not path through a spore cloud.
- Ranged keep casting on the boss from outside the leech radius during Foul
  Spores.

## Impact Assessment

  with thousands of bots?
  - [x] Minimal impact

  The added work exists only for bots inside The Underbog, and inside that
  instance it is one cached boss lookup per trigger/multiplier until Hungarfen
  is engaged. It does not scale with global bot count.

  - [x] Yes

  During the Hungarfen fight it overrides targeting (mushrooms are excluded and
  aoe is suppressed for dps) and movement (vacate the cloud, retreat from Foul
  Spores), because the default behavior wipes the group. The override is
  confined to that fight and mirrors the existing per-dungeon strategies.

  - [x] Yes

  It adds a self-contained `Ai/Dungeon/UB/` unit plus one registration line each
  in the existing dungeon strategy/action/trigger/value context files. The
  complexity is isolated to the new directory.

## AI Assistance

- [x] Yes

  Used for code generation, iteration on the retreat/vacate geometry, and
  documentation. The behavior was refined against the live fight, and all logic
  has been reviewed and is owned by the contributor.

## Final Checklist

- [x] Any new bot dialogue lines are translated. (N/A: no dialogue added.)
- [x] Any code ported/adapted from another project is attributed. (N/A.)
- [x] Documentation updated if needed. (N/A: no config or commands added.)

### DIFF
--- a/src/Ai/Dungeon/UB/UBActions.cpp
+++ b/src/Ai/Dungeon/UB/UBActions.cpp
@@ -1,0 +1,69 @@
+/*
+ * This file is part of the mod-playerbots module for AzerothCore. See AUTHORS file for Copyright
+ * information; released under GNU GPL v2 license, redistribute/modify under version 2 of the License,
+ * or (at your option) any later version.
+ */
+
+#include "UBActions.h"
+#include "Playerbots.h"
+#include "UBShared.h"
+
+#include <cmath>
+
+using namespace UnderbogHungarfen;
+
+bool UBRetreatFromFoulSporesAction::Execute(Event /*event*/)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
+    if (!boss)
+        return false;
+
+    float const safeDistance = MaxEffectRadius(SPELL_FOUL_SPORES, FOUL_SPORES_RADIUS_FALLBACK) + FOUL_SPORES_MARGIN;
+    float const currentDistance = bot->GetDistance2d(boss);
+    if (currentDistance >= safeDistance)
+        return false;
+
+    float const moveDist = safeDistance - currentDistance + 1.0f;
+    float const awayAngle = boss->GetAngle(bot);
+
+    GuidVector const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
+
+    for (float delta : { 0.0f, float(M_PI / 8), float(-M_PI / 8), float(M_PI / 4), float(-M_PI / 4),
+                         float(3 * M_PI / 8), float(-3 * M_PI / 8), float(M_PI / 2), float(-M_PI / 2) })
+    {
+        float const angle = awayAngle + delta;
+        float dx = bot->GetPositionX() + cos(angle) * moveDist;
+        float dy = bot->GetPositionY() + sin(angle) * moveDist;
+        float dz = bot->GetPositionZ();
+        if (!bot->GetMap()->CheckCollisionAndGetValidCoords(bot, bot->GetPositionX(), bot->GetPositionY(),
+                                                            bot->GetPositionZ(), dx, dy, dz))
+            continue;
+
+        if (RetreatPathUnsafe(bot, mushrooms, dx, dy))
+            continue;
+
+        if (MoveTo(bot->GetMapId(), dx, dy, dz, false, false, true, true,
+                   MovementPriority::MOVEMENT_COMBAT))
+            return true;
+    }
+
+    return MoveAway(boss, moveDist);
+}
+
+bool UBVacateSporeCloudAction::Execute(Event /*event*/)
+{
+    float const dangerRange = MushroomDangerRange(bot);
+    GuidVector const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
+    Creature* mushroom = GetNearestDangerousMushroom(bot, mushrooms, dangerRange);
+    if (!mushroom)
+        return false;
+
+    Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
+    if (botAI->IsTank(bot) && boss && boss->GetVictim() == bot)
+    {
+        float const currentDistance = bot->GetDistance2d(mushroom);
+        return MoveAway(mushroom, dangerRange - currentDistance + 2.0f);
+    }
+
+    return FleePosition(mushroom->GetPosition(), dangerRange);
+}
--- a/src/Ai/Dungeon/UB/UBActions.cpp
+++ b/src/Ai/Dungeon/UB/UBActions.cpp
@@ -6,8 +6,8 @@
 
 #include "UBActions.h"
 #include "Playerbots.h"
+#include "Timer.h"
 #include "UBShared.h"
-
 #include <cmath>
 
 using namespace UnderbogHungarfen;
@@ -66,4 +66,60 @@ bool UBVacateSporeCloudAction::Execute(Event /*event*/)
     }
 
     return FleePosition(mushroom->GetPosition(), dangerRange);
+}
+
+bool UBClearUnderbatBackAction::Execute(Event /*event*/)
+{
+    if (!bot->IsAlive() || botAI->IsTank(bot))
+        return false;
+
+    GuidVector const& attackers = AI_VALUE_REF(GuidVector, "attackers");
+    Creature* bat = GetNearestUnderbatInLashRange(bot, attackers);
+    if (!bat)
+        return false;
+
+    bool const throttled = _lastReposition && GetMSTimeDiffToNow(_lastReposition) < UNDERBAT_REPOSITION_COOLDOWN;
+
+    bool const melee = botAI->IsMelee(bot);
+    if (melee)
+    {
+        Unit* rally = UnderbatRallyUnit(bot, attackers);
+        bool const parked = rally && bot->GetExactDist(rally) <= UNDERBAT_RALLY_TOLERANCE;
+
+        if (rally && !parked)
+        {
+            if (throttled)
+                return false;
+
+            float const rx = rally->GetPositionX();
+            float const ry = rally->GetPositionY();
+            float const rz = rally->GetPositionZ();
+            if (SpotClearOfUnderbats(bot, attackers, rx, ry, rz) &&
+                MoveTo(bot->GetMapId(), rx, ry, rz, false, false, true, true,
+                       MovementPriority::MOVEMENT_COMBAT))
+            {
+                _lastReposition = getMSTime();
+                return true;
+            }
+        }
+
+        if (!GetLashingUnderbat(bot, attackers))
+            return false;
+    }
+    else if (bot->IsNonMeleeSpellCast(true) && !GetLashingUnderbat(bot, attackers))
+        return false;
+
+    if (throttled)
+        return false;
+
+    float const clearDistance = UNDERBAT_LASH_RANGE + UNDERBAT_LASH_MARGIN;
+    float const gap = bot->GetDistance2d(bat);
+    if (gap >= clearDistance)
+        return false;
+
+    if (!MoveAway(bat, clearDistance - gap + 1.0f))
+        return false;
+
+    _lastReposition = getMSTime();
+    return true;
 }
--- a/src/Ai/Dungeon/UB/UBActions.cpp
+++ b/src/Ai/Dungeon/UB/UBActions.cpp
@@ -26,7 +26,7 @@ bool UBRetreatFromFoulSporesAction::Execute(Event /*event*/)
     float const moveDist = safeDistance - currentDistance + 1.0f;
     float const awayAngle = boss->GetAngle(bot);
 
-    GuidVector const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
+    auto const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
 
     for (float delta : { 0.0f, float(M_PI / 8), float(-M_PI / 8), float(M_PI / 4), float(-M_PI / 4),
                          float(3 * M_PI / 8), float(-3 * M_PI / 8), float(M_PI / 2), float(-M_PI / 2) })
@@ -53,16 +53,19 @@ bool UBRetreatFromFoulSporesAction::Execute(Event /*event*/)
 bool UBVacateSporeCloudAction::Execute(Event /*event*/)
 {
     float const dangerRange = MushroomDangerRange(bot);
-    GuidVector const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
+    auto const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
     Creature* mushroom = GetNearestDangerousMushroom(bot, mushrooms, dangerRange);
     if (!mushroom)
         return false;
 
-    Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
-    if (botAI->IsTank(bot) && boss && boss->GetVictim() == bot)
+    if (PlayerbotAI::IsTank(bot))
     {
-        float const currentDistance = bot->GetDistance2d(mushroom);
-        return MoveAway(mushroom, dangerRange - currentDistance + 2.0f);
+        Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
+        if (boss && boss->GetVictim() == bot)
+        {
+            float const currentDistance = bot->GetDistance2d(mushroom);
+            return MoveAway(mushroom, dangerRange - currentDistance + 2.0f);
+        }
     }
 
     return FleePosition(mushroom->GetPosition(), dangerRange);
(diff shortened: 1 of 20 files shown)


jrad7:
Hungarfen is a 100% wipe on heroic with no combat strat and a coin flip on normal with a full bot group. This should get most groups past him.

src/Ai/Dungeon/UB/UBMultipliers.cpp
```diff
@@ -0,0 +1,49 @@
...
+        dynamic_cast<AttackAction*>(action))
+        return 1.0f;
+
+    if (dynamic_cast<MovementAction*>(action) || dynamic_cast<CastReachTargetSpellAction*>(action))
+        return 0.0f;
+
+    return 1.0f;
+}
+
+float HungarfenMushroomIgnoreMultiplier::GetValue(Action* action)
```
brighton-chi, line 35: I don't think you need this multiplier for anything if you have the target exclusion in the strategy.

brighton-chi (commented):
Cool, looking forward to giving it a try. Currently, beating him is an aggravating clown fest of running around and cycling bots through stay/follow/flee.

src/Ai/Dungeon/UB/UBMultipliers.cpp
jrad7: the exclusion is stopping them from targeting the mushroom, but this prevents them from considering mushrooms as valid AOE/Cleave targets.

brighton-chi: Oh interesting. I looked into this and it seems like it's a gap that should be filled with AppendTargetExclusions. AoeTrigger pulls from the raw attackers list; I thought it operated on the same pathway as the OnAttackers actions but it doesn't. Well, that's a matter for a separate PR if somebody feels like dealing with it =P

Celandriel:
> Cool, looking forward to giving it a try. Currently, beating him is an aggravating clown fest of running around and cycling bots through stay/follow/flee.

How did this go?

src/Bot/Engine/BuildSharedValueContexts.cpp
```diff
@@ -6,10 +6,12 @@
 
 #include "AiObjectContext.h"
 #include "ValueContext.h"
-#include "Ai/Dungeon/Mech/MechValueContext.h"
+#include "MechValueContext.h"
+#include "UBValueContext.h"
```
jrad7, line 9: Wanted to leave a note, I changed the Mech include here to not use the full path since that seems to be the norm.

brighton-chi:
So I just tried to test this but I forgot to tune my damage down from my normal 60 these days so the boss just melted and I don't feel like I got a good grasp on the mechanics. I'll try again this weekend.

One thing though is I think bots ignoring the mushrooms shouldn't be limited to during the encounter. The mushrooms don't despawn when the boss dies so as soon as he dies the bots go ham on the mushrooms. This is obviously not too hard to prevent as a player, but in the spirit of wanting dungeons to be able to be run by bots on their own (easy enough even now with selfbotting), I think bots should inherently ignore them.

If grid searching for mushrooms during the whole instance poses a performance concern, then one way to gate it might be to gate the target exclusion after a GetCreatureBySpawnIdStore() check for Hungarfen's dbguid and the bot being within some range of Hungarfen. It's not ideal, but I think that way of doing it is super cheap. Although I guess Hungarfen can be "gathered" after he is killed so maybe that won't work well. Another alternative might be a gate based on being within some distance of a point in the center of Hungarfen's platform. But also I suspect having 5 bots just search for the mushrooms in the instance is not a big deal.

Also, the next time I test I am just going to teleport to Hungarfen. Those frontal cleaves from the Underbats are just cancer for bots.

jrad7:
For me, the majority of runs end at Hungarfin on heroic while testing, nothing else in the zone was killing runs that I noticed. I didn't see underbats being a problem, but it shouldn't be too much trouble to account for if that's the next big party wiper.

For the mushrooms, I can add their spore cloud to out-of-combat aoe avoidance, then it shouldn't matter if they try to dps or not they'll just move if it explodes.

brighton-chi:
Yeah as long as you have a multiplier to keep them out.

Watching the run, my Rogue (in SWP gear with like 10k HP) was constantly dying. And then ranged would get killed too just if they ended up in range for one reason or another.

jrad7:
Pushed some updates:

- When in combat with an underbat, melee will rally on top of the tank, and casters will move out of the lash range. I'm seeing almost zero deaths to this now, unless they get unlucky timing when the melee tries to run to the tank.
- mushrooms should no longer be targeted to kill after hungarfun dies
- throttled the mushroom scan by a lot, it was doing a scan every 200ms that was way too aggressive. It now scans every 1 sec while Hungarfin is alive, and eventually stops completely 30 secs after Hungarfin dies.

src/Ai/Dungeon/UB/UBActions.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    float const safeDistance = MaxEffectRadius(SPELL_FOUL_SPORES, FOUL_SPORES_RADIUS_FALLBACK) + FOUL_SPORES_MARGIN;
+    float const currentDistance = bot->GetDistance2d(boss);
+    if (currentDistance >= safeDistance)
+        return false;
+
+    float const moveDist = safeDistance - currentDistance + 1.0f;
+    float const awayAngle = boss->GetAngle(bot);
+
+    GuidVector const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
```
brighton-chi: GuidVector is probably a good candidate for use of auto

src/Ai/Dungeon/UB/UBActions.cpp
```diff
@@ -0,0 +1,125 @@
...
+bool UBVacateSporeCloudAction::Execute(Event /*event*/)
+{
+    float const dangerRange = MushroomDangerRange(bot);
+    GuidVector const& mushrooms = AI_VALUE_REF(GuidVector, "ub mushrooms");
+    Creature* mushroom = GetNearestDangerousMushroom(bot, mushrooms, dangerRange);
+    if (!mushroom)
+        return false;
+
+    Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
+    if (botAI->IsTank(bot) && boss && boss->GetVictim() == bot)
```
brighton-chi: We should use scope resolution for all role checks.

src/Ai/Dungeon/UB/UBShared.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include "Define.h"
+#include "ObjectGuid.h"
+
+class AiObjectContext;
+class Creature;
+class Player;
+class Unit;
+
+namespace UnderbogHungarfen
```
brighton-chi, line 18: typical practice is not to indent namespaces

src/Ai/Dungeon/UB/UBShared.h
```diff
@@ -0,0 +1,83 @@
...
+    constexpr uint32 SPELL_GROW        = 31698;
+    constexpr uint32 SPELL_SPORE_CLOUD = 34168;
+
+    constexpr float SPORE_CLOUD_RADIUS_FALLBACK = 8.0f;
+    constexpr float FOUL_SPORES_RADIUS_FALLBACK = 20.0f;
+
+    constexpr float SPORE_CLOUD_MARGIN = 1.5f;
+    constexpr float FOUL_SPORES_MARGIN = 2.0f;
+
+    constexpr uint32 GROW_STACKS_DANGER = 8;
```
brighton-chi: Not particularly relevant at all but this and MUSHROOM_IDLE_SCAN_SKIPS can be uint8

src/Ai/Dungeon/UB/UBTriggers.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+bool UBFoulSporesTrigger::IsActive()
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
+    return boss && boss->HasAura(SPELL_FOUL_SPORES);
+}
+
+bool UBSporeCloudDangerTrigger::IsActive()
+{
+    if (!bot->IsAlive())
```
brighton-chi: BOT_STATE_DEAD won't run dungeon strategies so you don't need checks for if the bot is alive for triggers

src/Ai/Dungeon/UB/UBMultipliers.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "Playerbots.h"
+#include "ReachTargetActions.h"
+#include "UBActions.h"
+#include "UBShared.h"
+
+using namespace UnderbogHungarfen;
+
+float HungarfenFoulSporesMultiplier::GetValue(Action* action)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "hungarfen");
```
brighton-chi: Better for performance to check the actions first I think

src/Ai/Dungeon/UB/UBShared.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+                if (bot->GetExactDist(member) > UNDERBAT_RALLY_MAX_RANGE)
+                    continue;
+
+                return member;
+            }
+        }
+
+        Unit* victim = bat->GetVictim();
+        if (victim && victim != bot && victim->IsAlive() &&
```
brighton-chi: Victim can't be dead, no?

src/Ai/Dungeon/UB/UBShared.cpp
```diff
@@ -0,0 +1,271 @@
...
+                continue;
+
+            if (!bat->HasInArc(UNDERBAT_SAFE_ARC, &spot))
+                return false;
+        }
+
+        return true;
+    }
+
+    namespace
```
brighton-chi: Probably easier to follow if the anon namespace goes to the top. I don't think you need anything else before it to compile.

src/Ai/Dungeon/UB/UBShared.cpp
```diff
@@ -0,0 +1,271 @@
...
+        return !hungarfen || !hungarfen->IsAlive();
+    }
+
+    float MushroomDangerRange(Player* bot)
+    {
+        return MaxEffectRadius(SPELL_SPORE_CLOUD, SPORE_CLOUD_RADIUS_FALLBACK) + bot->GetCombatReach() +
+               SPORE_CLOUD_MARGIN;
+    }
+
+    bool IsMushroomDangerous(Creature* mushroom)
```
brighton-chi: I think this can go in the anon namespace

brighton-chi (commented):
Looking at your stuff always gives me ideas, and it's way cleaner than what I do too. It's nice I get to keep learning, although it drives me kind of nuts too to feel like  I need to keep changing thigns =P

jrad7:
> Looking at your stuff always gives me ideas, and it's way cleaner than what I do too. It's nice I get to keep learning, although it drives me kind of nuts too to feel like I need to keep changing thigns =P

Thanks for the kind words, credit goes to AI, though. To be honest, the real value here is your reviews. I make it reference your reviews here on github while planning and writing, so in a way it is "learning" from you!

I'll run through the issues you found here tonight, thank you!

src/Ai/Dungeon/UB/UBMultipliers.cpp
brighton-chi: Well, to be more specific, I think the general cost runs: combat state/class (both are like nothing), role (not including ones requiring group iteration), action, find target, and roles requiring group iteration. A constant aura is probably good to go early too. Part of the consideration I think though is how much does the check exclude. Like a check that excludes only tanks is less useful than a check that excludes everybody except for tanks. It matters for actions to some degree too, like how common is the action and also how many are being suppressed, but that's getting too much into the weeds I think.

One thing I know for sure is a spec check is terrible. It's really expensive so generally to be avoided at all. It also means that role checks that can sweep in humans can be very expensive. Not a problem for multipliers, but for general group scans (like who in the group has an aura), the cost of checking a role is suddenly way worse.

src/Ai/Dungeon/UB/UBActions.cpp
jrad7: good call, fixed

src/Ai/Dungeon/UB/UBActions.cpp
jrad7: fixed

src/Ai/Dungeon/UB/UBMultipliers.cpp
jrad7: got it, check actions first

src/Ai/Dungeon/UB/UBShared.cpp
jrad7: fixed

src/Ai/Dungeon/UB/UBShared.cpp
jrad7: makes sense lol

src/Ai/Dungeon/UB/UBShared.cpp
jrad7: fixed

src/Ai/Dungeon/UB/UBShared.h
jrad7: fixed

src/Ai/Dungeon/UB/UBShared.h
jrad7: switched to uint8

src/Ai/Dungeon/UB/UBTriggers.cpp
jrad7: got it, dropping the guard

jrad7 (commented):
I think i covered everything

src/Ai/Dungeon/UB/UBValueContext.h
```diff
@@ -0,0 +1,92 @@
...
+            CheckHungarfenGone();
+
+        return mushrooms;
+    }
+
+private:
+    void CheckHungarfenGone()
+    {
+        if (_hungarfen.IsEmpty())
+            _hungarfen = UnderbogHungarfen::FindHungarfenGuid(bot);
```
kadeshar, line 67: What if you kill boss and make dungeon reset and make second attempt. Then _hungarfen dont contains dead boss guid?

src/Bot/PlayerbotAI.cpp
```diff
@@ -1660,6 +1660,9 @@ void PlayerbotAI::ApplyInstanceStrategies(uint32 mapId, bool tellMaster)
         case 544:
             strategyName = "magtheridon";  // Magtheridon's Lair
             break;
+        case 546:
+            strategyName = "tbc-ub";  // Coilfang Reservoir: The Underbog
```
kadeshar: : The Underbog is redundant in comment

kadeshar:
Conflicts to resolve

src/Ai/Dungeon/UB/UBValueContext.h
jrad7: good catch, fixed.

src/Bot/PlayerbotAI.cpp
jrad7: shortened to just "The Underbog"

brighton-chi: > : The Underbog is redundant in comment

Coilfang Reservoir is four instances: The Slave Pens, The Underbog, The Steamvault, and Serpentshrine Cavern. For Auchenai Crypts, Sethekk Halls, and The Eye, we've already listed out the instance hub in addition to the instance name so it seems whatever the approach is it should be consistent.

kadeshar: I thought is because boss name xD Sorry :D Then should stay Coilfang Reservoir: The Underbog

kadeshar (approved):

PR #2605, opened 2026-07-27
Fix/equip persistence on downlevel

Celandriel:
## Pull Request Description

Allows persistence to be enabled and not have issues while downleveling. By checking if the new level is lower than the previous one we wipe the gear. 

## Feature Evaluation

Use level brackets with peristence on, and bots whole levels drop should now have their gear reset as well. 

N/A

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Solution offered by Claude, seems consistent with a minimal change. 

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Bot/Factory/PlayerbotFactory.cpp
+++ b/src/Bot/Factory/PlayerbotFactory.cpp
@@ -633,12 +633,6 @@ void PlayerbotFactory::Randomize(bool incremental)
         ClearSkills();
         ClearSpells();
         ResetQuests();
-        // Equipment persistence applies to leveling up only. When a bot is taken down a level
-        // (level brackets, level reset, skip) its gear does not belong at the new level, so it is
-        // always cleared: InitEquipment() below hands out nothing but the starting outfit under
-        // level 5 and never strips what is already worn, which would otherwise leave a level 1 bot
-        // running around in level 80 epics. This runs before GiveLevel(), so bot->GetLevel() is
-        // still the level the bot is coming from.
         if (!sPlayerbotAIConfig.equipAndSpecPersistence ||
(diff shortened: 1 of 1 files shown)


src/Bot/Factory/PlayerbotFactory.cpp
```diff
@@ -634,7 +634,7 @@ void PlayerbotFactory::Randomize(bool incremental)
         ClearSpells();
         ResetQuests();
         if (!sPlayerbotAIConfig.equipAndSpecPersistence ||
-            level < uint32(sPlayerbotAIConfig.equipAndSpecPersistenceLevel))
+            level < uint32(sPlayerbotAIConfig.equipAndSpecPersistenceLevel) || level < bot->GetLevel())
         {
             ClearAllItems();
         }
```
NoxMax, line 642: tangential, but might as well get rid of the brackets around clearallitems and resetTalents while you're there.

NoxMax (changes requested):

Celandriel (commented):
For multi-line conditionals I think its a gray area on whether to use brackets or not. I think preference is to keep them.

NoxMax:
Up to you if you think an exception is useful here, but I advice against keeping it here if the concern is clarity. In instances like this I would say on their own `ClearAllItems();` and `bot->resetTalents(true);` above it are obvious as the statement. Only concern that happens sometimes is if you get something like:

```
    if (conditionA ||
    ConditionB)
        statement1;
    -some other code on the very next line-
```
and I do come across these sometimes, but honestly just a blank line after `statement1;` makes everything obvious without needing brackets.

kadeshar (approved):

PR #2607, opened 2026-07-27
fix food/drink delay

Yuof:
## Pull Request Description

Three fixes on the cheat-food rest path (`EatAction` / `DrinkAction`, on by default via `AiPlayerbot.BotCheats`).

1. The action applies aura `25990`, which restores 5% per 2s tick, and then slept for `18000ms * missing% / 100` (`12000ms` in a battleground). At 0% mana that is 18s of sleep against 40s of regen, so the bot stood up half full, re-triggered, and sat back down. The wait is now `ceil(missing% / 5) * 2s`, floored at one tick. The battleground constant goes with it, since the aura ticks at the same rate there.
2. Only time clears `nextAICheckDelay`, and the switch to the combat engine happens inside `AttackAction`, which the engine cannot reach while it is asleep. A bot that sat down to drink slept through the start of the fight. `UpdateAI` now zeroes a pending delay when the bot is in combat off the combat engine.
3. `StopMoving()` stops the spline but leaves the movement generator, which re-issues movement and breaks the seated aura. `GetMotionMaster()->Clear(false)` runs first, on the existing `isMoving()` branch.

## Feature Evaluation

- **Minimum logic**: the same arithmetic as before with different constants, one extra call on a branch that already existed, one `if` in `UpdateAI`.
- **Processing cost**: per bot per tick, a test of a `uint32` member that is 0 for any bot not sleeping. Sleeping bots also cost a unit-flag read and a pointer compare. No allocations, no lookups. Resting wakes up less often than it used to: one sit per meal instead of one per 18s.

## How to Test the Changes

Default config, one bot.

1. Drop a mana user to ~10% mana out of combat and let it drink. **Before**: stands up half full and re-sits, several cycles. **After**: one sit, stands at full.
2. Let a bot sit to eat, then pull a mob onto it. **Before**: stays seated until the delay expires. **After**: engages on the next tick.
3. Trigger the meal while the bot is moving (follow, or reach melee). **Before**: it drifts and the aura falls off. **After**: it stops and stays seated.

## Impact Assessment

    - - [ ] No, not at all

    One `uint32` test per bot per `UpdateAI`, short-circuiting for every bot that is not sleeping. Offset by far fewer rest wake-ups.

    - - [x] Yes (**explain why**)

    The cheat-food path is default. Bots now stay seated for the whole meal, drop the rest of the wait when combat starts, and stop properly when they sit. Same regen, delivered in one cycle instead of several.

    One branch in `UpdateAI`, one call inside an existing `if`. `NonCombatActions.cpp` loses more lines than it gains.

## AI Assistance

Used for tracing the delay/aura mismatch and drafting this description. Every line reviewed and understood.

## Final Checklist

- - [x] Any code ported/adapted from another project is attributed. (n/a, original)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (n/a, no config or command surface changed)

## Notes for Reviewers

- The wake-up guard clears *any* pending delay while the bot is in combat off the combat engine, not just a resting one. The delay is what blocks `AttackAction` in every case.

### DIFF
--- a/src/Ai/Base/Actions/NonCombatActions.cpp
+++ b/src/Ai/Base/Actions/NonCombatActions.cpp
@@ -7,6 +7,8 @@
 #include "NonCombatActions.h"
 #include "Event.h"
 #include "Playerbots.h"
+#include <algorithm>
+#include <cmath>
 
 namespace
 {
@@ -45,22 +47,16 @@ bool DrinkAction::Execute(Event event)
 
         if (bot->isMoving())
         {
+            bot->GetMotionMaster()->Clear(false);
             bot->StopMoving();
-            // botAI->SetNextCheckDelay(sPlayerbotAIConfig->globalCoolDown);
-            // return false;
         }
         bot->SetStandState(UNIT_STAND_STATE_SIT);
         botAI->InterruptSpell();
 
-        // float hp = bot->GetHealthPercent();
-        float mp = bot->GetPowerPct(POWER_MANA);
-        float p = mp;
         float delay;
 
-        if (!bot->InBattleground())
-            delay = 18000.0f * (100 - p) / 100.0f;
-        else
-            delay = 12000.0f * (100 - p) / 100.0f;
+        // 25990 restores 5% per 2s tick; wait whole ticks so the drink finishes.
+        delay = std::max(1.0f, std::ceil((100.0f - bot->GetPowerPct(POWER_MANA)) / 5.0f)) * 2 * IN_MILLISECONDS;
 
         botAI->SetNextCheckDelay(delay);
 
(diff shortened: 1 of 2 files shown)


Celandriel:
For this I started looking at porting over the cmango reaction engine, which should eliminate the food and drink issue.

https://github.com/Celandriel/mod-playerbots/tree/addReactionEngine

It's an AI port of the cmango, but I haven't had time to test it at all yet.

Yuof:
> For this I started looking at porting over the cmango reaction engine, which should eliminate the food and drink issue.
> 
> https://github.com/Celandriel/mod-playerbots/tree/addReactionEngine

This looks more comprehensive than my workaround for the non-combat engine switch. 

The delay calculation changes in this PR are still valid (and necessary to prevent cancel/resume spam of the food/drink action).

Aligning both is just removal of one if block, so it should not matter too much in which order they get merged.

Yuof:
I adjusted the 'delay' shape back to the separate assign, this way it will prevent merge conflicts with your branch.

brighton-chi:
So this would be really helpful for me to know: what is the path that would put a bot in combat if you pull a mob by it? The only way I'm aware of bots entering combat is by issuing AttackAction::Attack() (but I'm sure that's not the whole picture). What causes the bot to pass IsInCombat() while still drinking?

Yuof:
> So this would be really helpful for me to know: what is the path that would put a bot in combat if you pull a mob by it? The only way I'm aware of bots entering combat is by issuing AttackAction::Attack() (but I'm sure that's not the whole picture). What causes the bot to pass IsInCombat() while still drinking?

bot kills mob
bot sits down to drink
another mobs wanders closer and attacks

brighton-chi:
Well of course. I mean what is the code path that causes IsCombat() to pass here? It doesn't pass the same way for a bot as it does a real player since there isn't a client to enforce being in combat in the same way.

Edit: Or is the drinking interrupt you're implementing supposed to come into play only if the bot is directly attacked (not on account of group members engaging in combat, except presumably with boss fights where zone combat is activated)?

Yuof:
> Well of course. I mean what is the code path that causes IsCombat() to pass here? It doesn't pass the same way for a bot as it does a real player since there isn't a client to enforce being in combat in the same way.
> 
> Edit: Or is the drinking interrupt you're implementing supposed to come into play only if the bot is directly attacked (not on account of group members engaging in combat, except presumably with boss fights where zone combat is activated)?

I was pretty confused but that edit finally cleared it up! Yes this is for allowing the bot to protect itself after getting attacked. Without this it will just not defend itself while the timer is running.

kadeshar (approved):

PR #2612, opened 2026-07-29
fix(CORE/LFG): keep random bots in the dungeon queue

viik2k:
## Pull Request Description

With `AiPlayerbot.RandomBotJoinLfg` enabled, random bots queue for the Random Dungeon finder but a
group essentially never forms.

`LfgStrategy` wires two trigger nodes that fight each other: `random` -> `lfg join`, and
`seldom` -> `lfg leave`. `seldom` is a `RandomTrigger` with a 300s period, so roughly every five
minutes each queued bot is told to leave. `LfgLeaveAction::Execute` only refuses when the bot is
already past `LFG_STATE_QUEUED` (invited, in a proposal, in the dungeon) -- there is no guard for the
ordinary "still waiting to be matched" state, and the leave path emits no log line. Bots are
therefore evicted while simply waiting, so `LFGQueue::CheckCompatibility` never sees a
role-complete (tank + healer + 3x DPS), level-compatible pool that holds still long enough to match.

This adds one early return to `LfgLeaveAction::Execute`: while `RandomBotJoinLfg` is on, a random bot
does not leave a queue it deliberately joined. Nine added lines, six of them comment; nothing else in
the file changes.

Note this is a correctness fix, not a realism one -- the current behaviour makes an
enabled-by-default feature never reach its terminal state. It also *removes* work rather than adding
it on the path that used to leave (see below).

## Feature Evaluation

**Minimum logic.** One early return, reusing state that already exists. No new config option, no new
state on the bot, no timers, no new trigger, action or strategy, no change to `isUseful()`. The two
terms in the condition are an existing config bool and the existing `IsRandomBot` helper.

I considered two smaller variants and rejected both:
- Deleting the `seldom` -> `lfg leave` trigger node from `LfgStrategy` is a smaller diff and strictly
  cheaper, but it removes the only mechanism that ever drains a bot from the queue, including for
  bots that legitimately should leave.
- Guarding on `LFG_STATE_QUEUED` alone (the commented-out `HasStrategy("lfg", ...)` idea left in the
  function) would apply to non-random bots too, and would leave `LfgLeaveAction` unable to leave any
  queue at all given the existing `> LFG_STATE_QUEUED` guard above it.

**Processing cost.** `LfgLeaveAction::Execute` is reached from exactly one place: the `seldom`
trigger node in `LfgStrategy`. `lfg leave` is registered as an action/trigger name in
`WorldPacketActionContext` / `WorldPacketTriggerContext` but no strategy wires it as a trigger node,
so the 300s `RandomTrigger` is the only live caller. That is on the order of one evaluation per bot
per 300s, not per tick.

When it runs, the added cost is one bool read, and then only if that bool is true, `IsRandomBot()`:
a `CharacterCache` account lookup, `IsInRandomAccountList`, and a `std::find` over `currentBots`
(linear in the bot count). The condition is ordered with the config bool first so the linear scan is
skipped entirely whenever the feature is off. Also note `AiFactory` only attaches the `lfg` strategy
when `IsRandomBot(player) && randomBotJoinLfg`, so in practice the guard is near-tautological on this
path; it is kept as a cheap correctness backstop for a bot that got `+lfg` manually.

On the branch that now returns early, the change is net negative cost: it skips a `new WorldPacket`
allocation plus a `CMSG_LFG_LEAVE` session queue, and it removes the subsequent re-join through
`random` -> `lfg join`. The join/leave packet churn that used to happen every ~300s per queued bot
goes away.

## How to Test the Changes

Setup used (`playerbots.conf` / `worldserver.conf`):

```
AiPlayerbot.RandomBotJoinLfg = 1
AiPlayerbot.MinRandomBots = 380
AiPlayerbot.MaxRandomBots = 400
LFG.Location.All = 1
Instance.IgnoreRaid = 1
Instance.IgnoreLevel = 1
DungeonFinder.OptionsMask = 5
Logger.lfg = 2      # optional, to watch queue joins
```

Enough bots above level 15 to fill a 5-man across roles is the only real requirement; the effect is
easier to see with a few hundred.

1. Start the realm and let bots log in and level-spread as usual.
2. Watch the standing queue. With `Logger.lfg=2` you get `queues LFG, Dungeon as ...` on join.
   Before the change, bots accumulate and then quietly disappear from the queue; after it they stay.
3. Verify the terminal state rather than the queue log -- a join line is not proof a group formed:

```sql
SELECT COUNT(*) FROM characters.character_instance;
SELECT COUNT(*) FROM characters.instance;
```

Expected: before the change these stay at 0 indefinitely. After it, they become non-zero once the
first group forms and teleports in.

4. To confirm the escape hatch still works, set `AiPlayerbot.RandomBotJoinLfg = 0` and reload -- bots
   mid-queue fall through the new guard and leave normally.

## Impact Assessment

    - - [ ] No, not at all

The guard itself is cheap and runs on a 300s per-bot trigger, and it removes packet churn as
described above. I am flagging **moderate** because of a second-order effect I want to be upfront
about rather than have a reviewer find it:

`PlayerbotAI::AllowActive()` treats a non-`LFG_STATE_NONE` state as a reason to keep a bot fully
active ("bot is in LFG queue -- stay active", `src/Bot/PlayerbotAI.cpp`). Today the `seldom` leave
bounds how long any bot holds that state to ~300s. With this change, a queued bot holds it until it
gets a group, so a bot waiting a long time stays pinned to full activity instead of falling back to
throttled activity. On a realm with thousands of bots and a wide level spread, low-population level
brackets could keep a meaningful number of bots pinned active.

In other words the old leave was partly acting as a pressure valve, and this patch removes it. That
tradeoff is a design call I do not think I should make alone, so options if you would rather bound
it: leave only after the bot has been queued beyond some threshold, or exclude LFG-queued bots from
the `AllowActive` exemption once they pass one. Happy to implement whichever you prefer -- I did not
want to add state or a config option speculatively.

    - - [ ] No
    - - [x] Yes (**explain why**)

`AiPlayerbot.RandomBotJoinLfg` defaults to `true`, so this is the default path. Random bots at level
15+ now remain in the Random Dungeon queue instead of being pulled out roughly every 300s. The
intent is that the default behaviour reaches its terminal state (a group forms) instead of looping
forever; nothing new is enabled, and the feature can still be turned off.

    - - [ ] No

One new branch: an early return in a function that already opens with an early return on adjacent
state, reusing an existing config flag and an existing helper. No new abstraction, no new state, no
new config to keep in sync, so the added maintenance surface is about as small as a branch gets --
but it is a branch, so I have answered Yes.

## AI Assistance

- - [ ] No

Yes, and substantially, so I would rather spell it out than undersell it.

- **Investigation and diagnosis:** AI-assisted. I ran the realm, noticed dungeons never popped, and
  used an assistant to trace the call path (`LfgStrategy` trigger nodes -> `LfgLeaveAction` -> the
  missing `LFG_STATE_QUEUED` guard) and to check where else `lfg leave` and `IsRandomBot` are
  referenced.
- **The code:** the nine-line hunk was AI-drafted, then reviewed line by line by me. It is small
  enough that I can account for every term: what `LFG_STATE_QUEUED` means versus the states above
  it, why the config bool is tested first, and what happens when the feature is switched off
  mid-queue.
- **This description:** AI-assisted drafting, including the `AllowActive()` interaction in the impact
  section, which the assistant surfaced and I verified by reading the function.
- **What I am less sure of:** whether pinning queued bots active is acceptable at several-thousand-bot
  scale. I have not measured `pmon` before/after. My verification is behavioural on a single 400-bot
  realm, not a performance benchmark -- that is exactly why I marked moderate impact above and asked
  for direction. Tell me what you want measured and I will measure it.

 MaNGOS, another module)?

Original. No files added, no headers touched. The commit carries a `Co-Authored-By:` trailer for the
AI assistant per the disclosure above, not for any upstream project.

## Final Checklist

Notes on the above, so no tick is doing more work than it should:

- **Performance box deliberately left unticked.** It is understood and explained, but not *tested* in
  the sense this repo means -- I have not run a `pmon` before/after. I did not want to tick a
  measurement I have not taken. Point me at the setup you consider representative and I will run it.
- Dialogue / GPLv2 header / provenance boxes are ticked as not-applicable: no new bot text, no new
  files, no ported code.
- Documentation: no new or changed config option, so no `.conf.dist` or wiki change is needed.
- Compiler warnings: the modified file already includes and uses both `sPlayerbotAIConfig` and
  `RandomPlayerbotMgr`, so no new includes were needed; the CI build matrix covers this commit.

## Notes for Reviewers

- **Re-targeted to `test-staging`** and the branch was force-pushed after @Celandriel's review, so the
  diff is now a single commit, purely additive (+9/-0), with the commit message following
  `.git_commit_template.txt`. Sorry for the initial `master` target and missing template.
- **Empirical result**, same realm and config, only this patch changed: standing queue depth in a
  comparable ~15-20 minute post-restart window went from 14 to 80+ bots, and the `instance` /
  `character_instance` tables went from 0 rows in 4+ days to a first completed group (Ragefire
  Chasm, map 389) with a second group passing `LFG roles checked` shortly after. Before this, the
  realm had produced zero dungeon instances in 4 days across 3 restarts.
- **One honest caveat on that measurement:** the realm run used the same guard with the two operands
  in the opposite order (`IsRandomBot(bot) && randomBotJoinLfg`). I reordered them for the
  short-circuit described above; both terms are side-effect-free reads so it is semantically
  identical, but the behavioural numbers above come from the pre-reorder build. The CI matrix builds
  this exact commit.
- The `AllowActive()` point under Impact Assessment is the part I would most like a maintainer
  opinion on. If you would rather bound queue dwell time than remove the leave outright, I am happy
  to rework it that way -- I just did not want to invent a config option unprompted.
- I left the commented-out `HasStrategy("lfg", BOT_STATE_NON_COMBAT)` block in the function alone
  rather than delete it, to keep the diff purely additive. Say so if you would prefer it removed now
  that a real guard exists in its place.


Celandriel:
Use PR template.

viik2k:
Sorry! will open a new PR using the template.

Celandriel:
No need to open new, just replace the text.
Thank you.

kadeshar (approved):

Issue #2614, opened 2026-07-29
Playerbots learning all professions instead of only two primary professions


Wishmaster117:
Bug Report — Playerbots learning all professions instead of only two primary professions
Bug Description
Random playerbots incorrectly learn all available primary professions, even though WotLK rules limit characters to two primary trade skills.
Secondary professions (Cooking, Fishing, First Aid) behave correctly and are learned as expected.
The issue occurs despite MaxPrimaryTradeSkill = 2 being properly configured in both worldserver.conf and playerbots.conf.
This indicates that the problem is not configuration‑related, but likely a logic or regression issue inside Playerbots’ profession‑learning system.

Commit Hash
Playerbots Module: 3fa1c1e49f8f1324b72461e576bce7c89b0a6521 (master)

AzerothCore Playerbot Core: bf25eae704f5fa2cb6c9fc2458f96ec099c03a8e

How To Reproduce Bug
Start a server using the commits listed above.

Ensure the following settings are active:

MaxPrimaryTradeSkill = 2

AiPlayerbot.AutoLearnTrainerSpells = 1

AiPlayerbot.EnableNewRpgStrategy = 1

Create random bots (rndbots) at a low level so they level up naturally.

Observe their profession list as they gain levels.

Bots will eventually learn every primary profession, exceeding the configured limit of two.

Expected Behavior
Random bots should never learn more than two primary professions, matching WotLK rules and respecting the MaxPrimaryTradeSkill = 2 configuration.
They should only learn secondary professions freely.

Screenshots

<img width="1411" height="1183" alt="Image" src="https://github.com/user-attachments/assets/5b431a8b-646f-4dd9-b283-da132b65b77f" />

<img width="1438" height="940" alt="Image" src="https://github.com/user-attachments/assets/87b2d09d-23f8-4b1d-af35-603ddada588c" />

Modules
Only Playerbots module is used:

Playerbots Module @ 3fa1c1e49f8f1324b72461e576bce7c89b0a6521

No other modules known to conflict with Playerbots are installed.

Playerbot Settings
Random bots enabled

Auto-learn trainer spells enabled

New RPG strategy enabled

Standard scaling and population settings (not performance‑related)

System
OS: Ubuntu Server (Linux)

Hardware: Standard server hardware (not performance‑related)

Celandriel:
This one had dropped off my radar and I never finished testing the fix.

[...]

PR #2615, opened 2026-07-29
Playerbots learning all professions instead of only two primary professions  #2614

Wishmaster117:
## Pull Request Description

This PR prevents random bots from accumulating more than two primary professions, while still
respecting a lower AzerothCore `MaxPrimaryTradeSkill` limit if the server is configured below two.

The issue can happen while bots level through normal gameplay and RPG activity. Initial profession
selection respects the configured limit, but later maintenance and trainer-learning paths could keep or
learn professions outside the selected pair.

The change now:

- Detects the primary professions already known by the bot.
- Reconciles them with the persisted `firstSkill` and `secondSkill` values.
- Keeps no more than the configured number of primary professions.
- Removes extra primary professions during profession maintenance.
- Repairs missing or inconsistent persisted profession values.
- Prevents trainer learning from adding an unselected primary profession.
- Keeps secondary professions such as Cooking, Fishing, and First Aid available.
- Does not affect normal class spell learning.

The goal is to keep profession selection stable and predictable throughout the lifetime of a random bot.

## Feature Evaluation

The minimum required logic is to identify the primary professions already known by the bot, limit that
list to `AiPlayerbot.MaxPrimaryTradeSkill`, and persist the resulting selection.

Profession spells offered through maintenance or a tradeskill trainer are then accepted only when they
belong to one of the selected primary professions. Secondary professions remain allowed.

The implementation reuses the existing `firstSkill`, `secondSkill`, skill initialization, and trainer
learning flows instead of adding a separate profession-selection system.

The new logic does not run every AI tick.

Profession reconciliation runs during existing initialization or maintenance operations, such as bot
creation, refresh, and level-up maintenance. It checks a small fixed set of primary profession skill IDs.

The trainer filter adds a lightweight profession check while the existing trainer spell list is already
being iterated. It does not add another trainer scan, database query, or periodic world scan.

The cost therefore scales with existing initialization and trainer actions, not with the normal bot
update loop.

## How to Test the Changes

Tested on a fresh AzerothCore database with 500 active random bots.

Configuration used:

# worldserver.conf
MaxPrimaryTradeSkill = 2

# playerbots.conf
AiPlayerbot.AllowLearnTrainerSpells = 1
AiPlayerbot.AutoLearnTrainerSpells = 1
AiPlayerbot.EnableNewRpgStrategy = 1

### Basic population test

1. Start a population of random bots.
2. Allow the bots to initialize and log in.
3. Count primary profession skill lines for each bot.
4. Verify that no bot has more than `AiPlayerbot.MaxPrimaryTradeSkill` primary professions.

With 500 active bots, all initialized bots had exactly two primary professions. The query looking for
characters with more than two primary professions returned zero rows.

### Level-up maintenance test

1. Select a random bot with two existing professions.
2. Record its profession skills and persisted `firstSkill` / `secondSkill` values.
3. Increase the bot's level.
4. Save the characters.
5. Verify that the same two professions remain and progress normally.
6. Verify that no third profession is added.

The tested bot kept Mining and Jewelcrafting while leveling from level 54 to level 63.

### Excess profession cleanup test

1. Stop the worldserver.
2. Manually add a third primary profession to the test bot in `character_skills`.
3. Start the worldserver.
4. Trigger level-up maintenance.
5. Save the characters.
6. Verify that the injected third profession is removed.
7. Verify that the original selected profession pair remains.

Tailoring was injected as a third profession and was removed after the next level-up maintenance.
Mining and Jewelcrafting remained unchanged.

### Inconsistent persisted selection test

1. Keep the bot's real profession skills unchanged.
2. Replace `firstSkill` and `secondSkill` with two different primary profession IDs.
3. Start the worldserver.
4. Trigger level-up maintenance.
5. Save the characters.
6. Verify that the real professions are kept.
7. Verify that the persisted values are repaired to match the real professions.

The persisted values were changed to Blacksmithing and Tailoring while the bot actually had Mining and
Jewelcrafting.

After maintenance, the bot still had Mining and Jewelcrafting, and the persisted values were repaired to:

```text
firstSkill  = 186
secondSkill = 755
```

### Missing persisted selection test

1. Delete the bot's `firstSkill` and `secondSkill` events.
2. Keep its real profession skills unchanged.
3. Start the worldserver.
4. Trigger level-up maintenance.
5. Save the characters.
6. Verify that `firstSkill` and `secondSkill` are recreated from the real professions.

Both events were recreated correctly as Mining and Jewelcrafting.

### Trainer learning test

1. Use a bot that already has the configured maximum number of primary professions.
2. Group the bot with a player.
3. Place the player and bot next to a trainer for a different primary profession.
4. Target the trainer.
5. Trigger the bot's trainer action.
6. Save the characters.
7. Verify that the trainer's profession was not added.

A bot with Mining and Jewelcrafting was tested at a Blacksmithing trainer. Blacksmithing was not learned.

### Final database check

Run a database-wide query counting primary professions per character and filter for counts above the
configured limit.

Expected result:

```text
0 rows
```

## Impact Assessment

    - [ ] No, not at all
    - [x] Minimal impact (**explain below**)

There is no new per-tick processing.

The added work happens only during existing initialization, maintenance, level-up, or trainer actions.
It checks a small fixed set of profession skill IDs and adds a lightweight condition inside the existing
trainer spell iteration.

    - [ ] No
    - [x] Yes (**explain why**)

Random bots are now prevented from keeping or learning more primary professions than allowed by
`AiPlayerbot.MaxPrimaryTradeSkill`.

Bots that already have a valid profession selection should keep the same behavior. Bots with missing,
inconsistent, or excessive profession data are normalized during maintenance.

    - [ ] No
    - [x] Yes (**explain below**)

It adds explicit checks for primary versus secondary professions and whether a trainer spell belongs to
the bot's selected profession set.

The logic is kept in shared profession helper functions so initialization and trainer learning use the
same rules instead of maintaining separate profession-limit implementations.

## AI Assistance

- [ ] No
- [x] Yes (**explain below**)

AI assistance was used to help investigate the different profession-learning paths,  possible edge
cases, organize the test plan, and improve the wording of this PR description.

I reviewed the final implementation, compiled it locally, and manually tested the behavior with a
500-bot population and targeted database manipulation tests. I understand the submitted changes and
take responsibility for them.

 MaNGOS, another module)?
- [ ] Yes (**name the project and the original author(s) below**)

No code was copied or adapted from another project.

## Final Checklist

- [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).

No new bot dialogue, source files, configuration options, or commands were added. No code was ported from
another project.

## Notes for Reviewers

This change targets the profession accumulation problem described in issue #2614.

The main points to review are:

- Reconciliation of actual profession skills with `firstSkill` and `secondSkill`.
- Cleanup behavior when a bot already has more primary professions than configured.
- Repair behavior when persisted profession values are missing or inconsistent.
- Filtering of tradeskill trainer spells without affecting class trainers.
- Continued availability of secondary professions.
- Avoiding any new work in the regular AI update loop.

The change compiled successfully. The final database-wide check returned no characters with more than
two primary professions.


Wishmaster117:
Something wrong with this PR?

Celandriel:
No.. Just bandwidth for review is low.. Sorry...

kadeshar (dismissed):

src/Bot/Factory/PlayerbotFactory.h
```diff
@@ -164,6 +174,8 @@ class PlayerbotFactory
     static std::pair<uint16, uint16> ChooseProfessionPair(std::vector<WeightedProfessionPair> const& professionPairs);
     static bool HasProfessionPair(std::vector<WeightedProfessionPair> const& professionPairs,
```
Celandriel:
```suggestion

```
Dead Code

src/Bot/Factory/PlayerbotFactory.cpp
Celandriel:
```suggestion
```
Delete

Celandriel (changes requested):
Drop dead code

src/Bot/Factory/PlayerbotFactory.cpp
Wishmaster117: Removed

src/Bot/Factory/PlayerbotFactory.h
Wishmaster117: My bad, forget to remode this old one when refactored

kadeshar (approved):

Celandriel (approved):

brighton-chi:
It looks like rather than adding a new config option for max professions, which is how I interpreted the PR description, this relies on the number allowed in WorldServer.

I think an independent config is important. I have the config set to 11 in WorldServer, but I don't want bots to have more than 2. Me having 11 is for convenience--I've intentionally not allowed myself to have the combat benefit of more than 2 professions, including by deleting the passive abilities from my character. But with the WorldServer config being used, bots can have the combat benefit of more than 2.

Wishmaster117:
> It looks like rather than adding a new config option for max professions, which is how I interpreted the PR description, this relies on the number allowed in WorldServer.
> 
> I think an independent config is important. I have the config set to 11 in WorldServer, but I don't want bots to have more than 2. Me having 11 is for convenience--I've intentionally not allowed myself to have the combat benefit of more than 2 professions, including by deleting the passive abilities from my character. But with the WorldServer config being used, bots can have the combat benefit of more than 2.

Thanks for pointing this out. I checked the merged implementation again and I also noticed that the PR description is misleading here.

`AiPlayerbot.MaxPrimaryTradeSkill` is not actually a playerbots config option. I accidentally added the `AiPlayerbot.` prefix when documenting the test setup. The core setting used is `MaxPrimaryTradeSkill`.

The implementation itself already keeps the bot limit separate in the case you describe:

    maxPrimaryTradeSkills = min(2, CONFIG_MAX_PRIMARY_TRADE_SKILL)

So with `MaxPrimaryTradeSkill = 11`, random bots are still capped at 2 primary professions.

The core value only acts as a lower ceiling:
- core = 11 -> bots = 2
- core = 2 -> bots = 2
- core = 1 -> bots = 1

The profession selection is also capped when building the selected profession list, excess existing profession skills are removed during maintenance, and trainer/available-spell learning is filtered against the selected professions.

So your specific setup with 11 professions allowed for players should still keep random bots at 2.

An independent playerbots config could still make sense if we wanted the bot limit to be configurable independently rather than fixed at 2, but it isn't required to keep bots at 2 when the core limit is higher. I think that bots should stay Blizzlike.

I'll correct the PR description because the config name there is definitely wrong. Thanks for catching it.

PR #2616, opened 2026-07-30
feat(rogue): use riposte in the combat dps rotation

Yuof:
## Pull Request Description

`CastRiposteAction` was registered as the action `"riposte"`, but nothing could reach it: no trigger of that name existed and no strategy referenced it. Combat rogues never used Riposte.

This copies the warrior proc-window pattern. `RiposteAvailableTrigger` is a `CAN_CAST_TRIGGER`, same as `revenge` and `overpower`, and it goes into `DpsRogueStrategy` at `ACTION_HIGH + 4`, the priority `overpower` already uses in `ArmsWarriorStrategy`. The parry window and every other cast condition stay with `Spell::CheckCast`, which `CanCastSpell` runs anyway. `AiFactory` maps only `ROGUE_TAB_COMBAT` to the `dps` strategy, which is the tree Riposte is talented in.

None of the premade combat builds use Riposte, so random bots on a stock config never reach the trigger. This is for bots that got their talents from a player: alts and selfbots.

## Feature Evaluation

- **Minimum logic**: one factory entry and one trigger node, plus a trigger class that is a bare macro expansion. No new condition is written.
- **Processing cost**: one `CanCastSpell` per tick, and only for combat-spec rogues in combat. A rogue without the talent returns on the cached spell-id lookup. Same cost as the warrior triggers.

## How to Test the Changes

Take a combat-spec rogue bot talented into Riposte and give it a mob that attacks it, so it can parry. Riposte shows up in the combat log after a parry. Before this change it never did. Assassination and subtlety run `melee`, not `dps`, so they still never cast it.

## Impact Assessment

    One more trigger per tick for combat rogues in combat, the same as the warrior equivalents.

    - - [x] Yes (**explain why**)

    Combat rogues now use Riposte when they parry. It was dead code before.

    One trigger node. The trigger class has no body.

## AI Assistance

Used for tracing the unreachable action and drafting this description. Every line reviewed and understood.

## Final Checklist

- - [x] Any code ported/adapted from another project is attributed. (n/a, original)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (n/a, no config or command surface changed)

### DIFF
--- a/src/Ai/Class/Rogue/RogueTriggers.h
+++ b/src/Ai/Class/Rogue/RogueTriggers.h
@@ -43,6 +43,12 @@ public:
     BladeFuryTrigger(PlayerbotAI* botAI) : BoostTrigger(botAI, "blade fury") {}
 };
 
+class RiposteAvailableTrigger : public SpellCanBeCastTrigger
+{
+public:
+    RiposteAvailableTrigger(PlayerbotAI* botAI) : SpellCanBeCastTrigger(botAI, "riposte") {}
+};
+
 class RuptureTrigger : public DebuffTrigger
 {
 public:
--- a/src/Ai/Class/Rogue/RogueAiObjectContext.cpp
+++ b/src/Ai/Class/Rogue/RogueAiObjectContext.cpp
@@ -63,6 +63,7 @@ public:
     RogueTriggerFactoryInternal()
     {
         creators["kick"] = &RogueTriggerFactoryInternal::kick;
+        creators["riposte"] = &RogueTriggerFactoryInternal::riposte;
         creators["rupture"] = &RogueTriggerFactoryInternal::rupture;
         creators["slice and dice"] = &RogueTriggerFactoryInternal::slice_and_dice;
         creators["hunger for blood"] = &RogueTriggerFactoryInternal::hunger_for_blood;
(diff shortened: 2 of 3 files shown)


src/Ai/Class/Rogue/RogueTriggers.h
```diff
@@ -43,6 +43,8 @@ class BladeFuryTrigger : public BoostTrigger
     BladeFuryTrigger(PlayerbotAI* botAI) : BoostTrigger(botAI, "blade fury") {}
 };
 
+CAN_CAST_TRIGGER(RiposteAvailableTrigger, "riposte");
```
brighton-chi: To align with the rest of the Rogue triggers, it's probably best not to declare it with the macro, which is probably a practice we should get away from anyway for consistency since macros for triggers are the minority case. I don't know if there's really any benefit to tying riposte to SpellCanBeCastTrigger anyway as opposed to having it just be in the default actions for Combat.

But for sure the Rogue strategies in general are a dumpster fire. I think @ThePenguinMan96 was working on redoing them.

Celandriel: I agree with moving away from macro.

Celandriel (changes requested):

kadeshar (approved):

PR #2617, opened 2026-07-30
feat(RandomBot): make periodic stats log interval configurable

Yuof:
## Pull Request Description

The periodic random-bot stats dump in `RandomPlayerbotMgr::UpdateAIInternal` is hardcoded to 300s, with no way to change it or turn it off.
Adds `AiPlayerbot.RandomBotPrintStatsInterval`, in seconds, defaulting to the old hardcoded 300. `0` short-circuits the condition, so the dump and its per-bot pass never run.

## Feature Evaluation

  One config read, one `uint32` member, `300` in the existing `if` replaced by it. The off switch is one more term on the same condition, not a new branch.

  None per bot. `UpdateAIInternal` is the single manager update, not a per-bot or per-trigger path. One integer test added per manager tick. `0` removes the O(bots) `PrintStats()` pass.

## How to Test the Changes

1. Option absent: `Random Bots Stats: N online` still logs every ~300s, unchanged.
2. Set to `30`: logs every ~30s. First dump lands one interval late: existing arming behavior, `printStatsTimer` starts at 0.
3. Set to `0`: never logs.

Needs `AiPlayerbot.RandomBotAutologin = 1`, restart between changes.

## Impact Assessment

    - - [x] No, not at all

    Manager-level, once per `RandomPlayerbotMgr` update. Nothing scales with bot count. No pmon before/after taken. The change sits outside the per-bot tick pmon measures.

    - - [x] No

    Default 300 is the hardcoded value. Log output only.

    - - [x] No

    Same `if`, one more term. Option follows the interval options around it.

## AI Assistance

Conf comment and this description. Change is four lines; all reviewed and understood.

 MaNGOS, another module)?

## Final Checklist

- - [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note). (n/a, original)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (conf comment added)

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -1382,6 +1382,11 @@ AiPlayerbot.MaxRandomBotRandomizeTime = 1209600
 # Default: 60
 AiPlayerbot.RandomBotsPerInterval = 60
 
+# Interval (in seconds) between periodic random bot statistics dumps in the log
+# (online count, level/race/class distribution, roles, status). Set to 0 to disable.
+# Default: 300
+AiPlayerbot.RandomBotPrintStatsInterval = 300
+
 # Minimum and maximum seconds after death before a bot revives
 # Defaults: 60 (min), 300 (max)
 AiPlayerbot.MinRandomBotReviveTime = 60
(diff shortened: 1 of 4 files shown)


NoxMax:
Neat idea, particularly that running the stats command manually when it's needed is often more useful than the auto one.

Celandriel (approved):

PR #2627, opened 2026-08-02
Implement Naxx Heigan Strategy

karlverik:
## Pull Request Description

This PR implements and restores the strategy for the Naxxramas Heigan the Unclean encounter.
My works Base on the existing commented-out code, re-implemented and refined ```HeiganDanceAction::CalculateSafe()```.
Reference the implementation pattern in OSAction.cpp, i use ```AI_VALUE(GuidVector, "nearest hostile npcs")``` to query nearest trap hidden npcs which only appear when eruption.

 And a little fix on Thaddius, add  multiplier to prevent bot use Disengage and Blink.
```cpp
 if (dynamic_cast<CastDisengageAction*>(action) || (dynamic_cast<CastBlinkBackAction*>(action)))
        return 0.0f;
```
## YouTube
[Heigan Test Video](https://www.youtube.com/watch?v=d1Bs0wwNmn0)
DPS was intentionally lowered during this test to extend combat up to 5 minutes, demonstrating that bots reliably dodge all eruptions throughout the entire fight.

## Feature Evaluation

- This is a standard boss strategy execution with standard action registration. It only contains basic boolean checks and minimal numerical calculations to cycle through pre-defined safe waypoints when an eruption trigger is detected.
- There is no complex logic or expensive pathfinding calculations involved—only basic state checks and quick arithmetic operations.

## How to Test the Changes

1.Form a raid group containing Tank, Melee, and Ranged bots.
2.Engage the boss to start combat.
3.Observe bot movement during both the platform phase and the dance phase when eruptions occur.

## Impact Assessment

    - - [ ] No, not at all
    - New triggers, multipliers, and actions are evaluated only when the Heigan strategy is active during combat with Heigan the Unclean. Zero performance overhead during regular combat or open-world gameplay.

    - - [ ] No
    - - [x] Yes (**explain why**)
    Bots follow strategies when combat with heigan.

    - - [x] No

## AI Assistance

- - [ ] No

- **Translation & Documentation:** As a non-native English speaker, I used AI to help refine and translate descriptions into English for clearer documentation.
- **Code Refactoring:** I used Claude to assist in rewriting the `helper` .

 MaNGOS, another module)?

Explanation: No code was ported from external projects. This PR cleans up and restores existing commented-out Heigan strategy code within the repo, with only the CalculateSafe() logic being rewritten to accurately determine safe zones.

## Final Checklist


src/Ai/Raid/Naxx/NaxxBossHelper.h
```diff
@@ -537,4 +537,105 @@ class ThaddiusBossHelper : public AiObject
...
+public:
+    const std::pair<float, float> platform = {2794.26f, -3706.67f};
+    const std::pair<float, float> waypoints[4] = {
+        {2794.88f, -3668.12f}, {2775.49f, -3674.43f}, {2762.30f, -3684.59f}, {2755.99f, -3703.96f}};
+
+    HeiganBossHelper(PlayerbotAI* botAI) : AiObject(botAI) {}
+
+    bool UpdateBossAI()
+    {
+        if (!bot->IsInCombat())
```
brighton-chi: The checks for IsInCombat, IsInWorld, and IsAlive aren't operative because none of them will be false if you use FindTargetValue to get Heigan, and this helper is called in actions that are gated behind FindTargetValue.

src/Ai/Raid/Naxx/NaxxBossHelper.h
```diff
@@ -537,4 +537,105 @@ class ThaddiusBossHelper : public AiObject
...
+        }
+
+        return true;
+    }
+
+    bool IsPlatformPhase() const { return _platform_phase; }
+    std::pair<float, float> const& GetSafeWaypoint() const { return waypoints[_curr_safe]; }
+
+private:
+    const uint32 ERUPTION_NPC_ENTRY = 12999;
```
brighton-chi: I think these variables should be static constexpr.

Also, AC codestyle has standard naming requirements for NPC and Spell ids (they start with NPC_ and SPELL_, respectively).

src/Ai/Raid/Naxx/NaxxBossHelper.h
```diff
@@ -537,4 +537,105 @@ class ThaddiusBossHelper : public AiObject
...
+    void NextSafe()
+    {
+        _curr_safe += _curr_dir;
+        if (_curr_safe == 3 || _curr_safe == 0)
+            _curr_dir = -_curr_dir;
+    }
+
+    bool HasEruptionNearby()
+    {
+        GuidVector npcs = *context->GetValue<GuidVector>("nearest hostile npcs");
```
brighton-chi: I would not use nearest hostile npcs. It's very performance heavy--it does nearest npcs and does the hostility check afterwards. So it ends up running all its checks for all units, including players (which is particularly bad in a raid).

The best method is probably GetCreatureListWithEntryInGrid here.

src/Ai/Raid/Naxx/NaxxTriggers.cpp
```diff
@@ -82,25 +82,25 @@ bool GrobbulusCloudTrigger::IsActive()
...
-//    return botAI->IsRanged(bot);
-//}
+bool HeiganMeleeTrigger::IsActive()
+{
+    Unit* heigan = AI_VALUE2(Unit*, "find target", "heigan the unclean");
+    if (!heigan)
+    {
+        return false;
+    }
+    return !botAI->IsRanged(bot);
```
brighton-chi: Role checks are better to go before finding the target since they are cheaper. Probably clearer to use IsMelee also, given what the trigger is for (IsMelee() just means !IsRanged(), which is kind of lousy and should be changed, but it is what it is right now).

src/Ai/Raid/Naxx/NaxxTriggers.cpp
```diff
@@ -82,25 +82,25 @@ bool GrobbulusCloudTrigger::IsActive()
...
-//    Unit* heigan = AI_VALUE2(Unit*, "find target", "heigan the unclean");
-//    if (!heigan)
-//    {
-//        return false;
-//    }
-//    return botAI->IsRanged(bot);
-//}
+bool HeiganMeleeTrigger::IsActive()
+{
+    Unit* heigan = AI_VALUE2(Unit*, "find target", "heigan the unclean");
```
brighton-chi: No need to define a variable for heigan since you don't use it. You can just check for him inside the if statement.

src/Ai/Raid/Naxx/NaxxMultipliers.h
```diff
@@ -19,14 +19,14 @@ class GrobbulusMultiplier : public Multiplier
...
-//public:
-//    virtual float GetValue(Action* action);
-//};
+class HeiganDanceMultiplier : public Multiplier
+{
+public:
+    HeiganDanceMultiplier(PlayerbotAI* ai) : Multiplier(ai, "helgan dance") {}
+
+public:
+    virtual float GetValue(Action* action);
```
brighton-chi: I know using virtual for GetValue is the pattern in this file, but we should change them all to use override instead (although it's actually not entirely consistent since GluthGenericMultiplier already uses override).

Anyway, Multiplier's GetValue is already virtual, so it's automatically virtual for all derived classes. Override will protect against mistakes like mispellings by forcing a compile error.

src/Ai/Raid/Naxx/NaxxTriggers.cpp
```diff
@@ -82,25 +82,25 @@ bool GrobbulusCloudTrigger::IsActive()
...
+}
+
+bool HeiganRangedTrigger::IsActive()
+{
+    Unit* heigan = AI_VALUE2(Unit*, "find target", "heigan the unclean");
+    if (!heigan)
+    {
+        return false;
+    }
+    return botAI->IsRanged(bot);
```
brighton-chi, line 97: Since roles are static members of PlayerbotAI, we should use PlayerbotAI::IsRanged(bot) and so on instead of botAI->, which misleadingly implies that what botAI is matters in this context. I need to go back and change these calls throughout the codebase, or at least in stuff I've worked on.

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
```diff
@@ -43,68 +43,69 @@ float GrobbulusMultiplier::GetValue(Action* action)
...
+float HeiganDanceMultiplier::GetValue(Action* action)
+{
+    Unit* boss = AI_VALUE2(Unit*, "find target", "heigan the unclean");
+    if (!boss)
+    {
+        return 1.0f;
+    }
+    bool platform_phase = boss->IsWithinDist2d(2794.26f, -3706.67f, 10.0f);
+    bool eruption_casting = false;
+    if (boss->HasUnitState(UNIT_STATE_CASTING))
```
brighton-chi: I think this block is better put in a helper like bool IsEruptionCasting().

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
```diff
@@ -43,68 +43,69 @@ float GrobbulusMultiplier::GetValue(Action* action)
...
+                // Fallback to name for custom spell data.
+                isEruption = botAI->EqualLowercaseName(info->SpellName[LOCALE_enUS], "eruption");
+            }
+            if (isEruption)
+            {
+                eruption_casting = true;
+            }
+        }
+    }
+    if (dynamic_cast<CombatFormationMoveAction*>(action) ||
```
brighton-chi: So for multipliers, the current pattern in the codebase is like how it is here, with encounter conditions (like finding the boss) checked before the actions. And that's reversed from how it should be. It's better for performance to check the actions first--then no need to get into all the encounter checks if the bot isn't even attempting the action that you're trying to block.

I'm in the process of flipping them for strategies I worked on, but anyway that should probably be the practice going forward. And if you need to check the same action twice in order to reverse the order, then you can just use a local variable for it. You can look at the Sunwell PR to see the way I tried to do it.

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
```diff
@@ -43,68 +43,69 @@ float GrobbulusMultiplier::GetValue(Action* action)
...
+        {
+            return 0.0f;
+        }
+        uint32 castTime = spellInfo->CalcCastTime();
+        if (castTime == 0 && !spellInfo->IsChanneled())
+        {
+            return 1.0f;
+        }
+    }
+    return 0.0f;
```
brighton-chi, line 71: So since this blocks everything at the end, I'd explicitly allow WipeAction through. That way, the "wipe" command will work even when the multiplier is in effect.

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
```diff
@@ -43,68 +43,69 @@ float GrobbulusMultiplier::GetValue(Action* action)
...
-//        }
-//        uint32 castTime = spellInfo->CalcCastTime();
-//        if (castTime == 0 && !spellInfo->IsChanneled())
-//        {
-//            return 1.0f;
-//        }
-//    }
-//    return 0.0f;
-//}
+float HeiganDanceMultiplier::GetValue(Action* action)
```
brighton-chi, line 44: I know you didn't add the include for UseMeetingStoneAction.h, but that seems unused and I think worth deleting while you're at it. I'm guessing it was to enable the "summon" command for something (maybe even Heigan).

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
```diff
@@ -43,68 +43,69 @@ float GrobbulusMultiplier::GetValue(Action* action)
...
+    {
+        Spell* spell = boss->GetCurrentSpell(CURRENT_GENERIC_SPELL);
+        if (!spell)
+        {
+            spell = boss->GetCurrentSpell(CURRENT_CHANNELED_SPELL);
+        }
+        if (spell)
+        {
+            SpellInfo const* info = spell->GetSpellInfo();
+            bool isEruption = NaxxSpellIds::MatchesAnySpellId(info, {NaxxSpellIds::Eruption10});
```
brighton-chi: I know you didn't name this constant, but there's only one Eruption spell (it's same for 10 and 25) so this should probably be renamed so it's not misleading.

src/Ai/Raid/Naxx/NaxxBossHelper.h
karlverik: Thanks for your reply! I'll review and address these points step by step. I'm actually quite new to C++ (only about 10 days in), so I've been mostly following existing patterns and might miss some best practices. Appreciate the guidance!

brighton-chi: I don't really know what I'm doing, either, so we'll all help each other as we go =P

src/Ai/Raid/Naxx/NaxxBossHelper.h
karlverik: I actually used '''GetCreatureListWithEntryInGrid''' initially. However, after reading the previous PR discussions regarding Naxx strategies, I noticed the recommendation against directly querying the AZ core, which is why I switched to the current approach.

src/Ai/Raid/Naxx/NaxxTriggers.cpp
karlverik: My bad, I got a bit lazy here. These were originally commented-out lines, so I assumed they were fine and just copied and used them directly. I'll properly fix this!

src/Ai/Raid/Naxx/NaxxBossHelper.h
brighton-chi: Ah I think there was a misunderstanding. There's nothing wrong with using core functions. The module itself uses tons of them. All the distance measurements, for example.

The issue with the original Naxx strategies was that they relied on modifications to the actual boss scripts in the Playerbots fork of AC. So what we can't do is edit AC code in order to enable a Playerbots strategy. In practice, we've discovered more things over time that make it wholly unnecessary anyway.

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
karlverik: I kept the original code in NaxxMultipliers.cpp especially for Heigan since I wasn't entirely sure about these multipliers and saw that it mostly returns 1.0f in normal combat.

That said, I feel the spell part might be unnecessary here—since Heigan isn't the one casting the eruptions, we can't really fetch specific spell info through these methods.

src/Ai/Raid/Naxx/NaxxMultipliers.cpp
karlverik: I noticed significant logic duplication in the multipliers. They should be refactored alongside the helper to avoid repeating these checks.

src/Ai/Raid/Naxx/NaxxBossHelper.h
```diff
@@ -535,4 +535,99 @@ class ThaddiusBossHelper : public AiObject
...
+        _curr_safe += _curr_dir;
+        if (_curr_safe == 3 || _curr_safe == 0)
+            _curr_dir = -_curr_dir;
+    }
+
+    bool QueryEruptionNearby()
+    {
+        std::list<Creature*> eruptionNpcs;
+        bot->GetCreatureListWithEntryInGrid(eruptionNpcs, NPC_HEIGAN_ERUPTION, 100.0f);
+        return !eruptionNpcs.empty();
```
kadeshar, line 621: That return true in heigan fight? Can you run debugger on that?

src/Ai/Raid/Naxx/NaxxBossHelper.h
```diff
@@ -535,4 +535,99 @@ class ThaddiusBossHelper : public AiObject
...
+
+    HeiganBossHelper(PlayerbotAI* botAI) : AiObject(botAI) {}
+
+    bool UpdateBossAI()
+    {
+        Unit* boss = AI_VALUE2(Unit*, "find target", "heigan the unclean");
+        if (!boss)
+            return false;
+
+        if (_unit != boss)
```
kadeshar, line 553: What this check cover?
_unit != boss is true after wipe?

src/Ai/Raid/Naxx/NaxxBossHelper.h
```diff
@@ -535,4 +535,99 @@ class ThaddiusBossHelper : public AiObject
...
+        {
+            Reset();
+            _unit = boss;
+        }
```
kadeshar, line 557: Pattern from rest of BossHelper's shouldnt be reproduce also? 
Like:
Image

brighton-chi: I think this came from my comment. At call sites (at least for Heigan), the helper is gated by find target so all that stuff is dead code.

kadeshar: Nice finding, we need to fix that in other helpers :D

src/Ai/Raid/Naxx/NaxxBossHelper.h
karlverik: > What this check cover? _unit != boss is true after wipe?

I want to run reset() after every fight, whether it's a win or a wipe.

src/Ai/Raid/Naxx/NaxxBossHelper.h
karlverik: > That return true in heigan fight? Can you run debugger on that?
```cpp        
LOG_INFO("playerbots", "QueryEruptionNearby: empty = {} (found {} NPCs)", eruptionNpcs.empty(),
                 eruptionNpcs.size());
```
image
image


Yes, it will return true when find eruption npcs and return false in other time.

Celandriel:
Thank you for the contribution but #2671 was merged.

PR #2628, opened 2026-08-03
Rewrite Nightbane and Update Karazhan Code

brighton-chi:
## Pull Request Description

This PR is a rewrite of Karazhan, which includes a reworked strategy for Nightbane and other more minor tweaks for several bosses.

The major change for Nightbane is that hardcoded coordinates are no longer used during the ground phase. To accomplish this:
- When the tank picks up Nightbane, the tank will position between Nightbane and the balcony so that Nightbane faces off the balcony, instead of pulling Nightbane through two hardcoded points like before.
- Ranged now reads Charred Earth locations by scanning the dynamic objects and moves in a coordinated stack, observing certain min/max distances from Nightbane, rather than relying on the previous hardcoded 3-position cycle.

I’ve removed RTI marking entirely except for selective usage of Skull. This means the strategy won’t screw with existing RTI assignments and is not reliant on marks being placed for bots to target the correct enemy (besides Skull, which I’ve allowed any bot in the raid to mark).

The majority of changes, however, are not intended to be functional but to improve the quality of the code and align style. I’m not going to pretend I’m at all proficient with C++, but I can at least make some improvements compared to before. Changes include:
- Swapping from maps to private member variables where possible to limit the impact on the global namespace
- Adding an out-of-combat method to prune maps
- Reordering multipliers to check actions first before conditions such as finding the boss (a change that is better for performance)
- Directly accessing bot role functions (e.g., IsTank, IsDps), which are static members of PlayerbotAI, instead of using an instance pointer, which misleadingly suggests that whatever botAI is matters when calling those functions.
- Switching to enum class and East-coast const per AC codestyle
- Reorganizing helpers and data by boss so the code is easier to follow and maintain
- Generally consolidating code in many places, with the purpose of making it less complicated and duplicative

## Feature Evaluation

The overall structure of the strategy remains the same, save that I added a target exclusion for Nightbane during the flight phase (based on the system I previously implemented for the open Sunwell PR).

## How to Test the Changes

Run Karazhan, make sure bosses are beatable (preferably with heavily nerfed player damage/healing). Focus particularly on Nightbane. Make sure performance impact is acceptable.

## Impact Assessment

    - - [x] No, not at all
  
I have not done a detailed breakdown of numbers, but I believe overall performance is improved over the prior version.

    - - [ ] No
    - - [x] Yes (**explain why**)
 
Only with the “karazhan” strategy enabled and in the Karazhan raid.

    - - [x] No

## AI Assistance

- - [ ] No

For complicated math and the heavy changes to Netherspite beam code. Those changes were to allow the red beam dance without using maps and fix a bug with the timing of the beam blocker yells.

 MaNGOS, another module)?

## Final Checklist


Celandriel:
Here is what im getting from claude 
**CRITICAL**
- **[A4] Renamed action reference with no matching creator** — `KaraStrategy.cpp` (curator block): `NextAction("the curator spread ranged to mitigate arcane sear", ACTION_RAID)` but the only registered key is `creators["the curator spread ranged"]` (KaraActionContext.h unchanged there; the class also self-names "the curator spread ranged"). Silent failure: ranged never spread on Curator. Fix: use the existing name or rename the creator key + class name consistently.
- **[D4 + logic] Unused `moveX`/`moveY` in `NightbaneGroundPhaseTanksPositionBossAction::Execute`** — KaraActions.cpp (diff ~2246–2253): computes `maxMoveDist`/`moveDist`/`moveX`/`moveY` then calls `MoveTo(KARA_MAP_ID, destX, destY, ...)`. Every sibling action (Maiden, Wolf, Curator, Attumen) moves via `moveX/moveY`; here the clamp is dead and `-Werror` CI fails on the unused variables. Fix: `MoveTo(..., moveX, moveY, ...)`.

**WARNING**
- **[B11] Unsynchronized cross-thread/cross-instance state, retained in rewritten code** — `KaraHelpers.cpp`: function-local `static ObjectGuid currentRed/Green/BlueBlocker` in `GetCurrentBeamBlockers()` are shared by ALL Karazhan instances across map threads (data race + blocker identity leaking between raids); ditto the namespace-scope `unordered_map<uint32, time_t>` timers mutated from map threads. Pre-existing, but the PR rewrote these functions — key statics by instanceId at minimum and note the threading caveat.

**NOTE**
- P2: several mechanics sit above the emergency-heal band (existing file convention), and trackers moved to `ACTION_EMERGENCY + 10` (=100); teleport failsafe at 99 — conditional, acceptable, but documents drift from the RAID band.
- `GetAttumenMounted`: `constexpr uint32 searchRadius = 50.0f;` — should be `float`.
- Ranged-leader pick doesn't filter to bots (`GET_PLAYERBOT_AI`); a human ranged player can become the stack target.
- 0.2f/0.5f stack tolerances with `MOVEMENT_FORCED` + `InterruptSpell()` risk jitter/cast-starvation while the leader moves.
- Unused enum entries retained (`NPC_TERESTIAN_ILLHOOF`, `NPC_DEMON_CHAINS`, `NPC_KILREK`).

Celandriel (approved):

PR #2631, opened 2026-08-03
BWL Bot Strategies: fire resistance for the remaining bosses

raijyan:
## Pull Request Description

Razorgore and Vaelastrasz already register the shared fire resistance trigger/action pair, but the bosses after them do not, so bots stop buffing for fire partway through the raid. Broodlord's Blast Wave (23331), Firemaw's Flame Buffet (23341) and the Fire Nova (23462) that Flamegor's Frenzy triggers are all fire damage.

Registers the existing BossFireResistanceTrigger / BossFireResistanceAction for Broodlord Lashlayer, Firemaw and Flamegor. No new logic, no new classes - the same wiring the two earlier bosses already use.

Ebonroc is deliberately left out: his kit is Shadow Flame, Shadow of Ebonroc and Wing Buffet, which is shadow and physical only.

## Feature Evaluation

Minimum logic: three registrations of an existing trigger/action pair, plus three trigger nodes. No new code paths.

Processing cost: three additional trigger evaluations per tick for bots on map 469, identical in cost to the two already present for Razorgore and Vaelastrasz.

## How to Test the Changes

Take a bot raid into BWL and engage Broodlord, Firemaw or Flamegor. Bots should apply fire resistance the same way they already do for Razorgore and Vaelastrasz. Ebonroc is unchanged.

## Impact Assessment

    - - [ ] No, not at all

Three extra trigger evaluations per tick for bots on map 469, same cost as the two already there.

    - - [ ] No
    - - [x] Yes (**explain why**)

Bots in BWL now buff fire resistance for three more bosses; nothing outside map 469.

    - - [x] No

## AI Assistance

- - [ ] No

Yes; mechanical wiring of an existing trigger/action pair to three more bosses. Damage schools were verified against Spell.dbc before deciding which bosses belong in the list.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

Second of the BWL series. Ebonroc was dropped from the original draft of this as his Shadow Flame entirely shadow, not fire, so fire resistance would not help against him!

### DIFF
--- a/src/Ai/Base/Trigger/BossAuraTriggers.cpp
+++ b/src/Ai/Base/Trigger/BossAuraTriggers.cpp
@@ -12,13 +12,13 @@
 
 bool BossFireResistanceTrigger::IsActive()
 {
+    // Check if bot is paladin first: cheap check that filters out most of the raid
+    if (bot->getClass() != CLASS_PALADIN)
+        return false;
+
     // Check boss and it is alive
     Unit* boss = AI_VALUE2(Unit*, "find target", bossName);
     if (!boss || !boss->IsAlive() || boss->IsFriendlyTo(bot))
-        return false;
-
-    // Check if bot is paladin
-    if (bot->getClass() != CLASS_PALADIN)
         return false;
 
     // Check if bot have fire resistance aura
@@ -63,13 +63,13 @@ bool BossFireResistanceTrigger::IsActive()
 
 bool BossFrostResistanceTrigger::IsActive()
 {
+    // Check if bot is paladin first: cheap check that filters out most of the raid
+    if (bot->getClass() != CLASS_PALADIN)
+        return false;
+
     // Check boss and it is alive
     Unit* boss = AI_VALUE2(Unit*, "find target", bossName);
     if (!boss || !boss->IsAlive() || boss->IsFriendlyTo(bot))
-        return false;
-
-    // Check if bot is paladin
-    if (bot->getClass() != CLASS_PALADIN)
         return false;
 
     // Check if bot have frost resistance aura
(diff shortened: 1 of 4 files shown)


brighton-chi:
I'm on record as disliking the boss aura actions because forced strategy swaps in general should be avoided. The change persists after the encounter, and it takes away flexibility during the encounter. What if I don't want to use up a Paladin's slot on Shadow Resistance Aura? It doesn't stack with Priest's Shadow Resistance. And Fire/Frost Resistance Auras don't stack with corresponding Shaman totems. The other thing is this strategy could also have been implemented with the same level of functionality but avoiding any forced strategy swap by incorporating a multiplier.

That being said, I understand you're not responsible for this overall design =P so none of that is directed toward you personally. But what should probably be added at a minimum if these strategies are going to be touched is to move the class check to the top of each trigger in BossAuraTriggers.cpp. It's a very cheap check and immediately filters out the vast majority of bots in the raid so at least we can lessen the performance impact.

dillyns:
I also agree that these boss resistance strategies should be rethought as a whole, unrelated to this PR. I am not a fan of ever adding or removing strategies to the bots which users have configured themselves, and like its stated above they just generally don't work well. I have paladins switching to shadow resistance aura in MC all the time when I already have priest shadow resist buffs up, and then they don't switch back after the boss fights are done.

I do think the idea has value though, especially with the recent push for bots to run raids with no supervision, but the current design needs an overhaul imo.

raijyan:
Completely agree! It's not a robust system as it is and in an ideal world I'd have the encounters having some flag for damage types and the bots work out the cheapest way to satisfy the resistances and coverage from the available class tools on engagement of the entity. Then all future encounters are just a generic tagging job or resolved straight from the damage types in the DB automatically (and it'd work better for "weird" groups like players 5 manning with scaling).

I'll hoist the class checks and validate it all when I'm back in BWL Saturday, working on the boss strats and validating some of my trash-handling improvements.

raijyan:
Moved the cheaper XXResistanceTrigger::IsActive() escapes to the top.

Validated in BWL by having FR aura after Broodlord, stripping the strategy `co -fire resistance`, swapping all paladins to `cast devotion aura` and seeing the FR aura removed. Then pulling Firemaw, which returned the FR aura strategy and buff.

Leaving the aura cleanup out of this one, though I do already have something for it in the AQ40 set.

kadeshar (approved):

PR #2636, opened 2026-08-06
feat: pvp dispel behavior

zacharis278:
## Pull Request Description

Simple improvement to dispel mechanics when bots are in pvp. Right now bots have a near-instant reaction time to recognize and dispel debuffs on team members. This isn't really realistic and locks down players using DoT classes in a way that doesn't feel great. Additionally bots don't recognize dangerous debuffs like Vampiric Touch and Unstable Affliction and will instantly dispel them every time and off themselves.

I'm sure this could be 'smarter' but I opted for an easy solution that doesn't add any new state or expensive checking. 

Changes:

1. (Only in BG/Arena) Each dispel check now has a percentage chance to execute each tick. So the longer the aura stays up the higher chance it eventually gets dispelled. See table below.
2. Buffs that would be harmful to dispell are identified the chance a bot makes a "mistake" and dispells it anyway are much lower than the default behavior.

Assuming combat hasn't ended this is the chance the bot would act.

| Elapsed | Ordinary debuff | Dangerous debuff (UA / Vampiric Touch) |
|---|---|---|
| 1s | 25.0% | 1.0% |
| 2s | 43.75% | 2.0% |
| 3s | 57.8% | 2.9% |
| 5s | 76.3% | 4.9% |
| 7s | 86.7% | 6.8% |
| 10s | 94.4% | 9.6% |

## Feature Evaluation

Negligible cost here. I just added a string compare / random call to an existing check. 

## How to Test the Changes

Ran a few battlegrounds with this and it feels about right.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

See description. ONLY impacts in pvp maps

    - - [ ] No

Bot can now choose to skip dispelling based on urand() check. See description.

## AI Assistance

- - [ ] No

Claude code was used for code generation. I'm not a C++ dev but the change here is small enough to easily understand.

 MaNGOS, another module)?

## Final Checklist


zacharis278:
will reopen with staging as base

PR #2638, opened 2026-08-07
Fix/infinite gathering loop

humra:
## Pull Request Description

I noticed several bots getting stuck in an infinite loop (observed for about 5 hours) of locating a herb/ore, going to gather it, not picking up the item if their bag space is over 80% full, then gathering it again.

## Feature Evaluation

When determining is the looting possible it checks the bag space is low and if any of the 4 affected skill IDs are present. If both are true it returns false and so ends further checks.
Processing cost is minimal, it is a single if statement comparing uint32 values. I tested it by running 900 bots on a mid-range laptop and did not notice any difference.

## How to Test the Changes

Bots with learned skinning, herbalism, mining or engineering have to exist, along with a non-null chance of them gathering appropriate herbs.
While bot's bag is <80% full the bot should gather the profession item if it encounters it as usual.
If the bag is >80% full the bot should ignore profession gathering nodes.
I tested it by following the bots around and changing their bag space available.

## Impact Assessment

    - - [X] No, not at all

    - - [ ] No
    - - [X] Yes (**explain why**)
It will change how and when the bot will gather herbalism, mining, skinning or engineering nodes.

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist


brighton-chi:
Funny. I think the exception should be for HasRealPlayerMaster() though, not HasActivePlayerMaster().

Edit: Actually on second thought, we probably shouldn't let a selfbot-led group create an exception. So maybe it should be HasActivePlayerMaster OR IsRealPlayer.

Although are self bots also subject to the 80% rule? If so I guess we'd have to limit this to HasActivePlayerMaster...

humra:
I think the self bots are also subject to the rule because LootAction checks if (!botAI->HasActivePlayerMaster() && AI_VALUE(uint8, "bag space") > 80)
I also changed my check because I was using >= instead of >, it should now match the rest of the checks.

kadeshar:
Fix compilation errors before review

humra:
> Fix compilation errors before review

There was a missing reference after pulling from master, fixed it now

src/Mgr/Item/LootObjectStack.cpp
```diff
@@ -315,6 +314,20 @@ bool LootObject::IsLootPossible(Player* bot)
     if (go && go->HasFlag(GAMEOBJECT_FLAGS, GO_FLAG_INTERACT_COND) && !go->ActivateToQuest(bot))
         return false;
 
+    //Prevents bot from getting stuck in an infinite loop of
+    //gathering herb/ore/skin -> bag too full, don't pick up -> gather again
```
kadeshar, line 318: But what if object is skinnable and also have normal loot. Then still bot will not loot that target when bagUsage > 80?

humra: AzerothCore does not mark object as skinnable until all the regular loot is removed from it.
It's the void Creature::AllLootRemovedFromCorpse() in Creature.cpp

kadeshar (approved):

PR #2639, opened 2026-08-07
Let bots cast during Kael'thas's Gravity Lapse

brighton-chi:
## Pull Request Description

PlayerbotAI::CastSpell() returns false for any bot that satisfies IsFlying() (flying or disabled gravity movement flag). I assume there's a good reason for this, probably due to some exploiting concerns, but in any case, that's a problem during the Kael'thas encounters in TK and MgT.

Kael casts Gravity Lapse, which throws players up into the air, and players take consistent damage during that period. Bots should be able to cast during those periods, just like players with clients. This PR adds exceptions to the IsFlying() early returns in the CastSpell() overloads for unit- and ground-targeted spells for bots that have one of the Gravity Lapse auras.

## Feature Evaluation

Add two spellid checks to the IsFlying() check in two PlayerbotAI::CastSpell() overloads, after checking IsFlying() itself. The third overload is just a wrapper for the others that takes a string for the spell so it doesn't need its own exception.

## How to Test the Changes

1. Fight Kael'thas in TK or MgT
2. Wait until he uses his Gravity Lapse mechanic (in both cases, after he reaches 50% HP)
3. Confirm that bots can still cast in flight

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist


kadeshar:
Conflict to resolve

brighton-chi:
> Conflict to resolve

Done.

src/Bot/PlayerbotAI.cpp
```diff
@@ -56,8 +56,13 @@
 #include <sstream>
 #include <string>
 
+namespace
+{
 constexpr uint32 SPELL_TITAN_GRIP = 49152;
 constexpr uint32 SPELL_DK_FROST_PRESENCE = 48263;
+constexpr uint32 SPELL_GRAVITY_LAPSE_TK = 39432;
+constexpr uint32 SPELL_GRAVITY_LAPSE_MGT = 44224;
```
kadeshar: I dont know this boss but that spell shouldnt be check?

Image

brighton-chi: Oh good call, I think you're probably right. I'll check ingame and update if that's the case.

brighton-chi: Changed to 44226. I think 44227 might also be applied, but 44226 is the one that has a visible icon on players.

kadeshar: Ok i think so far is ok :) AI review was: "Players there carry 44219-44223 / 44226 / 44227, none is 44224."

kadeshar (approved):

PR #2642, opened 2026-08-07
Nomenclature: Use the word "bot" instead of "randombot" when referring to all bot accounts

NoxMax:
## Pull Request Description

This touches non-config instances only to provide some clarity, as such the changes are minor and straightforward. Done in 3 commits, and their messages should be obvious.

## Feature Evaluation

## How to Test the Changes

Compiled and ran cleanly on my end. Double check the same thing on your end.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Searching for any missing instances.

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -89,7 +89,7 @@ AiPlayerbot.RandomBotAutologin = 1
 AiPlayerbot.MinRandomBots = 500
 AiPlayerbot.MaxRandomBots = 500
 
-# Randombot accounts
+# Bot accounts
 # Set to 0 to automatically calculate needed accounts based on MaxRandomBots, EnablePeriodicOnlineOffline
 # and its ratio, and AddClassAccountPoolSize. Set manually if you want to create more accounts than needed.
 # If the manual value is less than the required, the system will override it with the required value.
@@ -105,7 +105,7 @@ AiPlayerbot.AddClassAccountPoolSize = 50
 # Do not change this prefix while there are existing bot accounts.
 AiPlayerbot.RandomBotAccountPrefix = "rndbot"
 
-# Delete all randombot accounts
+# Delete all bot accounts
 # To apply this, set the number to 1 and run the Worldserver. Once deletion is complete, if you would
 # like to recreate randombots, set the number back to 0 and rerun the Worldserver.
 AiPlayerbot.DeleteRandomBotAccounts = 0
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -89,7 +89,7 @@ AiPlayerbot.RandomBotAutologin = 1
 AiPlayerbot.MinRandomBots = 500
 AiPlayerbot.MaxRandomBots = 500
 
-# Bot accounts
+# Randombot accounts
 # Set to 0 to automatically calculate needed accounts based on MaxRandomBots, EnablePeriodicOnlineOffline
 # and its ratio, and AddClassAccountPoolSize. Set manually if you want to create more accounts than needed.
 # If the manual value is less than the required, the system will override it with the required value.
@@ -105,7 +105,7 @@ AiPlayerbot.AddClassAccountPoolSize = 50
 # Do not change this prefix while there are existing bot accounts.
 AiPlayerbot.RandomBotAccountPrefix = "rndbot"
 
-# Delete all bot accounts
+# Delete all randombot accounts
 # To apply this, set the number to 1 and run the Worldserver. Once deletion is complete, if you would
 # like to recreate randombots, set the number back to 0 and rerun the Worldserver.
 AiPlayerbot.DeleteRandomBotAccounts = 0
--- a/src/Ai/Base/Actions/ChangeStrategyAction.cpp
+++ b/src/Ai/Base/Actions/ChangeStrategyAction.cpp
@@ -48,7 +48,7 @@ bool ChangeNonCombatStrategyAction::Execute(Event event)
     std::string const text = event.getParam();
 
     uint32 account = bot->GetSession()->GetAccountId();
-    if (sPlayerbotAIConfig.IsInRandomAccountList(account) && botAI->GetMaster() &&
+    if (sPlayerbotAIConfig.IsInBotAccountList(account) && botAI->GetMaster() &&
         !botAI->GetMaster()->CanBeGameMaster())
     {
         if (text.find("loot") != std::string::npos)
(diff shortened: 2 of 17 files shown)


NoxMax:
Merge conflict fix.

src/Ai/Base/Actions/ChangeStrategyAction.cpp
```diff
@@ -48,7 +48,7 @@ bool ChangeNonCombatStrategyAction::Execute(Event event)
     std::string const text = event.getParam();
 
     uint32 account = bot->GetSession()->GetAccountId();
-    if (sPlayerbotAIConfig.IsInRandomAccountList(account) && botAI->GetMaster() &&
```
kadeshar, line 51: I was a little bit out of project but there was discussion about that? 
IsInBotAccountList sound for me like list with account which only have bots. But technically is not true. Because i can manually create account and create alts bot there and never use as normal characters.
Places like TrainerAction.cpp will be confusing
Image
But maybe that is only my pov and most coders agree with concept from this PR

conf/playerbots.conf.dist
```diff
@@ -105,7 +105,7 @@ AiPlayerbot.AddClassAccountPoolSize = 50
 # Do not change this prefix while there are existing bot accounts.
 AiPlayerbot.RandomBotAccountPrefix = "rndbot"
 
-# Delete all randombot accounts
+# Delete all bot accounts
 # To apply this, set the number to 1 and run the Worldserver. Once deletion is complete, if you would
 # like to recreate randombots, set the number back to 0 and rerun the Worldserver.
 AiPlayerbot.DeleteRandomBotAccounts = 0
```
kadeshar, line 111: Changing description without changing config key in multiple places not only here

src/Bot/PlayerbotMgr.cpp
```diff
@@ -577,7 +577,7 @@ void PlayerbotHolder::OnBotLogin(Player* const bot)
     //     // bot->TeleportTo(master);
     // }
     uint32 accountId = bot->GetSession()->GetAccountId();
-    bool isRandomAccount = sPlayerbotAIConfig.IsInRandomAccountList(accountId);
```
kadeshar, line 580: Not fully renamed. Variable still is named isRandomAccount

NoxMax: I kept staring at this and how this happened, then I noticed it's red. That's the delete diff not the add 😂

conf/playerbots.conf.dist
NoxMax: If it was up to me, I would. But dropping the word "Random" from all configs that are not exclusively about randombots, is a thing that needs to be thoroughly discussed and there would be a need to plan a wide announcement so people don't get confused about why the old keys are not working. We could maybe look into introducing a script that renames things in the user's conf file, we can make it accept new and old config styles, regardless there's a bunch of ways that this can be done.

So I purposely didn't touch the keys here. I want to make things a bit clearer in code in this minor PR, and key edits without a plan would cause confusions.

src/Ai/Base/Actions/ChangeStrategyAction.cpp
NoxMax: I kinda see your point, but at the same time I think among the coders and the general users when someone says "a bot account", then they are referring to a server created account. Even with the case of an account that just has altbots, at the very top of the conf file we define them as "characters created on player accounts". So I don't think there would really be a confusion here.

src/Bot/PlayerbotMgr.cpp
kadeshar: Sorry in wrong column i put comment

Image

src/Ai/Base/Actions/ChangeStrategyAction.cpp
kadeshar: Im not sure is correct logic. Bot = altbot or randombot. 
Randombot have 2 types -> rndbot and addclass (personally would say that should be named like Generated and Manual or something like this).
For me both name IsInRandomAccountList and IsInBotAccountList are wrong.
Table is named randomBotAccounts then something like IsInRandomBotAccounts/IsInRandomBotAccountList is more correctly. 
Renaming method without renaming table (and even rest like configs) is confusing.

NoxMax: Wait, which table? There is no random accounts table. There is `playerbots_random_bots` which I can rename here to `playerbots_bot_events` for a more honest description. But there is no other table with the word "random" in it.

There is `playerbots_account_type` however, which I do plan to make the source of global truth that answers questions about bot accounts and bot types. It would even make IsRandomBot a cheaper check.

Right now most, but not all the truths, center around `sPlayerbotAIConfig.randomBotAccountPrefix` and I hate how much has been built on top of that prefix. But I do have a plan of action to strip down responsibilities from it, and thankfully given that this is not a commonly adjusted config, it would be less of an issue if they key is renamed. And of course whatever I am going to do would be backwards compatible with any existing account. The main takeaway however is that that is part of a larger PR that I am planning in October, and the changes here is setting up groundwork for it.

kadeshar: Sorry, yea i mean sPlayerbotAIConfig.randomBotAccounts.
I will back to PR when i get feedback from other admin's because for me is not something that our two should decide.

NoxMax: For what it's worth, I think the matter of renaming the configs is doable. The safest bet is adding a framework to PlayerbotAIConfig so that it recognizes a config that has been renamed, and the list of old names it might be known as in the user's conf file. But I would do that in its own PR.

kadeshar: Admins consensus about that PR is to not change that. "Bot" word is too wide for that

NoxMax:
Update for latest PR merges.

kadeshar:
obraz

Based on this comment i think this PR should be close

NoxMax:
Well technically that's kinda the point. A bot account is any bot account, and calling all bot accounts randombot accounts is straightup wrong. Regardless, the whole point is to have less confusion, and if the current system is less confusing while also kinda misleading, then I guess it is what it is. Though mull it a bit more over before coming to a final decsion.

kadeshar:
That is not my personal decision but active admins. I will leave that PR opened for one week if someone want to say something more about that.

PR #2644, opened 2026-08-08
Rewrite Tempest Keep: The Eye Strategies

brighton-chi:
## Pull Request Description

This PR is a rewrite of Tempest Keep, which includes functional revisions to all bosses and many (hopefully) non-functional changes to improve code in the same vein as the recent Karazhan rewrite.

The main functional changes are:

**Crimson Hand Centurions**
-	The prior logic was actually non-functional. It works now (Mages interrupt Arcane Flurry with Polymorph).
Al’ar
-	Each platform now has a tank position and a melee dps position, which are set up so that melee dps can attack from behind Al’ar.
-	At 5% hp in phase 1, melee dps jumps off the platform so they don’t get hit by the explosion when Al’ar “dies” in phase 1.
-	Fixed phase 2 avoidance with respect to spreading before Dive Bomb and moving away from Al’ar afterwards to avoid the Rebirth.
-	Fixed phase 2 tanking of Embers so there isn’t a swap after one dies.

**Void Reaver**
-	Ranged bots no longer initially spread around the boss upon the pull. It was dangerous and not that helpful.
-	Instead, ranged only dodges Arcane Orbs until 90% HP. The initial Orbs naturally result in ranged splitting into at least a couple of clumps. After that, they enable a very slight spread from nearby players, which helps keep them from clumping up too much over the duration of the fight.

**High Astromancer Solarian**
-	The main tank stays on her the whole time, including when she reappears with the priests, since she technically is tankable despite spending most of her time spamming spells on random targets.
-	Melee bots still split between priests, but no RTI marking is used now for their targeting
-	Removed all coded ranged movement (besides to run away during Wrath of the Astromancer). I found it to be more constricting than helpful.

**Kael’thas**
-	Gravity Lapse phase is totally redone. It’s a bit more of a discussion so I put this in the notes at the end.
-	Fixed legendary weapon equipment logic. It now no longer matters whether the bot considers the weapon to be an upgrade. It should also ensure that Combat Rogues put the sword in their MH and dagger in their OH, and it should resolve issues that could arise for bots that use OH-only or MH-only weapons (or have them in their inventory).
-	DKs now loot/equip the proper weapons. I don’t know shit about DKs. I literally had them looting the shield before. I never imagined they would be unable to equip them. FML.
-	Changed the handling of the weapon phase quite a bit. Notably, the bow and axe were moved up in priority, and also melee targeting is contingent on the target not being too close to the axe.

**Re RTI Marking**
I’ve removed RTI marking entirely except for selective usage of Skull and, in cases where dps must be split, Cross. No marks actually impact bot targeting. They are purely visual markers to guide players. This means the strategy won’t screw with existing RTI assignments and is not reliant on marks being placed for bots to target the correct enemy. It also is more friendly for player tanks.

## Feature Evaluation

The overall structure of the strategy remains the same, except that I added a couple of target exclusions to TKStrategy.cpp.

## How to Test the Changes

Run Tempest Keep, make sure bosses are beatable (preferably with heavily nerfed player damage/healing via IP). Focus on whether the primary mechanics mentioned above work, with a particular focus on Kael’thas (bots equipping the legendary weapons properly and Gravity Lapse actually resulting in bots being airborne and flying to spread). Al’ar is also a point of focus with respect to handling of Embers and movement overall.

## Impact Assessment

    - - [x] No, not at all
 
I have not done a detailed breakdown of numbers, but I believe overall performance is similar to or slightly improved over the prior version.

    - - [ ] No
    - - [x] Yes (**explain why**)
 
Only with the “tempestkeep” strategy enabled and in the Tempest Keep raid.

    - - [x] No

## AI Assistance

- - [ ] No

AI helped substantially with implementing the method to replicate and handle Gravity Lapse and with the custom equip method for the legendary weapons. The rest I mainly wrote myself with some AI assistance to save time.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

The bottom line is that Gravity Lapse does not properly affect bots, and my objective with this PR was not just to make avoidance work but also try to simulate the actual mechanic. The tl;dr reason for why the spell doesn’t work is that flight flags and gravity simulation are done on the client side for players. The slightly more detailed version is that the server sends a flight packet (via SetCanFly()), but there is no corresponding packet handler in Playerbots. There are handlers for teleporting and knockbacks and roots, but for whatever reason one has not been implemented for flight (maybe because players flying without mounts is quite rare). Instead, Playerbots, through UpdateMovementState, directly applies the can fly/disable gravity/flying movement flags upon the first update once a flying aura is detected.

Thus, what happens is Kael casts Gravity Lapse, putting a flying aura on the bots and hitting them with a knockback. Because there is no flight packet handler, instead of stopping somewhere in midair, bots complete the entire knockback arc and land on the ground, Then, at that point the next time the bot tries to move at all, the movement flags get updated and flying/disable gravity flags are added because a flying aura is present. The result is that bots end up flying just above the ground. Then, when the aura lapses after 30 seconds, a packet is sent again, but again the mod doesn’t handle it. Since the aura lapsing in and of itself has no effect on movement flags, bots stay airborne. Only once the bots move again is UpdateMovementState called, clearing the flags.

The new Gravity Lapse method does the following: (1) it makes the bots actually airborne, and move airborne through flying, while the aura is effective, and (2) it drops the bots to the ground as soon as the aura lapses. It does this by setting the flight movement flags directly when the aura is present, rather than waiting for UpdateMovementState to do it on the next movement, and by discarding the knockback’s movement generator so bots start moving immediately instead of riding the arc back down. From there, spreading happens in 3D space in a semi-coordinated manner. When the aura lapses, the action clears the flight flags and issues an actual fall via MoveFall. It then clears the leftover falling movement flag once they land (needed because a player normally clears it by sending MSG_MOVE_FALL_LAND, which a bot never does, and FWIW I think the lack of that packet is also the reason bots don’t take fall damage).

I also have PR# 2639 open in conjunction with this one to allow bots to cast while flying (in this limited circumstance).


brighton-chi:
Latest commit is big but very minor on the functional side. There are a couple of bug fixes, and the remainder is just trying to made code improvements. I'm not touching this anymore, other than to address issues/comments.

brighton-chi:
This and Gruul are getting a little unwieldy to keep open with other changes going on. Will reopen once other things are settled. I'll keep SWP open just in case people want to use it.

PR #2594, opened 2026-07-24
BWL Wyrmguard tanks imrprovements and add Ebonroc swap

dillyns:
## Pull Request Description

Improvements to Wyrmguard tank triggers. Tank also use Free Action Potions if they have them.
Add Tank swap for Shadow of Ebonroc.

## Feature Evaluation

- Describe the **minimum logic** required to achieve the intended behavior.
- Describe the **processing cost** when this logic executes across many bots.

## How to Test the Changes

Test 1:
Give a tank a Free Action Potion.
Go to BWL and pull a Wyrmguard. Tank should use the Free Action Potion when it is the current Wyrmguard target.

Test 2:
Go to BWL and pull Ebonroc.
When Shadow of Ebonroc is cast, one of the current offtanks should taunt Ebonroc

## Impact Assessment

- Does this change increase per-bot/per-tick processing or risk scaling poorly with thousands of bots?
    - - [x] No, not at all
    - - [ ] Minimal impact (**explain below**)
    - - [ ] Moderate impact (**explain below**)

- Does this change modify default bot behavior?
    - - [x] No
    - - [ ] Yes (**explain why**)

- Does this change add new decision branches or increase maintenance complexity?
    - - [x] No
    - - [ ] Yes (**explain below**)

## AI Assistance

Was AI assistance used while working on this change?
- - [x] No
- - [x] Yes (**explain below**)

Copilot was used for method stub generation. All code was thoroughly reviewed.

## Code Provenance / Attribution

Was any code in this PR copied or adapted from a sister / upstream project (e.g. CMaNGOS playerbots,
 MaNGOS, another module)?
- - [x] No, all code in this PR is original
- - [ ] Yes (**name the project and the original author(s) below**)

## Final Checklist

- - [x] Stability is not compromised.
- - [x] Performance impact is understood, tested, and acceptable.
- - [x] Added logic complexity is justified and explained.
- - [x] Any new bot dialogue lines are translated.
- - [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).
- - [x] New source files use the GPLv2 header.
- - [x] Documentation updated if needed (Conf comments, WiKi commands).
- - [x] New and modified files do not introduce new compiler warnings.

PR #2646, opened 2026-08-08
BWL Bot Strategies: warlock pack focus and technician spread for trash

raijyan:
## Pull Request Description

Each Blackwing Warlock (12459) in the BWL trash packs casts Demon Portal (22372) every 30-45s, and each portal summons an Enraged Felguard every 30s until the warlock dies (killing it despawns its portals). Left to the generic target pickers the bots chew on whatever is nearest, so the pack never shrinks while a warlock is alive and the felguards stack up.

Three pieces, all gated on the pack NPCs being in combat within 35 yards:

- Warlock pack anchor: every dps prioritises a live warlock. Melee split evenly across the live warlocks by GUID ordinal, so every bot derives the same split with no shared state and the warlocks fall together. Ranged burn the nearest one.
- Warlock pack focus multiplier: zeroes the generic target pickers for dps while a warlock is alive, so the summoned felguards can't drag anyone off the kill.
- Technician spread: the technicians' Bomb (22334) is thrown at random raiders and splashes 5 yards, so ranged and healers keep a 6-yard gap via weighted repulsion. Only genuine clumps move, and steps are paced to one committed move per bot per 1.5s so the camp repositions deliberately instead of thrashing.

The strategy is only auto-applied inside Blackwing Lair (map 469).

## Feature Evaluation

- Minimum logic: two triggers, one attack action, one movement action and one multiplier, all on existing base classes, plus two grid-search helpers. In the steady state (bot already burning a warlock) all three sites share a cached current-target check and skip the grid search entirely.
- Processing cost: combat and role flags exit first, so bots outside BWL pack combat pay a couple of boolean checks. Grid searches only run for dps actually fighting the pack, and only until a warlock is acquired. Measured numbers under Impact Assessment.

## How to Test the Changes

1. Take a bot raid into BWL and pull any trash pack containing Blackwing Warlocks and Blackwing Technicians.
2. Watch the dps: they should switch to the warlocks on pull, melee spreading across several warlocks rather than piling one, and ignore the summoned felguards until the last warlock drops, at which point normal target picking resumes.
3. Watch the ranged/healers: they should open up roughly 6-yard gaps while technicians are alive. Bots already spaced should stand still and keep casting, only clumps move.
4. The decision trace in Playerbots.log shows the new entries as `bwl warlock pack anchor`, `bwl attack warlock pack anchor` and `bwl technician spread`.

## Impact Assessment

    - - [ ] No, not at all

Measured with the standardized pmon procedure from this template (BotActiveAlone = 100, botActiveAloneSmartScale = 0, 41 fully active bots: a 40-bot raid parked in BWL plus 1 random bot). Parked baseline Total was 38.4s over the 5-minute window (0.361ms avg per AI update). During a full pack fight (18 warlocks, 92 technicians respawned for the test) the average rose to 0.459ms. The new trigger, action and multiplier rank below pmon's visible table floor in both windows, while pre-existing BWL triggers chart at 0.1-0.3%.

    - - [ ] No
    - - [x] Yes (**explain why**)

Only inside BWL, and only while these specific trash NPCs are in combat within 35 yards: dps target selection locks to warlocks and ranged/healers de-clump. Bosses and everything outside map 469 are untouched, and the lock releases itself when the last warlock dies.

    - - [ ] No

Two trigger nodes and one multiplier added to the existing BWL raid strategy. Self-contained in the BWL raid files, no new config, no changes to shared machinery.

## AI Assistance

- - [ ] No

Code generation and log analysis were done with AI assistance (Claude, via Claude Code, co-author trailer on the commit). I reviewed all of it and validated it live across four pack pulls in a 40-bot raid, iterating on what the decision traces showed (the spread pacing and the steady-state early-outs both came out of those traces). The pmon measurements above were run on the final code.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

The behaviour claims were checked against the data rather than folklore: the warlock SAI casts Demon Portal (22372) on a 30-45s timer, the portal NPC (14081) summons via serverside aura 22391 triggering 22392 "Summon Enraged Felguard" every 30s, and the warlock's On Just Died script despawns its portals. Bomb (22334) is a 5-yard fire AoE cast at a random hostile per the technician SAI.

The spread started life re-deciding every tick (235 moves in 12 seconds across the camp in the first live trace) and was paced to one committed step per bot per 1.5s after that showed up as visible zigzagging. The early-outs cut roughly 4-5k redundant grid searches per pack and dropped target acquisitions from 45 to 32, since bots now hold an acquired warlock instead of reshuffling when another one dies.

Wyrmguard handling is untouched (#2594 owns that space), and the suppression room hatcher priority is deliberately left out for a separate PR.

-----------------------

Let me know if you prefer more (like this) or less verbosity in PRs etc. As I progress my tooling and understanding of the codebase hopefully I can settle in and help out more confidently. :)

### DIFF
--- a/src/Ai/Raid/BWL/BWLTriggers.cpp
+++ b/src/Ai/Raid/BWL/BWLTriggers.cpp
@@ -87,6 +87,33 @@ bool BwlNefarianFearWardTrigger::IsActive()
 
 // Trash
 
+static constexpr float BWL_TRASH_SEARCH_RANGE = 35.0f;
+
+bool BwlWarlockPackAnchorTrigger::IsActive()
+{
+    // Each warlock's Demon Portal (22372) keeps summoning Enraged Felguards,
+    // and its death despawns its portals, so every dps burns a warlock whenever
+    // one is alive - not the technicians and not the summoned felguards.
+    if (!bot->IsInCombat() || !PlayerbotAI::IsDps(bot))
+        return false;
+
+    if (IsTargetingLiveWarlock(botAI))
+        return true;
+
+    return FindNearestInCombat(bot, BlackwingLairNPCs::NPC_BLACKWING_WARLOCK, BWL_TRASH_SEARCH_RANGE) != nullptr;
+}
+
+bool BwlTechnicianSpreadTrigger::IsActive()
+{
+    // Bomb (22334) is thrown at random raiders; a clumped raid multiplies
+    // every bomb. Melee share their target's reach, so only ranged/healers
+    // spread.
+    if (!bot->IsInCombat() || !PlayerbotAI::IsRanged(bot) || PlayerbotAI::IsTank(bot))
+        return false;
+
+    return FindNearestInCombat(bot, BlackwingLairNPCs::NPC_BLACKWING_TECHNICIAN, BWL_TRASH_SEARCH_RANGE) != nullptr;
+}
+
 bool BwlDeathTalonWyrmguardTankTrigger::IsActive()
 {
     return PlayerbotAI::IsTank(bot) && AI_VALUE2(Unit*, "find target", "death talon wyrmguard");
--- a/src/Ai/Raid/BWL/BWLActionContext.h
+++ b/src/Ai/Raid/BWL/BWLActionContext.h
@@ -29,6 +29,8 @@ public:
 
         creators["bwl use hourglass sand"] = &RaidBwlActionContext::bwl_use_hourglass_sand;
         creators["bwl nefarian fear ward"] = &RaidBwlActionContext::bwl_nefarian_fear_ward;
+        creators["bwl attack warlock pack anchor"] = &RaidBwlActionContext::bwl_attack_warlock_pack_anchor;
+        creators["bwl technician spread"] = &RaidBwlActionContext::bwl_technician_spread;
         creators["bwl death talon wyrmguard tank move away"] = &RaidBwlActionContext::bwl_death_talon_wyrmguard_tank_move_away;
         creators["bwl death talon wyrmguard ranged move away"] = &RaidBwlActionContext::bwl_death_talon_wyrmguard_ranged_move_away;
     }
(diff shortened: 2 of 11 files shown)


NoxMax:
Notwithstanding the conflicts to resolve (probably from the include ordering PR) for the future when you make a branch for a PR, make it out of test-staging. That way it's most up to date, and doesn't have the merge commits from master.

raijyan:
Sorry, was used to my work where everything is worktrees automatically off the current main. Hopefully resolved?

dillyns:
Just curious why you wouldn't focus down the techs first before switching to warlocks. Normally the techs would get kited by some ranged dps if they arent going to get bombed first, but thats probably beyond the bots capaabilities.

raijyan:
The risk of the technician bombs was entirely defused (pun intended) by the small spread and the mages were already spam sheeping them which I thought might cause issues on a fixed target and the warlocks are also immune to sheep/etc. 
On my first run in bad gear I found two issues - first the bombs exploded my entire raid (full stack = guaranteed overlap), and you might not kill them fast enough to prevent that. Second I was guaranteed to have multiple portals open each would further split the bots, allowing even more portals to open. At worst on a very bad pull I ended up with 10-15 felguards, which meant even after a run back I had to wait several minutes for them to all despawn before I could safely pull again. 
The felguards also have very high HP pools so any that slip through significantly lengthens the encounter!

I did an an extra test (completely intentionally) of pulling two packs at once and ended up still with a very clean kill and zero felguards.

src/Ai/Raid/BWL/BWLTriggers.cpp
```diff
@@ -87,6 +87,33 @@ bool BwlNefarianFearWardTrigger::IsActive()
 
 // Trash
 
+static constexpr float BWL_TRASH_SEARCH_RANGE = 35.0f;
```
kadeshar, line 90: 3 times same variable. Unify that

kadeshar (changes requested):

PR #2648, opened 2026-08-09
fix(dungeons): stop the generic cc strategy from firing in 5-man dung…

jrad7:
## Pull Request Description

Bots in dungeons break CC. One bot sheeps a mob, the next bot hits it, sheep heals the mob, mage keeps
spamming sheep. In 5 man dungeons I've never seen CC be useful, as it is simply not yet supported in a generic way
with the "cc" combat strat.

So this PR turns off the generic `cc` strategy while a bot is in a 5-man.

I did it by having `Engine::Init()` skip strategies that say they're suppressed:

```cpp
Strategy* strategy = i->second;
if (strategy->IsSuppressed())
    continue;
```

and one implementation on the base class:

```cpp
bool Strategy::IsSuppressed() { return botAI->IsInNonRaidDungeon() && getName() == "cc"; }
```

Nothing gets added to or removed from the engine, which matters for a few reasons:

* Open world, raids and BGs behave exactly like they do now.
* `co +cc` / `co -cc` still work normally and the saved strategy string is never touched, so a master's
  setting can't get clobbered.
* Instance strategies (`tbc-mgt`, `wotlk-uk`, etc) have their own CC triggers under their own names, so
  a dungeon script that wants a sheep still gets one.
* No new call sites needed. `Init()` already runs on login, on strategy changes, on combat state flips,
  and after a far teleport through `Reset(true)`.

Second thing in here: Dragon's Breath and Blast Wave move out of `MageCcStrategy` into
`FireMageStrategy` and `FrostFireMageStrategy`. They're panic buttons for when something is in your
face, not crowd control, and they were fire-spec only anyway. `AiFactory` adds `fire`/`frostfire` for
the same spec the old `MAGE_TAB_FIRE` check tested, so the same bots end up with the same triggers.

What actually goes quiet in a 5-man:

| Class | Spells |
|---------|--------------------------------------|
| Mage | Polymorph |
| Priest | Shackle Undead |
| Paladin | Turn Undead |
| Hunter | Scare Beast, Freezing Trap |
| Rogue | Stealth into Sap |
| Druid | Cyclone, Hibernate, Entangling Roots |
| Warlock | Banish, Fear |

Warrior, Shaman and DK don't have a `cc` strategy so nothing changes for them.

Heads up on one thing: Banish and Cyclone make the target immune to damage, so bots can't actually break
those two. My reasoning above doesn't really cover them. I still think they're not worth casting in a
5-man, but if you'd rather I leave them alone I can.

## Feature Evaluation

Minimum logic: one `continue` in the `Engine::Init()` loop, one virtual on `Strategy`, and one map check
on `PlayerbotAI`. 48 lines added across 8 files, 5 of those lines in `src/Bot`. No new members, no new
state to keep in sync, no config read, nothing added to the trigger or action tables.

I originally wrote this the obvious way, pulling `cc` off the engine on the way into a dungeon and
putting it back on the way out. That needed two bools, five call sites, and a fix in
`PlayerbotRepository::Save` so the suppression didn't get written into the bot's saved strategy list. It
was more code and it had a bug where a bot logging in inside a dungeon could get its saved `-cc`
overwritten. This version has neither problem.

Cost: nothing on the tick path. `IsSuppressed()` is only called from `Engine::Init()`, and the tick path
(`UpdateAIInternal` -> `DoNextAction` -> `currentEngine->DoNextAction()`) never calls `Init()`. A bot
sitting in a dungeon pays nothing per tick.

Per `Init()` it's one call per strategy, so about 15-20 on a combat engine. I put the cheap check first
on purpose: `IsInNonRaidDungeon()` is `sMapStore.LookupEntry(id)`, a bounds check and an array read, no
allocation. Outside a dungeon it stops there and never calls `getName()`, which is where most bots are
most of the time. Inside a dungeon you also get one short string per strategy for the name compare.

For context, `Init()` already deletes and reallocates every `TriggerNode` and `Multiplier` on the
engine, so 18 array reads on top of that isn't going to show up.

`Init()` doesn't run any more often than it did before. It fires on login, on each combat/non-combat/dead
engine switch, on `co +x` / `-x`, on spec change or randomize, and after a far teleport. I specifically
did not add an `Init()` call in `HandleTeleportAck`, because `Reset(true)` a few lines down already
re-inits all three engines and doing it twice per teleport would be wasteful with a lot of random bots
zipping around.

## How to Test the Changes

You need a mage in the party, and humanoid mobs since that's what Polymorph works on.

1. Outdoors, pull a humanoid pack. Mage should sheep one. `co ?` shows `cc`.
2. Go into any 5-man and pull a humanoid pack. No sheep. `co ?` still shows `cc`, because the strategy
   is still on the engine, it's just not doing anything.
3. Leave and pull outdoors again. Sheep comes back on its own, no `co +cc`, no relog.
4. Do step 1 again inside a raid and inside a BG. Should be identical to current test-staging.
5. Master settings: `co -cc` outside, zone into a dungeon, zone out, `cc` should still be off. Then
   `co +cc` outside, log out inside a dungeon, log back in, leave, `cc` should still be on. This is the
   case my first attempt got wrong so it's worth checking.
6. Heroic is the same map type so it behaves the same. Quick spot check.
7. As fire or frostfire mage, let something get in melee range in a dungeon. Dragon's Breath and Blast
   Wave should still go off.

## Impact Assessment

    - - [x] No, not at all

Nothing on the tick path, and `Init()` isn't called any more often than before. The added work scales
with `Init()` calls, not with ticks or party size or number of targets. In a dungeon the engine actually
builds fewer triggers than it does today, since the `cc` ones get skipped.

    - - [ ] No
    - - [x] Yes (**explain why**)

Yes, that's the whole point. By default bots stop casting the spells in the table above while in a
non-raid dungeon. Open world, raids, BGs and arenas are unchanged, and so is every instance strategy.

There's no config option for it right now. I left it as a plain default because a CC the group instantly
breaks is a bug everywhere it happens, not a preference. If you want a toggle I'll add one, see the
notes at the bottom.

    - - [ ] No

One branch in `Engine::Init()` and one virtual on `Strategy`. Two things I'd want a second opinion on:

1. `Strategy::IsSuppressed()` sounds generic but the only implementation hardcodes `"cc"`, so the base
   class now knows about one specific strategy name. The alternative was a guard clause in all seven
   `*CcStrategy::InitTriggers`, which I had working first. I went with the single place because any
   future `cc` strategy gets it for free, but I'm not attached to it.
2. The `continue` skips the whole strategy, not just its triggers, so it also skips `GetType()`,
   `HasTargetExclusions()`, `InitMultipliers()` and `actionNodeFactories`. I checked all seven `cc`
   strategies and none of them use any of that (all have `: Strategy(botAI) {}` constructors and no
   overrides), so right now it's the same thing. If someone later adds a multiplier or a factory to a
   `cc` strategy it would quietly get dropped.

## AI Assistance

- - [ ] No

Yes, for review and cleanup passes.

It went through a few rewrites. First version added and removed the strategy from the engine, which had
the save bug I mentioned above. Second version put a guard clause in each of the seven cc strategies,
which worked but was seven copies of the same line. Third is what's here. A perf pass after that caught
a redundant `Engine::Init()` I'd added in `HandleTeleportAck` and got the `&&` in `IsSuppressed()` the
right way round so the cheap check runs first.

I went through the call sites myself for the frequency and cost claims above rather than taking them on
faith.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

8 files, +48 / -8:

| File | What |
|---|---|
| `src/Bot/Engine/Engine.cpp` | skip suppressed strategies in `Init()` |
| `src/Bot/Engine/Strategy/Strategy.h` | `virtual bool IsSuppressed();` |
| `src/Bot/Engine/Strategy/Strategy.cpp` | the one line implementation |
| `src/Bot/PlayerbotAI.h` / `.cpp` | `IsInNonRaidDungeon()` and the `DBCStores.h` include |
| `src/Ai/Class/Mage/Strategy/GenericMageStrategy.cpp` | Dragon's Breath / Blast Wave out |
| `src/Ai/Class/Mage/Strategy/FireMageStrategy.cpp` | and in |
| `src/Ai/Class/Mage/Strategy/FrostFireMageStrategy.cpp` | same |

On the map check, `MapEntry::IsNonRaidDungeon()` is `map_type == MAP_INSTANCE`. That's per map, not per
difficulty, so heroics are covered by the same check and raids, BGs and arenas are excluded.

If you want this behind a config, I'd suggest:

```
AiPlayerbot.DungeonSuppressedStrategies = "cc"
```

and have `IsSuppressed()` check that set instead of the literal. That gets the hardcoded `"cc"` out of
`Strategy` too, and an empty string gives you current behavior back. Happy to push that if you'd prefer
it.

Also happy to split the mage Dragon's Breath / Blast Wave bit into its own PR if you'd rather look at
them separately, it stands on its own either way.

### DIFF
--- a/src/Ai/Base/Trigger/GenericTriggers.cpp
+++ b/src/Ai/Base/Trigger/GenericTriggers.cpp
@@ -577,7 +577,23 @@ bool IsNotFacingTargetTrigger::IsActive()
 
 bool HasCcTargetTrigger::IsActive()
 {
-    return AI_VALUE2(Unit*, "cc target", getName()) && !AI_VALUE2(Unit*, "current cc target", getName());
+    Unit* rtiCcTarget = nullptr;
+    if (botAI->IsInNonRaidDungeon())
+    {
+        rtiCcTarget = AI_VALUE(Unit*, "rti cc target");
+        if (!rtiCcTarget)
+            return false;
+    }
+
+    return IsCcTargetFree(AI_VALUE2(Unit*, "cc target", getName()), rtiCcTarget);
+}
+
+bool HasCcTargetTrigger::IsCcTargetFree(Unit* ccTarget, Unit* rtiCcTarget)
+{
+    if (!ccTarget || (rtiCcTarget && ccTarget != rtiCcTarget))
+        return false;
+
+    return !AI_VALUE2(Unit*, "current cc target", getName());
 }
 
 bool NoMovementTrigger::IsActive() { return !AI_VALUE2(bool, "moving", "self target"); }
--- a/src/Ai/Base/Trigger/GenericTriggers.h
+++ b/src/Ai/Base/Trigger/GenericTriggers.h
@@ -686,6 +686,9 @@ public:
     HasCcTargetTrigger(PlayerbotAI* botAI, std::string const name) : Trigger(botAI, name) {}
 
     bool IsActive() override;
+
+protected:
+    bool IsCcTargetFree(Unit* ccTarget, Unit* rtiCcTarget);
 };
 
 class NoMovementTrigger : public Trigger
--- a/src/Ai/Base/Trigger/RtiTriggers.cpp
+++ b/src/Ai/Base/Trigger/RtiTriggers.cpp
@@ -30,7 +30,7 @@ bool RtiCcTrigger::IsActive()
 
     Unit* ccTarget = AI_VALUE2(Unit*, "cc target", getName());
     if (ccTarget && ccTarget == rtiCcTarget)
-        return HasCcTargetTrigger::IsActive();
+        return IsCcTargetFree(ccTarget, rtiCcTarget);
 
     return botAI->CanCastSpell(getName(), rtiCcTarget);
 }
--- a/src/Ai/Class/Mage/Strategy/FireMageStrategy.cpp
+++ b/src/Ai/Class/Mage/Strategy/FireMageStrategy.cpp
@@ -56,6 +56,23 @@ void FireMageStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
             }
         )
     );
+
+    triggers.push_back(
+        new TriggerNode(
+            "enemy too close for spell",
+            {
+                NextAction("dragon's breath", ACTION_INTERRUPT + 1)
+            }
+        )
+    );
+    triggers.push_back(
+        new TriggerNode(
+            "enemy is close",
+            {
+                NextAction("blast wave", ACTION_INTERRUPT)
+            }
+        )
+    );
 }
 
 // Combat strategy to run to melee for Dragon's Breath and Blast Wave
--- a/src/Ai/Class/Mage/Strategy/FrostFireMageStrategy.cpp
+++ b/src/Ai/Class/Mage/Strategy/FrostFireMageStrategy.cpp
@@ -54,4 +54,21 @@ void FrostFireMageStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
             }
         )
     );
+
+    triggers.push_back(
+        new TriggerNode(
+            "enemy too close for spell",
+            {
+                NextAction("dragon's breath", ACTION_INTERRUPT + 1)
+            }
+        )
+    );
+    triggers.push_back(
+        new TriggerNode(
+            "enemy is close",
+            {
+                NextAction("blast wave", ACTION_INTERRUPT)
+            }
+        )
+    );
 }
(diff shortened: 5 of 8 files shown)


brighton-chi:
Do Banish and Cyclone still cast in instances if the RTI cc mark is placed on a target?

jrad7:
yes it does block the rti cc mark in 5 mans, re-working this so we can keep that behavior

jrad7:
rti moon targets work again, this moved everything into the trigger, which sounds scary, as it is run every tick.

But this should be an overall small performance boost everywhere.

The question we are answering here is:  "Which enemy should I crowd control?" ... to get that answer, it has to walk every mob in the fight and run a "can I cast" check on each.

the RtiCcTrigger's ( banish/cyclone) used to ask the question twice in a row by accident. Now it is asked once, and that is a small savings everywhere.

Inside dungeons, we now ask the cheap question first: "Did someone mark a target?" ... usually there is no target marked, so we rarely ever have to ask the more expensive question, "Which enemy should I crowd control?"

kadeshar (approved):

PR #2653, opened 2026-08-11
Add Ulduar Xt-002 boss strategy

karlverik:
## Pull Request Description

Add Ulduar Xt-002 boss strategy.

Since XT-002 Deconstructor currently lacks a boss strategy, bots are unable to progress further in Ulduar. This PR adds the XT-002 boss strategy to enable proper progression.

### Test Video
[Youtube](https://youtu.be/Vlk47QwQfHk)

## Feature Evaluation

Added the boss strategy for XT-002 Deconstructor. Logic includes basic  triggers,only include light and gravity bomb positioning checks, and add priority  target. 

Zero cost outside combat with XT-002. During combat, targeting and standard positioning are simple O(1) checks . The only dynamic scan is a 100-yard Grid lookup for Void Zone NPCs inside the Gravity Bomb movement action, which executes only for the single bot affected by the debuff, keeping the overall performance cost negligible across hundreds of bots.

## How to Test the Changes

**1.**  Form a raid group with playerbots and enter Ulduar.

**2.** Engage XT-002 Deconstructor.

**3.** Observe bot behavior during the encounter:

**4.** Check if Main Tank and ranged bots take their designated initial positions.

**5.**  Verify that bots afflicted with Searing Light and Gravity Bomb move away from the group correctly, and that Gravity Bomb targets actively avoid existing Void Zones.

**6.**  Verify targeting priority when ads spawn (Life Spark > Boombot > Scrapbot > Pummeller > Heart > Boss).

## Impact Assessment

    - - [ ] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Used Claude for code review and assisting with documentation. All implementation logic was thoroughly reviewed, verified, and tested prior to submission.

 MaNGOS, another module)?

## Final Checklist


noisiver:
Isn't the recommended way to handle targeting to use marks? Also, the heart, how exactly does it work out? If you nuke it down you initiate hard mode. Are you forcing one or the other?

brighton-chi:
I have zero knowledge of this fight, but my preferred approach these days is not to use marks, except skull/cross for guidance but with their targeting effect disabled. That makes it so that you don't have to change players' RTI assignments, and also I think it confuses people to use too many marks because they don't know what they mean (whereas skull/cross have common meaning).

That being said, if you don't use marks, just issuing attack can make bots oscillate targets because it fights with their normal targeting logic. The simplest way is to disable, via multiplier, DpsAssistAction and/or TankAssist. If you want to disable dotting up secondary targets, you would separately disable CastDebuffSpellOnAttackerAction. Then if you want to disable AoE, you need to separately disable Action::ActionThreatType::Aoe.

brighton-chi:
Alternatively, if you need to disable acquisition of some targets but still want to allow bots to use normal targeting logic between other targets, you can use AppendTargetExclusions() to filter out whatever you don't want them to consider. That's the simplest way to have tanks still pick up multiple targets while ignoring specified ones.

karlverik:
> Isn't the recommended way to handle targeting to use marks? Also, the heart, how exactly does it work out? If you nuke it down you initiate hard mode. Are you forcing one or the other?

I'm quite new to C++ and Playerbots, so I hadn't considered using standard target assignments. I'm converting this PR to a draft for now. I plan to study guides for other raid encounters as well and rewrite this.

Destroying the Heart is indeed what triggers Hard Mode. Mechanically, the boss behaves almost the same in both Normal and Hard Mode, but I didn't take into account the risk of failing to destroy the Heart during the first time and accidentally breaking it in a later , which would reset the boss back to full health.

karlverik:
> I have zero knowledge of this fight, but my preferred approach these days is not to use marks, except skull/cross for guidance but with their targeting effect disabled. That makes it so that you don't have to change players' RTI assignments, and also I think it confuses people to use too many marks because they don't know what they mean (whereas skull/cross have common meaning).
> 
> That being said, if you don't use marks, just issuing attack can make bots oscillate targets because it fights with their normal targeting logic. The simplest way is to disable, via multiplier, DpsAssistAction and/or TankAssist. If you want to disable dotting up secondary targets, you would separately disable CastDebuffSpellOnAttackerAction. Then if you want to disable AoE, you need to separately disable Action::ActionThreatType::Aoe.

My intention was to have ranged bots prioritize adds within their range. Since the ranged stack stays in a fixed position, I used attack range to restrict their targeting. The reason I didn't use multipliers is that multipliers aren't registered for the Ulduar raid as a whole, so I went with this approach for now.

And i have a question: if no target is specified, do bots naturally prioritize units with lower health? In the Ignis encounter earlier, I noticed bots were attacking the summoned even though no combat strategy was for Ignis.

brighton-chi:
> > I have zero knowledge of this fight, but my preferred approach these days is not to use marks, except skull/cross for guidance but with their targeting effect disabled. That makes it so that you don't have to change players' RTI assignments, and also I think it confuses people to use too many marks because they don't know what they mean (whereas skull/cross have common meaning).
> > That being said, if you don't use marks, just issuing attack can make bots oscillate targets because it fights with their normal targeting logic. The simplest way is to disable, via multiplier, DpsAssistAction and/or TankAssist. If you want to disable dotting up secondary targets, you would separately disable CastDebuffSpellOnAttackerAction. Then if you want to disable AoE, you need to separately disable Action::ActionThreatType::Aoe.
> 
> My intention was to have ranged bots prioritize adds within their range. Since the ranged stack stays in a fixed position, I used attack range to restrict their targeting. The reason I didn't use multipliers is that multipliers aren't registered for the Ulduar raid as a whole, so I went with this approach for now.
> 
> And i have a question: if no target is specified, do bots naturally prioritize units with lower health? In the Ignis encounter earlier, I noticed bots were attacking the summoned even though no combat strategy was for Ignis.

Yeah, Gromleq somehow did all of Ulduar without knowing anything about multipliers. Kind of impressive actually. But when it is redone, a multiplier system does need to be implemented for sure. I'm not sure if you're planning a greater overhaul or what. I think some bosses, particularly Yogg-Saron, rely heavily on cheats. And Algalon is not implemented.

Target prioritization differs a bit between ranged, combo melee, and other melee. Generally speaking, they do prioritize lowest health, though range also plays a factor (like melee bots will go for lowest health in melee range rather than moving to attack a lower health target not in melee range). I think casters also do some sort of calculation of how fast a target is dying and pull off when it is about to die, presumably so they don't waste a cast.

PR #2654, opened 2026-08-12
Implement free slot check for random battlegrounds

CheckYourFax:
## Pull Request Description

Adds a minimal Random Battleground capacity check to `BattleGroundJoinAction`.

Random Battlegrounds use `BATTLEGROUND_RB` as their queue/template type. This is not re-evaluated later. The existing bot queue calculation therefore continues to use the RB template's 10v10 capacity instead of the real capacity of the battleground.

This change checks existing active Random BG instances and uses `GetFreeSlotsForTeam()` to determine whether the bot can fill an available slot.

The existing queue logic is unchanged for all other BG types since those have a consistent static `TeamSize`. (read note at bottom of PR)

## Feature Evaluation

    - For `BATTLEGROUND_QUEUE_RB`, inspect the already-active battleground instances.
    - Select instances whose `GetBgTypeID()` is `BATTLEGROUND_RB` and whose bracket matches.
    - If an existing Random BG has a free slot for the bot's team, allow the bot to join.
    - If no existing Random BG has free capacity, fall through to the existing queue logic unchanged. This means the queu
    - No map lookup is necessary because the actual map/type is irrelevant; `GetFreeSlotsForTeam()` provides the capacity needed by the decision.

    - The additional work is a small iteration over the currently active battleground instances, not over players or bots.
    - Each matching instance performs two inexpensive checks and one `GetFreeSlotsForTeam()` call.
    - The number of active battleground instances is expected to be very small compared with the number of playerbots.
    - No database access, player iteration, allocation, or map lookup is introduced.
    - The loop exits immediately when a matching Random BG with available capacity is found.

## How to Test the Changes

1. Start the server with the modified playerbot module.
2. Queue a real player for a Random Battleground.
3. Allow the playerbot system to populate the Random BG.
4. Verify that once AzerothCore creates the actual Random BG instance, bots continue filling available team slots instead of using the RB template's 10v10 capacity.
5. Test with a Random BG that has more than a 10v10 capacity.
6. Verify that a full Random BG does not cause bots to treat its RB template capacity as additional available slots.
7. If multiple Random BG instances are active, verify that bots can fill free slots in any matching instance.
8. Verify that the existing non-Random-BG queue behavior is unchanged.
9. Verify no extraneous BG instances are spawned without any real players.

Expected behavior:

- Before an instance exists, the existing Random BG queue logic is unchanged.
- After a Random BG instance exists, `GetFreeSlotsForTeam()` determines whether there is capacity for the bot's team.
- A full existing Random BG is not treated as having additional 10v10 template slots.
- Non-Random BG behavior remains unchanged.

## Impact Assessment

    - [x] No, not at all

  The check runs as part of the existing battleground queue evaluation, which already runs approximately every 35 seconds. It does not introduce a new timer, trigger, or per-bot tick. The additional work only iterates over the currently active battleground instances and checks their available slots. It does not iterate over players or bots and performs no database or map queries.

    - [x] No
    - [ ] Yes (**explain why**)

    The change only corrects Random BG capacity handling when an actual Random BG instance already exists. The existing queue behavior is preserved.

    - [x] No
    - [ ] Yes (**explain below**)

    A small Random BG-specific check is added without changing the existing queue calculation or introducing new state.

## AI Assistance

- [ ] No
- [x] Yes (**explain below**)

AI was used for source-code analysis, comparing the relevant `mod-playerbots` and AzerothCore battleground implementations, identifying the appropriate existing APIs, and reviewing the proposed minimal change.

The final implementation was reviewed against the source code and intentionally kept minimal. The contributor understands and owns the submitted change.

MaNGOS, another module)?
- [ ] Yes (**name the project and the original author(s) below**)

The implementation is original. It uses existing APIs from the AzerothCore  (`GetActiveBattlegrounds()`, `GetBgTypeID()`, and `GetFreeSlotsForTeam()`), but no code was copied or ported from another project.

## Final Checklist

- [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).

## Notes for Reviewers

One design question for review: Should we determine available capacity directly with `GetFreeSlotsForTeam()`, or should we first resolve the actual battleground type with `GetBgTypeID(true)` and handle capacity based on the resulting BG type?

All PR text above is AI-assisted (manually checked and edited).

**Last note:** Maybe `GetFreeSlotsForTeam()` can be used for all queue types? It would make the bot queue behavior much more consistent, since `GetFreeSlotsForTeam()` also takes invited but not-yet-accepted players into account. That seems more robust than using a fixed `TeamSize` value.

### DIFF
--- a/src/Ai/Base/Actions/BattleGroundJoinAction.cpp
+++ b/src/Ai/Base/Actions/BattleGroundJoinAction.cpp
@@ -306,9 +306,6 @@ bool BGJoinAction::shouldJoinBg(BattlegroundQueueTypeId queueTypeId, Battlegroun
             if (bg->GetFreeSlotsForTeam(teamId) > 0)
                 return true;
         }
-
-        // All RB instances are full. If a real player is waiting, allow bots to populate a new queue.
-        return activeBgQueue != 0;
     }
 
     if (teamId == TEAM_ALLIANCE)
--- a/src/Ai/Base/Actions/BattleGroundJoinAction.cpp
+++ b/src/Ai/Base/Actions/BattleGroundJoinAction.cpp
@@ -294,6 +294,21 @@ bool BGJoinAction::shouldJoinBg(BattlegroundQueueTypeId queueTypeId, Battlegroun
     uint32 activeBgQueue = sRandomPlayerbotMgr.BattlegroundData[queueTypeId][bracketId].activeBgQueue;
     uint32 bgInstanceCount = sRandomPlayerbotMgr.BattlegroundData[queueTypeId][bracketId].bgInstanceCount;
 
+    // Check if queue type is random, they have no static TeamSize.
+    if (queueTypeId == BATTLEGROUND_QUEUE_RB)
+    {
+        for (Battleground const* bg : sBattlegroundMgr->GetActiveBattlegrounds())
+        {
+            if (bg->GetBgTypeID() != BATTLEGROUND_RB ||
+                bg->GetBracketId() != bracketId)
+                continue;
+
+            if (bg->GetFreeSlotsForTeam(teamId) > 0)
+                return true;
+        }
+    }
+
+    
     if (teamId == TEAM_ALLIANCE)
     {
         if ((bgAllianceBotCount + bgAlliancePlayerCount) < TeamSize * (activeBgQueue + bgInstanceCount))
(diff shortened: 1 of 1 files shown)


CheckYourFax:
Current PR has one issue: It only works properly on  1 active RBG. afterwards bots stop queuing for any player making a new random BG. Only when the first RBG is over do bots queue again. I can't find a reliable fix.

specific BGs are unaffected.

It is better than the current implementation which is just entirely broken, but any help fixing the ability to have multiple RBG's populate properly without bots flooding the queue would be nice.

CheckYourFax:
I'm going with a different approach that doesn't use the free slot check since it doesn't work correctly with the current bookkeeping code in RandomPlayerbotmgr.cpp

It would need a full rewrite that doesn't depend on static values and just use `GetFreeSlotsForTeam()` to simplify whether a bot should join a BG queue. For now a simpler bug fix is warranted.

For now I'm going to go with a simpler fix.

PR #2658, opened 2026-08-14
Rename, Move, and Modify RaidBossHelpers

brighton-chi:
## Pull Request Description

Since RaidBossHelpers is also used for dungeons and trash strategies, I've renamed it EncounterHelpers and moved it to the /src/Util/ folder. I wrapped the functions in a new EncounterHelpers namespace and then updated includes and namespace usage where any relevant functions are used.

Other changes:

- Added boolean functions for IsDpsCooldownAction, IsTauntAction, and IsAoeThreatAction for usage in multipliers. This can cut down a lot on includes in multiplier files in addition to making the code cleaner. I implemented them for Black Temple as an example since BT was already using the constituent actions, and I intend to use them in other strategies as well going forward.
- Cleaned up SetRtiTarget, as half the function was inoperative, and removed a parameter and updated call sites. I eventually intend to get rid of forcing RTI changes entirely and probably doing away with this helper at that point.
- Removed unneeded PlayerbotAI* botAI parameters from GetGroupMainTank and GetGroupAssistTank and updated call sites accordingly

## Feature Evaluation

Minimum logic is described above. Should be no relevant impact on processing cost, absent a tiny bit of improved efficiency with the BT multipliers.

## How to Test the Changes

Right now, the only potential functional regression would be in BT so if there are problems noticed there it would be worth flagging.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [ ] No

I've never really understood this question, but this does require using the namespace for the helpers going forward (either with using namespace like me because I'm lazy or each time you call a function from the helper file). It's necessary though as the file grows to cordon off all the classless functions.

## AI Assistance

- - [ ] No

Assistance with modifying SetRtiTarget

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Raid/BT/BTMultipliers.cpp
+++ b/src/Ai/Raid/BT/BTMultipliers.cpp
@@ -8,66 +8,51 @@
 #include "BTActions.h"
 #include "BTHelpers.h"
 #include "ChooseTargetActions.h"
-#include "DKActions.h"
-#include "DruidActions.h"
-#include "DruidBearActions.h"
 #include "DruidShapeshiftActions.h"
+#include "EncounterHelpers.h"
 #include "FollowActions.h"
 #include "HunterActions.h"
 #include "MageActions.h"
-#include "PaladinActions.h"
 #include "PriestActions.h"
 #include "ReachTargetActions.h"
 #include "RogueActions.h"
 #include "ShamanActions.h"
-#include "WarlockActions.h"
-#include "WarriorActions.h"
 #include "WipeAction.h"
+#include <array>
+#include <ctime>
 
 using namespace BlackTempleHelpers;
+using namespace EncounterHelpers;
 
-static bool IsDpsCooldownAction(Action* action)
+// General
+
+// Illidan is handled separately
+float BlackTempleDelayDpsCooldownsMultiplier::GetValue(Action* action)
 {
-    return dynamic_cast<CastHeroismAction*>(action) ||
-           dynamic_cast<CastBloodlustAction*>(action) ||
-           dynamic_cast<CastMetamorphosisAction*>(action) ||
-           dynamic_cast<CastAdrenalineRushAction*>(action) ||
-           dynamic_cast<CastBladeFlurryAction*>(action) ||
-           dynamic_cast<CastIcyVeinsAction*>(action) ||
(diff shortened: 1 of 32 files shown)


src/Ai/Raid/BT/BTMultipliers.cpp
```diff
@@ -8,67 +8,52 @@
...
-           dynamic_cast<CastAvengingWrathAction*>(action) ||
-           dynamic_cast<CastElementalMasteryAction*>(action) ||
-           dynamic_cast<CastFeralSpiritAction*>(action) ||
-           dynamic_cast<CastFireElementalTotemAction*>(action) ||
-           dynamic_cast<CastFireElementalTotemMeleeAction*>(action) ||
-           dynamic_cast<CastForceOfNatureAction*>(action) ||
-           dynamic_cast<CastArmyOfTheDeadAction*>(action) ||
-           dynamic_cast<CastSummonGargoyleAction*>(action) ||
-           dynamic_cast<CastBerserkingAction*>(action) ||
-           dynamic_cast<CastBloodFuryAction*>(action);
```
kadeshar, line 52: You dont move that to new IsDpsCooldownAction?

kadeshar:
Conflict to resolve

src/Ai/Raid/BT/BTMultipliers.cpp
brighton-chi: Are you referring to the racials? I forgot about those. I'll add them.

brighton-chi: Did the update on my phone. I think this should work but lmk if I messed something up.

kadeshar: You fix code on phone? xD For me looks ok. Waiting only on checks.

brighton-chi: Not at home and I'm motivated to get this one merged so I can update Sunwell and TK and then PR a rewrite of Hyjal that I'm done with.

I'm starting to get into using values, though I definitely can't without AI right now. But I think that will be good as performance is always a big consideration with strategies.

kadeshar (approved):

PR #2663, opened 2026-08-15
feat(mgt): add the tbc-mgt strategy for Magisters' Terrace

jrad7:
## Pull Request Description

Adds a dungeon strategy for Magisters' Terrace, covering the Sunblade and Wretched trash and all three fights that have a mechanic: Selin Fireheart, Priestess Delrissa, and Kael'thas Sunstrider. Vexallus needs nothing beyond the default rotations.

By default a bot group loses the instance in the same three places. It lets Selin drain his Fel Crystals and heal mana to full. It lets Delrissa's retinue heal, fear and polymorph its way through the party while the bots spread damage across every add. And it stands in everything Kael'thas puts on the floor, chases the Phoenix out of the fight, and has no answer to Gravity Lapse.

**Selin Fireheart and the Sunblade trash**

- Selin drains a Fel Crystal to refill his mana and follow with Fel Explosion, so the active crystal is killed at emergency priority. While one stands, assist, attacker debuffs and DPS AoE are zeroed for everyone but the healer, so the party burns it instead of trickling onto it.
- Selin evades if the fight leaves his chamber. Flee and run-away are zeroed inside the room, and a bot that ends up past the door walks itself back in a step at a time.
- Sunblade Physicians drop a Magic Dampening Field that silences casting inside 5yd. Bots step clear of it; a tank only moves far enough to keep hold of the pull.
- Sunblade Magisters cast Arcane Nova centred on themselves, so ranged bots clear 16yd. Melee stay in, as players do.
- A Sunblade Mage Guard more than 12yd from its victim throws glaives instead of meleeing, so the bot it fixed on closes the gap.
- Wretched that reach a crystal and drink Fel Infusion enrage; the tank taunts the enraged one off whoever it grabbed.
- DPS focus fire is ranked by what a mob costs to leave alive (Physician, Blood Knight, Sister of Torment, Warlock, Magister, ...) and the pick is latched until it dies, so bots stop rotating targets every time threat shuffles.
- Interrupts are ranked the same way (Deadly Embrace, Holy Light, Arcane Nova, Forked Lightning, Incinerate, ...). A dedicated interrupt is always allowed; a stun or silence is only spent on the top two casts, so those cooldowns are still up when a Deadly Embrace or a heal lands.

**Priestess Delrissa**

- The retinue gets its own focus order, High Explosive Sheep first, then the healers, then the casters, Delrissa herself last, because the fight is decided by killing the support and not the boss.
- And its own interrupt order (fear and polymorph, then Flash Heal and Lesser Healing Wave, then Summon Imp). Each bot takes a stable slot into that list by group order, so four bots kick four different casters instead of racing for the same one.
- A shaman drops Tremor Totem for the fight and the other earth totems are suppressed, since they share the slot and would otherwise overwrite the only counter to Delrissa's fear.
- The retinue's pets are added to the strategy's target exclusions, so target selection never spends the fight on a respawning pet.

**Kael'thas Sunstrider**

- Flame Strike and the burning Phoenix are read off the trigger creature and the dynamic objects, and bots step out to a scored, reachable spot rather than the generic flee position.
- Melee never target the Phoenix and never chase it: it is excluded from target selection outright, and while it burns across Kael's melee ring the step-out is scored on the ring itself so they keep hitting the boss. Getting behind the target is zeroed while a Phoenix burns there, since it walks a bot straight back through the fire it just left.
- Phoenix Eggs are killed on sight before they rehatch; ranged also finish an engaged Phoenix, and everyone switches to Kael while Shock Barrier is up so the barrier breaks before the Pyroblast lands.
- Kael's interrupts are Pyroblast first, then Fireball, and are skipped entirely under Shock Barrier where they cannot land.
- Kael turns immune between phases; assist, debuffs and DPS AoE are suppressed while he does, so the party does not spin on an untargetable boss.
- Gravity Lapse gets a full station layout: tank and melee hold the two spots by Kael, healer and ranged spread behind, and each bot walks its route in legs, re-deciding per leg. Arcane Spheres are dodged on their predicted sweep with a lead, a sphere that chases is kited by scored turn with commit hysteresis so bots do not jitter between headings, and Power Feedback burn patches are costed into every candidate spot. Formation, follow and reach-target are zeroed for the duration so nothing drags a lapsed bot off its station.

It follows the existing per-dungeon pattern, so it is inert everywhere except inside Magisters' Terrace.

## Feature Evaluation

Sixteen triggers with their actions and ten multipliers, installed only on map 585. Each one exits on the cheapest check first: in combat, then a threat-list or `find target` lookup for the encounter it belongs to, so nothing past the first branch runs during the rest of the instance. Every grid search behind them is shared through a per-bot calculated value on a 200-500ms interval, so the cost on the tick is one lookup and not one search per action evaluated. Hazard sets are gathered once per action and reused across scoring instead of being re-queried per candidate.

Nothing runs outside Magisters' Terrace, so the open-world population is unaffected. Inside, the per-bot per-tick cost is a handful of cached, range-bounded grid searches behind early-outs. The scoring loops (escape spots, lapse routes, sphere dodges) run only for a bot that is actually standing in a hazard or lapsed, for the few bots in the dungeon.

## How to Test the Changes

With a group covering tank, healer, and DPS, clear Magisters' Terrace on Normal and Heroic.

Expected:
- Selin: the party swaps to the active Fel Crystal as soon as it lights up, and no bot flees out of the chamber and resets him.
- Trash: ranged step out of Arcane Nova, everyone steps out of the Dampening Field, a Mage Guard at range gets closed on, an enraged Wretched gets taunted, and the pull dies in priority order rather than whatever each bot happened to tag.
- Delrissa: the sheep dies first, the healers next, Delrissa last; fears, polymorphs and heals get kicked, and different bots kick different casters.
- Kael'thas: no bot stands in a Flame Strike; melee never run off after a Phoenix; eggs die before they hatch; the party switches to Kael under Shock Barrier.
- Gravity Lapse: bots take their stations rather than clumping, keep clear of the Arcane Spheres, and resume the fight when it ends.

## Impact Assessment

    - - [ ] No, not at all

The added work exists only for bots inside Magisters' Terrace, so it does not scale with global bot count. Within the instance it is a small number of cached, bounded grid searches behind early-outs, with the heavier scoring gated on a bot actually being in a hazard.

    - - [ ] No
    - - [x] Yes (**explain why**)

Inside this instance it overrides targeting and movement: focus order and target exclusions during the trash and the Delrissa fight, hazard step-outs and Gravity Lapse stations during Kael'thas, and suppression of the competing movement and AoE actions while those are active. The default behavior loses these fights. The override is confined to map 585 and mirrors the existing per-dungeon strategies.

    - - [ ] No

It adds a self-contained `src/Ai/Dungeon/MgT/` unit plus one registration line each in the existing dungeon strategy, context and `ApplyInstanceStrategies` files. The complexity is isolated to the new directory and follows the established layout.

## AI Assistance

- - [ ] No

Used for code generation, iteration on the escape-spot, lapse-route and sphere-dodge geometry, and documentation. The behavior was refined against the live fights across repeated runs on both difficulties, and all logic has been reviewed and is owned by the contributor.

 MaNGOS, another module)?

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (N/A: no dialogue added.)
- - [x] Any code ported/adapted from another project is attributed. (N/A: all original.)
- - [x] Documentation updated if needed (Conf comments, WiKi commands). (N/A: no config or commands added.)

## Notes for Reviewers

- Prior art, so the patterns here are not new to review: the healing-exempt AoE suppression is the Erekem/Violet Hold shape from #2570; the kite and hazard-scoring geometry is the same approach as the Sepethrea kite in #2570; and the focus exclusions are built on the `AppendTargetExclusions` hook added in #2604.
- `TargetValueExclusionType::Attacker` feeds only `AttackerWithoutAuraTargetValue`. The DPS and tank pickers use their own types and `FindTargetStrategy` uses `None`, so the focus exclusions cannot starve tank targeting or the boss `find target` lookups.
- Bots do not float during Gravity Lapse. The encounter grants flight through the client, and a bot never acks it, so it takes the auras and the damage but stays on the floor. The lapse stations are therefore ground positions, and `LapseActive()` keys on the auras and Kael's channel rather than on any movement flag.
- New files under `src/Ai/Dungeon/MgT/` follow the same convention as the existing dungeon strategy files: GPLv2 header, header guard, no inline comments, and a single anonymous namespace in `MgTShared.cpp` for the entries, spell IDs and tuning constants.
- Melee can run up to four Phoenix grid searches per 200-500ms window during the Kael'thas fight. Folding them behind one shared list value was considered and left out to match the neighbouring strategies, which scan per value; it is an easy follow-up if a maintainer prefers it.
- Expect the usual additive collision with sibling dungeon strategy PRs: one line each in `DungeonStrategyContext.h`,
`TbcDungeonActionContext.h`, `TbcDungeonTriggerContext.h`, `BuildSharedActionContexts.cpp`, `BuildSharedTriggerContexts.cpp`, `BuildSharedValueContexts.cpp`, and the strategy list plus map case in `PlayerbotAI.cpp`.

### DIFF
--- a/src/Ai/Dungeon/MgT/MgTActions.cpp
+++ b/src/Ai/Dungeon/MgT/MgTActions.cpp
@@ -1,0 +1,191 @@
+/*
+ * This file is part of the mod-playerbots module for AzerothCore. See AUTHORS file for Copyright
+ * information; released under GNU GPL v2 license, redistribute/modify under version 2 of the License,
+ * or (at your option) any later version.
+ */
+
+#include "MgTActions.h"
+#include "AiObjectContext.h"
+#include "EncounterHelpers.h"
+#include "MgTShared.h"
+#include "Playerbots.h"
+#include <algorithm>
+#include <cmath>
+#include <string>
+#include <vector>
+
+using namespace EncounterHelpers;
+
+bool MgTEscapeAction::Execute(Event)
+{
+    if (IsWaitingForLastMove(MovementPriority::MOVEMENT_COMBAT))
+        return false;
+
+    auto const& spots = AI_VALUE_REF(MagistersTerrace::EscapeSpots, _value);
+    for (MagistersTerrace::EscapeSpot const& spot : spots)
+    {
+        if (MoveTo(bot->GetMapId(), spot.x, spot.y, spot.z, false, false, false, false,
+                   MovementPriority::MOVEMENT_COMBAT, true, false))
+            return true;
+    }
+
+    return false;
+}
+
+bool MgTFocusAction::Execute(Event)
+{
+    Unit* target = botAI->GetUnit(AI_VALUE(ObjectGuid, _value));
+    if (!target)
+        return false;
+
+    if (IsMechanicTrackerBot(bot, MagistersTerrace::MAP_MAGISTERS_TERRACE))
+        MarkTargetWithSkull(bot, target);
+
+    if (AI_VALUE(Unit*, "current target") == target)
+        return false;
+
+    return Attack(target);
+}
+
+bool MgTKillCrystalAction::Execute(Event)
+{
+    Unit* crystal = botAI->GetUnit(AI_VALUE(ObjectGuid, "mgt crystal target"));
+    if (!crystal || AI_VALUE(Unit*, "current target") == crystal)
+        return false;
+
+    return Attack(crystal);
+}
+
+bool MgTReturnToRoomAction::Execute(Event)
+{
+    float x = MagistersTerrace::SELIN_REGROUP_X;
+    float y = bot->GetPositionY();
+    MagistersTerrace::ClampIntoRoom(x, y);
+
+    float const bx = bot->GetPositionX();
+    float const by = bot->GetPositionY();
+    float const bz = bot->GetPositionZ();
+
+    float const dist = bot->GetExactDist2d(x, y);
+    if (dist < 1.0f)
+        return false;
+
+    float const step = std::min(8.0f, dist);
+    float const moveX = bx + ((x - bx) / dist) * step;
+    float const moveY = by + ((y - by) / dist) * step;
+
+    float moveZ = bot->GetMapHeight(moveX, moveY, bz + MagistersTerrace::GROUND_SEARCH_UP, true,
+                                    MagistersTerrace::GROUND_SEARCH_DOWN);
+    if (std::fabs(moveZ - bz) > MagistersTerrace::GROUND_TIER_STEP)
+        moveZ = bz;
+
+    return MoveTo(bot->GetMapId(), moveX, moveY, moveZ, false, false, false, false,
+                  MovementPriority::MOVEMENT_COMBAT, true, false);
+}
+
+bool MgTCloseOnMageGuardAction::Execute(Event)
+{
+    Unit* guard = botAI->GetUnit(AI_VALUE(ObjectGuid, "mgt mage guard target"));
+    if (!guard)
+        return false;
+
+    return ReachCombatTo(guard, 0.0f);
+}
+
+bool MgTPriorityInterruptAction::Execute(Event)
+{
+    Unit* target = botAI->GetUnit(AI_VALUE(ObjectGuid, "mgt interrupt target"));
+    if (!target)
+        return false;
+
+    uint8 const urgency = MagistersTerrace::GetInterruptUrgency(target);
+    if (!urgency)
+        return false;
+
+    return !MagistersTerrace::CastInterrupt(botAI, target, urgency).empty();
+}
+
+bool MgTTauntEnragedWretchedAction::Execute(Event)
+{
+    Unit* wretched = botAI->GetUnit(AI_VALUE(ObjectGuid, "mgt enraged wretched"));
+    if (!wretched)
+        return false;
+
+    for (std::string const& spell : MagistersTerrace::TauntSpellNames())
+    {
+        if (botAI->CanCastSpell(spell, wretched) && botAI->CastSpell(spell, wretched))
+            return true;
+    }
+
+    return false;
+}
+
+bool MgTDelrissaInterruptAction::Execute(Event)
+{
+    auto const& order = AI_VALUE_REF(GuidVector, "mgt delrissa interrupt order");
+    for (ObjectGuid const guid : order)
+    {
+        Unit* target = botAI->GetUnit(guid);
+        if (!target)
+            continue;
+
+        uint8 const urgency = MagistersTerrace::GetRetinueInterruptUrgency(target);
+        if (!urgency)
+            continue;
+
+        if (!MagistersTerrace::CastInterrupt(botAI, target, urgency).empty())
+            return true;
+    }
+
+    return false;
+}
+
+bool MgTDelrissaTremorTotemAction::Execute(Event)
+{
+    return botAI->CanCastSpell("tremor totem", bot) && botAI->CastSpell("tremor totem", bot);
+}
+
+bool MgTKaelInterruptAction::Execute(Event)
+{
+    Unit* target = MagistersTerrace::GetKaelInterruptTarget(bot);
+    if (!target)
+        return false;
+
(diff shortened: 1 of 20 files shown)


jrad7:
Kael'thas fight: https://www.youtube.com/watch?v=omQdDzfSDD4

Bots are in Sunwell gear, but Kael has x3 HP to make the fight last a couple of gravity lapses.

Celandriel:
RaidBossHelpers no longer exists. Need to change to EncounterHelpers

src/Ai/Dungeon/MgT/MgTStrategy.cpp
```diff
@@ -0,0 +1,93 @@
...
+        NextAction("mgt return to room", ACTION_EMERGENCY + 10) }));
+
+    triggers.push_back(new TriggerNode("mgt crystal active", {
+        NextAction("mgt kill crystal", ACTION_EMERGENCY + 8) }));
+
+    triggers.push_back(new TriggerNode("mgt in dampening field", {
+        NextAction("mgt leave dampening field", ACTION_EMERGENCY + 6) }));
+
+    triggers.push_back(new TriggerNode("mgt priority interrupt", {
+        NextAction("mgt priority interrupt", ACTION_INTERRUPT + 5) }));
```
Celandriel: Any reason for +5 and lower down +6?

src/Ai/Dungeon/MgT/MgTActions.cpp
```diff
@@ -0,0 +1,193 @@
+/*
+ * This file is part of the mod-playerbots module for AzerothCore. See AUTHORS file for Copyright
+ * information; released under GNU GPL v2 license, redistribute/modify under version 2 of the License,
+ * or (at your option) any later version.
+ */
+#include "MgTActions.h"
+
```
brighton-chi, line 6: I think Nox centralized includes so that there aren't any white lines between them, whether the associated header or standard library includes.

jrad7:
rebased and fixed both issues

kadeshar:
Conflicts to resolve

src/Ai/Dungeon/MgT/MgTActions.cpp
```diff
@@ -0,0 +1,194 @@
...
+    bool move = false;
+
+    if (MagistersTerrace::GetGravityLapseKite(bot, world, spot, _kite))
+    {
+        move = true;
+    }
+    else if (MagistersTerrace::GetGravityLapseDodge(bot, world, spot))
+    {
+        move = true;
+    }
```
kadeshar: If MagistersTerrace::GetGravityLapseKite || MagistersTerrace::GetGravityLapseDodge

src/Ai/Dungeon/MgT/MgTShared.cpp
```diff
@@ -0,0 +1,1936 @@
...
+void CollectLapseWorld(Player* bot, LapseWorld& out)
+{
+    std::vector<Unit*> units;
+    CollectCreaturesByEntry(bot, { NPC_ARCANE_SPHERE, NPC_PHOENIX }, std::max(SPHERE_SCAN, PHOENIX_SCAN), units);
+
+    for (Unit* unit : units)
+    {
+        if (unit->GetEntry() == NPC_ARCANE_SPHERE)
+        {
+            if (bot->GetExactDist2d(unit) <= SPHERE_SCAN)
```
kadeshar: Merge if's

src/Ai/Dungeon/MgT/MgTShared.cpp
```diff
@@ -0,0 +1,1936 @@
...
+                float const hy = ax * sinT + ay * cosT;
+
+                float const x = bx + hx * KITE_LEG;
+                float const y = by + hy * KITE_LEG;
+
+                float const runway = KiteRunway(bx, by, hx, hy, healer, outside);
+
+                if (requireFullLeg)
+                {
+                    if (runway < KITE_LEG)
```
kadeshar: Merge if's

jrad7:
merged the if's, nice finds

kadeshar (approved):

PR #2666, opened 2026-08-17
feat: expose non-mutating chat command classifier

FIocker:
Adds a non-mutating PlayerbotAI::IsChatCommand API, using the existing trigger and item-link recognition rules. This lets integrations classify playerbot commands without executing them.


kadeshar:
PR target branch should be test-staging. Fill PR template.

PR #2670, opened 2026-08-18
Configs: Framework to rename configs while honouring the old names

NoxMax:
## Pull Request Description

Playerbots has the word "random" in several configs that are not exclusive to randombot. Simply renaming those to an appropriate name is not viable as it would break the user's custom values in their .conf file. Here I introduce a framework where configs can be renamed, while still honouring the user's value under the old name.

Attached are screenshots showing terminal output, as a result of renaming several configs. The config in this demonstration through these screenshots is the change of `DeleteRandomBotAccounts` to `DeleteBotAccounts`

1. First screenshot, DeleteRandomBotAccounts = 0
2. Second screenshot, DeleteRandomBotAccounts = X (invalid value)
3. Third screenshot, DeleteBotAccounts = X (invalid value)
4. Fourth screenshot, DeleteBotAccounts = 1
5. Fifth screenshot, missing config

Server behaviour is as expected in all cases.

1 DeleteRandomBotAccounts=0
2 DeleteRandomBotAccounts=X
3 DeleteBotAccounts=X
4 DeleteBotAccounts=1
5 Missing Config

## Feature Evaluation

State is at minimum given the objective. In the future we might add a terminal warning for config's in the user's file that are not recognized at all, so the user would get a note about dead non-existing configs such as `AiPlayerbot.GoblinMode = 1`.

## How to Test the Changes

These are the configs that have been renamed in this PR, by dropping "random" from them:
```
RandomBotAccountPrefix
RandomBotAccountCount
DeleteRandomBotAccounts
RandomBotRandomPassword
RandomGearQualityLimit
RandomGearScoreLimit
RandomGearLoweringChance
RandomBotQuestIds
RandomBotQuestItems
RandomBotSpellIds
RandomBotTalk
RandomBotEmote
RandomBotSuggestDungeons
RandomBotSayWithoutMaster
RandomBotShowHelmet
RandomBotShowCloak
```
Whatever your current values are for them in your .conf file, they should still be accepted. And if you remove the word "Random" in .conf, they should continue to work and you don't get a terminal warning.

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [ ] No
Minor. The change of config styles in `PlayerbotAIConfig` should be very obvious to a dev about how to add a new config or rename one.

## AI Assistance

- - [ ] No

Had Claude double check that I am only renaming configs that should actually have "random" dropped from them, no more or less.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

This is part of a larger series of PRs, starting with #2642, to remove the word "random" from where it doesn't belong. Having "random" everywhere is a legacy from an older simpler time when this project was just randombots and nothing else. But today there's also addclass bots, altbots, selfbots, and whatever else will come in the future. This rename direction would take months to accomplish and discussions to be had about renaming things in the codebase, but I believe bit by bit this can be done.


NoxMax:
Updated with the newest configs.

noisiver:
I don't think getting rid of `random` everywhere is the best way to go. It introduces a lot of confusion about what bots are covered in the options, and it goes against what cmangos names the playerbots config options too. They use `random` in the option names that are specifically for random bots.

I feel doubt about adding a framework to handle options that had their name changed. It's just extra code that doesn't do much in the end, options are rarely renamed and when they are the user can be told about it.

I don't know what the other maintainers feel about this but I'd like to know.

NoxMax:
Oh this is not at all about getting rid of "random" everywhere and more about getting rid of instances where it is not just about randombots, account related configs are chief among them. Conversely something like `RandomBotAutologin`, is very much a randombot exclusive config so it doesn't get renamed.

conf/playerbots.conf.dist
```diff
@@ -89,12 +89,12 @@ AiPlayerbot.RandomBotAutologin = 1
 AiPlayerbot.MinRandomBots = 500
 AiPlayerbot.MaxRandomBots = 500
 
-# Randombot accounts
+# Bot accounts
```
kadeshar, line 92: https://github.com/mod-playerbots/mod-playerbots/pull/2642#discussion_r3835834907 I think that also apply here

NoxMax: On that PR I undid all changes to conf.dist, instead having them all here. Just to not mix things up.

kadeshar:
https://github.com/mod-playerbots/mod-playerbots/pull/2642#discussion_r3835834907 That i think apply also here.

NoxMax:
Grom you mentioned that comment last week here too, I am a bit confused at your meaning. #2642 no longer edits any of the configs, I undid that. All the config edits now happen in this PR (#2670) only. Their comments are only editted here, AND their keys are editted (while respecting the old ones) and only the non-randombot-execlusive keys are renamed. The ones that are randombot execlusive are not renamed because they are accurate.

kadeshar:
Im mean changes like this:
obraz
are covered via statement
obraz

NoxMax:
Ah okay, you mean the idea in general. Any reason why not to use a wide term, if the config does indeed cover everything and not just randombots? I mean if it's about all bots, why not just use "bot"?

Also Rev mentioned something about "getting rid of random everywhere", which not at all what is happening here, in case there is this confusion in the discussion among admins.

kadeshar:
Because bot is wide range covering also selfbot and altbot.

NoxMax:
For the account configs, I can get the gray area of it, though I would argue that a "bot account" is clear to be any account that is created by the server, which is what these configs affects, and saying that it's a "randombot account config" is kinda misleading. That it implies an execlusivity to some particular thing when it cover more than that.

But then you have things like `RandomBotQuestIds`, which other than addclass, it also hits selfbots and altbots if you give them `maintenance`.

kadeshar:
Dont extend my comment to all changes in PR. I directly pointed in comment my issue. If config affecting all bots i agree that shouldnt have randombot in name.
obraz

NoxMax:
Far from me to try misconstrue what you say; sorry if it seemed that way. Okay, so what about addressing them categorically, and then and I can drop random from some and leave it in others. There are 16 configs that I renamed here, and I understand that some of them affect two typs of bots, some three, some all four. So they're not on equal grounds.

I can go back and double-check the paths for each according to their effect on randombot, addclass bot, altbot, and/or selfbot. And we can make a better judgment call on them individually. Does that sound good?

PR #2031, opened 2026-01-18
Restore Naxx Strategies without core dependencies

Wishmaster117:
### Summary
This PR restores the Naxxramas raid strategies that were removed in commit 686fe513b25bbb20ccfcc89f08ee8c4b498a263e .
The reintroduced logic is core‑friendly (no AzerothCore script headers or internal boss AI/EventMap dependencies), and the Naxxramas actions have been refactored into per‑boss files for better maintainability.

### Motivation
The previous removal was meant to avoid core modifications and unblock upstreaming.
This PR brings the strategies back while adhering to that requirement, using only observable state and mod‑playerbots helpers.

### What’s included

- Re‑enabled the Naxxramas strategies previously removed.
- Replaced core script header dependencies with observable checks (auras, casts, unit flags, flight state, etc.).
- Split the Naxxramas action logic into per‑boss source files to avoid a “god file” and ease future maintenance.
- Minor, non‑intrusive behavior improvements aligned with existing helpers.

### Future work
Some strategies may still require refinement or more advanced handling later.
This PR focuses on restoring the baseline logic without core dependencies, while keeping changes minimal and safe.

**Any contributions are welcome to further improve and fine‑tune the Naxxramas strategies.**

### Testing
Tested in some Naxx boxx.
No server crash and boss killed :D

Note: I'll make another PR with revised scripts when this one are merged

PR #2671, opened 2026-08-18
Naxx: re-enable Heigan safety dance with a time-based fight clock

jbmartino:
## Pull Request Description

Re-enables the Heigan the Unclean "safety dance" in the Naxxramas strategy, which has shipped commented out since #2031 (and was removed before that in #1961 because it depended on core script internals).

The #2031 rewrite tried to advance the safe zone whenever it saw the **boss** cast Eruption (29371). That can never fire: in AzerothCore the eruption is cast by the floor GameObjects (`instance_naxxramas.cpp` `HeiganEruptSections` -> `go->CastSpell(nullptr, SPELL_ERUPTION)`), never by Heigan himself. So bots stood in the eruptions.

This PR keeps the "observable state only, no core changes" constraint and instead drives the dance from a fight clock:

- `boss_heigan.cpp` uses fixed constants (identical for 10/25): slow dance 90 s with eruptions at 15 s then every 10 s (8 total); fast dance 45 s with eruptions at 7 s then every 4 s (10 total); safe section bounces 3,2,1,0,1,2,3,...
- During the fast dance Heigan channels **Plague Cloud (29350)**, a 45 s self-aura applied 1 s after his teleport. Its remaining duration is an exact clock for the phase, and because phase lengths are constant, the following slow dance is exact too (`slowStart = fastStart + 45 s`). Only the very first slow dance is anchored on "first bot tick after the pull", and any latency there errs in the safe direction (bots leave a spot later, never earlier).
- The clock lives in one small `static` map keyed by boss GUID (mutex-guarded, same precedent as the RS/Ulduar helpers) so a bot that dies and is battle-rezzed, or joins late, picks the schedule back up instead of restarting it from "now".
- Safe waypoint = f(phase, elapsed) with a 300 ms margin (the boss `TaskScheduler` fires up to a tick late); after a phase's last eruption bots pre-move to the first spot of the next phase, which makes the transitions seamless.
- Roles: melee/off-tanks always dance; the bot main tank dances once it has aggro (Heigan starts on his platform, so the tank is free to go pick him up); ranged/healers stand on the platform during the slow dance (Plague Cloud radius 21 y, platform is >= 38 y from the dance spots) and leave after its 8th eruption; everyone dances the fast dance.
- `HeiganDanceMultiplier` gates casts by `CanStandStillFor(castTime + 500 ms)` (instants always allowed) and suppresses formation/flee/blink/disengage moves and reach/charge actions while dancing (Heigan stands inside his own Plague Cloud during the fast dance).

The module's original waypoints were kept; they were checked against the core's `GetEruptionSection` classifier (module waypoint 0..3 = core sections 3..0). Spell facts (45 000 ms duration, 21 y radius) were confirmed from `Spell.dbc`.

Files: `src/Ai/Raid/Naxx/` only. New `NaxxBossHelper.cpp` (the clock and waypoint math), `HeiganBossHelper` in `NaxxBossHelper.h`, `NaxxActions_Heigan.cpp` rewritten, the Heigan trigger/action/multiplier declarations and their registrations uncommented and updated.

## Feature Evaluation

- **Minimum logic:** per bot per tick while fighting Heigan: one `find target` lookup (as every other Naxx trigger), one `GetAura(29350)`, one distance check, one hash-map lookup under a mutex, integer arithmetic for the safe waypoint. The multiplier does a handful of `dynamic_cast`s per candidate action, like the existing raid multipliers.
- **Processing cost:** the trigger only exists on map 533 (instance strategy) and returns immediately when the bot is not in combat with Heigan. There is no per-tick pathing beyond the moves the bot would issue anyway; `MoveInside` returns false once at the spot, so no repeated sub-yard moves. Nothing runs for bots outside Naxxramas.

## How to Test the Changes

Setup: `AiPlayerbot.ApplyInstanceStrategies = 1` (default), a 10-man group of you plus 9 bots with melee, ranged and healers, Naxxramas 10, clear to Heigan.

1. Pull with the tank (human or bot). Expected: melee bots stand on the first spot (west of the platform) and step to the next section every 10 s starting at 15 s; ranged/healers stand on the platform and heal/DPS from there; at ~85 s ranged leave the platform for the first fast-dance spot; at 90 s Heigan teleports and every bot moves every 4 s from 7 s on; at 135 s everyone is back on the first slow-dance spot; repeat until dead. Nobody should take Eruption damage.
2. Optional: kill a bot mid-fight and battle-rez it. Expected: it rejoins the correct spot immediately instead of standing on waypoint 0.
3. Optional: wipe or evade, then re-pull. Expected: the schedule restarts cleanly (state is dropped after 30 s without observation).

### Test scenarios performed and results

Environment: AzerothCore Playerbot branch `092e9ba6ff8d`, module at `a7b885d2` plus this change, Docker (Linux), `MapUpdate.Threads = 4`, 250 random bots online, `mod-autobalance` on dist defaults.

| # | Scenario | Result |
|---|---|---|
| 1 | Compile the module (full worldserver image build) | Builds; zero compiler warnings in `src/Ai/Raid/Naxx/` |
| 2 | Boot the worldserver with the change | Clean boot ("World Initialized"), 250 random bots logged in, no new log errors |
| 3 | 10-man Naxxramas, human blood-DK tank plus 9 bots (mixed melee, ranged and healers), pull Heigan and kill him | **Kill, zero deaths.** DBM: "Heigan the Unclean down after 3 minutes and 28.00 seconds" (208 s = slow dance 0 to 90 s, fast dance 90 to 135 s, second slow dance to the kill), so both phases and both phase transitions were exercised. Observed: melee danced from 15 s on the 10 s schedule; ranged and healers stood on the platform and cast from there; ranged left the platform at ~85 s; every bot danced the 4 s fast phase; all bots back on the first slow-dance spot at 135 s. Screenshot of the kill attached below. |
| 4 | Bots outside Naxxramas during the same session | No behavior change (trigger only exists on map 533) |

10-man kill (DBM timer bottom left, all raid frames alive):

WoWScrnShot_081726_141711

Not yet exercised (would appreciate testers): bot main tank pulling (code path is there: the tank dances once it has aggro), battle-rez rejoin mid-fight, wipe and re-pull, 25-man.

## Impact Assessment

    - - [ ] No, not at all

Only bots in Naxxramas and in combat with Heigan pay anything (see Feature Evaluation). Bots elsewhere are unaffected.

    - - [ ] No
    - - [x] Yes (**explain why**)

Bots in the Heigan encounter now dance instead of standing still; that is the previously commented-out, intended behavior of the Naxx strategy.

    - - [ ] No

The schedule constants mirror `boss_heigan.cpp`; if the core script's timings ever change, the constants in `HeiganBossHelper` need to follow. They are documented in one place at the top of the class.

## AI Assistance

- - [ ] No

The investigation (git history, core script and Spell.dbc reading) and the code were produced with Claude Code, driven and reviewed by me. I have read and understand the full change and tested it live as described above.

## Final Checklist

- - [x] Documentation updated if needed (Code comments, Conf comments, WiKi commands). (behavior described in code comments; no conf changes)

## Notes for Reviewers

- The static fight state is guarded by a mutex because bot AI runs from `OnPlayerAfterUpdate` inside `Map::Update`, and with `MapUpdate.Threads > 1` two Naxxramas instances can tick concurrently.
- `HeiganDanceAction` clears `last movement` before re-issuing a `MOVEMENT_COMBAT` move when the safe waypoint changes; otherwise `IsWaitingForLastMove` would keep a bot walking toward the section that is about to erupt.
- Known limit: a healer more than 40 y from the tank could in theory step off the platform to get in range during the slow dance (platform to far waypoint is ~38 to 40 y). Not observed in testing.

### DIFF
--- a/src/Ai/Raid/Naxx/Action/NaxxActions.h
+++ b/src/Ai/Raid/Naxx/Action/NaxxActions.h
@@ -62,33 +62,25 @@ private:
     float distance;
 };
 
+// One action for both roles: melee (and tanks) dance every phase, ranged/healers wait on the platform during the
+// slow dance and only dance the fast one.
 class HeiganDanceAction : public MovementAction
 {
 public:
-    HeiganDanceAction(PlayerbotAI* ai, std::string const name = "heigan dance")
-        : MovementAction(ai, name), helper(ai) {}
+    HeiganDanceAction(PlayerbotAI* ai, bool ranged)
+        : MovementAction(ai, ranged ? "heigan dance ranged" : "heigan dance melee"), helper(ai), ranged(ranged)
+    {
+    }
+    bool Execute(Event event) override;
 
-protected:
+private:
     // Move to (near) the given dance waypoint. Returns true while a move had to be issued.
     bool MoveToWaypoint(uint32 index, float distance);
     bool MoveToPlatform(float distance);
 
     HeiganBossHelper helper;
+    bool ranged;
     int32 lastWaypoint = -1;
-};
-
-class HeiganDanceMeleeAction : public HeiganDanceAction
-{
-public:
-    HeiganDanceMeleeAction(PlayerbotAI* ai) : HeiganDanceAction(ai, "heigan dance melee") {}
-    bool Execute(Event event) override;
-};
-
-class HeiganDanceRangedAction : public HeiganDanceAction
-{
-public:
-    HeiganDanceRangedAction(PlayerbotAI* ai) : HeiganDanceAction(ai, "heigan dance ranged") {}
-    bool Execute(Event event) override;
 };
 
 class ThaddiusAttackNearestPetAction : public AttackAction
--- a/src/Ai/Raid/Naxx/Action/NaxxActions.h
+++ b/src/Ai/Raid/Naxx/Action/NaxxActions.h
@@ -62,56 +62,34 @@ private:
     float distance;
 };
 
-//class HeiganDanceAction : public MovementAction
-//{
-//public:
-//    HeiganDanceAction(PlayerbotAI* ai) : MovementAction(ai, "heigan dance")
-//    {
-//        this->last_eruption_ms = 0;
-//        this->platform_phase = false;
-//        ResetSafe();
-//        waypoints.push_back(std::make_pair(2794.88f, -3668.12f));
-//        waypoints.push_back(std::make_pair(2775.49f, -3674.43f));
-//        waypoints.push_back(std::make_pair(2762.30f, -3684.59f));
-//        waypoints.push_back(std::make_pair(2755.99f, -3703.96f));
-//        platform = std::make_pair(2794.26f, -3706.67f);
-//    }
-//
-//protected:
-//    bool CalculateSafe();
-//    void ResetSafe()
-//    {
-//        curr_safe = 0;
-//        curr_dir = 1;
-//    }
-//    void NextSafe()
-//    {
-//        curr_safe += curr_dir;
-//        if (curr_safe == 3 || curr_safe == 0)
-//        {
-//            curr_dir = -curr_dir;
-//        }
-//    }
-//    uint32 last_eruption_ms;
-//    bool platform_phase;
-//    uint32 curr_safe, curr_dir;
-//    std::vector<std::pair<float, float>> waypoints;
-//    std::pair<float, float> platform;
-//};
-//
-//class HeiganDanceMeleeAction : public HeiganDanceAction
-//{
-//public:
-//    HeiganDanceMeleeAction(PlayerbotAI* ai) : HeiganDanceAction(ai) {}
-//    virtual bool Execute(Event event);
-//};
-//
-//class HeiganDanceRangedAction : public HeiganDanceAction
-//{
-//public:
-//    HeiganDanceRangedAction(PlayerbotAI* ai) : HeiganDanceAction(ai) {}
-//    virtual bool Execute(Event event);
-//};
+class HeiganDanceAction : public MovementAction
+{
+public:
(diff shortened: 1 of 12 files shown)


jbmartino:
Heads-up for maintainers: this overlaps with #2627 by @karlverik, which I only noticed after opening this. Both restore the commented-out Heigan strategy and touch the same files, so only one can land. Summarising the difference so you can compare them side by side:

- **How eruptions are tracked.** #2627 detects each eruption after the fact by scanning the grid for the temporary World Trigger creatures (entry 12999) the floor gameobjects spawn, and steps a per-bot safe index. This PR derives the safe spot from a fight clock: the Plague Cloud aura's remaining duration is an exact clock for the fast dance and, since the phase lengths are constants in `boss_heigan.cpp`, for the following slow dance too. The clock is shared per boss GUID.
- **What that buys.** Bots can move *before* an eruption instead of after it, so ranged/healers leave the platform at 85 s (before the teleport and Plague Cloud) rather than running out through the cloud, healers get an explicit "safe to cast for N ms" window on each slow-dance step, and a bot that ticks late, dies, or is battle-rezzed lands on the correct spot immediately instead of being off by one until the next phase. Per-tick cost is one aura lookup plus a hash-map lookup rather than a grid creature scan.
- **What it costs.** More code (a small shared registry behind a mutex, since bot AI runs inside `Map::Update` and multiple Naxx instances can tick concurrently), and the schedule constants mirror the core script, so if AzerothCore ever changed Heigan's timings they would need updating. #2627 is simpler and independent of the timings.

I have no strong attachment to which one goes in. If you prefer the #2627 approach I am happy to close this, or to port over any pieces from here that are useful (the early platform exit for ranged, the shared clock, the cast gating). If you prefer this one, I will keep it rebased on `test-staging` and address review.

kadeshar (approved):

PR #2678, opened 2026-08-20
Remove GenericAction and copy actions to AttackAction and PetsAction

brighton-chi:
## Pull Request Description

GenericActions is an unhelpful name for a file that contains four actions that don't have any apparent reason to be there. I deleted GenericActions.cpp/h and moved its contents. I moved MeleeAction to AttackAction.cpp/h (AttackAction is the parent class, and the files include other derived classes of AttackAction). And I moved TogglePetSpellAutoCastAction, PetAttackAction, and SetPetStanceAction to PetsAction.cpp/h, which seems like a good fit.

I copied everything over verbatim, even the bad formatting and inconsistencies, to make this as easy as possible.

I also updated includes. Most includes of GenericActions.h actually didn't need it at all so mostly I just deleted them. I confirmed everything compiles.

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

To search for unused includes of GenericActions.h

 MaNGOS, another module)?

## Final Checklist


NoxMax:
Changes to includes are also self-contained as far as I can tell. When merged, worth posting about it in the dev channel to anyone that has an open PR using GenericActions.

brighton-chi:
Sounds good. There are other candidates too for files to be consolidated as some contain very little content, but this is the one that seems most nonsensical to exist.

src/Ai/Base/Actions/PetsAction.cpp
```diff
@@ -304,3 +338,177 @@ bool PetsAction::Execute(Event event)
...
+        if (shouldApply != isAutoCast)
+        {
+            pet->ToggleAutocast(spellInfo, shouldApply);
+            toggled = true;
+        }
+    }
+
+    // Debug message if pet spells have been toggled and debug is enabled
+    if (toggled && sPlayerbotAIConfig.petChatCommandDebug == 1)
+        botAI->TellMaster("Pet autocast spells have been toggled.");
```
kadeshar, line 402: That require translation

src/Ai/Base/Actions/PetsAction.cpp
```diff
@@ -304,3 +338,177 @@ bool PetsAction::Execute(Event event)
...
+        target->SetReactState(react);
+        CharmInfo* charmInfo = target->GetCharmInfo();
+        // If the creature has a CharmInfo, set the player-visible stance as well
+        if (charmInfo)
+            charmInfo->SetPlayerReactState(react);
+    }
+
+    // If debug is enabled in config, inform the master of the new stance
+    if (sPlayerbotAIConfig.petChatCommandDebug == 1)
+        botAI->TellMaster("Pet stance set to " + stanceText + " (applied to all pets/guardians).");
```
kadeshar, line 511: That require translation

kadeshar (changes requested):

src/Ai/Base/Actions/PetsAction.cpp
brighton-chi: I just copied over the old code with no changes. You want me to update it?

kadeshar: If that is no problem for you. I though i removed all hardcoded texts without translations.

NoxMax: likely were overlooked because these two fall under petChatCommandDebug

kadeshar (approved):

brighton-chi:
I didn't add code for translation yet. I only fixed the compile error from the merge.

kadeshar:
> I didn't add code for translation yet. I only fixed the compile error from the merge.

I know, translation is optional. If you add i will reapprove

PR #2680, opened 2026-08-20
Rework Spell Interrupt Calls in Actions

brighton-chi:
## Pull Request Description

This PR reworks how the module calls spell interrupts through the action/trigger system.

While reexamining Vashj to address some regressions raised in Discord, I was thinking about why I could not get core passing to actually deliver the core (and thus used a hack in the strategy of creating the core on recipient bots). The answer is that ```Unit::InterruptNonMeleeSpells(true)``` cancels a completed but unresolved cast, and unfortunately that impacts many strategies (including others of mine, oops). The reason is its first parameter, ```bool withDelayed```, will cancel the effect of projectiles in flight if true. This is obviously not something that a human player could do, so it doesn’t make any sense for it to happen to bots. ```PlayerbotAI::InterruptSpell()```, which is also commonly used in the codebase, has the same problem.

I would divide what this PR does into two things, and I can split them up if desired (but they’re related).

### Change Interrupt Calls
The simplest correction would be to change all of the calls to InterruptNonMeleeSpells(true) and InterruptSpell() to ```Unit::InterruptNonMeleeSpells(false)```. However, I think that is not the best course of action. Instead, this PR changes the calls to ```Unit::CastStop()```. InterruptNonMeleeSpells(false) and CastStop() are behaviorally equivalent. They just visit spell slots in different orders and have different parameters. I think CastStop() is better because:
-  CastStop() hardcodes ```withDelayed = false```. No call site for InterruptNonMeleeSpells in Playerbots should use true for withDelayed. Thus, using CastStop() as convention going forward protects against screwups resulting from the unneeded parameter.
- CastStop() is a more appropriate name for what is actually happening:
  - InterruptNonMeleeSpells, by its name, does not interrupt “melee” spells. Taken at face value, that suggests that it would not interrupt melee-range spells (or at least that’s what a “melee” spell would mean to me colloquially). However, a “melee” spell for this purpose means a queued spell that hits on your next auto-attack (like Heroic Strike or Raptor Strike). Neither function would interrupt such spells.
  - InterruptNonMeleeSpells is not really “interrupting” anything. In practice it stops only two things: (1) an active channel (like Volley or Blizzard) and (2) auto-shot/wanding. This is because ```PlayerbotAI::UpdateAI()``` returns early when a spellcast is pending so you cannot cancel a pending spellcast via an action with any function. That’s why I implemented ```PlayerbotAI::RequestSpellInterrupt()``` a while back, which allows pending spellcasts to be interrupted by calling it outside of the action system. Thus, “CastStop,” which does the same thing, better describes what is actually happening.
 
Note: Both InterruptNonMeleeSpells and CastStop are capable of cancelling pending casts; it’s just due to the structure of UpdateAI that they cannot do so for bots. I believe calling them on pets actually can cancel pending casts. I've never verified it, but logically it should work.

### Remove ```PlayerbotAI::InterruptSpell()```
So what does this function do?
- It interrupts delayed spells, like InterruptNonMeleeSpells(true).
- It does not interrupt auto-shots or wanding. I think ideally auto-shots are not interrupted with respect to boss avoidance, but this is very minor, and for the most part boss strategies have been interrupting them anyway with InterruptNonMeleeSpells.
- It does interrupt queued melee spells. It’s unclear to me why this would be needed. I think it’s generally a bad thing as it relates to boss avoidance, but it’s also minor.
- The most notable difference is it sends its own spell failure packets, unconditionally (even if an interrupt is not successful). A successful interrupt already results in the packets being sent by Spell::cancel. I don’t know why we would need to falsify spell failure packets. If someone can think of a good reason, that would be great, but otherwise it seems illogical.

I’ve looked at all the call sites for this, and nothing seems to make sense for why an AC interrupt function would not suffice in those situations. Thus, because of the problem of withDelayed = true, I’ve changed all the calls of PlayerbotAI::InterruptSpell() as well. And after that, it has no calls left, so I’ve proposed just removing it with this PR so there isn’t confusion by having an unused (and seemingly not useful) interrupt function.

### Regarding ```bool withDelayed```
There are certainly uses for interrupting delayed spells, but from what I understand, the purpose is primarily to destroy the object on the backend (like when a player leaves the world or teleports). It’s used in the core upon teleporting and when leaving the world. Those aspects are relevant to bots for sure, but they’re handled by the core already. 

The only place in Playerbots that stands out to me as potentially relevant is the knockback packet handler. Based on some back-and-forth with Claude, I understand the core issues two types of knockbacks. One is a spell effect, which cancels spells, including in-flight projectiles. It’s surprising to me as that seems like the wrong behavior, but in any case, that’s a knockback from a spell effect that is independent of the packet. The other is a knockback from a script. In such cases for players, the cancellation comes from client-reported movement rather than from the knockback itself. Bots obviously move server-side, which is why the handler needs to interrupt directly for them. So removing InterruptSpell() possibly changes knockback behavior in that script-based knockbacks seemingly previously would have cancelled in-flight projectiles and now do not. I don’t know if that cancellation was the right behavior in the first place. It’s worth keeping in mind though.

### Other Changes:
- Some InterruptNonMeleeSpells and PlayerbotAI::InterruptSpell calls are gated by a ```bot->IsNonMeleeSpellCast()``` check. I’ve removed them. 

    First, because in some cases the argument choices for IsNonMeleeSpellCast don’t make any sense. For example, BattleGroundTactics.cpp has:

    ```
    if (bot->IsNonMeleeSpellCast(false, true, true, false, true))
        bot->InterruptNonMeleeSpells(true);
    ````

    The guard's arguments exclude channels and auto-shots, so the only thing that can satisfy it is a pending cast (which, as noted above, can't happen inside an action). The interrupt therefore never runs at all. Removing the guard is a functional change that lets the interrupt actually do something.

    Second, even if the arguments were set up in a sensible way, the condition, applied solely to an interrupt call, provides nothing of value (not even for performance).
- Utgarde Keep had a commented-out strategy to stop casting against Ingvar. I don’t know that boss to understand why it is relevant, but based on comments, it seems Bobblybook, two years ago and with a lot better attention to detail and knowledge than me, correctly observed that the action couldn’t actually interrupt a pending cast due to the action being queued behind the current spell. So there’s no need for that specific strategy (and associated action/trigger plumbing) to remain. If UK’s strategy were to be revisited with a desire to implement an interrupt, it could not go through the trigger/action wiring anyway.
- Some comments in ICC, MC, and Onyxia strategies implied you could interrupt a casting spell. Since you can’t, I edited the comments to specify the interrupts are for channeling spells.
- I modified UpdateAI to replace InterruptSpell() with ```bot->InterruptSpell(currentSpell->GetCurrentContainer())```. The existing version actually works fine since it’s properly guarded to cancel just the pending cast or channel. But I got rid of InterruptSpell() so it’s not available anymore, and the revised code, which cancels whatever is currently casting, better reflects the purpose of RequestSpellInterrupt().
- In the interest of pushing across a more fulsome fix for Vashj that just the spell interrupt fix, I modified two TeleportTo() calls to NearTeleportTo() in blocks where spell interrupts were called. Happy to pull these out, but the changes are definitely correct and are needed to make the fight completable with the strategy so I think it's worthwhile to toss in when the code is right there.

## Feature Evaluation

I think the minimum logic is covered sufficiently above. I don't think there is any relevant impact on processing cost.

## How to Test the Changes

I think this one kind of just needs to be played around with a bit. The boss strategies aren't really where I'd focus because I tested the effect of CastStop() on interrupting Mind Flay in Sethekk, and it works properly. Maybe it's worth running through ICC since I removed guards there. I think the most useful cases are to make sure nothing unexpected happens during knockbacks and when bots are forced to move (like with a "follow" order) while they are channeling or shooting/wanding.

## Impact Assessment

    - - [x] No, not at all

    - - [ ] No
    - - [x] Yes (**explain why**)

Certain spell interrupts now will have slight differences, as described above, as relating to in-flight projectiles and auto-shots/wanding.

    - - [x] No

I think this should simplify things going forward.

## AI Assistance

- - [ ] No

I went back-and-forth with Opus a lot on this one to try to understand more about the spellcasting system, like spell slots and spell states, and the various interrupt functions. I pressed it in pretty much every way I could think of with respect to the changes to chase down potential adverse effects.

The actual edits are all my own. They're not exactly rocket science.

 MaNGOS, another module)?

## Final Checklist

### ```PlayerbotAI::Reset()``` in Boss Strategies
I did not change calls to botAI->Reset() in boss strategies, even though I think that function is the wrong approach, whether that be for emergency avoidance or to stop attacking/casting. It doesn’t do anything you can’t do by using directly relevant functions like CastStop() or AttackStop() or ```MotionMaster::Clear()```. It goes farther by putting the bot in its non-combat engine and doing irrelevant things like clearing whispers and lfg proposals. But I don’t want to broaden the PR any more, and many of these calls are in WotLK strategies, and I don’t want to guess at what was intended. I can fix the usages in TBC content as I go with strategy rewrites.


kadeshar (dismissed):

kadeshar:
Conflict to resolve

brighton-chi:
> Conflict to resolve

Done. Related to Naxx PR.

kadeshar (approved):

PR #2681, opened 2026-08-20
Fix worldserver crash: guard chat command dispatch against half-initialized bots

BBGNB:
## Pull Request Description

`RandomPlayerbotMgr::HandleCommand` iterates every playerbot on each player chat message and calls `GET_PLAYERBOT_AI(bot)->HandleCommand` without checking the returned pointer. During a mass bot login flood, the bot map can contain entries whose `PlayerbotAI` is not attached yet (or already detached during logout); dispatching a chat command into one of them dereferences null and crashes the worldserver.

We hit this three times in one morning on a 3000-bot realm:

```
SIGSEGV at offset 0x258
Stack: PlayerbotAI::CheckLevelFor
    <- PlayerbotAI::HandleCommand
    <- RandomPlayerbotMgr::HandleCommand
    <- OnPlayerCanUseChat
```

This PR adds guard clauses only:

- Per bot: skip entries that are not fully in world, are mid-teleport, or return null from `GET_PLAYERBOT_AI` before dispatching.
- Sender side: return early if `fromPlayer` is null, not in world, mid-teleport, or has no map (we also crashed in `LevelFor(fromPlayer)` when chatting immediately after a GM teleport).

## Feature Evaluation

- **Minimum logic:** four pointer/state checks on the sender (once per chat message) and three checks per bot, plus a null check on the AI pointer that the code was already fetching. No new state, no new configuration.
- **Processing cost:** runs only on the player-chat path, which already iterates all bots. Per bot it adds two boolean flag reads on top of the existing map lookup; bots that cannot process commands are now skipped earlier, so in the flood case it does strictly less work than before.

## How to Test the Changes

1. Realm with a large bot pool (we run 3000 online, `MapUpdate.Threads = 4`).
2. Restart the worldserver and log a real player in while bots are still mass-logging-in.
3. Speak repeatedly in General (or any channel that fans out to bots) during the flood.
4. Without the guards, the worldserver can segfault when the iteration reaches a half-initialized bot; with them, chat is dispatched only to bots with an attached AI.

Note this is a **timing-dependent race**: the crash requires a chat message to reach a map entry inside the window between bot insertion and AI attachment, so a non-repro on a given machine/config is expected — see reviewer notes below.

## Impact Assessment

    - - [x] No, not at all

(Two flag reads per bot on the chat path only; the AI pointer fetch already existed and is now reused instead of fetched for the call.)

    - - [ ] No
    - - [x] Yes (**explain why**)

Only in the degenerate case: bots with no attached AI (mid-login/logout) no longer receive chat commands — previously that was a crash, not a behavior. Fully initialized bots behave identically.

    - - [x] No

(Guard clauses at the top of an existing function.)

## AI Assistance

- - [ ] No

The crash was diagnosed and the patch written with Claude (Claude Code) operating our realm, from the live crash stacks. The change was reviewed by the realm operator and verified in production: settled chat, teleport-then-chat, and sustained chat through a full 3000-bot login flood with zero crashes since (previously three crashes in one morning under the same load). The logic is fully understood: it is null/lifecycle guarding on an existing dispatch loop.

## Final Checklist

- - [x] Documentation updated if needed. (comments in the diff document the race)

## Notes for Reviewers

On reproducibility: the window only exists while the playerbot map holds an entry whose `PlayerbotAI` is not attached, so hitting it needs chat fan-out to land inside that window — heavily dependent on login-flood size, hardware timing, and thread scheduling. We could hit it reliably only during the initial mass login after a restart, with a player actively chatting through the flood. Even where the race never fires, `GET_PLAYERBOT_AI` returning null is possible by its contract, and the unguarded call is a latent null-deref; the guard costs nothing on the hot path. We can provide the full crash dumps and our exact playerbots.conf on request.


NoxMax:
Please follow the PR template format and instruction. That being said, can anyone else recreate this crash? I spammed general during logins in similar circumstances as the PR, tried with and without `DisabledWithoutRealPlayer`, no crash at all. `MapUpdate.Threads =10`.

BBGNB:
Thanks for the quick look — the description now follows the template (and the base is retargeted to test-staging per the repo rule).

On reproduction: it's a narrow timing race, so a non-repro is very plausible — the crash needs a chat fan-out to reach a playerbot map entry inside the window where its PlayerbotAI isn't attached (mid-login/logout). We only ever hit it during the initial mass login after a worldserver restart (3000-bot pool, `MapUpdate.Threads = 4`, Apple M4/arm64 build), with a player chatting through the flood — three SIGSEGVs in one morning under those exact conditions, zero since the guards. Hardware timing and thread count will move the window, which likely explains the difference with your `MapUpdate.Threads = 10` attempt.

Independent of repro, `GET_PLAYERBOT_AI` can return null by its contract, and the unguarded `->HandleCommand` is a latent null-deref on a path that runs for every player chat message. The guards cost two flag reads per bot on that path.

We'll follow up shortly with the exact playerbots.conf (including `DisabledWithoutRealPlayer`) and the full crash stacks from the affected realm.

kadeshar:
You tried only test crash when only added check that botAI is initialized?
```
PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
        if (!botAI)
            continue;
```
In theory other checks are already in code.

BBGNB:
Follow-up with the promised runtime data from the affected realm.

**Crash records** — three crashes on 2026-08-20 (07:27:06, 07:28:54, 07:32:04), all inside the post-restart login-flood window, with identical stacks:

```
EXC_BAD_ACCESS (KERN_INVALID_ADDRESS) at offset 0x258
PlayerbotSecurity::CheckLevelFor
  <- PlayerbotAI::HandleCommand
  <- RandomPlayerbotMgr::HandleCommand
  <- PlayerbotsPlayerScript::OnPlayerCanUseChat
  <- WorldSession::HandleMessagechatOpcode
```

The 0x258 offset is a member read through a null/half-initialized AI pointer — exactly the contract violation the guard covers.

**Exact configuration at time of crash:**
- `MinRandomBots = MaxRandomBots = 3000` (since raised to 4000, still crash-free with the guard)
- `AiPlayerbot.DisableWithoutRealPlayer = 0`, login/logout delays 30/300
- `RandomBotsPerInterval = 60`
- `MapUpdate.Threads = 4`
- Apple M4 (arm64), Release build, AzerothCore rev 111ce6600dc3

On reproducibility: the window opens when general-chat fan-out races the mass login. Your Threads=10 x86 box likely just never loses that race — timing differences between platforms and thread counts move the window. Zero recurrences here since the guards, under sustained higher load than the crashing config.

Happy to provide the full crash report files if useful.

kadeshar:
That didnt answered my question. 
Your PR making a lot of checks:
obraz
I asked that simple is not enough:
```
PlayerbotAI* botAI = GET_PLAYERBOT_AI(bot);
        if (!botAI)
            continue;
 ```

File: src/Bot/RandomPlayerbotMgr.cpp [hunk 1]
```diff
@@ -2482,10 +2482,24 @@ bool RandomPlayerbotMgr::HandlePlayerbotConsoleCommand(ChatHandler* /*handler*/,
 
 void RandomPlayerbotMgr::HandleCommand(uint32 type, std::string const text, Player* fromPlayer, std::string channelName)
 {
+    // The sender needs guarding too: chatting right after a (GM/panel) teleport
```
Celandriel: Comment clean up. Generally these should be minimal, and ideally human written.

PR #2684, opened 2026-08-20
Fix Blade Flurry trigger and strategy

brighton-chi:
## Pull Request Description

God the Rogue files are so bad. To recap, the "dps" strategy applies to all Rogues, and Assassination additionally gets "melee" for stuff like Mutilate and HFB.

Right now, Blade Flurry is present in the "dps" and the "aoe" strategies. It is triggered by "light aoe" in both, and then it separately is also triggered by "blade flurry" in the dps strategy.

Problem? There is no "blade flurry" trigger registered. It is BladeFuryTrigger (???) and has the "blade fury" string key. Thus, the ability is currently not used outside of aoe situations.

I've renamed the class to BladeFlurryTrigger and fixed the associated registrations/keys. I deleted the TriggerNode that fires based on light aoe in DpsRogueStrategy so that it sits only in RogueAoeStrategy. I moved the TriggerNode that fires based on blade flurry in DpsRogueStrategy and moved it to RogueBoostStrategy (it already inherits from BoostTrigger). 

I also tweaked the priorities. BF should be very high priority in AoE situations so I bumped it to ACTION_HIGH + 4 in the aoe strategy, which basically displaces everything that isn't a stealth move or defensive cooldown. However, it shouldn't be above SnD, so I moved SnD up to ACTION_HIGH + 5. In the boost strategy, I also made it ACTION_HIGH + 4, which puts it above AR at ACTION_HIGH + 2 (BF should be used first since it spends energy while AR restores it). I did 4 because it's unoccupied otherwise. That does put it over EA at ACTION_HIGH + 3, but it's getting too crammed, and to fix that I would need to start moving a ton of shit so we'll just live with this tiny inefficiency. The Rogue files badly need redoing anyway.

## Feature Evaluation

Minimum logic described above. Processing cost is just the additional evaluation of one more BoostTrigger, which always existed but was not properly implemented.

## Impact Assessment

    - - [ ] No, not at all

One more BoostTrigger.

    - - [ ] No
    - - [x] Yes (**explain why**)

Rogues will properly use BF as a dps cooldown in single-target situations when boost is active (most importantly, boss fights).

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Class/Rogue/RogueAiObjectContext.cpp
+++ b/src/Ai/Class/Rogue/RogueAiObjectContext.cpp
@@ -77,12 +77,10 @@ public:
         creators["off hand weapon no enchant"] = &RogueTriggerFactoryInternal::off_hand_weapon_no_enchant;
         creators["tricks of the trade on main tank"] = &RogueTriggerFactoryInternal::tricks_of_the_trade_on_main_tank;
         creators["adrenaline rush"] = &RogueTriggerFactoryInternal::adrenaline_rush;
-        creators["blade fury"] = &RogueTriggerFactoryInternal::blade_fury;
+        creators["blade flurry"] = &RogueTriggerFactoryInternal::blade_flurry;
     }
 
 private:
-    static Trigger* adrenaline_rush(PlayerbotAI* botAI) { return new AdrenalineRushTrigger(botAI); }
-    static Trigger* blade_fury(PlayerbotAI* botAI) { return new BladeFuryTrigger(botAI); }
     static Trigger* kick(PlayerbotAI* botAI) { return new KickInterruptSpellTrigger(botAI); }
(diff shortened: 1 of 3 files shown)


kadeshar (approved):

Yuof:
One of the problems with blade flurry currently is that the "light aoe" trigger check for 1+ targets next to the current target. This lead to the rogue using blade flurry during engage and losing half the buff while running into melee.

It really need its own trigger condition that checks for melee range.

brighton-chi:
I remember looking into that a while back. I don't remember the exact distance that the targets have to be for the light aoe trigger but it's too far to hit both with Blade Flurry. The Rogue code overall needs a ton of work.

PR #2690, opened 2026-08-22
Test staging

kadeshar:
Maintenace PR

Changelog:
- Bots can refresh missing or expiring buffs through the new `rebuff` command.
- Ready checks can optionally make out-of-combat bots finish buffing before confirming readiness.
- Paladin blessings are not refreshed by the new rebuff behavior.
- In Molten Core, bots focus Golemagg instead of his Core Ragers until Golemagg is nearly defeated.
- In the Golemagg fight, healers position themselves to better cover both tanks.
- In the Golemagg fight, melee back away before Magma Splash becomes lethal, while casters avoid moving into melee range.
- In the Majordomo Executus encounter, bots focus the adds rather than attacking Majordomo.
- In Molten Core, bots that fall into lava move toward nearby safe ground.
- On Shazzrah, ranged bots reposition only when they are actually too close to Arcane Explosion.
- Arena teams are automatically restored or created as needed when bots log in.
- Bots no longer use generic crowd-control spells in five-player dungeons, while this behavior remains available elsewhere.
- No player-visible behavior change is intended from the encounter-helper cleanup.
- In Blackwing Lair, bots apply fire resistance buffs for Broodlord Lashlayer, Firemaw, and Flamegor.
- Selfbots can initiate trades and receive items from altbots without offering an item in return.
- Random bots treat selfbots like regular players when trading, including normal trade-value checks and refusal rules.
- Refused trades now close properly, allowing another trade to be started immediately.

AzerothCore fork: https://github.com/mod-playerbots/azerothcore-wotlk/tree/test-staging

### DIFF
--- a/src/Ai/Raid/BT/BTHelpers.cpp
+++ b/src/Ai/Raid/BT/BTHelpers.cpp
@@ -5,8 +5,10 @@
  */
 
 #include "BTHelpers.h"
+#include "EncounterHelpers.h"
 #include "Playerbots.h"
-#include "RaidBossHelpers.h"
+
+using namespace EncounterHelpers;
 
 namespace BlackTempleHelpers
 {
@@ -104,7 +106,7 @@ std::unordered_map<ObjectGuid, TankPositionState> shahrazTankStep;
 
 TankPositionState GetShahrazTankPositionState(PlayerbotAI* botAI, Player* bot)
 {
-    Player* mainTank = GetGroupMainTank(botAI, bot);
+    Player* mainTank = GetGroupMainTank(bot);
     if (!mainTank)
         return TankPositionState::Unknown;
 
--- a/src/Ai/Raid/Hyjal/Util/HyjalHelpers.cpp
+++ b/src/Ai/Raid/Hyjal/Util/HyjalHelpers.cpp
@@ -5,10 +5,12 @@
  */
 
 #include "HyjalHelpers.h"
+#include "EncounterHelpers.h"
 #include "Playerbots.h"
-#include "RaidBossHelpers.h"
 #include "Timer.h"
 #include <algorithm>
+
+using namespace EncounterHelpers;
 
 namespace HyjalSummitHelpers
 {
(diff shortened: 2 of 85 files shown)


src/Ai/Raid/BT/BTHelpers.cpp
```diff
@@ -104,7 +106,7 @@ std::unordered_map<ObjectGuid, TankPositionState> shahrazTankStep;
 
 TankPositionState GetShahrazTankPositionState(PlayerbotAI* botAI, Player* bot)
 {
```
NoxMax, line 108: Drop botAI, keep `PlayerbotAI*`.

src/Ai/Raid/Hyjal/Util/HyjalHelpers.cpp
```diff
@@ -165,7 +167,7 @@ std::unordered_map<ObjectGuid, bool> isBelowManaThreshold;
 
 TankPositionState GetKazrogalTankPositionState(PlayerbotAI* botAI, Player* bot)
 {
```
NoxMax, line 169: Drop botAI, keep `PlayerbotAI*`.

src/Ai/Raid/Hyjal/Util/HyjalHelpers.cpp
```diff
@@ -206,7 +208,7 @@ RainOfFireData* GetActiveAzgalorRainOfFire(uint32 instanceId)
 
 TankPositionState GetAzgalorTankPositionState(PlayerbotAI* botAI, Player* bot)
 {
```
NoxMax, line 210: Drop botAI, keep `PlayerbotAI*`.

NoxMax (changes requested):
Minor edits. Should be added to this batch of test-staging.

brighton-chi:
Probably we should do one PR to fix all the usages of the PlayerbotAI static functions, and I'm sure that will reveal more unneeded PlayerbotAI* botAI parameters.

kadeshar:
That are minor things which not blocks test-staging merge. Should be applied separate PR.

Celandriel (approved):

PR #2702, opened 2026-08-24
Edits to Auchindoun dungeon strategies

brighton-chi:
## Pull Request Description

Updates to Sethekk Halls do not impact the strategy; they only modify the name of the shared file to match codebase convention and tweak some multipliers for better performance. Note that the non-combat engine check at the top of multipliers, which very cheap and effective, cannot be used for any multipliers that can result in bots dropping targets since that will make them go into their non-combat engines, even during combat. Hence it is intentional that the check is not applied to a couple of the multipliers.

Updates to Auchenai Crypts include some functional changes. Namely, I had some issues with bots running too far to escape the hazard or with getting out of range of the tank and then being unable to properly reach the tank due to tension between reaching spell range and returning to the hardcoded ranged position. I fixed these things by reducing the hazard radius and movement distance and also adding an early return based on the config's spell distance.

Other Auch Crypts changes are to improve efficiency (e.g., reordering the multiplier and using FindNearestCreature instead of GetCreatureListWithEntryInGrid to avoid constructing a list), trim unneeded code (e.g., make more use of early returns, remove unused includes), or just follow better practices (e.g., use the :: operator for the PlayerbotAI static functions, add missing standard library includes).

## Feature Evaluation

Just tweaks within existing files that don't involve any structural-type changes. Performance should be slightly better, though I wouldn't expect it to be noticeable.

## How to Test the Changes

Only Auch Crypts should have different behavior, which can be observed by fighting Shirrak (preferably on Heroic with player nerfs).

## Impact Assessment

    - - [x] No, not at all

Should be minimally better.

    - - [ ] No
    - - [x] Yes (**explain why**)

Changes to AC strategies described above.

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist


brighton-chi:
Over time I've settled on the same logic for calculating tank movement for pretty much all scenarios so there is high value in making that a general encounter helper. I've decided to implement that helper across raid strats I've worked on, but in the interest of getting the helper reviewed in a small PR, I put the open raid PRs all on draft and incorporated the helper into this one, with one use in each of the dungeon strategies. I'll reopen once the helper is merged.

Tl;dr is this helper calculates the incremental movement (which I think should always be used for tanks) and also toggles backwards movement based on whether the destination is on the opposite side of the bot from the boss (or whatever is being tanked). I included some other things in comments so anybody who uses this hopefully can understand how it works.

Celandriel (approved):

PR #2703, opened 2026-08-25
Large warnings cleanup

Celandriel:
## Pull Request Description

Warnings clean up

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/conf/playerbots.conf.dist
+++ b/conf/playerbots.conf.dist
@@ -509,6 +509,22 @@ AiPlayerbot.TellWhenMissingBuffReagents = 1
 # Default: 300 (5 minutes)
 AiPlayerbot.MissingBuffReagentMessageCooldown = 300
 
+# When a ready check is issued out of combat, bots refresh missing/short buffs before reporting ready.
+# A bot that still has buff work it can perform stays not-ready until the ready check ends; this
+# includes missing reagents (the cast is attempted and fails). Uses the always-on "force rebuff"
+# strategy: bots with strategy sets saved before this feature keep their saved set - reset their
+# strategies or use "nc +force rebuff" to opt them in. The "rebuff" chat command starts the same
+# buff pass without a ready check and works regardless of this setting.
+# Default: 0 (disabled)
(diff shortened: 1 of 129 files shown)


NoxMax:
Are these Windows based warnings? Haven't had any other on Linux other than the few botAI for a couple of raids from the latest batch of PRs.

Celandriel:
> Are these Windows based warnings? Haven't had any other on Linux other than the few botAI for a couple of raids from the latest batch of PRs.

Im trying to work through #2557 and whats coming up there.

NoxMax:
Does 2557 mark new things as warnings (in an actual server compile) that wouldn't have been marked before? Also I don't know what git is doing over there with the merge that it would want to re-apply all these other older commits.

Celandriel:
It looks like its ubuntu based actually Windows seems to have passe the check (Assuming I implemented it correctly.)

NoxMax:
Interesting. You might've then set it to be more sensitive than the default level for warnings that would appear on a real compile. I don't think we want that. Whatever warnings CI fires, should be the same warnings a user gets by default on their server.

kadeshar (approved):

CheckYourFax:
> Interesting. You might've then set it to be more sensitive than the default level for warnings that would appear on a real compile. I don't think we want that. Whatever warnings CI fires, should be the same warnings a user gets by default on their server.

It causes more warnings for gcc. I get zero warnings on clang.

Have to also take into account that clang targets gcc 16's binutils and libstdc++ on 26.04 (they are pulled when you install clang) while gcc 15.2 is used for the real gcc build. There's bound to be some difference there. 

Also I don't think warnings on gcc should be taken that seriously, since it's only really useful for debugging compiler bugs, since clang is used by default upstream.

Celandriel:
Can you share your list? I have to do another one as there are still some others.

Issue #1454, opened 2025-07-20
Prevent bots getting realm first achievements


TRPB:
**Is your feature request related to a problem? Please describe.**

Bots can currently get achievements as they progress which makes sense, however, there are some which should really be reserved for players only: Realm first achievements.

**Describe the solution you'd like**

An option to either:

- Disable realm first achievements explicitly (preferred) but otherwise keep them the same
- Disable bots getting any "Feat of Strength" achievement (probably easier to implement?)
- Disable all bot achievements
- Make playerbots `gm on` so they don't get achievements

**Additional context**

I assume this would need to patch into azerothcore.

Here is a list of achievement IDs for realm firsts:

```
456,"Realm First! Obsidian Slayer"
457,"Realm First! Level 80"
458,"Realm First! Level 80 Rogue"
459,"Realm First! Level 80 Warrior"
460,"Realm First! Level 80 Mage"
461,"Realm First! Level 80 Death Knight"
462,"Realm First! Level 80 Hunter"
463,"Realm First! Level 80 Warlock"
464,"Realm First! Level 80 Priest"
465,"Realm First! Level 80 Paladin"
466,"Realm First! Level 80 Druid"
467,"Realm First! Level 80 Shaman"
1400,"Realm First! Magic Seeker"
1402,"Realm First! Conqueror of Naxxramas"
1404,"Realm First! Level 80 Gnome"
1405,"Realm First! Level 80 Blood Elf"
1406,"Realm First! Level 80 Draenei"
1407,"Realm First! Level 80 Dwarf"
1408,"Realm First! Level 80 Human"
1409,"Realm First! Level 80 Night Elf"
1410,"Realm First! Level 80 Orc"
1411,"Realm First! Level 80 Tauren"
1412,"Realm First! Level 80 Troll"
1413,"Realm First! Level 80 Forsaken"
1414,"Realm First! Grand Master Blacksmith"
1415,"Realm First! Grand Master Alchemist"
1416,"Realm First! Cooking Grand Master"
1417,"Realm First! Grand Master Enchanter"
1418,"Realm First! Grand Master Engineer"
1419,"Realm First! First Aid Grand Master"
1420,"Realm First! Grand Master Angler"
1421,"Realm First! Grand Master Herbalist"
1422,"Realm First! Grand Master Scribe"
1423,"Realm First! Grand Master Jewelcrafter"
1424,"Realm First! Grand Master Leatherworker"
[...]

PR #2716, opened 2026-08-27
Block realm firsts for every bot session, not only the random pool

pjerra:
### The problem

`OnPlayerBeforeAchievementComplete` only blocks realm firsts for bots that
`RandomPlayerbotMgr::IsRandomBot()` or `IsAddclassBot()` answers true for.
Both are narrow:

```cpp
bool RandomPlayerbotMgr::IsRandomBot(ObjectGuid::LowType bot)
{
    ObjectGuid guid = ObjectGuid::Create<HighGuid::Player>(bot);
    if (!sPlayerbotAIConfig.IsInRandomAccountList(sCharacterCache->GetCharacterAccountIdByGuid(guid)))
        return false;

    if (std::find(currentBots.begin(), currentBots.end(), bot) != currentBots.end())
        return true;

    return false;
}
```

The character's account has to be in the random account list **and** the GUID has
to be in the live random pool. Any module that logs its own bots in on its own
accounts fails both tests, so its bots are treated as real players by the
achievement system and can take the realm firsts.

This is not hypothetical. A server owner installed a city-population module
(400 bots on dedicated accounts) and reported that the bots took every realm
first on the realm. He found this guard and assumed it covered them; it does
not, because the bots are neither random nor addclass.

### The change

Gate on the session flag instead, which is the property the guard actually
wants:

```cpp
if (player->GetSession() && player->GetSession()->IsBot() &&
    (achievement->flags & (ACHIEVEMENT_FLAG_REALM_FIRST_REACH | ACHIEVEMENT_FLAG_REALM_FIRST_KILL)))
    return false;
```

- Strict superset of the old condition: random and addclass bots log in on bot
  sessions, so everything blocked before stays blocked.
- Covers bots any other module logs in, without playerbots needing to know about
  them.
- **Selfbots are unaffected.** `IsSelfBot()` is a human's own character with a
  PlayerbotAI whose master is itself; that is a real session, not a bot session.
- Nothing else changes: the realm-first flag test is untouched, ordinary
  achievements are still allowed for bots.

### Behaviour change to be aware of

Characters logged in with `.bot add` (a player's own alt, driven by the AI on a
bot session) can no longer take a realm first. That looks like the intent of
this guard to me — an AI-controlled character taking a realm first is the thing
being prevented — but it is a real change from today, so flagging it explicitly.
If you would rather keep those eligible, say so and I will narrow the condition
to exclude bots with a real player master.

### Related

Follows the intent of #1454 (realm firsts should not go to bots), which this
guard was the answer to. The gap is only in *which* bots it recognises.

### Verification

Compiles against an AzerothCore Playerbot branch with this module built as a
static module (GCC, `MODULES=static`):

```
[1/1] Building CXX object modules/CMakeFiles/modules.dir/mod-playerbots/src/Script/Playerbots.cpp.o
```

Not yet run on a live realm — the downstream server that hit the original
problem is the natural place to confirm it, and I can report back once it has.


dillyns:
What about altbots? I feel like they should be getting realm firsts especially for things like raids

Celandriel:
> What about altbots? I feel like they should be getting realm firsts especially for things like raids

Agree with this.

pjerra:
> > What about altbots? I feel like they should be getting realm firsts especially for things like raids
> 
> Agree with this. 

I do too, I just didnt think about it when I did this. I fixed another problem on a mod when I did this.

src/Script/Playerbots.cpp
```diff
@@ -253,7 +253,13 @@ class PlayerbotsPlayerScript : public PlayerScript
...
     bool OnPlayerBeforeAchievementComplete(Player* player, AchievementEntry const* achievement) override
     {
-        if ((sRandomPlayerbotMgr.IsRandomBot(player) || sRandomPlayerbotMgr.IsAddclassBot(player)) &&
+        // Every bot session, not only the random and addclass pools: IsRandomBot()
+        // needs the character's account to be in the random account list and the
+        // GUID to be in the live pool, so bots a module logs in on its own
+        // accounts pass this guard and take the realm firsts from real players.
+        // A selfbot is unaffected -- it plays on the human's own session, which is
+        // not flagged as a bot session.
+        if (player->GetSession() && player->GetSession()->IsBot() &&
```
kadeshar: Delete comment, change is pretty obvious

kadeshar:
While applying changes change also destination branch. All PR should target test-staging branch.
Also comment to this code is redundant. Is pretty simple

PR #2719, opened 2026-08-28
Enhance RemoveAura method to support multiple auras removal

dillyns:
## Pull Request Description

allows the "ra" command to remove a list of aura as well as auras that are not provided by the bot.

## Feature Evaluation

## How to Test the Changes

When in a group with a bot, or as a selfbot you can use ra with a comma separated list to remove auras.
You can give spell ids, spell links, or aura names.
example to remove a list of worldbuffs: 
"ra 22888,15366,Spirit of Zandalar"

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Copilot for code exploration and suggestions. All code reviewed and modified where necessary before commit.

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Bot/PlayerbotAI.cpp
+++ b/src/Bot/PlayerbotAI.cpp
@@ -4280,9 +4280,47 @@ void PlayerbotAI::InterruptSpell()
 
 void PlayerbotAI::RemoveAura(std::string const name)
 {
-    uint32 spellid = aiObjectContext->GetValue<uint32>("spell id", name)->Get();
-    if (spellid && bot->HasAura(spellid))
-        bot->RemoveAurasDueToSpell(spellid);
+    if (name.empty())
+        return;
+
+    std::stringstream auraList(name);
+    std::string auraName;
+    while (std::getline(auraList, auraName, ','))
+    {
+        trim(auraName);
+        if (auraName.empty())
+            continue;
+
+        PlayerbotChatHandler handler(bot);
+        uint32 spellid = handler.extractSpellId(auraName);
+
+        if (!spellid)
+        {
+            std::wstring requestedName;
+            if (!Utf8toWStr(auraName, requestedName))
+                continue;
+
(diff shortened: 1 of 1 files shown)


src/Bot/PlayerbotAI.cpp
```diff
@@ -4276,9 +4280,47 @@ void PlayerbotAI::InterruptSpell()
...
+    std::stringstream auraList(name);
+    std::string auraName;
+    while (std::getline(auraList, auraName, ','))
+    {
+        trim(auraName);
+        if (auraName.empty())
+            continue;
+
+        PlayerbotChatHandler handler(bot);
+        uint32 spellid = handler.extractSpellId(auraName);
```
kadeshar, line 4295: You verified that works with multiword spells? That method use method extractSpellIdFromLink which suggest that you cant use string name. I think that take only first word from string which for "Spirit of Zandalar" will equal "Spirit" and provide wrong id.

src/Bot/PlayerbotAI.cpp
```diff
@@ -4276,9 +4280,47 @@ void PlayerbotAI::InterruptSpell()
...
+
+            for (Unit::AuraApplicationMap::const_iterator itr = bot->GetAppliedAuras().begin();
+                 itr != bot->GetAppliedAuras().end(); ++itr)
+            {
+                AuraApplication const* application = itr->second;
+                if (!application || !application->GetBase() || !application->GetBase()->GetSpellInfo())
+                    continue;
+
+                SpellInfo const* spellInfo = application->GetBase()->GetSpellInfo();
+                if (Utf8FitTo(spellInfo->SpellName[LOCALE_enUS], requestedName))
```
kadeshar, line 4313: That means that spell name contains requested name. Should be equal

PR #2723, opened 2026-08-28
Gruul Bug Fixes & Improvements

brighton-chi:
## Pull Request Description

The prior version of Gruul had a really bad implementation of the Shatter spread that made it practically useless. First, for the trigger, it used the version of the Ground Slam aura that covered only the first 2s of the sequence (the player is slowed over the course of 7s, frozen at 7s, and then Shatter hits at 9.7s). Second, for the action, it called FleePosition(), which looks for a position from which the bot can attack, meaning it would send all melee back into melee range of Gruul.

Then, Gruul's Shatter itself causes a bug via its interaction with MovementAction::MoveTo(). The tl;dr is that MoveTo has a built in delay based on the speed of the prior movement, and because the immobilization from Gruul's Shatter survives for 300ms _after_ the end of the Shatter sequence, MoveTo() tries to construct a spline during that period, calculates an infinite movement delay due to movement speed being 0, clamps that to a 5s max, and makes it so that bots stand around for 5s after Shatter before any movement is accepted again. This was fixed via a multiplier.

Maulgar also had a bug for the selection of the Kiggler moonkin tank where I had defined a local variable for spec outside of the group loop using bot, meaning that the group loop was actually checking the bot's spec. It still worked if you set the assistant flag on a moonkin, but it wouldn't work properly otherwise.

I added an out-of-encounter reset of variables, a method to clear RTI icons out of encounters, and a Fear Ward action for Maulgar. I also renamed actions/triggers that were incorrectly named or just generally misleading.

Along with this, I made various cleanups that I overlooked last time (like looping a second time to get tank targets for Hunter misdirects, or completely overcomplicating/using redundant logic the Gruul ranged spread action).

## Feature Evaluation

Two new trigger/action pairs, one new multiplier. Processing cost should be better, particularly due to the removal of several redundant group loops.

## How to Test the Changes

Go to Gruul's Lair and kill the bosses. Bring a moonkin and don't assign any assistant flags. The moonkin should tank Kiggler. Confirm that Shatter spreading works properly on Gruul.

## Impact Assessment

    - - [x] No, not at all

Performance should be better.

    - - [ ] No
    - - [x] Yes (**explain why**)

Fixed/improved strategies in Gruul's Lair with the "gruulslair" strategy applied.

    - - [x] No

## AI Assistance

- - [ ] No

I had AI assist with figuring out the issue that was causing bots not to move after Shatter. Everything else I did myself, as my original objective was to do this without AI, but I got stuck on that one issue.

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Raid/Gruul/GruulActionContext.h
+++ b/src/Ai/Raid/Gruul/GruulActionContext.h
@@ -15,15 +15,13 @@ class RaidGruulsLairActionContext : public NamedObjectContext<Action>
 public:
     RaidGruulsLairActionContext()
     {
+        // General
+        creators["gruul's lair reset encounter states"] =
+            &RaidGruulsLairActionContext::gruuls_lair_reset_encounter_states;
+
         // High King Maulgar
-        creators["high king maulgar main tank attack maulgar"] =
-            &RaidGruulsLairActionContext::high_king_maulgar_main_tank_attack_maulgar;
-
-        creators["high king maulgar first assist tank attack olm"] =
-            &RaidGruulsLairActionContext::high_king_maulgar_first_assist_tank_attack_olm;
-
-        creators["high king maulgar second assist tank attack blindeye"] =
-            &RaidGruulsLairActionContext::high_king_maulgar_second_assist_tank_attack_blindeye;
+        creators["high king maulgar melee tanks position bosses"] =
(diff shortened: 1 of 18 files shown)


kadeshar:
Conflict to resolve

PR #2733, opened 2026-09-01
fix(gundrak): rework the Slad'ran encounter

jrad7:
Branch: `fix/gundrak-sladran-encounter` (base: `test-staging`)

---

## Pull Request Description

Three things were wrong with the Slad'ran fight, and the snake wrap search was
expensive.

**The tank chased the summons.** `FindTankTargetSmartStrategy` ranks untanked units
above everything else, so every new snake outranked Slad'ran. The tank ran off to
taunt it, dragged the boss along behind, and then the next spawn outranked that one.
The snakes are melee mobs that walk to the party on their own, so the tank only has
to stand still. `SladranMultiplier` now zeroes `TankAssistAction` and
`ReachTargetAction` for snakes further than 12yd out, and `SladranTankHoldTrigger`
puts the tank back on the boss when nothing is in reach.

**The group was spread out.** The generic combat formation scattered the dps and
healers, so the snakes split up chasing them. Formation moves are zeroed for
non-tanks and `SladranStackOnTankTrigger` walks them to within 10yd of the tank, so
the snakes arrive in one pile. Flee, run away, move random and move from group are
zeroed too.

**Dps AoE was off.** The multiplier zeroed anything declaring
`ActionThreatType::Aoe` for the whole fight. That enum is a threat hint, not an AoE
flag, and it barely matches the abilities that matter here. Dropped.

**The snake wrap search.** It read the uncached `"possible targets"` value, so every
call swept the grid out to sight range with the `IsPossibleTarget` filter, from two
triggers plus once per action in the multiplier. It is now a cached value on a 200ms
interval, gated on combat, scanning by creature entry. Line of sight is still
raycast because `Attack()` fails without it.

## Feature Evaluation

Minimum logic: the multiplier bails on `find target slad'ran` outside the fight.
Inside it, two `dynamic_cast` chains and a distance check. The triggers are predicate
reads, and the wrap trigger's real work sits behind the cached value.

Cost: lower than before. A per-tick sight-range grid sweep becomes one
entry-filtered scan per bot every 200ms. Bots outside Gundrak pay nothing, since
`NamedObjectContext` only builds a value when something reads it and only this
strategy reads this one.

## How to Test the Changes

Five bots (tank, healer, 3 dps) in Gundrak, normal or heroic. Pull Slad'ran and watch
the summons at 90% and 50% (75% heroic).

- Tank stays on the boss, picks up only snakes that reach it, returns to the boss
  when one dies.
- Boss stays put instead of touring the room.
- Dps and healer stay within ~10yd of the tank, and the snakes arrive together.
- Dps use AoE on the pile.
- A wrapped player gets freed: the nearest dps with line of sight kills the Snake
  Wrap, and only one bot goes for each.

Run `playerbot pmon stack` during the fight. The snake wrap trigger should no longer
be near the top.

## Impact Assessment

- **Increases per-bot/per-tick processing?** No, it reduces it.
- **Modifies default behavior?** Yes, but only during Slad'ran and only for bots
  running `wotlk-gd`. Tank targeting, non-tank positioning and the AoE suppression
  all change for that fight.
- **New decision branches?** Yes: two triggers, two actions, one value, and a tank
  branch in the existing multiplier. The shared predicates live in a
  `GundrakSladran` namespace so the trigger, action and multiplier cannot drift
  apart.

## AI Assistance

Yes, for the code and review pass. The diagnosis and the code were checked
against the source and the core's `boss_slad_ran.cpp`.

All original.

## Final Checklist

- [x] Changes are understood and tested for server stability and performance impact.
- [x] Any new bot dialogue lines are translated. (none added)
- [x] New source files use the GPLv2 header. (`GDValueContext.h`)
- [x] Documentation updated if needed. (none needed)

## Notes for Reviewers

IDs checked against `creature_template`: 29680 Viper, 29713 Constrictor, 29742 Snake
Wrap. `SPELL_SNAKE_WRAP` 55126 matches `boss_slad_ran.cpp`.

Wrap assignment is the nearest eligible dps with line of sight, ties broken by lower
GUID, so exactly one bot goes for each wrap.

Stacking the group on the tank does put everyone inside Poison Nova's radius.
`AvoidPoisonNovaAction` still runs during the cast (the multiplier only zeroes the
other movement actions), so they step out and back. The alternative is the snakes
splitting up to chase spread-out casters.

### DIFF
--- a/src/Ai/Dungeon/GD/GDActionContext.h
+++ b/src/Ai/Dungeon/GD/GDActionContext.h
@@ -17,11 +17,15 @@ class WotlkDungeonGDActionContext : public NamedObjectContext<Action>
         WotlkDungeonGDActionContext() {
             creators["avoid poison nova"] = &WotlkDungeonGDActionContext::avoid_poison_nova;
             creators["attack snake wrap"] = &WotlkDungeonGDActionContext::attack_snake_wrap;
+            creators["slad'ran stack on tank"] = &WotlkDungeonGDActionContext::sladran_stack_on_tank;
+            creators["slad'ran tank hold"] = &WotlkDungeonGDActionContext::sladran_tank_hold;
             creators["avoid whirling slash"] = &WotlkDungeonGDActionContext::avoid_whirling_slash;
         }
     private:
         static Action* avoid_poison_nova(PlayerbotAI* ai) { return new AvoidPoisonNovaAction(ai); }
         static Action* attack_snake_wrap(PlayerbotAI* ai) { return new AttackSnakeWrapAction(ai); }
+        static Action* sladran_stack_on_tank(PlayerbotAI* ai) { return new SladranStackOnTankAction(ai); }
+        static Action* sladran_tank_hold(PlayerbotAI* ai) { return new SladranTankHoldAction(ai); }
         static Action* avoid_whirling_slash(PlayerbotAI* ai) { return new AvoidWhirlingSlashAction(ai); }
 };
 
--- a/src/Ai/Dungeon/GD/GDActions.cpp
+++ b/src/Ai/Dungeon/GD/GDActions.cpp
@@ -26,27 +26,28 @@ bool AvoidPoisonNovaAction::Execute(Event /*event*/)
 
 bool AttackSnakeWrapAction::Execute(Event /*event*/)
 {
-    Unit* boss = AI_VALUE2(Unit*, "find target", "slad'ran");
+    Unit* snakeWrap = GundrakSladran::GetAssignedSnakeWrap(botAI);
+    if (!snakeWrap) { return false; }
+
+    if (AI_VALUE(Unit*, "current target") == snakeWrap) { return false; }
+
+    return Attack(snakeWrap);
+}
+
+bool SladranStackOnTankAction::Execute(Event /*event*/)
+{
+    Player* tank = GundrakSladran::GetStackTank(botAI);
+    if (!tank) { return false; }
+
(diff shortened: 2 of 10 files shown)


jrad7:
https://streamable.com/9yat40

Shows the party stacked, letting the adds come to them. AoE
on the adds. Run out during poison nova.

Celandriel (approved):

Ben19882:
Its buged now. It gets many many snakes now please check

noisiver:
If the encounter itself is now suffering from a bug that's not something that is fixed here.

Ben19882:
> If the encounter itself is now suffering from a bug that's not something that is fixed here.

Its after the latest test-branch + dungeonclear

Please check it

Ben19882:
> > If the encounter itself is now suffering from a bug that's not something that is fixed here.
> 
> Its after the latest test-branch + dungeonclear
> 
> Please check it.

It could be the core ofc. i will try to find more info

jrad7:
Nothing in this PR can spawn a snake. They naturally spawn every few seconds until the fight is over.

2 vipers spawn every 8 seconds, and 3 constrictors spawn every 3-5 seconds.

Ben19882:
> 2 vipers spawn every 8 seconds, and 3 constrictors spawn every 3-5 seconds.

Yeah well now its a lot quicker then 
2 vipers spawn every 8 seconds, and 3 constrictors spawn every 3-5 seconds.

But i will check core script.

Thanks anyway!

Issue #2423, opened 2026-05-27
Make bots use skill books to learn spells


noisiver:
**Is your feature request related to a problem? Please describe.**
The bots won't use skill books - like `Manual of Eviscerate IX` or `Codex: Prayer of Fortitude` - to learn those spells. They won't do it even when told to use the item.

**Describe the solution you'd like**
It could be done by either making them do it automatically when they get one for their class or when telling them via `u [item-link]` as long as they don't already know the spell.

**Describe alternatives you've considered**

**Additional context**

HennyWilly:
This would be nice to have.

This is particularly relevant for mod-individual-progression. For example, Hunters can learn Tranquilizing Shot either at level 61 or by using the [Tome of Tranquilizing Shot](https://www.wowhead.com/classic/item=16665/tome-of-tranquilizing-shot) dropped by Lucifron.

This is a problem for level 60 bots: Flamegor in BWL can gain Frenzy, which causes him to spam [Fire Nova](https://www.wowhead.com/classic/spell=23462/fire-nova). The intended way to remove Frenzy is with Tranquilizing Shot, but bots cannot learn the spell at level 60 because they cannot use the tome.

Being able to tell a bot to use a skill book, or having it automatically use one when appropriate, would solve this and would be useful for similar cases in progression modules.

PR #2736, opened 2026-09-01
BWL Boss Strategies: Broodlord, Black Drakes, and Nefarian

HennyWilly:
## Pull Request Description

The following changes should complement and not collide with #2594 and #2646.

Adds rudimentary strategies for the remaining BWL bosses that previously had no bot-specific behavior: **Broodlord Lashlayer**, **Firemaw**, **Ebonroc**, **Flamegor**, and improved **Nefarian** positioning.
The changes for Nefarian were necessary for beating the boss (at least for me). The other ones make the fights easier.
 
For mod-individual-progression: Flamegor may be tricky if no Hunter knows Tranquilizing Shot. See #2423. 

- **Broodlord Lashlayer**: Ranged bots now move to a safe distance instead of standing close to the boss to avoid [Blast Wave](https://www.wowhead.com/classic/spell=23331). A new `BroodlordTankMultiplier` prevents tanks from swapping to corrupted whelps during the encounter, keeping aggro on the boss while DPS handles whelps independently.
- **Firemaw / Ebonroc / Flamegor**: Non-tanking bots standing in the boss's frontal breath cone now reposition out of it (reuses the existing `RearFlankAction` infrastructure).
- **Nefarian**: Non-tanks now position at the rear flank in order to avoid [Cleave](https://www.wowhead.com/classic/spell=19983), [Shadow Flame](https://www.wowhead.com/classic/spell=22539), and [Tail Lash](https://www.wowhead.com/classic/spell=23364).
- **Minor fix**: `BwlDeathTalonWyrmguardRangedMoveAwayAction` now skips repositioning if the bot is the target's current victim, preventing the bot from kiting the mob.

## Feature Evaluation

Each new trigger/action pair follows the existing pattern: a trigger checks whether the bot is a non-tank and whether the boss is active, and the action performs a simple distance/arc check followed by a `MoveAway` or repositioning call. No new timers, pathfinding, or persistent state are introduced. The `IsCorruptedWhelp()` helper is a simple NPC entry comparison. The breath-avoidance logic reuses the existing `RearFlankAction` base class.

Cost is minimal and comparable to existing BWL triggers: each trigger is a single `find target` lookup plus a `GetVictim()` comparison. The `MoveAway` / rear-flank repositioning only fires for bots currently in the breath cone or out of melee range, which is a small subset of the raid. The `BroodlordTankMultiplier` adds one `IsCorruptedWhelp` entry check per tank on `TankAssistAction` -> negligible.

## How to Test the Changes

1. **Setup**: Enter BWL with a raid containing multiple bots (tanks, healers, ranged DPS, melee DPS).
2. **Broodlord Lashlayer**: Pull Broodlord. Verify that ranged bots move to ~18 yd distance and stay there. Verify that tanks do NOT swap to corrupted whelp aggro. Only DPS bots should engage whelps.
3. **Firemaw / Ebonroc / Flamegor**: Pull each boss. Observe that non-tanking bots that are inside the frontal breath cone reposition to the side/behind. The current tank should remain stationary.
4. **Nefarian**: Pull Nefarian. During phase 2, non-tank bots should move to the rear flank of the boss.
5. **Death Talon Wyrmguard**: Verify that the existing ranged move-away behavior still works and that a bot being targeted by a Wyrmguard does not kite it.

## Impact Assessment

    - - [ ] No, not at all

New triggers add minimal costs checks per bot per tick: A `find target` lookup and a `GetVictim()` comparison. The breath-avoidance action only fires for bots inside the cone, a small subset. No new timers or persistent state.

    - - [ ] No
    - - [x] Yes (**explain why**)

Ranged bots now reposition during Broodlord and black drake encounters instead of standing near the boss. Tanks no longer swap to whelps during Broodlord. Non-tanks position next to Nefarian. These are improvements to existing BWL raid behavior.

    - - [ ] No

Adds new trigger, action, and multiplier classes. All follow existing BWL patterns and are organized in the same `BWL*` files, so maintenance overhead is contained.

## AI Assistance

- - [ ] No

Only for creating this PR.

 MaNGOS, another module)?

## Final Checklist

## Notes for Reviewers

- The `BwlBlackDrakeAvoidBreathAction` reuses `RearFlankAction` with a configurable `minAngle` (default `ANGLE_45_DEG`), so the same trigger/action pattern applies to all three black drakes with minimal code duplication.
- The Broodlord tank multiplier intentionally returns `0.0f` (block) for `TankAssistAction` targeting a whelp, but returns `1.0f` (allow) for non-whelp targets. This handles edge cases where elites may respawn or add during the encounter.


isitsimon:
Amazing! Does this only leave chromaggus with no strat now in BWL?
Great work

HennyWilly:
Thanks.

You could count the handling of `Affliction Bronze` as a strategy for Chromaggus.
Theoretically, bots could LoS the breaths or only use specific spells depending on the currently active Shimmer effect. But fighting Chromaggus never gave me any trouble.

IMHO the Death Talon Wyrmguards are still by far the most difficult encounter here.

isitsimon:
Yeah I think bots LoSing the breaths will make the fight feel way more complete. Thats my perspective anyway. You've done an awesome job 😁

HennyWilly:
Yeah, I can see your point. My main goal with this PR was to make the BWL encounters manageable for me, and I think these changes get them to that point. I'll probably leave Chromaggus as-is for now, though. Maybe someone else will pick up the LoS idea at some point. Thanks for the kind words!

src/Ai/Raid/BWL/BWLMultipliers.cpp
```diff
@@ -61,6 +61,34 @@ float VaelastraszTankMultiplier::GetValue(Action* action)
...
+float BroodlordTankMultiplier::GetValue(Action* action)
+{
+    if (!PlayerbotAI::IsTank(bot))
+        return 1.0f;
+
+    Unit* boss = AI_VALUE2(Unit*, "find target", "broodlord lashlayer");
+    if (!boss)
+        return 1.0f;
+
+    if (dynamic_cast<TankAssistAction*>(action))
```
brighton-chi: I don't think this multiplier entirely works, though it probably sort of does. It zeroes tank assist if the bot targets a whelp, but that doesn't cause the bot to consider the next highest-priority target. It would just keep the bot on its current target. So I think this keeps bots from picking up whelps, but it also keeps them from picking up elites unless the elites were already considered higher priority than the whelps by tank assist. The other issue would be if a tank picks up an elite and it is killed but whelps are up. The tank probably wouldn't go back to the boss until the whelps were dead because the whelps would likely be considered highest priority since they're untanked. Probably the tank would just idle then until the whelps were killed.

To make this sort of thing fully work, you have to exclude the target from being considered at all. That's what AppendTargetExclusions exists for. But of course it's totally your call whether this works well enough and it's not worth putting in a different system for what sounds like a fringe situation.

Also, typo in the comment ("thread").

src/Ai/Raid/BWL/BWLStrategy.cpp
```diff
@@ -30,16 +30,26 @@ void RaidBwlStrategy::InitTriggers(std::vector<TriggerNode*>& triggers)
...
         NextAction("bwl flamegor fire resistance", ACTION_RAID) }));
+    triggers.push_back(new TriggerNode("bwl flamegor not victim", {
+        NextAction("bwl flamegor avoid breath", ACTION_MOVE + 4) }));
 
     triggers.push_back(new TriggerNode("bwl affliction bronze", {
         NextAction("bwl use hourglass sand", ACTION_RAID) }));
 
-    triggers.push_back(new TriggerNode("bwl wild magic", {
+    triggers.push_back(new TriggerNode("bwl nefarian positioning", {
+        NextAction("rear flank", ACTION_MOVE + 4) }));
```
brighton-chi: Rear flank moves to the current target's rear flank. Does this make everybody try to get behind the adds in phase 3?

src/Ai/Raid/BWL/BWLMultipliers.cpp
HennyWilly: Okay, you recently added `AppendTargetExclusions`. I wasn't aware of this feature.

I added `RaidBwlStrategy::AppendTargetExclusions`, and it seems to work just fine. At least, I didn't encounter any issues during my testing.
Does this approach make sense, or is there a better way to handle this? I checked for `TargetValueExclusionType::Tank`.
I also noticed that `RaidMcStrategy` checks for `TargetValueExclusionType::Attacker`. I'm not entirely sure what that means in this context, though.

src/Ai/Raid/BWL/BWLStrategy.cpp
HennyWilly: You’re right. `RearFlankAction` checks the "current target", and that target can be a Bone Construct during phase 3.

I had to do a bit of refactoring to make `RearFlankAction` extensible. Also there is now a new `BossRearFlankAction` that can target a specific boss instead of the current target. The original `RearFlankAction` should behave exactly as before, so existing strategies remain unchanged.

Let me know if changes outside of the BWL code are okay for this PR, or if this should be split into a separate PR.

brighton-chi: If it's not okay then I'm in trouble because I do this sort of thing all the time =P

IMO things don't move quickly enough here to make it wise to do intertwined things piecemeal, but obviously my opinion is not really relevant.

src/Ai/Raid/BWL/BWLMultipliers.cpp
brighton-chi: I didn't look at that in MC so I'm not sure what it's doing there, but attackers is the third targeting path (I think consisting only of members of the poorly named CastDebuffSpellOnAttackersAction) that is for dotting up secondary targets. It doesn't run through tank or DPS assist (incidentally, that is also why bots will dot up cc'd targets and break them). Maybe it's useful if you don't want bots to dot up whelps because they die too fast so it's a waste. But it's probably not really important here.

There's actually a fourth path of sorts that I missed, which is AoE. AppendTargetExclusions really should include a way to disregard certain targets when calculating number of mobs to justify AoE, and it doesn't. But I digress.

src/Ai/Raid/BWL/BWLActions.h
```diff
@@ -58,6 +58,27 @@ class BwlVaelastraszMoveAwayAction : public MovementAction
...
+};
+
+// Firemaw / Ebonroc / Flamegor
+
+class BwlBlackDrakeAvoidBreathAction : public BossRearFlankAction
+{
+public:
+    BwlBlackDrakeAvoidBreathAction(PlayerbotAI* botAI, std::string const& bossName,
+                                   float minAngle = ANGLE_45_DEG)
+        : BossRearFlankAction(botAI, bossName, 0.0f, minAngle, M_PI) {}
```
kadeshar, line 78: How that works with ranged/healers. They go near boss and dont go away from him instantly?

src/Ai/Raid/BWL/BWLTriggers.cpp
```diff
@@ -54,6 +54,25 @@ bool BwlVaelastraszBurningAdrenalineTrigger::IsActive()
     return bot->HasAura(static_cast<uint32>(BlackwingLairSpells::SPELL_BURNING_ADRENALINE));
 }
 
+// Broodlord Lashlayer
+
+bool BwlBroodlordRangedTrigger::IsActive()
+{
+    return PlayerbotAI::IsRanged(bot) && AI_VALUE2(Unit*, "find target", "broodlord lashlayer");
```
kadeshar, line 61: This trigger always fire for ranged in this encounter which is bad. Should also check distance which even dont is checked in Action isUseful but in Execute

Image

brighton-chi: Sometimes I would do a trigger that is this basic. The problem is the check might need to be made in the action, and if it's an expensive check, it's undesirable to do it in both the action and trigger and result in duplication. But I do agree here that the trigger can be narrower. The check to exclude the victim could be in the trigger.

Distance could be checked in the trigger or in isUseful, but it will need to be done in the execute regardless so if you move that out you are duplicating. I don't think that one is expensive though so not a huge deal if duplicated in the trigger.

I generally don't think isUseful is, well, useful for raid actions. It's good for many class spells because all you are doing is calling the base class's execute and maybe passing a string, so then you override isUseful() if your spell needs different checks. But with raid strategies, you are never relying on a base class's execute so all your logic needs to be in the execute anyway.

HennyWilly: Makes sense.
I moved the target check from `BwlBroodlordRangedMoveAwayAction::isUseful` into `BwlBroodlordRangedTooCloseTrigger::IsActive`. The trigger also contains the distance check.

src/Ai/Raid/BWL/BWLActions.h
HennyWilly: Well, if non-tank bots are in front of the boss, they move behind or next to the boss.
For ranged/healers, they then move away again, e.g. by blinking (mage), disengaging (hunter), or simply fleeing.

As far as I can tell, this happens relatively quickly.
The intention is to use the same positioning logic as RearFlankAction. The bots may then get back to their preferred range.

So this should be the same behavior as with the "normal" RearFlankAction.

src/Ai/Raid/BWL/BWLActions.cpp
```diff
@@ -264,6 +264,32 @@ bool BwlVaelastraszMoveAwayAction::MoveAlongFleeDirection(Unit const* boss, floa
...
+
+bool BwlBlackDrakeAvoidBreathAction::isUseful()
+{
+    Unit const* boss = GetTarget();
+    if (!boss)
+        return false;
+
+    // Only reposition while standing in the frontal breath cone.
+    // No tail-swipe check: unlike RearFlankAction there is no rear danger zone here.
+    return boss->HasInArc(2.f * minAngle, bot);
```
brighton-chi: I know the formatting you have already exists elsewhere, but this should be 2.0f per AC code standards iirc.

src/Ai/Raid/BWL/BWLStrategy.cpp
```diff
@@ -5,7 +5,10 @@
  */
 
 #include "BWLStrategy.h"
+
```
brighton-chi: I believe when Nox redid the include style, he had there not be any blank lines between includes.

brighton-chi (commented):
Sorry this took a while. Nothing important to flag.

src/Ai/Raid/BWL/BWLActions.cpp
HennyWilly: Done 👍

src/Ai/Raid/BWL/BWLStrategy.cpp
HennyWilly: And done 👍

src/Ai/Raid/BWL/BWLStrategy.cpp
```diff
@@ -56,3 +68,43 @@ void RaidBwlStrategy::InitMultipliers(std::vector<Multiplier*>& multipliers)
     multipliers.push_back(new VaelastraszTankMultiplier(botAI));
     multipliers.push_back(new VaelastraszBurningAdrenalineMultiplier(botAI));
 }
+
+void RaidBwlStrategy::AppendTargetExclusions(GuidSet& exclusions, TargetValueExclusionType type)
+{
+    if (type != TargetValueExclusionType::Tank)
+        return;
+
+    AiObjectContext* context = botAI->GetAiObjectContext();
```
kadeshar, line 77: That context is used to something?

kadeshar: Nvm, @brighton-chi pointed me why.

kadeshar (approved):

PR #2576, opened 2026-07-15
chore: Standerdize include order and make headers self-contained

NoxMax:
## Pull Request Description

Standerdize the order of includes to follow [the AC code standard](https://github.com/azerothcore/wiki/blob/master/docs/cpp-code-standards.md#includes). And more importantly, the headers are made to be self-contained. 

I broke the changes in this PR into a bunch of commits, and each commit message should describe what it does.

## Feature Evaluation

- Describe the **minimum logic** required to achieve the intended behavior.
- Describe the **processing cost** when this logic executes across many bots.
No effect on processing.

## How to Test the Changes

Everything should compile correctly. There should be no issues up to the current state of test-staging. There might be a need for an ameding commit to address PRs that are merged into test-staging later.

## Impact Assessment

- Does this change increase per-bot/per-tick processing or risk scaling poorly with thousands of bots?
    - - [x] No, not at all
    - - [ ] Minimal impact (**explain below**)
    - - [ ] Moderate impact (**explain below**)

- Does this change modify default bot behavior?
    - - [x] No
    - - [ ] Yes (**explain why**)

- Does this change add new decision branches or increase maintenance complexity?
    - - [x] No
    - - [ ] Yes (**explain below**)

## AI Assistance

Was AI assistance used while working on this change?
- - [ ] No
- - [x] Yes (**explain below**)

Search and identify files that needed addressing. And made it create hundreds of temp testing files to check self-containment status.

## Code Provenance / Attribution

Was any code in this PR copied or adapted from a sister / upstream project (e.g. CMaNGOS playerbots,
 MaNGOS, another module)?
- - [x] No, all code in this PR is original
- - [ ] Yes (**name the project and the original author(s) below**)

## Final Checklist

- - [x] Stability is not compromised.
- - [x] Performance impact is understood, tested, and acceptable.
- - [x] Added logic complexity is justified and explained.
- - [x] Any new bot dialogue lines are translated.
- - [x] Any code ported/adapted from another project is attributed (project + author(s), `Co-authored-by:` trailer + in-file note).
- - [x] New source files use the GPLv2 header.
- - [x] Documentation updated if needed (Conf comments, WiKi commands).
- - [x] New and modified files do not introduce new compiler warnings.

PR #2738, opened 2026-09-02
fix(pmon): measure every calculated value, not just the target pickers

jrad7:
## Pull Request Description

pmon's Value block is missing most of what it should report. PERF_MON_VALUE is
only recorded in `UnitCalculatedValue::Get()`, which covers 35 of the 208 values
registered in ValueContext, all of them target pickers. `AttackersValue`, the
`NearestUnitsValue` classes and every bool and counter value are not measured.

`CalculatedValue` had the same instrumentation commented out and
`SingleCalculatedValue` had it live. #2576 removed both when it dropped the
`PerfMonitor.h` include from Value.h.

Calling `sPerfMonitor.start()` from `CalculatedValue::Get()` is too expensive to
put back. `start()` returns early when the monitor is off, but it takes `name` by
value and `getName()` also returns by value, so a std::string is built and
destroyed on every value read before `start()` can decline.

`CalculatedValue` now looks up its `PerformanceData` once, caches the pointer and
times `Calculate()` with an RAII scope. The enabled check is re-read on every
call, so turning the monitor on arms values that already exist. With the monitor
off it is one bool check. `UnitCalculatedValue::Get()` still uses `start()`
because it is the only value that pushes onto the performance stack.

Also included:

- `PerfMonitorOperation` and the new scope share one clock read and one record
  helper instead of duplicating the min/max/total accumulation.
- `PerfMonitor.h` includes `<string>`, which its own signatures use.

## Feature Evaluation

Minimum logic: with the monitor off, one bool read per value evaluation. With it
on, one map lookup per value per bot for the lifetime of that value object, then
two clock reads and a locked accumulate per evaluation. That is the same cost the
target pickers already pay.

Processing cost across many bots: the monitor-off path adds a call to
`PerfMonitor::IsEnabled()` and a branch in `CalculatedValue::Get()` and
`RefGet()`. No allocation and no string construction. The measuring path is
opt-in behind `AiPlayerbot.PerfMonEnabled`.

## How to Test the Changes

1. Set `BotActiveAlone = 100` and `botActiveAloneSmartScale = 0`. Leave
   `AiPlayerbot.PerfMonEnabled` at 0.
2. Start the server and wait for the bots to log in.
3. Run `playerbot pmon toggle`.
4. Five minutes later, run `playerbot pmon stack`.

The Value block should list values other than target pickers, `attackers` and the
`nearest ...` values in particular. Its `Value : Total` line reads much higher
than before because it sums the whole metric instead of a sixth of it. The last
`Total` line does not change, it is still driven by
`PlayerbotAI::UpdateAIInternal`.

Values nest, so a value that reads another value inside `Calculate()` has the
inner time counted in both rows. That was already true of the target pickers.

## Impact Assessment

  with thousands of bots?
    - - [ ] No, not at all

One bool check per value read while the monitor is off.

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

No new bot decision logic. The one added branch is the monitor's own enabled
check, covered above.

## AI Assistance

- - [ ] No

Used for the timing helper refactor and for drafting this description. Reviewed
and understood, including the decision to cache the `PerformanceData` pointer
rather than restore `start()` on the base template.

PerfMonitor.cpp and PerfMonitor.h are ported from CMaNGOS playerbots and carry
that attribution in their file headers. Those headers are untouched.

## Final Checklist

- - [x] Documentation updated if needed.

## Notes for Reviewers

Three files: PerfMonitor.h, PerfMonitor.cpp, Value.h.

`CalculatedValue::GetPerfData()` caches a `PerformanceData*` for the lifetime of
the value object. That is safe because `PerfMonitor::Reset()` zeroes the counters
in place and never erases the map or frees entries, and std::map pointers are
stable. If Reset() ever starts deleting, the cache has to go too.

### DIFF
--- a/src/Bot/Debug/PerfMonitor.cpp
+++ b/src/Bot/Debug/PerfMonitor.cpp
@@ -15,10 +15,62 @@
 #include "PerfMonitor.h"
 #include "Playerbots.h"
 
+namespace
+{
+std::chrono::microseconds Now()
+{
+    return (std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::high_resolution_clock::now()))
+        .time_since_epoch();
+}
+
+void RecordSample(PerformanceData* data, uint64 elapsed)
+{
+    std::lock_guard<std::mutex> guard(data->lock);
+    if (elapsed > 0)
+    {
+        if (!data->minTime || data->minTime > elapsed)
+            data->minTime = elapsed;
+
+        if (!data->maxTime || data->maxTime < elapsed)
+            data->maxTime = elapsed;
+
+        data->totalTime += elapsed;
+    }
+
+    ++data->count;
+}
+}  // namespace
+
+bool PerfMonitor::IsEnabled() { return sPlayerbotAIConfig.perfMonEnabled; }
+
+PerformanceData* PerfMonitor::GetOrCreate(PerformanceMetric metric, std::string const& name)
+{
+    std::lock_guard<std::mutex> guard(lock);
+    PerformanceData*& pd = data[metric][name];
+    if (!pd)
+    {
+        pd = new PerformanceData();
+        pd->minTime = 0;
+        pd->maxTime = 0;
+        pd->totalTime = 0;
+        pd->count = 0;
+    }
+
+    return pd;
+}
+
+PerformanceData* PerfMonitor::acquire(PerformanceMetric metric, std::string const& name)
+{
+    if (!IsEnabled())
+        return nullptr;
+
+    return GetOrCreate(metric, name);
+}
+
 PerfMonitorOperation* PerfMonitor::start(PerformanceMetric metric, std::string const name,
                                                        PerformanceStack* stack)
 {
-    if (!sPlayerbotAIConfig.perfMonEnabled)
+    if (!IsEnabled())
         return nullptr;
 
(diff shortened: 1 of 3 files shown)


src/Bot/Debug/PerfMonitor.cpp
```diff
@@ -15,10 +15,62 @@
 #include "PerfMonitor.h"
 #include "Playerbots.h"
 
+namespace
+{
+std::chrono::microseconds Now()
+{
+    return (std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::high_resolution_clock::now()))
+        .time_since_epoch();
+}
```
kadeshar, line 24: Why this changes overlap changes from #2737 ? I dont know what to review

jrad7: Yeah, both of the PR's needed the same code. If you'd rather I combine the PR's, that can happen.

If #2737 merges, then this PR becomes much smaller, just a Value.h update.

kadeshar: That is your decision. But shouldnt be open 2 PR with same code.

Celandriel (approved):

PR #2740, opened 2026-09-02
feat(pmon): dump the performance monitor to JSON alongside the table

jrad7:
## Pull Request Description

`.playerbots pmon`, `pmon tick` and `pmon stack` now also write the same data to
a JSON file next to the server logs, and log the path they wrote to. The printed
table in the console is unchanged.

The reason is that the table in the console is a summary, not a dump. It only prints a row when
it is over 0.1% of total, averages 0.25ms or more, or has had a single call over
1ms; on the total view it also drops every `Total` row that isn't
`UpdateAIInternal`. That is the right filter for reading in a console, but it
means anything cheap-but-very-frequent never shows up, and without the full stack
flag the stack names get truncated at the first `|`. Comparing two profiling runs
means eyeballing two walls of text, and anything that wants to chart or diff pmon
output has to scrape the log format.

The JSON has every row, unfiltered, with untruncated names: raw microsecond
totals, call counts, min/max, average, percentage, per-tick time and calls per
tick, sorted most expensive first, plus the denominators the percentages are
derived from (`fullTickCount`, `fullTickTotalTime`, `updateAiTotalTime`) so the
numbers can be recomputed or rescaled. Per-metric aggregates are included for
Trigger/Value/Action/RndBot. The `Total` metric gets its rows but no aggregate —
its scopes nest inside each other, so summing them double-counts.

Files are `<LogsDir>/pmon_total.json` for `pmon`/`pmon stack` and
`<LogsDir>/pmon_tick.json` for `pmon tick`, overwritten each time.

The metric-name switch (`PERF_MON_VALUE` -> `"Value"` etc.) was already
duplicated across both `PrintStats` branches and the dump needed it a third time,
so it moved into a single `MetricName` helper. No behavior change there.

Written by hand with `std::ostringstream` rather than a library: AzerothCore has
no JSON dependency, and its Boost floor (1.74) predates `boost::json` (1.75).
The output is a handful of scalar types and one string field, which is escaped.

## Feature Evaluation

- **Minimum logic:** the dump snapshots the existing `PerfMonitor` data map under
  the same lock `PrintStats` uses, copies each row's counters under that row's
  own lock, sorts, formats and writes one file. It reads state that is already
  being collected; it adds no new measurement points and stores nothing extra.
- **Processing cost across many bots:** none. The code only runs when an
  administrator types a `pmon` command, and it runs once on the calling thread,
  not per bot and not per tick. The work is proportional to the number of
  distinct measured names (the same set `PrintStats` already walks), not to the
  bot count.

## How to Test the Changes

1. Boot a server with bots and enable the monitor: `.playerbots pmon toggle`.
2. Let it run a few minutes so there is something to see.
3. Run `.playerbots pmon stack`. The console table prints as before, followed by
   a line like `Performance monitor dump written to
   /path/to/logs/pmon_total.json`.
4. Open that file and confirm it is valid JSON (`python3 -m json.tool
   pmon_total.json`, or `jq . pmon_total.json`).
5. Spot-check a few rows against the table: `percent`, `avgTime` (microseconds,
   so the table's ms column x1000) and `count` should match the printed values,
   and rows that the table filtered out should be present in the file.
6. Run `.playerbots pmon tick` and confirm `pmon_tick.json` appears with
   `"mode": "tick"`.
7. `.playerbots pmon reset` then `.playerbots pmon` — the file is rewritten from
   the cleared counters.

Some things worth checking deliberately:

- With the monitor toggled off and no data collected, the command writes nothing
  and returns cleanly rather than emitting a truncated file.
- If `LogsDir` is unset the file lands in the server's working directory; if it
  is not writable the command logs `Performance monitor could not write <path>`
  as an error and carries on.

## Impact Assessment

  with thousands of bots?
    - - [x] No, not at all

  Nothing runs on the bot update path. The only code that executes outside a
  `pmon` command is the extracted `MetricName` helper, which is called from the
  same two places the inline switch was called from — both inside `PrintStats`.

    - - [x] No

    - - [x] Yes

  One new method, `PerfMonitor::DumpJson`, plus small file-local helpers for JSON
  escaping and number formatting. The maintenance cost worth naming is that the
  JSON is now a second output format to keep in step with the counters: if a new
  field is added to `PerformanceData`, it should be added here too. The printed
  format is deliberately untouched, since it is parsed by external tooling.

## AI Assistance

- - [x] Yes

AI was used for the JSON writer and the helper extraction, and to draft this
description. I reviewed all of it: the locking follows the existing `PrintStats`
pattern, the escaping and the division-by-zero guards were checked by hand, and
the output was validated as parseable JSON and checked against the printed table.
The commit carries a `Co-Authored-By` trailer.

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (None added.)
- - [x] New source files use the GPLv2 header. (No new files.)

## Notes for Reviewers

The printed table format is intentionally byte-for-byte unchanged — external
tooling parses it, so any reformatting there would be a breaking change.

The `Total` metric having rows but no aggregate is deliberate, not an oversight.
Its scopes are nested (`FullTick` contains `UpdateAIInternal`, and so on), so an
aggregate line would be a meaningless sum. Note that the existing total-view
table does print one; the JSON omits it rather than reproducing a number a
consumer could easily misread.

`avgTime`, `minTime`, `maxTime` and `timePerTick` are all microseconds, matching
the monitor's internal unit. The table converts to ms/s for display; the JSON
does not convert anything, so consumers can scale as they like. `timeUnit` in
the header states this.

### DIFF
--- a/src/Bot/Debug/PerfMonitor.cpp
+++ b/src/Bot/Debug/PerfMonitor.cpp
@@ -13,7 +13,101 @@
  */
 
 #include "PerfMonitor.h"
+#include "Config.h"
 #include "Playerbots.h"
+#include <cmath>
+#include <ctime>
+#include <fstream>
+#include <iomanip>
+#include <sstream>
+
+namespace
+{
+std::string MetricName(PerformanceMetric metric)
+{
+    switch (metric)
+    {
+        case PERF_MON_TRIGGER:
+            return "Trigger";
+        case PERF_MON_VALUE:
+            return "Value";
+        case PERF_MON_ACTION:
+            return "Action";
+        case PERF_MON_RNDBOT:
+            return "RndBot";
+        case PERF_MON_TOTAL:
+            return "Total";
+        default:
+            return "?";
+    }
+}
+
+std::string JsonEscape(std::string const& value)
+{
+    std::ostringstream out;
+    for (char c : value)
+    {
+        switch (c)
+        {
+            case '"':
+                out << "\\\"";
+                break;
+            case '\\':
+                out << "\\\\";
+                break;
+            case '\b':
+                out << "\\b";
+                break;
+            case '\f':
+                out << "\\f";
+                break;
+            case '\n':
+                out << "\\n";
+                break;
+            case '\r':
+                out << "\\r";
+                break;
+            case '\t':
+                out << "\\t";
+                break;
+            default:
+                if (static_cast<unsigned char>(c) < 0x20)
+                    out << "\\u" << std::hex << std::setw(4) << std::setfill('0') << static_cast<uint32>(c)
+                        << std::dec;
+                else
+                    out << c;
+                break;
+        }
+    }
+
+    return out.str();
+}
+
+std::string JsonNumber(double value)
+{
+    if (!std::isfinite(value))
+        value = 0.0;
+
+    std::ostringstream out;
+    out << std::fixed << std::setprecision(3) << value;
+
+    return out.str();
+}
+
+std::string JsonRatio(double value, double total)
+{
+    return JsonNumber(total > 0.0 ? value / total : 0.0);
+}
+
+std::string JsonPath(bool perTick)
+{
+    std::string dir = sConfigMgr->GetOption<std::string>("LogsDir", "", false);
+    if (!dir.empty() && dir.back() != '/' && dir.back() != '\\')
+        dir.push_back('/');
+
+    return dir + (perTick ? "pmon_tick.json" : "pmon_total.json");
+}
+}  // namespace
 
 PerfMonitorOperation* PerfMonitor::start(PerformanceMetric metric, std::string const name,
                                                        PerformanceStack* stack)
@@ -82,28 +176,7 @@ void PerfMonitor::PrintStats(bool perTick, bool fullStack)
         {
             std::map<std::string, PerformanceData*> pdMap = i->second;
 
-            std::string key;
-            switch (i->first)
-            {
-                case PERF_MON_TRIGGER:
-                    key = "Trigger";
-                    break;
-                case PERF_MON_VALUE:
-                    key = "Value";
-                    break;
-                case PERF_MON_ACTION:
-                    key = "Action";
-                    break;
-                case PERF_MON_RNDBOT:
-                    key = "RndBot";
-                    break;
-                case PERF_MON_TOTAL:
-                    key = "Total";
-                    break;
-                default:
-                    key = "?";
-                    break;
-            }
+            std::string const key = MetricName(i->first);
 
             std::vector<std::string> names;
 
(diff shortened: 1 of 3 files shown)


jrad7:
Dumping to json so all of our wonderful vibecoded dashboards can consume it more easily.

src/Bot/Debug/PerfMonitor.cpp
```diff
@@ -255,6 +308,164 @@ void PerfMonitor::PrintStats(bool perTick, bool fullStack)
...
+    std::string const path = JsonPath(perTick);
+    std::ofstream file(path.c_str(), std::ios::out | std::ios::trunc);
+    if (!file)
+    {
+        LOG_ERROR("playerbots", "Performance monitor could not write {}", path);
+        return;
+    }
+
+    file << out.str();
+    file.close();
```
kadeshar, line 470: Image

Streaming file and this statement can be problematic. File can be empty or partially when external dashboard will try to read file.

src/Bot/Debug/PerfMonitor.cpp
```diff
@@ -255,6 +308,164 @@ void PerfMonitor::PrintStats(bool perTick, bool fullStack)
...
+
+    LOG_INFO("playerbots", "Performance monitor dump written to {}", path);
```
kadeshar, line 492: Without check is can be not true.

src/Bot/Debug/PerfMonitor.cpp
```diff
@@ -255,6 +308,164 @@ void PerfMonitor::PrintStats(bool perTick, bool fullStack)
...
+            uint64 typeTotalTime = 0;
+            uint64 typeMinTime = 0;
+            uint64 typeMaxTime = 0;
+            uint64 typeCount = 0;
+            for (Sample const& row : metric.second)
+            {
+                typeTotalTime += row.totalTime;
+                typeCount += row.count;
+                if (!typeMinTime || (row.minTime && typeMinTime > row.minTime))
+                    typeMinTime = row.minTime;
```
kadeshar, line 411: That is not different from table approach?

src/Bot/Debug/PerfMonitor.cpp
jrad7: Changed the write to a temp file, then an atomic rename.

src/Bot/Debug/PerfMonitor.cpp
jrad7: This is fixed, we now check for a failed write to the temp file and for a failed rename before declaring success.

src/Bot/Debug/PerfMonitor.cpp
jrad7: Changed to more closely match what the console table does.

kadeshar (approved):

PR #2747, opened 2026-09-03
Refactor bot movement and travel, enabling node system

Celandriel:
## Pull Request Description

Hi all. This is a big one. 
The main reason for this PR is to enable the travel node system to allow bots to travel long distances, cross map, etc. What ended up what a wholesale revision of the approach by Bash and I. 

There are many changes, and Ive tested them as much as I can. 

To view the updated maps you can use https://github.com/Celandriel/wowmin/tree/map-compare which allows you to load them and see the paths. 

##MMAPS
The travel nodes were generated using updated mmaps from our core, not the stock values. It will work fine if users dont update, they just dont get the advantages of the slope or fixes in a few zones. Im working on a companion PR, but they can go separately. 

## Feature Evaluation

The feature is off by default (It should be for a while until we are happy with the results.)

- TBD need to rerun a comparison here. 

## How to Test the Changes

Selfbot with rpg enabled, or just look around at what bots are doing. 

## Impact Assessment

    - - [ ] No, not at all
This will have some performance impact.

    - - [ ] No
    - - [x] Yes (**explain why**)
Not be default, but it should by changing how pathing in any single place is done. 

    - - [ ] No
Travel.. Complicated. Many branches. 

## AI Assistance

- - [ ] No

Specification development, implementation, extraction of node data and conversion to update sql. 

 MaNGOS, another module)?
- - [ ] No, all code in this PR is original
- - [x] Yes (**name the project and the original author(s) below**)

cmangoes project.

## Final Checklist

## Summary by CodeRabbit

- **New Features**
  - Added automated client-data extraction for maps, collision data, and movement data.
  - Added travel-node generation and zone debugging commands with visual markers.
  - Added configurable graph-based travel routing and transport-aware movement.
  - Improved bot teleport handling and transport journey resumption.
  - Added movement and pathfinding CSV logs.

- **Bug Fixes**
  - Improved pathfinding around water, cross-map travel, docks, and transports.
  - Improved movement holds and quest interaction pauses.
  - Improved travel-node reachability and path validation.

- **Documentation**
  - Added required server configuration instructions for mod-playerbots.


brighton-chi:
image

dying

Celandriel:
> image
> 
> dying

Haha yea, the SQLs are pretty large.

kadeshar:
Ok, now how review that xD

Celandriel:
> Ok, now how review that xD

Read in this order

- LastMovementValue.h/.cpp. Shared movement state. Small.
- MovementActions.h. The numbered design block describes my intent
- MovementActions.cpp in pipeline order: MoveTo, MoveTo2, ResolveMovePath, HandleSpecialMovement with WaitForTransport and BoardTransport, then DispatchMovement. Read MoveTo2 whole, not as a diff.
- TravelNode.cpp. Runtime first (GetFullPath, GetNodeRoute, the TravelPath methods), then generation (generatePaths and what it calls), then Init.
- TravelMgr.cpp. Mostly deletions. Look at getPathStepFrom, getPathFromPath, GetPathingCreature.
- Callers: NewRpgBaseAction::MoveFarTo, the transport auto-sync guard in PlayerbotAI.cpp, the GM edit paths in DebugAction.cpp and PlayerbotCommandScript.cpp.
-

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -2936,4 +2354,989 @@ bool MoveAwayFromPlayerWithDebuffAction::Execute(Event /*event*/)
...
+    }
+
+    bool const generatePath = !bot->IsFlying() && !bot->isSwimming() &&
+                              !bot->IsInWater() && !bot->IsUnderWater();
+
+    bot->ClearEmoteState();
+    if (!bot->IsStandState())
+        bot->SetStandState(UNIT_STAND_STATE_STAND);
+    if (bot->IsNonMeleeSpellCast(true))
+        bot->InterruptNonMeleeSpells(true);
```
jrad7: I think this will cancel spell casts that the engine would normally allow

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -878,36 +520,48 @@ bool MovementAction::IsMovingAllowed(WorldObject* target)
...
         return false;
 
-    // float distance = ServerFacade::instance().GetDistance2d(bot, target);
-    // if (!bot->InBattleground() && distance > sPlayerbotAIConfig.reactDistance)
-    //     return false;
-
     return IsMovingAllowed();
 }
 
+// Priority check for a new move request: requires equal or higher priority,
```
jrad7, line 526: This says it requires equal or higher priority, but the priority checks below don't test for equality.

src/Ai/Base/Value/LastMovementValue.cpp
```diff
@@ -9,67 +9,71 @@
...
-    lastFollow = nullptr;
     lastAreaTrigger = 0;
     lastFlee = 0;
     nextTeleport = 0;
     msTime = 0;
-    lastdelayTime = 0;
     priority = MovementPriority::MOVEMENT_NORMAL;
+    holdStartMs = 0;
+    holdDurationMs = 0;
+    lastTransportEntry = 0;
```
jrad7, line 40: is it ok that lastCompletedTransportEntry isn't being cleared here? I think if it's not cleared, they can never board that transport again.

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -834,15 +474,17 @@ bool MovementAction::ReachCombatTo(Unit* target, float distance)
...
 
-    // if (bot->GetExactDist(tx, ty, tz) <= shortenTo)
-    //     return false;
-
     path.ShortenPathUntilDist(G3D::Vector3(tx, ty, tz), shortenTo);
     G3D::Vector3 endPos = path.GetPath().back();
-    return MoveTo(target->GetMapId(), endPos.x, endPos.y, endPos.z, false, false, false, false,
-                  MovementPriority::MOVEMENT_COMBAT, true);
+
+    bool moved = MoveTo(target->GetMapId(), endPos.x, endPos.y, endPos.z, false, true, false, false,
```
jrad7: I think the 6th param here is 'react' and that was changed to true. Does this mean every melee chase tick will rerun the route pipeline?

src/Ai/Base/Actions/MovementActions.cpp
```diff
@@ -2936,4 +2354,989 @@ bool MoveAwayFromPlayerWithDebuffAction::Execute(Event /*event*/)
...
+                              !bot->IsInWater() && !bot->IsUnderWater();
+
+    bot->ClearEmoteState();
+    if (!bot->IsStandState())
+        bot->SetStandState(UNIT_STAND_STATE_STAND);
+    if (bot->IsNonMeleeSpellCast(true))
+        bot->InterruptNonMeleeSpells(true);
+
+    for (auto& pt : points)
+        bot->UpdateAllowedPositionZ(pt.x, pt.y, pt.z);
```
jrad7: these points came out of the navmesh so it's guaranteed to be walkable ground, i think this can cause z to be below where it is intended in rare cases where it could default to the base ground.

To contrast, `bot->UpdateAllowedPositionZ();` is used on line 423, but it is taking in hand-rolled coords, so it's needed there since the dest could be in the middle of a hill

src/Ai/Base/Actions/MovementActions.cpp
brighton-chi: Yeah we should use CastStop() unless there is a particular reason to be cancelling already-cast but unresolved spells. The guard also isn't needed.

src/Ai/Base/Actions/MovementActions.cpp
Celandriel: I set it back to false,  I think I was trying to figure out when it should happen, and React is a bad name for what it does.

Celandriel:
@coderabbitai full review

Celandriel:
@coderabbit full review

PR #2750, opened 2026-09-04
Create GenericRogueStrategy class & refresh poisons in combat

brighton-chi:
## Pull Request Description

I mainly just wanted to get Rogues to refresh poisons in combat because it kills their DPS to lose them mid-encounter, but there wasn't a good place to put the strategies because Rogues, unlike every other class in the mod, don't have a generic combat strategy. So I went ahead and added one to put the combat poison strategies in. This file/strategy also is the better place to house many of the strategies that are currently in the spec-based files (like Evasion, Cloak of Shadows, Sprint, etc.), but I'm not dealing with that right now and probably not ever.

I set the priority for applying poisons to higher than any offensive ability (including Slice and Dice) but below defensive ones. I think this is justified since poisons constitute (at level 70 anyway) 20-25% of Combat's dps and 35-40% of Assassination's dps (and also enables Assassination's finisher, so basically if poisons drop off mid-combat for an Assassination Rogue, they're getting outdps'd by the Prot Warrior).

I have no desire to refactor the class, but after this is merged, I'll do a pass to rename the specs from "dps" and "melee" in a separate PR.

## Feature Evaluation

  - New GenericRogueStrategy.cpp and GenericRogueStrategy.h files
  - New GenericRogueStrategy class that inherits from CombatStrategy
  - DpsRogueStrategy (i.e., Combat spec) and AssassinationRogueStrategy now inherit from GenericRogueStrategy
  - One ActionNode and two TriggerNodes in GenericRogueStrategy to apply MH instant and OH deadly poisons in combat (fallback to OH instant if no deadly is available, not because I think this is a good idea but because this matches the non-combat strategy)
  - Some formatting/naming cleanups

Edit: Forgot, I also changed the conf to take the chance of an rndbot being sub to 0 since the spec isn't implemented. I tuned to raise Combat a bit and lower Assassination a bit since Assassination is not viable before level 62 (or 63?), and Combat is the better leveling/questing spec always. I made it 60/40; probably it should be even higher in favor of Combat, but I didn't want to rock the boat too much.

## How to Test the Changes

Easiest way is to log into a Rogue character, remove poisons on weapons, attack a mob, and turn on selfbot. They should put their poisons back on and otherwise keep fighting like normal.

## Impact Assessment

    - - [ ] No, not at all

The trigger wasn't previously evaluated in combat, but this is not going to make any difference.

    - - [ ] No
    - - [x] Yes (**explain why**)

Man I just told you.

    - - [x] No

## AI Assistance

- - [ ] No

I had AI take a pass after I finished and was told I unnecessarily included GenericRogueStrategy.h in RogueAiObjectContext.cpp so thanks for that value add

 MaNGOS, another module)?

## Final Checklist

### DIFF
--- a/src/Ai/Class/Rogue/Action/RogueActions.cpp
+++ b/src/Ai/Class/Rogue/Action/RogueActions.cpp
@@ -81,118 +81,52 @@ bool CastTricksOfTheTradeOnMainTankAction::isUseful()
 
 bool UseDeadlyPoisonAction::Execute(Event /*event*/)
 {
-    std::vector<std::string> poison_suffixs = {" IX", " VIII", " VII", " VI", " V", " IV", " III", " II", ""};
-    std::vector<Item*> items;
-    std::string poison_name;
-    for (std::string& suffix : poison_suffixs)
+    std::vector<Item*> const items =
+        AI_VALUE2(std::vector<Item*>, "inventory items", "Deadly Poison");
+    for (Item* const item : items)
     {
-        poison_name = "Deadly Poison" + suffix;
-        items = AI_VALUE2(std::vector<Item*>, "inventory items", poison_name);
-        if (!items.empty())
-        {
-            break;
-        }
-    }
-    if (items.empty())
-    {
-        return false;
-    }
-    Item* const itemForSpell = bot->GetItemByPos(INVENTORY_SLOT_BAG_0, EQUIPMENT_SLOT_OFFHAND);
-    return UseItem(*items.begin(), ObjectGuid::Empty, itemForSpell);
-    // return UseItemAuto(*items.begin());
-}
+        // This check is needed only if there is a name match. I think the only one is Handbook
+        // of Deadly Poison V, which might be restored by IP. Keeping this here just in case.
+        if (item->GetTemplate()->Class != ITEM_CLASS_CONSUMABLE)
+            continue;
 
-bool UseDeadlyPoisonAction::isPossible()
-{
-    std::vector<std::string> poison_suffixs = {" IX", " VIII", " VII", " VI", " V", " IV", " III", " II", ""};
-    std::vector<Item*> items;
-    std::string poison_name;
-    for (std::string& suffix : poison_suffixs)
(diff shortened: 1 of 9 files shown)


brighton-chi:
Oh, also, inheritance through CombatStrategy instead of MeleeCombatStrategy is intentional. Only DKs inherit through MeleeCombatStrategy. Warriors don't. As far as I can tell, this costs Rogues direct access to reach melee, but they have it built it as a default action for both specs anyway (and that's basically the pattern for melee specs).

I don't really care though if the pref is to go through MeleeCombatStrategy then I can make that change.

Celandriel (commented):
Claude flagged `UseInstantPoisonAction::isPossible()` as potentially being too heavy in a combat engine.
The three suggestions it had were 
 (a) give the poison-name lookups a real checkInterval (poisons in bags do not change tick to tick); (b) resolve the highest available rank once and cache the Item*/id rather than re-deriving it from a string every call; (c) at minimum, make the trigger go inactive when isPossible() has failed recently so the no-poison case does not spin.

brighton-chi:
Lol those actions are really fucking bad. You want me to try to do that as part of this PR or just strip the in-combat poison, get this restructuring in, then do a PR to add in-combat poison in and redo the poison actions as part of that? I kinda feel like the latter makes more sense.

Edit: Actually, lemme see how simple it would be to do just something in the meantime.

brighton-chi:
Ok, two changes for now: 
- Get rid of the duplicated checks in isPossible(), but override anyway because the base class's isPossible() will otherwise do its own item count lookup.
- Collapse all the rank-based lookups into a single one per action since parseItems already sorts by highest rank.

This is just a bandaid. I think poisons need reworking generally to support other types, allow configurable MH/OH strategies, and have different defaults for BGs/arenas. And along with that can come a better system overall, like having a value as you noted.

I did some basic testing of in/out of combat and having different poison ranks/types in the inventory, and everything seemed fine.

Celandriel (approved):

PR #2751, opened 2026-09-05
Gruul Bug Fixes & Improvements

brighton-chi:
## Pull Request Description

Reopening now that the Auchindoun PR with helper updates is merged. As noted before, this fixes some bugs:

1. The prior version of Gruul had a really bad implementation of the Shatter spread that made it practically useless. First, for the trigger, it used the version of the Ground Slam aura that covered only the first 2s of the sequence (the player is slowed over the course of 7s, frozen at 7s, and then Shatter hits at 9.7s). Second, for the action, it called FleePosition(), which looks for a position from which the bot can attack, meaning it would send all melee back into melee range of Gruul.
2. Then, Gruul's Shatter itself causes a bug via its interaction with MovementAction::MoveTo(). The tl;dr is that MoveTo has a built in delay based on the speed of the prior movement, and because the immobilization from Gruul's Shatter survives for 300ms after the end of the Shatter sequence, MoveTo() tries to construct a spline during that period, calculates an infinite movement delay due to movement speed being 0, clamps that to a 5s max, and makes it so that bots stand around for 5s after Shatter before any movement is accepted again. This was fixed via a multiplier.
3. Maulgar also had a bug for the selection of the Kiggler moonkin tank where I had defined a local variable for spec outside of the group loop using bot, meaning that the group loop was actually checking the bot's spec. It still worked if you set the assistant flag on a moonkin, but it wouldn't work properly otherwise.
4. The Warlock to Fel Stalker Banish assignment didn't work properly so I fixed that.
5. The initial ranged spread was not limited to just ranged when assigning spots; melee didn't take part since they were excluded from the trigger, but it did mean the ranged spread was not even.

Also, this time around, I made some much more significant structural changes that I think are good to reflect across all raids. Notably, I'm using `InstanceScript::IsEncounterInProgress()` to read whether a boss encounter is active. This is much more reliable and cheaper than using combat or threat list checks.* This is used to gate the action that erases maps and clears RTI icons (the latter also being subject to being out of combat so you can still mark trash), and the trigger for such action is also throttled. Then, I made intermediate classes for triggers and multipliers so that all actions that are solely for boss fights are gated behind IsEncounterInProgress. This has a significant reduction on performance hit when the raid is out of combat or fighting trash. I've tested with pmon, and it basically makes the triggers invisible. During a boss fight, there is no benefit of this, but there is no cost either as the check is super cheap.

Additionally, I implemented values for the role assignments of the Krosh mage tank and Kiggler moonkin tank and the grid search of Fel Stalkers. I used 1000 ms cache intervals for both.

*Caveat: `InstanceScript::IsEncounterInProgress()` functioning is reliant on the AC boss script actually properly setting the encounter to IN_PROGRESS. For TBC, this is true for all of them except for Terestian Illhoof (totally missing the relevant code for some reason), Illidari Council (has a totally different structure than other raids, I think due to the HP sharing of the council members), the very beginning of Kil'jaeden (he's not set to IN_PROGRESS until the first of the 3 Hands of the Deceivers dies in Phase 1), and Magtheridon (not even going to get into this one because his boss script is totally fucked up).

## Feature Evaluation

One new action/trigger pair. One new multiplier. Consolidated melee tank trigger/actions from 3 into a single 1. Added value file with 3 values. New intermediate classes as described above.

## How to Test the Changes

Run Gruul's Lair. The strategy should work the same mostly, except that it should be much better for Gruul's Shatter.

## Impact Assessment

    - - [x] No, not at all

No, it should be a significant performance improvement (with respect to the strategy, which is still going to be a drop in the bucket compared to thousands of rndbots running around).

    - - [ ] No
    - - [x] Yes (**explain why**)

With the gruulslair strategy.

    - - [x] No

The instance value system already exists thanks to jrad, and raids can leverage the same system as he has used for dungeons.

## AI Assistance

- - [ ] No

I relied on AI a lot more heavily this time to make so many big structural changes, but I've reviewed and done many passes over the code, and ultimately I think this is going to be a much better template for structuring going forward.

 MaNGOS, another module)?

## Final Checklist


brighton-chi:
jrad ran some testing, 100 runs, 100 HKM kills but only 98 Gruul. There were some issues with Cave In so I decided to go ahead and make its own method and multiplier.

I pulled Gruul with my damage nerfed to practically nothing and went AFK. The bots made it all the way to 30 stacks of Growth and then wiped shortly thereafter, though attrition started happening at 20+, and this was with SWP gear so they have a lot more stam than they would with appropriate gear. But that was fun; I didn't even know he could get this big.

Screenshot 2026-09-06 001325

Celandriel (approved):

PR #2754, opened 2026-09-05
fix(Bot): re-anchor m_lastFallZ so LFG can teleport bots into a dungeon

terry90:
## Pull Request Description

Bots queued through the Dungeon Finder often fail to teleport into the instance. On my realm it
was two or three of the four bots on every single pop, with the same bots standing where they
were while the rest of the party zoned in. Nothing ever brought them in afterwards.

`LFGMgr::TeleportPlayer` refuses them with `LFG_TELEPORTERROR_FALLING`:

```cpp
else if (player->IsFalling() || player->HasUnitState(UNIT_STATE_JUMPING))
    error = LFG_TELEPORTERROR_FALLING;
```

The catch is that `Player::IsFalling()` does not test movement flags. It is

```cpp
return GetPositionZ() < m_lastFallZ && !IsInFlight();
```

and it shadows the non-virtual `Unit::IsFalling()`, so a `Player*` call site like the one above
gets the Z comparison rather than the flag test.

`m_lastFallZ` is maintained by the client. Its only continuous updater is
`Player::UpdateFallInformationIfNeed`, and that function has exactly one caller in the core,
inside `WorldSession::HandleMovementOpcodes`. A real player re-anchors the value on every
movement packet they send, which is why a human in the party is never refused.

Bots send no movement opcodes. They move by spline `Relocate`, which never touches
`m_lastFallZ`, so the value stays frozen at whatever last wrote it. Every writer a bot can reach
stores a Z at or above the ground it then stands on:

| writer | value stored | result |
|---|---|---|
| `RandomPlayerbotMgr::RandomTeleport` | `ground + 0.05f` | those 5 cm are shed on the first pathed step |
| `MotionMaster::MoveFall` | Z at the top of the fall | below it once landed |
| `CheckMountStateAction::CompleteDismount` | the pre-fall `startZ` | never re-anchored after landing |
| flight path landing | the flight master's platform | walked down from |

Once a bot is standing below its own frozen mark it reads as falling, and it stays that way
forever, because the only thing that would refresh the mark is the teleport it is being refused.

That the state is per bot is what makes it a subset of the party rather than all of it or none of
it. `MakeNewGroup` walks every member in one synchronous loop, and each one is an independent
question of whether that bot happens to be below its own mark. The ones that get in are the ones
still anchored at their feet, typically a bot that just logged in or that has been standing still.

It also gets worse over time. A bot that fails is left at `LFG_STATE_DUNGEON`, where
`LfgJoinAction` will not re-queue it (it requires `LFG_STATE_NONE`) and `LfgLeaveAction` will not
release it. Every failed pop takes its stranded bots out of the LFG pool until that group
disbands.

The fix gives bots the heartbeat the missing client would have sent:

```cpp
if (!bot->Unit::IsFalling())
    bot->SetFallInformation(0, bot->GetPositionZ());
```

`Unit::IsFalling()` is the real test, movement flags plus an unfinalized falling movespline.
Guarding on it means a bot that is genuinely falling keeps its reference, so `Player::HandleFall`
still computes fall damage off the correct height, while a bot that is standing or walking gets
its reference put back at its feet, which is what the client would have done.

`CheckMountStateAction::CompleteDismount` gets the same re-anchor after `HandleFall`. The
per-tick refresh already covers it a tick later, but it leaves that function correct on its own.
The comment at the top of that file already describes the same class of problem, bots missing the
client packets that would normally clean up after a fall.

I looked at fixing only the writers instead, since `RandomTeleport` is the loudest one. That does
not work. A bot that simply walks downhill from where it logged in ends up below its mark with no
teleport involved, so the refresh has to be periodic rather than tied to particular call sites.

I did not touch `LFGMgr` itself. The gate there is correct for a real player, and this module is
what stands in for the missing client. It already fabricates `MSG_MOVE_TELEPORT_ACK` in
`PlayerbotAI::HandleTeleportAck` for the same reason.

One thing reviewers may want to know: `UNIT_STATE_JUMPING`, the other half of that branch in
`LFGMgr`, is never set anywhere in the core. There is no `AddUnitState` for it, only three places
that read it. That is why the existing `ClearUnitState(UNIT_STATE_ALL_STATE)` calls in
`LfgActions.cpp` never helped with this.

## Feature Evaluation

Minimum logic is one branch and two assignments per bot per tick. The mark has to be fresh at
whatever arbitrary moment `LFGMgr::TeleportPlayer` reads it, and nothing in the module knows when
that moment is, so a periodic refresh is the smallest thing that works. There is no new state, no
new field and no config.

Processing cost per call: `Unit::IsFalling()` is a bitmask test on `m_movementInfo` plus, when
that misses, three inlined movespline checks (`Finalized`, `Initialized`, `isFalling`).
`SetFallInformation` is two member assignments in the header. No allocation, no lock, no map or
grid lookup, no DB access, nothing virtual.

It is placed after the early return in `UpdateAI` that already discards bots which are not in
world, are being teleported, or are logging out, so it never runs for a bot in an invalid state.
It is deliberately above the `CanUpdateAI()` gate, because the AI delay can be seconds long and
the value needs to track the bot's actual position rather than its decision cadence.

I have not measured it with `pmon`. If a maintainer wants that number before merging, say so and
I will produce it. My honest expectation is that it disappears into the noise next to everything
else `UpdateAI` does per tick, but that is an expectation, not a measurement.

## How to Test the Changes

Setup: `AiPlayerbot.RandomBotJoinLfg = 1` and enough random bots for the finder to fill a party.
My realm runs several hundred. You need one real player, because the module only lets an LFG match
proceed when a non-bot is in it.

1. Log in, queue for a random dungeon in the Dungeon Finder, accept the pop.
2. Count how many party members are actually inside the instance with you.

Before the change, two or three of the four bots stay outside, every time. After it, all four
arrive.

To see the server's own reasoning, raise the LFG logging. Note that setting this through an
environment variable alone will not work if you run in Docker: `ConfigMgr::GetKeysByString`
enumerates keys loaded from the config file, so a `Logger.*` key that exists only in the
environment is never enumerated and never read. Put it in `worldserver.conf`:

```
Logger.lfg=5,Console Server
```

and make sure the `Appender.Server` level is at least 5 as well, since the appender filters
separately from the logger. Then watch for these two lines, both from `LFGMgr::TeleportPlayer`:

```
Player [name] is being teleported in to map [id] ...
Player [name] could NOT be teleported in to map [id] ... Error: N
```

You should get one `is being teleported` line per party member and no `could NOT` lines. The
error ordinal on a refusal is worth reading if you see one: 1 is PLAYER_DEAD, 2 is FALLING, 3 is
IN_VEHICLE, 4 is FATIGUE, 6 is INVALID_LOCATION, 8 is COMBAT. Before this change every refusal I
saw was a 2.

What I actually ran: three pops on a live realm with the patched build, 15 `is being teleported
in to map` lines and zero refusals, five of five members each time, across Deadmines and Ragefire
Chasm.

## Impact Assessment

    - - [ ] No, not at all

It does add work per bot per tick, so I am not claiming zero. What it adds is a flag test and two
assignments, with no allocation, no locking and no lookups, in a function that already runs per
bot per tick and does considerably more than this. The cost is constant per bot and does not grow
with bot count beyond the linear term that `UpdateAI` already carries.

    - - [ ] No
    - - [x] Yes (**explain why**)

Bots that previously could not enter an LFG dungeon now enter it. I am marking this Yes rather
than No because the observable behavior does change, even though the intent is to restore
behavior that was meant to work rather than to add anything. There is no new option and nothing
to opt into. Bots that were already entering successfully are unaffected, since for those the
refresh writes the value the client-side logic would have arrived at anyway.

    - - [x] No

It adds one guard, not a decision path. No new strategy, trigger, action or value, and no new
state to keep consistent. The comment is longer than the code because the shadowing between
`Player::IsFalling` and `Unit::IsFalling` is the sort of thing that gets silently undone by a
later refactor if nobody wrote down why the qualification is there.

## AI Assistance

- - [ ] No

Yes, and it did most of the digging. I used Claude to trace the call graph from the symptom back
to the cause, which is where the value came from, which functions write it, and which of them a
bot can actually reach. It also drafted the two code lines, the comments and this description.

I checked the claims that the change rests on by reading the source myself: that
`Player::IsFalling` is the Z comparison and shadows the non-virtual `Unit::IsFalling`, that
`UpdateFallInformationIfNeed` has the single caller inside the movement handler, and that
`UNIT_STATE_JUMPING` is never set. The change is two lines and I understand what both of them do
and why they are where they are. It is built and running on my realm, and the testing described
above is mine.

The one thing I have reasoned about rather than observed is fall damage. The guard is meant to
leave a genuine fall alone so `HandleFall` still measures from the right height, and reading
`MoveFall` that holds, but I have not watched a bot take fall damage before and after to confirm
it. If a reviewer wants that checked before merge, that is fair.

 MaNGOS, another module)?

## Final Checklist

No dialogue lines and no new files, so those two are nothing to do rather than something done.
On warnings, see the note below about which branch I compiled.

## Notes for Reviewers

Where I compiled it. The change is committed here on top of `test-staging`, but the build I
tested is `master` at `b949b50b` plus this change, which is what my realm runs. That build
finished with no errors and no new warnings. I have not separately compiled `test-staging` at
`075242fd` with warnings promoted to errors, so if CI catches something there it will be
something I did not see.

On attribution of the fix itself, since the version bump could confuse it. My realm went from
module `2f7d9f7` to `b949b50b` in the same rebuild that took this patch, roughly eleven days of
commits, so I checked that none of it explains the result.
`b949b50b:src/Bot/PlayerbotAI.cpp` contains no `SetFallInformation` and no `IsFalling` at all,
`b949b50b:src/Ai/Base/Actions/CheckMountStateAction.cpp:285` still stores the pre-fall `startZ`,
and `RandomPlayerbotMgr.cpp:1673` is still `z = 0.05f + ground;` on both commits. On the core side
`Player::IsFalling`, the `LFGMgr` gate and the single caller of `UpdateFallInformationIfNeed` are
all unchanged in the version I am running. The upstream movement did not fix this.

Worth knowing separately: every other consumer of `Player::IsFalling()` is wrong for bots for the
same reason, quietly. The liquid status check in `Unit.cpp` is one of them. This PR fixes the
cause rather than the LFG symptom, so those should come right as well, but LFG is the only one I
have tested.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where bots could remain incorrectly marked as falling after landing or movement.
  - Bots now correctly retain genuine falling states for fall-damage handling.
  - Prevented outdated altitude data from causing persistent falling indicators.


Celandriel:
@coderabbitai full review

kadeshar (dismissed):

Celandriel:
Please greatly reduce the comments in the PR. They should be no more than 1 maybe 2 lines and only of absolutely necessary

Celandriel (approved):

PR #2766, opened 2026-09-09
Align Rogue strategy names with specs

brighton-chi:
## Pull Request Description

Combat Rogues have the "dps" strategy for their spec and Assassination/Subtlety Rogues have the "melee" strategy for their spec. The Combat Rogue strategy files are also called "DpsRogue." This is confusing so the files and strategies have been renamed to reflect the specs (with the ingame strategy being called "combat" instead of "dps" and "assassin" instead of "melee").

Placeholders were also added for Subtlety Rogue strategies. No strategies were added. After this PR, they'll get the "assassin" strategy, which is the same in effect as what they get now.

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [x] No

 MaNGOS, another module)?

## Final Checklist


Wishmaster117:
Hi if merged please tag with MultiBot need update please and wiki need update too I think.

src/Ai/Class/Rogue/RogueTriggers.cpp
```diff
@@ -25,12 +25,11 @@ bool UnstealthTrigger::IsActive()
     if (!botAI->HasAura("stealth", bot))
         return false;
 
-    return botAI->HasAura("stealth", bot) && !AI_VALUE(uint8, "attacker count") &&
+    return !AI_VALUE(uint8, "attacker count") &&
```
Celandriel: Confirm design intention. 

Before:
```
return botAI->HasAura("stealth", bot) && !AI_VALUE(uint8, "attacker count") &&
       (AI_VALUE2(bool, "moving", "self target") &&
        ((botAI->GetMaster() && IsDistanceGreaterThan(distance to group leader, 10.0f) &&
          AI_VALUE2(bool, "moving", "group leader")) ||
         !AI_VALUE(uint8, "attacker count")));

```
Two of the three removed checks really were dead: the repeated HasAura("stealth", bot) duplicated the early return two lines above, and dropping it was correct. The third was not. !AI_VALUE(uint8, "attacker count") is a required conjunct of the outer &&, so by the time control reaches the inner disjunction that same expression is already known true. The disjunction therefore evaluated to true unconditionally, and the whole trigger reduced to:

> stealthed and not being attacked and moving

The master/distance/leader-moving clause was the dead arm. The PR deleted the live arm and kept the dead one, so the trigger now reads:

> stealthed and not being attacked and moving and has a master and group leader more than 10 yards away and leader is moving

A solo rogue, or one whose leader is standing still or nearby, can no longer fire UnstealthTrigger at all. Consequence is contained by the same orphan as finding 1 (only StealthedRogueStrategy consumes the "unstealth" trigger, CombatRogueStrategy.cpp:310-316) and by GenericRogueNonCombatStrategy.cpp:49, which queues the unstealth action on the "often" trigger at 30.0f regardless. But the commit message says "delete dead checks" and the PR template answers "No" to does this change modify default bot behavior; both are wrong for this hunk.

Consider original semantics explicitly — `return !AI_VALUE(uint8, "attacker count") && AI_VALUE2(bool, "moving", "self target");`

Celandriel (commented):
Fix: CheckStealthAction still toggles the deleted "dps" strategy
src/Ai/Class/Rogue/Action/RogueActions.cpp:47-56

src/Ai/Class/Rogue/RogueTriggers.cpp
brighton-chi: Mm did I delete the wrong one? Sec

brighton-chi: This is why we should never stack all these checks into one return. So I am following what is being said. The question then is do we keep it as is (makes the master too far away and moving stuff useless), or do we delete the outer check (makes it actually operative but not required, which is surely what was intended)?

brighton-chi:
> Fix: CheckStealthAction still toggles the deleted "dps" strategy
> File: src/Ai/Class/Rogue/Action/RogueActions.cpp:47-56

I am confused about what that action is even doing and why it is for combat spec specifically. But figuring that out is for another day.

A lot of the stuff in dpsrogue (combatrogue) should be ported to the generic rogue strategy. But I wanted to get the files set up first.

Celandriel (approved):

brighton-chi:
Just to have this on record: we ended up actually making a functional change to the unstealth trigger. Previously, part of the trigger was dead, and now it will be active so Rogues will unstealth if they have a master and the master is too far away and moving.

PR #2775, opened 2026-09-11
feat(toc): rework mounted joust combat

jrad7:
## Pull Request Description

In the Trial of the Champion joust, bots used the vehicle spells but never steered their mounts. Charge only landed when a champion happened to be in range, and nobody went after unhorsed champions.

Bots now joust properly:

- Focus the lowest-health hostile and lead with Charge, spreading out to get charge range.
- Circle the target and Thrust while Charge recovers.
- Keep Defend up and use Shield-Breaker without delaying the next Charge.
- Trample unhorsed champions before they can remount.
- Respect the joust spells' real cooldowns.
- Head straight for a spare mount when unhorsed instead of stopping to fight on foot.

Other movement is blocked while mounted in combat so it can't pull the horse back into melee.

## Feature Evaluation

- **Minimum logic:** one action drives the mount (pick a target, charge, circle, defend), plus one multiplier that blocks competing movement while mounted and combat while walking to a spare mount.
- **Processing cost:** only applies to bots in Trial of the Champion. The arena is scanned once per tick while jousting and every 500 ms otherwise. The lance and mount triggers now run every 500 ms instead of every tick.

## How to Test the Changes

1. Enter Trial of the Champion (normal or heroic) with a bot party.
2. Start the Grand Champions event.
3. Bots should mount, charge from range, circle between charges, keep Defend stacked, and trample unhorsed champions until the joust ends. An unhorsed bot should run straight to a free mount without stopping to fight.

## Impact Assessment

    - [ ] No, not at all
    - [x] Minimal impact

  Only runs inside Trial of the Champion, and the trigger checks are cheaper than before.

    - [ ] No
    - [x] Yes

  Only during the Trial of the Champion joust.

    - [ ] No
    - [x] Yes

  The joust action gains its own movement logic, contained to the ToC strategy.

## AI Assistance

- [ ] No
- [x] Yes

Used for code generation, refactoring and this write-up. All changes were reviewed and tested in-game.

## Final Checklist

- [x] Changes are understood and tested for server stability and performance impact.
- [x] Any new bot dialogue lines are translated. (none added)
- [x] New source files use the GPLv2 header. (no new files)
- [x] Documentation updated if needed. (none needed)

## Notes for Reviewers

Changes are limited to `src/Ai/Dungeon/TOC/`.

### DIFF
--- a/src/Ai/Dungeon/TOC/TOCActions.cpp
+++ b/src/Ai/Dungeon/TOC/TOCActions.cpp
@@ -5,11 +5,100 @@
  */
 
 #include "TOCActions.h"
+#include "AttackersValue.h"
+#include "CellImpl.h"
 #include "GenericSpellActions.h"
-#include "NearestNpcsValue.h"
+#include "GridNotifiers.h"
+#include "GridNotifiersImpl.h"
+#include "Group.h"
 #include "ObjectAccessor.h"
 #include "Playerbots.h"
+#include "SpellMgr.h"
+#include "Timer.h"
 #include "Vehicle.h"
+#include <algorithm>
+#include <cmath>
+
+namespace
+{
+constexpr uint32 JOUST_IDLE_SCAN_MS = 500;
+constexpr float JOUST_SCAN_RANGE = 60.0f;
+constexpr float JOUST_THRUST_RANGE = 6.0f;
+constexpr float JOUST_CHARGE_MIN = 8.0f;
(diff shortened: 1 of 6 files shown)


jrad7:
full bot party jousting on TOC heroic: https://streamable.com/1el7qf

kadeshar:
Conflict to resolve

jrad7:
fixed conflicts

kadeshar (approved):

PR #2777, opened 2026-09-12
fix: stop cyclic craft chains from overflowing the stack

madosa-dev:
## Pull Request Description

`ItemUsageValue::IsItemNeededForUsefullSpell` asks for the `"item usage"` of the item a spell crafts, and that value's `Calculate()` comes straight back into `IsItemNeededForUsefullSpell` for the crafted item. Crafting chains can be cyclic, so this recursion never terminates.

Nothing breaks the cycle today. `"item usage"` is constructed with `checkInterval == 1`, so `CalculatedValue::Get()` takes the `checkInterval < 2` branch and recalculates on every access instead of memoising. A re-entrant call on the same value object therefore recurses again rather than returning the in-flight result. (Values with `checkInterval >= 2` are immune: they set `lastCheckTime` *before* calling `Calculate()`, so the nested call returns the cached value.)

A bot that has learned the alchemy transmutes runs the chain until its thread stack overflows, taking the worldserver down with `SIGSEGV`.

**The cycles are real data, not hypothetical.** Verified against `item_template` on a 3.3.5a world DB — every item below is `class 7` / `subclass 10` (Trade Goods / Elemental), which is exactly the `ITEM_CLASS_TRADE_GOODS` branch in `Calculate()` that reaches this function:

| Cycle | Items | Spells |
|---|---|---|
| 4-cycle | Essence of Earth (7076) → Water (7080) → Air (7082) → Fire (7078) → Earth | 17561, 17562, 17559, 17560 |
| 2-cycle | Essence of Earth (7076) ↔ Living Essence (12803) | 17566, 17565 |
| 2-cycle | Essence of Water (7080) ↔ Essence of Undeath (12808) | 17564, 17563 |
| 4-cycle | Primal Earth (22452) → Water (21885) → Air (22451) → Fire (21884) → Earth | 28567, 28569, 28566, 28568 |

### Evidence

Two crashes on the same day on a server with ~2000 bots. Both core dumps show the same repeating frame triple:

```
ItemUsageValue::Calculate()                    ItemUsageValue.cpp:49
  → IsItemNeededForUsefullSpell(item)          ItemUsageValue.cpp:688
    → AI_VALUE2(ItemUsage, "item usage", craftedItem)
      → CalculatedValue<ItemUsage>::Get()      Value.h:80
        → Calculate()  … repeats until SIGSEGV
```

Reading the `ItemTemplate*` arguments out of the dumps identifies the cycle in each case: the first crash was the Earth ↔ Living Essence 2-cycle, the second the four-essence cycle.

## Feature Evaluation

**Minimum logic:** one `thread_local std::unordered_set<uint32>` of the item ids currently being evaluated further up the stack. An RAII guard inserts the id on entry and erases it on exit; if the insert fails the id is already in flight, the cycle has closed, and the function returns `false`. That is the smallest construct that terminates the recursion without changing any non-cyclic result. `thread_local` rather than a member or a global because bots are updated from several threads and the set describes one call stack.

**Processing cost:** one hash insert plus one erase per call to `IsItemNeededForUsefullSpell`, on a set whose size is the current recursion depth (1 for the overwhelming majority of items — anything not in a craft cycle). The function is only reached from `ItemUsageValue::Calculate` for `ITEM_CLASS_TRADE_GOODS` / `MISC` / `REAGENT` items when the bot has no real-player master.

That cost is negligible next to what the function already does on the same call: iterating `SpellsUsingItem`, an `sSpellMgr->GetSpellInfo` per spell, and an `AI_VALUE2` lookup that formats a `std::string` key and hashes it into the context map. **The net change is a reduction**, because the recursion it cuts off was itself doing that work repeatedly, thousands of frames deep, before killing the process.

## How to Test the Changes

Reproduce on an unpatched build:

1. Start a world with bots enabled; no special config needed.
2. Take any bot with Alchemy and teach it both directions of one transmute cycle, e.g. `.learn 17566` (Earth to Life) and `.learn 17565` (Life to Earth). The four-essence cycle works too: `.learn 17559 17560 17561 17562`.
3. Put one of the cycle items in the bot's bags, or let it evaluate one — any path that triggers `"item usage"` for it (looting, vendor/AH evaluation, bag-space checks).
4. The worldserver dies with `SIGSEGV`. A backtrace of the core shows `IsItemNeededForUsefullSpell` / `Calculate` / `Get` repeating to the top of the stack.

With the patch, step 4 does not happen: the usage query returns normally and the bot keeps behaving as before.

Verifying the guard did not break non-cyclic items: items whose craft chain terminates (ore → bar, herb → potion) still resolve to the same `ItemUsage` as before, because the guard only ever triggers on an id that is already on the stack.

## Impact Assessment

    - - [ ] No, not at all

One hash-set insert and erase per `IsItemNeededForUsefullSpell` call, on a set that holds a single element outside of a craft cycle. It does not add a per-tick or per-trigger cost — the function is only reached from an `"item usage"` calculation that already performs several map lookups and string formats.

I have **not** run the standardised `pmon` comparison, so I am not claiming a measured number. I marked "minimal" rather than "not at all" because the change is not literally free. If you would like the `pmon` before/after run with `BotActiveAlone = 100` / `botActiveAloneSmartScale = 0`, say so and I will provide it.

    - - [ ] No
    - - [x] Yes (**explain why**)

Only for items inside a craft cycle, and only because the previous behavior was a crash. When a cycle closes, the item is reported as not needed for a useful spell. Concretely: a bot no longer keeps or buys elemental essences *for the sake of the transmute chain itself*. It still keeps them through `IsItemNeededForSkill`, quest usage, equip/vendor/AH evaluation and every other path in `Calculate`.

Returning `false` is the conservative resolution — the honest answer to "is this needed?" inside a cycle is undefined, and `false` avoids inventing a need that would make bots hoard four essences in a loop.

    - - [ ] No

One early-return branch at the top of one function, plus a small RAII guard class in an anonymous namespace in the same file. No new tunables, no config, no public API change. The comment above the guard names the concrete cycles so the next reader does not have to rediscover them from a core dump.

## AI Assistance

- - [ ] No

Yes, and substantially — I would rather over-disclose here than under-disclose.

- **Purpose:** crash triage and root-cause analysis (loading both core dumps in gdb, reading the repeating frames, extracting the `ItemTemplate` arguments to identify which items formed the cycle), then code generation for the fix and this description.
- **Which parts:** the whole 41-line diff was AI-generated, as was this PR text.
- **Review:** the diagnosis is not taken on trust — it is reproducible from the two core dumps, and the item/spell tables above were verified independently with SQL against `item_template` rather than recalled. (Worth noting: the first recalled set of transmute spell IDs was wrong and the DB query corrected it, which is exactly why they were checked.)
- **What I am confident about:** the mechanism, the cycle data, and why `checkInterval == 1` defeats the existing cache. **What I have not done:** a `pmon` measurement, or a long-running soak beyond the uptime reported below.

 MaNGOS, another module)?

Original. No file headers were touched.

## Final Checklist

## Notes for Reviewers

- **Targeting `test-staging`** per the template and the recent merge history.
- **Uptime so far:** the patched binary has been serving a live world with active bots for ~1.5 hours without recurrence, against two crashes in the preceding day on the same world. That is a useful signal, not a soak test, and I am presenting it as such.
- **Scope:** `ItemUsageValue.cpp:729` is the only self-recursive `"item usage"` lookup in the module. The other ~20 call sites across actions and values are one-level lookups that cannot re-enter `IsItemNeededForUsefullSpell`, so they are unaffected.
- **Alternative considered:** giving `"item usage"` a `checkInterval >= 2` would also break the recursion via the existing `lastCheckTime` mechanism, but that silently changes caching semantics for every consumer of the value and would mask the cycle rather than handle it. The guard is the narrower fix.


Celandriel:
Please use PR template

madosa-dev:
Sorry about that — I checked `.github/` for a template and missed that `PULL_REQUEST_TEMPLATE.md` sits in the repo root. The description now follows it in full.

Two things worth flagging while you review:

- The AI Assistance section is filled in honestly and is not a small "yes": the analysis and the diff were AI-assisted. The diagnosis is reproducible from the core dumps, and the item/spell cycle table was verified with SQL against `item_template` rather than recalled — the first recalled set of transmute spell IDs was wrong, which is why they were checked.
- I have not run the standardised `pmon` before/after, so I marked "Minimal impact" rather than "No, not at all" and said so explicitly instead of implying a measurement I did not take. Happy to run it if you want the numbers.

Also noted from the template that PRs go to `test-staging` — this one already targets it.

Celandriel (dismissed):

Celandriel (approved):

PR #2781, opened 2026-09-13
fix(loot): value quest source items via the core quest-item predicate

avirar:
## Pull Request Description

**TLDR:** Bots skipped quest source items (quest keys) on corpses because `IsItemUsefulForQuest` only credited direct objectives and CREATE_ITEM items. It now also consults the core's `Player::HasQuestForItem` - the same predicate the loot system uses to decide what drops - so quest keys are valued. The local objective loop stays: `HasQuestForItem` skips non-raid quests for raid-grouped players.

## Feature Evaluation

- Minimum logic: one call to `HasQuestForItem` added alongside the existing objective loop; event-driven.
- Processing cost: net-neutral (one extra predicate call).

## How to Test the Changes

1. Quest 1667 "Dead-Tooth Jack" (Elwynn Forest): bot loots Dead-Tooth's Key (6783) from the corpse (previously "Useless"), then opens the strongbox.
2. Regression: objective items still valued; a bot with the quest complete no longer values the key.

## Impact Assessment

    - - [x] No, not at all

    - - [x] Yes (quest source items now valued while the quest is incomplete)

    - - [x] No

## AI Assistance

- - [x] Yes - pi (coding agent) with Qwen3.8-27B-NVFP4 (vLLM), DeepSeek used for review; all code reviewed and verified by me.

## Final Checklist

- - [x] Documentation updated if needed. (code comments added)

## Notes for Reviewers

- Aligns the module's valuation with the core loot gating predicate (`needs_quest` -> `HasQuestForItem`).
- The local `RequiredItemId` loop is retained for raid-grouped players, whom the core predicate skips for non-raid quests.


src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -481,25 +481,24 @@ bool ItemUsageValue::IsItemUsefulForQuest(Player* player, ItemTemplate const* pr
     if (!botAI)
         return false;
 
+    // Use the core's canonical quest-item predicate - the same one the loot
```
Celandriel: Shorten

src/Ai/Base/Value/ItemUsageValue.cpp
```diff
@@ -481,25 +481,24 @@ bool ItemUsageValue::IsItemUsefulForQuest(Player* player, ItemTemplate const* pr
...
+    // system uses to decide whether a quest item is offered at all
+    // (LootTemplate::HasQuestDropForPlayer / needs_quest gating). It covers both
+    // direct objectives (RequiredItemId) and quest source items
+    // (ItemDrop/RequiredSource - e.g. the key to a quest chest) with the proper
+    // owned-count/MaxCount/stack checks. Keeping the bot's valuation on this
+    // predicate means it agrees with what actually drops in a loot window, and
+    // quest keys are no longer dismissed as "Useless".
+    if (player->HasQuestForItem(proto->ItemId))
```
Celandriel, line 487: Claude Find: `Player::HasQuestForItem` skips non-raid quests while the player is in a raid group (`PlayerQuest.cpp:2321`). Bots in an open-world raid (anyone running 5+ bots) will now rate objective items they're carrying as Useless, which shows up in `.qi` and trade and makes full-bag random bots destroy them first. Keeping the old required-item loop alongside the new call avoids that.

Celandriel: Minor: once the key is looted it goes back to Useless, because the owned count now equals the required count. So "no longer Useless" only holds while the key is still on the corpse. `turnIn=true` would fix that, but it overlaps the keep-count logic in #2789, so it's probably best settled there.

src/Ai/Base/Value/ItemUsageValue.cpp
avirar: Done - trimmed to three lines.

src/Ai/Base/Value/ItemUsageValue.cpp
avirar: Fixed - restored the loop; the core predicate skips non-raid quests for raid groups.

avirar: Agreed - deferring to #2789.

PR #2793, opened 2026-09-15
Use core modular database functionality

Celandriel:
## Pull Request Description

One big reason for our own core was database handling. Now the core has enabled modules to manage their own. These are the changes required to do this, as well as moving sql into statements that can be properly managed. 

## Feature Evaluation

## How to Test the Changes

New db, everything works as expected. 

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

 MaNGOS, another module)?

## Final Checklist


kadeshar (approved):

Hisha:
## Possible world-thread performance regression following Playerbots database migration

Hi! I've been testing the recent Playerbots database migration on my AzerothCore PTR and wanted to share a performance issue I've encountered, along with the troubleshooting results.

I'm running two separate AzerothCore instances on the same machine: a LIVE realm and a PTR realm. They use separate character, world, and Playerbots databases on the same MySQL server. Both have approximately 1,500 Playerbots.

After updating PTR to the newer AzerothCore Playerbot branch and mod-playerbots revisions, I began experiencing severe delays in chat, gossip, and other in-game actions.

### 1. Versions tested

**Newer PTR configuration (severe lag):**

* AzerothCore Playerbot branch: `7f12e89ee`
* mod-playerbots: `7bae1b5c`

**Known-good configuration (no noticeable lag):**

* AzerothCore Playerbot branch: `025ede00f`
* mod-playerbots: `9a12dc79`

The newer configuration experienced repeated world-update stalls of approximately 12–25 seconds.

After rolling PTR back to the known-good core and module revisions together, rebuilding, and restarting, the noticeable lag disappeared.

This was tested on the same PTR instance, using its existing databases and configuration.

### 2. What the debugger showed

While PTR was experiencing the lag, I captured the worldserver thread using GDB on multiple occasions.

The world thread was repeatedly observed waiting for a MySQL `COMMIT` initiated by Playerbots:

```text
World::Update()
  -> ScriptMgr::OnWorldUpdate()
  -> PlayerbotAIBase::UpdateAI()
  -> RandomPlayerbotMgr::UpdateAIInternal()
  -> RandomPlayerbotMgr::ProcessBot()
  -> RandomPlayerbotMgr::SetEventValue()
  -> PlayerbotsDatabasePool::CommitTransaction()
  -> ModuleDatabasePool::DirectCommitTransaction()
  -> MySQLConnection::ExecuteTransaction()
  -> MySQLConnection::CommitTransaction()
  -> MySQLConnection::Execute("COMMIT")
  -> mysql_real_query()
  -> SSL_read()
```

The captures involved different bots but showed the same transaction path.

I also captured the LIVE worldserver during the same period. It was responsive and was observed waiting for map-worker completion rather than a Playerbots database transaction.

Both instances use the same MySQL server, and I observed transient `waiting for handler commit` states on both LIVE and PTR. The important difference appears to be which thread is waiting for the database operation.

### 3. Difference in transaction behavior

The older implementation uses the core's `DatabaseWorkerPool<PlayerbotsDatabaseConnection>`.

Its `CommitTransaction()` enqueues a `TransactionTask`, allowing the calling thread to continue while the database worker executes the transaction.

The newer module-owned `PlayerbotsDatabasePool` instead implements:

```cpp
void CommitTransaction(PlayerbotsDatabaseTransaction transaction)
{
    DirectCommitTransaction(transaction);
}
```

Consequently, calls such as `RandomPlayerbotMgr::SetEventValue()` now execute the transaction synchronously.

The `SetEventValue()` call site itself appears unchanged between the two revisions, but the behavior of `CommitTransaction()` has changed from queued execution to direct execution.

My concern is that frequently executed Playerbots state updates are now able to block the world-update thread while waiting for MySQL.

### 4. Configuration findings

I also checked whether the new database implementation requires different connection or worker-thread settings.

The latest `playerbots.conf.dist` still documents:

```ini
PlayerbotsDatabase.WorkerThreads = 1
PlayerbotsDatabase.SynchThreads = 1
```

However, searching the latest module source found that `PlayerbotsDatabase.WorkerThreads` is only referenced in the configuration file.

The database initialization code reads `PlayerbotsDatabase.SynchThreads` and passes it to:

```cpp
PlayerbotsDatabase.SetConnectionInfo(dbString, synchThreads);
```

The module database pool also documents that it is synchronous and has no asynchronous queue.

Therefore, increasing `PlayerbotsDatabase.WorkerThreads` does not appear to restore the previous asynchronous transaction behavior.

### 5. Additional checks

To investigate whether this was simply an overloaded database or an unusually large PTR table, I compared both environments.

* Both Playerbots databases contain approximately 17,000–18,000 random-bot event rows.
* Both have approximately 1,500 `update` event rows.
* The `playerbots_random_bots` table has an index covering `(owner, bot, event)`.
* Both instances share the same MySQL server.
* MySQL's observed peak connection usage was well below `max_connections`.
* The PTR worldserver was not experiencing sustained CPU or memory saturation.

These observations do not rule out database latency as a contributing factor, but they suggest that the changed transaction execution path is particularly relevant.

### 6. Question / possible next step

I noticed that the new `ModuleDatabasePool` infrastructure initially deferred asynchronous execution support.

Was the change from queued to synchronous transaction execution an intentional part of the Playerbots migration, or is restoring asynchronous transaction support planned?

The module-owned database architecture itself seems reasonable. My concern is specifically that existing world-thread callers now wait for database writes that were previously queued.

I'm happy to provide additional GDB captures, configuration details, or run targeted tests on PTR if they would help.

Thanks for the work on Playerbots!

Celandriel:
We should do two things. Drop runtime SQL callers by preloading and then make the DB async. Ill start on the core update

PR #2796, opened 2026-09-15
fix(addclass): pick a random eligible bot instead of the first one

NSO73:
## Pull Request Description

`.playerbots bot addclass <class>` keeps adding the same character for a given class: add a mage, remove it, add a
mage again, and it is the same one.

The command walks `addclassCache`, an `std::unordered_set`, and returns the first eligible character. An
unordered_set is iterated in the same order as long as it is not modified, so the first eligible character is
always the same one.

This change walks a shuffled copy of the class pool (`Acore::Containers::RandomShuffle`) instead of the set itself.
The eligibility checks and the early return are unchanged.

## Feature Evaluation

  Copy the pool of the requested class into a vector and shuffle it before the existing loop. The first eligible
  character of a random permutation is a uniform pick among the eligible ones, so nothing else is needed: no new
  state, no config option, no change to the checks.

  It does not run per bot or per tick, only when a player types the command. Each call adds one copy and one
  shuffle of a single class and faction pool (500 characters in total with the default
  `AddClassAccountPoolSize = 50`, split across both factions and all classes). The loop still returns at the first
  eligible character, so the per-candidate checks, including the `characters` query made when a gender is
  requested, do not run over the whole pool.

## How to Test the Changes

1. Log in with a character allowed to use the command (`AiPlayerbot.AddClassCommand = 1`, or a GM account).
2. `.playerbots bot addclass mage`, note the name of the added bot, then `.playerbots bot remove <name>`.
3. Repeat step 2 several times.
4. Repeat steps 2-3 with a gender: `.playerbots bot addclass mage female`.

Expected: without this change, step 3 adds the same mage every time. With it, the name varies between calls, and
in step 4 every added bot is female.

Observed on a live 3.3.5a server running this change (Alliance master at level 30, stock
`AddClassAccountPoolSize = 50`): 15 consecutive `addclass mage` calls, each added bot removed before the next
call, returned **10 different characters** (the most frequent one 4 times). 8 calls of `addclass mage female`
returned 5 different characters, all of them female.

## Impact Assessment

    - - [x] No, not at all

  The change is only on the `addclass` chat command path, which no bot runs.

    - - [x] No
    - - [ ] Yes (**explain why**)

  Bot AI is untouched. Only the choice of which available character `addclass` logs in changes.

    - - [x] No

  No new branch: two lines before the existing loop, which is otherwise unchanged.

## AI Assistance

- - [ ] No

AI assistance was used to read the command's code path, write the change and draft this description (English is not
my first language). I reviewed the change, built it and tested it on a live server before submitting.

 MaNGOS, another module)?

## Final Checklist

- - [x] Any new bot dialogue lines are translated. (N/A: no new dialogue.)
- - [x] New source files use the GPLv2 header. (N/A: no new file.)
- - [x] Documentation updated if needed (Code comments, Conf comments, WiKi commands). (N/A: command syntax unchanged.)

## Notes for Reviewers

- Alternative considered: collect every eligible character, then pick one with `urand`. Not kept: it runs all the
  checks over the whole pool, including one `characters` query per candidate when a gender is given, where the
  shuffle keeps the early return.
- `Acore::Containers::RandomShuffle` is already used in `TravelMgr.cpp`.
- The numbers above come from a scripted 3.3.5a client (the core's e2e harness) repeating the add/remove cycle,
  not from clicking through a retail client.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved candidate selection for class assignment by randomizing the order in which eligible candidates are evaluated.
  * Existing eligibility checks, including gender, loading status, online status, and guild membership, remain unchanged.


NSO73:
@coderabbitai review

NoxMax:
Might not be a desired behaviour. If someone wants that same character always, they should use an alt, but if someone uses addclass anyway, edits it, the next time they call onto that class they get another char (this is all assuming a single person is using the server and the addclass bot doesn't get called by other players anyway).

So I don't know about this. If implemented it should be behind a config.

Celandriel:
I share Nox's sentiment. I think for the majority of players who want to use addclass bots they expect the same ones to be used. In multiplayer settings this could be slightly different, but i dont see this as a merge right now.

RuralRob:
I think this should be configurable with a config setting (default to the old behavior).

NSO73:
Thank you for your feedback.

I initially thought this was a design issue, but there was probably a misunderstanding between the purpose of the function and how I was using it.

I play solo locally, and I use the function to control the classes in my instance groups (direct class selection, avoiding multiple characters of the same class).
I was always getting the same bots every time I created a new group. I know that this doesn’t make any real difference apart from the characters’ names, but it somewhat breaks the idea of a simulated living world, and I thought that was a shame.

I’ve made the necessary commits so that my proposed functionality can be enabled or disabled through a configuration option.

Feel free to let me know if you have any other comments or suggestions.

PR #2798, opened 2026-09-17
fix(quest): mirror auto-accept quests, which send no accept opcode

steviecraycray:
## Pull Request Description

A quest with `QUEST_FLAGS_AUTO_ACCEPT` is taken by the core inside
`WorldSession::HandleQuestgiverQueryQuestOpcode`:

```cpp
if (quest->IsAutoAccept() && _player->CanAddQuest(quest, true))
    _player->AddQuestAndCheckCompletion(quest, object);
```

The client therefore never sends `CMSG_QUESTGIVER_ACCEPT_QUEST` for it, and
`masterIncomingPacketHandlers` only listens for that opcode. The bots have
nothing to mirror, so they stay silent: no refusal, no error, nothing.

67 quests carry the flag on a 3.3.5a world database, and they happen to be the
starting zone quests - Dwarven Outfitters, Rude Awakening, Kobold Camp Cleanup,
Investigate Echo Ridge, Report to Goldshire. At low level that is most of what a
player picks up, so it reads as "bots never accept quests at all".

This adds a second handler on the query opcode with its own action, which acts
only when the quest really carries the flag.

Two returns in `AcceptQuestAction::Execute` were also silent, which made a bot
that never received the packet look exactly like one that refused the quest.
Every other way out of `AcceptQuest` already tells the master what happened.

## Feature Evaluation

**Minimum logic.** One extra entry in the opcode-to-handler map, and an action
that reads two fields already present in the packet, looks the quest up, and
returns immediately unless `IsAutoAccept()`. Acceptance itself reuses
`QuestAction::AcceptQuest` unchanged.

**Processing cost across many bots.** It runs only when the master sends
`CMSG_QUESTGIVER_QUERY_QUEST`, that is when a player opens a quest dialog. It is
not on any tick, trigger evaluation or bot update path. For a quest without the
flag the action returns after one `GetQuestTemplate` lookup and one flag test.
Bots with no master queue nothing, because the packet never reaches them.

## How to Test the Changes

1. Any realm, one player with at least one bot in the group.
2. Walk to a starting zone quest giver with an auto-accept quest, for example
   Dwarven Outfitters (179) in Coldridge Valley, and open the quest.
3. Before: the player has the quest, the bots do not, and say nothing.
   After: each bot that can take it whispers `Accepted [Dwarven Outfitters]`.
4. Counter-check with an ordinary quest: open the dialog and decline. The bots
   must not take it. Accept it, and they mirror it as before.

## Impact Assessment

  with thousands of bots?
    - [x] No, not at all
    - [ ] Minimal impact
    - [ ] Moderate impact

Nothing runs per tick. The work is bound to a packet the master sends by hand.

    - [ ] No
    - [x] Yes

Bots now mirror auto-accept quests, which is what the existing `quest` strategy
already intends for every other quest. Ordinary quests are untouched.

    - [ ] No
    - [x] Yes

One action class of about 20 lines beside the existing one, and one map entry.

## AI Assistance

- [ ] No
- [x] Yes

Claude was used to trace the opcode path, to count the affected quests in the
world database, and to draft the code and this description. The diagnosis was
made by elimination against a live server: a probe on the packet handler showed
`CMSG_QUESTGIVER_QUERY_QUEST` (0x186) arriving while `CMSG_QUESTGIVER_ACCEPT_QUEST`
(0x189) never did, although the quest was in the player's log. Every claim above
was checked against the source or the database rather than assumed, and the
change was built and tested in game before submitting.

- [ ] Yes


steviecraycray:
Closing this: sending it here was my mistake, not a considered decision. I want to verify the impact on a stock setup properly before taking anyone's review time. Apologies for the noise.

PR #2799, opened 2026-09-17
fix(crash): guard the four unchecked sTalentTabStore lookups

steviecraycray:
## Pull Request Description

`Talent.dbc` can name a `TalentTab` id that `TalentTab.dbc` does not define, and
`sTalentTabStore.LookupEntry` then returns `nullptr`.

Eight call sites in `PlayerbotFactory.cpp` look a tab up. Four dereference the
result immediately:

| function | dereference |
|---|---|
| `InitPetTalents` | `talentTabInfo->petTalentMask` |
| `InitTalentsBySpecNo` | `talentTabInfo->tabpage` |
| `InitTalentsByParsedSpecLink` | `talentTabInfo->tabpage` |
| `InitTalentsByTemplate` | `talentTabInfo->tabpage` |

The other four already check and `continue`. This makes the remaining four
consistent with them - same guard, same indentation, nothing else.

Stock 3.3.5a DBCs do not appear to trip this, so on a vanilla setup it reads as
defensive. On a custom client whose `Talent.dbc` carries rows beyond the stock
`TalentTab` set it takes the whole worldserver down, and it does so as soon as
the factory walks the talent store, which happens for every bot that gets
randomized.

## Feature Evaluation

**Minimum logic.** Four two-line null checks, identical to the four already in
the same file.

**Processing cost across many bots.** One pointer comparison per talent row per
call, on paths that already do a DBC lookup and several comparisons per row. It
is not measurable, and where the check fires it saves the rest of the loop body.

## How to Test the Changes

On stock data there is nothing to observe - no row hits the guard, and bots
behave exactly as before, which is the intended outcome.

To reproduce the crash it guards against, add a row to `Talent.dbc` whose
`TalentTab` column points at an id that `TalentTab.dbc` does not contain, then
add any bot. Before: the worldserver dies inside `PlayerbotFactory` on the first
randomize. After: that talent row is skipped and the bot is built.

## Impact Assessment

  with thousands of bots?
    - [x] No, not at all
    - [ ] Minimal impact
    - [ ] Moderate impact

A null comparison on a path that already performs a store lookup per row.

    - [x] No
    - [ ] Yes

On data where every `TalentTab` resolves, the guard never fires.

    - [x] No
    - [ ] Yes

It removes an inconsistency: four of eight sites were already written this way.

## AI Assistance

- [ ] No
- [x] Yes

Claude was used to find the call sites and to draft this description. The
change itself is four copies of a guard that already exists in the same file.
The failure was hit in practice on a custom-client server, not inferred from
reading.

- [ ] Yes


steviecraycray:
Closing this: sending it here was my mistake, not a considered decision. I want to verify the impact on a stock setup properly before taking anyone's review time. Apologies for the noise.

PR #2803, opened 2026-09-18
Dungeon Clean-ups

brighton-chi:
## Pull Request Description

The displayed strategy name for WotLK dungeons uses the full name of the dungeon, which doesn't match the context key name, which uses "wotlk-" with an abbreviation. This causes the WotLK dungeon strategies not to be removed upon zoning out and also breaks the sibling strategy aspect of them, meaning adding a WotLK dungeon strategy doesn't remove other dungeon strategies. I changed them all to use the "wotlk-" name since that's more generally consistent and consistent with TBC strategy names.

UP and VH have the same name for an action, each registered under "stop attack," which I believe means they conflict. AI tells me that UP always wins because it's the last write in BuildSharedActionContext.cpp, meaning VH is using the UP version. I've changed the names.

I threw out a bunch of unneeded includes and redundant virtual keywords while I was at it, but that's not a full sweep, only the files I touched for other reasons.

I also added missing licenses to the Mechanar files.

## Feature Evaluation

## Impact Assessment

    - - [x] No, not at all

    - - [x] No
    - - [ ] Yes (**explain why**)

    - - [x] No

## AI Assistance

- - [ ] No

Asked AI about the impact of the strategy name discrepancies and duplicated "stop attack" registration.

 MaNGOS, another module)?

## Final Checklist


Celandriel:
- Conflict Resolve
- Renaming VH's "stop attack" changes Zuramat. Today VH ends up running UP's `DropTargetAction`, because both register the same name and UP wins. After this PR it runs `ZuramatStopAttackAction`, which only calls `AttackStop()`, and ranged/casters keep hitting him through Shroud of Darkness. Claude suggests have both bosses queue the existing global `"drop target"` action and delete both aliases plus `ZuramatStopAttackAction`.
- That also lets you keep the `ChooseTargetActions.h` include in `UPActionContext.h`. Removing it undoes part of #2576, and the header only compiles because of include order.
- The template says default behaviour doesn't change, but it does (the strategies are now removed, and Zuramat). A few spots were missed: `virtual` in `HoLStrategy.h:16-18`, the unused `Multiplier.h` include in `PoSStrategy.h:10`, and a stray space before `#ifndef` in `MechMultipliers.h:7`.

brighton-chi:
> - Conflict Resolve
> 
> - Renaming VH's "stop attack" changes Zuramat. Today VH ends up running UP's `DropTargetAction`, because both register the same name and UP wins. After this PR it runs `ZuramatStopAttackAction`, which only calls `AttackStop()`, and ranged/casters keep hitting him through Shroud of Darkness. Claude suggests have both bosses queue the existing global `"drop target"` action and delete both aliases plus `ZuramatStopAttackAction`.
> 
> - That also lets you keep the `ChooseTargetActions.h` include in `UPActionContext.h`. Removing it undoes part of #2576, and the header only compiles because of include order.
> 
> - The template says default behaviour doesn't change, but it does (the strategies are now removed, and Zuramat). A few spots were missed: `virtual` in `HoLStrategy.h:16-18`, the unused `Multiplier.h` include in `PoSStrategy.h:10`, and a stray space before `#ifndef` in `MechMultipliers.h:7`.
> 
> 

Makes sense, I'll address later. I think calling drop target is not a good way to do this btw but I'm not going to wade into the technicalities of the strategy. I'm good with going with the simplest way to preserve current (unintended lol) behavior.

Celandriel:
I'm not sure what the intended behavior is supposed to be, but I think the value is in knowing what it is if it was unintentional.

brighton-chi:
> I'm not sure what the intended behavior is supposed to be, but I think the value is in knowing what it is if it was unintentional.
> 

I would need somebody else to weigh in on what the mechanic is supposed to do in VH. If the intent is to get bots to entirely stop attacking everything, the UP action should work from an action perspective, and better than the inoperative VH one.

There may be issues with respect to associated multipliers or triggers that influence the effectiveness of the strategy as a whole with respect to this mechanic, but none of that is impacted here.

brighton-chi:
Updated, wasn't sure what was meant by keeping ChooseTargetActions.h in UPActionContext.h though. With aliases deleted, it seems to me that should be placed in UPStrategy.cpp. So that's what I did.

